Issue #11, opened 2025-12-23
feat: Add async context manager support to PipefyClient


adriannoes:
## Summary

Add async context manager support to [PipefyClient](cci:2://file:///Users/adrianno.esnarriaga/Coding/pipefy-mcp-server/src/pipefy_mcp/services/pipefy/client.py:12:0-76:85) and [BasePipefyClient](cci:2://file:///Users/adrianno.esnarriaga/Coding/pipefy-mcp-server/src/pipefy_mcp/services/pipefy/base_client.py:11:0-64:74) to allow reusing connections for multiple queries in batch operations.

## Motivation

This suggestion was raised in [PR #10 - Comment by @Danielmoraisg](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/10#discussion_r2642837306):

> We could make this into an async context manager so we can batch many queries under the same connection like this:
> ```python
> async with PipefyClient() as client:
>     card = await client.get_card(123)
>     await client.update_card(123, title="X")
>     await client.move_card_to_phase(123, 5)
> ```

## Proposed Usage

```python
# New usage (same connection for batch operations)
async with PipefyClient() as client:
    card = await client.get_card(123)
    await client.update_card(123, title="X")
    await client.move_card_to_phase(123, 5)

# Current usage (still works - backward compatible)
client = PipefyClient()
card = await client.get_card(123)

## Implementation Details
- Add __aenter__ and __aexit__ methods to BasePipefyClient
- Store session in self._session when using context manager
- Modify execute_query to reuse existing session if available
- Keep current behavior as fallback when not using async with

## Files to Modify:
- src/pipefy_mcp/services/pipefy/base_client.py
- src/pipefy_mcp/services/pipefy/client.py
- Add tests for context manager behavior

Danielmoraisg:
Closing as obsolete.

Both classes this issue targets are gone. `BasePipefyClient` no longer exists, and the SDK client is now `PipefyClient` / `PipefyEngine` / `Executors` in `packages/sdk/src/pipefy_sdk/client.py`.

The executor also went the other way on purpose. From `packages/sdk/src/pipefy_sdk/graphql_executor.py:121`:

> a fresh transport is opened per call so concurrent requests never share mutable transport state (avoids `TransportAlreadyConnected`)

An `async with client:` session shared across a batch reintroduces exactly that failure, so the proposed API is incompatible with the current design rather than merely unimplemented.

[...]

PR #14, opened 2025-12-23
Feature/add tool to search pipes

Danielmoraisg:
## Summary

This PR adds a new `search_pipes` tool that allows users to search for pipes across all organizations with intelligent fuzzy matching.

## Motivation

When working with Pipefy, users often only know the name (or partial name) of a pipe but need its ID to perform operations. This tool solves the common problem of "I know the pipe is called something like 'Sales Pipeline' but I don't know its ID."

### New Tool: `search_pipes`

- **Purpose**: Find a pipe's ID when you only know its name
- **Fuzzy Matching**: Uses [RapidFuzz](https://github.com/rapidfuzz/RapidFuzz) library with `WRatio` algorithm for intelligent, case-insensitive partial matching
- **Match Threshold**: 70% similarity score required for matches
- **Sorted Results**: Best matches appear first within each organization
- **Optional Filtering**: When no `pipe_name` is provided, returns all accessible pipes

### Files Changed

| File | Description |
|------|-------------|
| `pyproject.toml` | Added `rapidfuzz>=3.14.3` dependency |
| `pipe_service.py` | Implemented `search_pipes` method with fuzzy matching logic |
| `pipe_queries.py` | Added `SEARCH_PIPES_QUERY` GraphQL query |
| `client.py` | Exposed `search_pipes` through the main client |
| `pipe_tools.py` | Registered new MCP tool with proper annotations |
| `test_pipe_service.py` | Comprehensive test coverage for the new feature |

### Response Format

```json
{
  "organizations": [
    {
      "id": "123",
      "name": "My Organization",
      "pipes": [
        {
          "id": "456",
          "name": "Sales Pipeline",
          "description": "Track sales opportunities",
          "match_score": "75.2" # optional when using a search term
        }
      ]
    }
  ]
}
```

## Testing

Added parameterized tests covering:
- ✅ Returns all pipes when no filter provided
- ✅ Exact match ranking (exact matches appear first)
- ✅ Case-insensitive matching
- ✅ Partial matching
- ✅ Matches across multiple organizations
- ✅ Empty results when no matches found

### DIFF
--- a/src/pipefy_mcp/services/pipefy/base_client.py
+++ b/src/pipefy_mcp/services/pipefy/base_client.py
@@ -16,9 +16,7 @@ class BasePipefyClient:
     underlying `gql.Client` instance via constructor injection.
     """
 
-    def __init__(
-        self, schema: str | None = None, client: Client | None = None
-    ) -> None:
+    def __init__(self, schema: str | None = None, client: Client | None = None) -> None:
         """Create a base client.
 
         Args:
--- a/src/pipefy_mcp/services/pipefy/client.py
+++ b/src/pipefy_mcp/services/pipefy/client.py
@@ -41,9 +41,13 @@ class PipefyClient:
 
     async def move_card_to_phase(self, card_id: int, destination_phase_id: int) -> dict:
         """Move a card to a specific phase."""
-        return await self._card_service.move_card_to_phase(card_id, destination_phase_id)
+        return await self._card_service.move_card_to_phase(
+            card_id, destination_phase_id
+        )
 
-    async def update_card_field(self, card_id: int, field_id: str, new_value: Any) -> dict:
+    async def update_card_field(
+        self, card_id: int, field_id: str, new_value: Any
+    ) -> dict:
         """Update a single field of a card."""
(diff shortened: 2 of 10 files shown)


Danielmoraisg:
@adriannoes I've created more test cases, made the match score available to the llm, changed the match score to be configured in the tool if necessary and added it to the description. regarding the other points:

- Error handling on execute_query: I guess we can do it on https://github.com/gbrlcustodio/pipefy-mcp-server/issues/11 since we will have to touch the method anyway

- Typed Return Structure: I don't see many advantages of using pydantic for now as we don't do any validations and we just pass along the result up the chain back the llm. IMO we should keep it simple until we grow enough to start doing some validations or to improve type hints. does it make sense

- Pagination: the query we use is already paginated, I'd mess with it once we start having more use cases to cover more pages to find a specific pipe.

adriannoes (approved):
Looks great! All points from the previous review were addressed. Match scores are now exposed, the threshold is configurable, and the new test cases cover all the edge cases we discussed (accents, special characters, etc).

Handling error logic in Issue #11 and keeping return types as dicts for now is the right call to keep things simple.

PR #17, opened 2025-12-30
✨ feat: Improve `create_card` tool with pre-filled fields and client-side elicitation skip

gbrlcustodio:
This pull request improves the `create_card` tool by making it capable of having fields pre-filled by the LLM and skipping the elicitation flow if the client doesn't support it. Additionally, small refactoring was make to enhance maintainability and readability within the form model and pipe tools.

### DIFF
--- a/src/pipefy_mcp/tools/pipe_tools.py
+++ b/src/pipefy_mcp/tools/pipe_tools.py
@@ -36,7 +36,8 @@ class PipeTools:
 
             Args:
                 pipe_id: The ID of the pipe where the card will be created
-                fields: A dictionary of fields to pre-fill the card with.
+                fields: A dictionary of fields that can be pre-filled on the card.
+                        This argument should be provided when the LLM is aware of the intended values for certain fields.
                 required_fields_only: If True, only elicit required fields. Default: False
             """
             form_fields = await client.get_start_form_fields(


src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -18,40 +18,61 @@ def register(mcp: FastMCP):
         container = ServicesContainer.get_instance()
         client = container.pipefy_client
 
+        async def _create_card_interactive(
```
adriannoes: `_create_card_interactive` is defined inside `register` and this makes it hard to test in isolation. 

Consider moving it to a module-level function or a helper class.

src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -18,40 +18,61 @@ def register(mcp: FastMCP):
...
+            expected_fields = form_fields.get("start_form_fields", [])
+            await ctx.debug(f"Expected fields for pipe {pipe_id}: {expected_fields}")
+            await ctx.debug(f"Provided fields: {fields}")
+
+            if ctx.session.client_params.capabilities.elicitation:
+                return await _create_card_interactive(
+                    pipe_id, fields, expected_fields, ctx
+                )
+            else:
+                return await client.create_card(pipe_id, fields)
```
adriannoes: **Robustness**
Passing `None` to `create_card` might fail if the client expects a dict.

**Suggestion:** Default to empty dict if `None`.
```python
return await client.create_card(pipe_id, fields or {})
```

Danielmoraisg (approved):

src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -18,40 +18,61 @@ def register(mcp: FastMCP):
...
-
-            if result.action == "accept":
-                return await client.create_card(pipe_id, result.data.model_dump())
-
-            return {"error": "Card creation cancelled by user."}
+            expected_fields = form_fields.get("start_form_fields", [])
+            await ctx.debug(f"Expected fields for pipe {pipe_id}: {expected_fields}")
+            await ctx.debug(f"Provided fields: {fields}")
+
+            if ctx.session.client_params.capabilities.elicitation:
```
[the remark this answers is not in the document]
Danielmoraisg: I guess dricursor is complaining because on `create_card` tool ctx is indeed explicitly marked as optional with `| None`

Image

src/pipefy_mcp/tools/pipe_tools.py
gbrlcustodio: I’ve updated the implementation to align with the actual `fields` typing.

PR #39, opened 2026-03-11
fix: create transport per request to enable parallel tool calls

Danielmoraisg:
Closes #38

## Problem

When multiple MCP tools are invoked in parallel, they all call `execute_query()` in `BasePipefyClient`, which did `async with self.client as session:`. This calls `transport.connect()` on a **single shared `HTTPXAsyncTransport` instance**. The second concurrent call hit `TransportAlreadyConnected` because the transport only permits one active session at a time.

## Solution: fresh transport per request

Each `execute_query()` call now creates its own `HTTPXAsyncTransport` + `gql.Client`. Every concurrent request gets an isolated connection — no shared mutable state, true parallelism.

```python
async def execute_query(self, query, variables):
    transport = HTTPXAsyncTransport(
        url=self.settings.graphql_url,
        auth=self._auth,   # shared — reuses token cache
        timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
    )
    async with Client(transport=transport, fetch_schema_from_transport=False) as session:
        return await session.execute(query, variable_values=variables)
```

The `OAuth2ClientCredentials` instance (`self._auth`) is created **once per service** and shared across calls so `TokenMemoryCache` reuses the OAuth token — parallel requests don't each trigger a separate token refresh.

## Why not `asyncio.Lock`?

The lock-based approach (which was already in place as a stopgap) **serializes** all GraphQL requests: while one tool is waiting for the API, every other parallel tool call queues up behind it. This defeats the purpose of parallel invocation — latency becomes additive instead of concurrent. It's the correct guard for a shared resource, but the right fix is to remove the sharing, not to protect it.

## Why not a connection pool?

A pool (option 1 from the issue) keeps N clients alive and recycles them. It's the right choice when:
- Connection setup is expensive (e.g. TLS + schema introspection on every request).
- You want to cap outbound connections to the upstream API.

In our case `fetch_schema_from_transport=False` means there is no schema round-trip overhead. Each transport is a thin HTTPX client wrapper — construction is cheap. A pool adds complexity (acquire/release logic, pool sizing, leak detection) without a meaningful performance benefit here. The per-request approach is simpler and correct for the current load profile.

## Why not gql 4.x permanent session?

`gql` 4.x `connect_async(reconnecting=True)` keeps one long-lived WebSocket/HTTP session. This only works if the underlying transport allows **concurrent** `execute()` calls on the same session — HTTPX's async transport does not; calls are still serialized internally. It also ties us to a gql major-version upgrade with its own migration cost.

## Changes

| File | Change |
|---|---|
| `base_client.py` | Remove `client=` injection, `schema=` param, `_create_client()`. Create transport per `execute_query()` call. Share `_auth`. |
| `card_service.py` / `pipe_service.py` | Accept `settings: PipefySettings` instead of `client: Client`. |
| `client.py` | Pass `settings` to services. Remove `BasePipefyClient` import, `self.client` attr, `schema=` param. |
| `test_pipefy_base_client.py` | Remove obsolete injection tests. Mock at `execute_query` level via `patch`. |
| `test_card_service.py` / `test_pipe_service.py` / `test_pipefy_client.py` / `test_pipefy_facade.py` | Replace mock gql client injection with `service.execute_query = AsyncMock(...)`. |

## Test plan

- [x] `uv run pytest tests/` — all 221 tests pass
- [ ] Manually call two MCP tools simultaneously (e.g. `get_pipe` + `get_start_form_fields`) — both complete without `TransportAlreadyConnected`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pipefy_mcp/services/pipefy/base_client.py
+++ b/src/pipefy_mcp/services/pipefy/base_client.py
@@ -16,13 +16,17 @@ class BasePipefyClient:
     Creates a fresh transport per execute_query() call so parallel requests
     never share mutable transport state (avoids TransportAlreadyConnected).
     The OAuth2 auth instance is shared across calls to reuse the token cache.
+    Pass a pre-built auth instance to share it across multiple service instances
+    (e.g. from PipefyClient) so only one token cache exists for the whole client.
     """
 
     GRAPHQL_REQUEST_TIMEOUT_SECONDS: ClassVar[int] = 30
 
-    def __init__(self, settings: PipefySettings) -> None:
-        if settings is None:
-            raise ValueError("Settings must be provided to create a GraphQL client.")
+    def __init__(
+        self,
+        settings: PipefySettings,
+        auth: OAuth2ClientCredentials | None = None,
+    ) -> None:
         if settings.graphql_url is None:
             raise ValueError("GraphQL URL must be provided in settings.")
         if settings.oauth_url is None:
@@ -33,7 +37,7 @@ class BasePipefyClient:
             raise ValueError("OAuth client secret must be provided in settings.")
 
         self.settings = settings
-        self._auth = OAuth2ClientCredentials(
+        self._auth = auth or OAuth2ClientCredentials(
             token_url=settings.oauth_url,
             client_id=settings.oauth_client,
             client_secret=settings.oauth_secret,
@@ -51,6 +55,6 @@ class BasePipefyClient:
             timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
         )
         async with Client(
-            transport=transport, fetch_schema_from_transport=False
+            transport=transport, fetch_schema_from_transport=True
         ) as session:
             return await session.execute(query, variable_values=variables)
--- a/src/pipefy_mcp/services/pipefy/base_client.py
+++ b/src/pipefy_mcp/services/pipefy/base_client.py
@@ -13,74 +13,42 @@ from pipefy_mcp.settings import PipefySettings
 class BasePipefyClient:
     """Base infrastructure for Pipefy GraphQL operations.
 
-    This class centralizes GraphQL client creation so services can reuse a single
-    underlying `gql.Client` instance via constructor injection.
+    Creates a fresh transport per execute_query() call so parallel requests
+    never share mutable transport state (avoids TransportAlreadyConnected).
+    The OAuth2 auth instance is shared across calls to reuse the token cache.
     """
 
     GRAPHQL_REQUEST_TIMEOUT_SECONDS: ClassVar[int] = 30
 
-    def __init__(
-        self,
-        settings: PipefySettings | None = None,
-        schema: str | None = None,
-        client: Client | None = None,
-    ) -> None:
-        """Create a base client.
-
-        Args:
-            schema: Optional schema string to pass to `gql.Client`.
-            client: Optional pre-built `gql.Client` to reuse (preferred for shared wiring).
-
-        Raises:
-            ValueError: If both schema and client are provided, as schema would be ignored.
-        """
-        if schema is not None and client is not None:
-            raise ValueError(
-                "Cannot specify both 'schema' and 'client'. "
-                "When reusing an existing client, its schema is already configured."
-            )
-
-        self.settings = settings
-        self.client: Client = client or self._create_client(schema)
-
-    def _create_client(self, schema: str | None) -> Client:
-        """Create and configure a `gql.Client` using project settings.
-
-        Note: This preserves the current behavior from `PipefyClient._create_client`.
-        """
-
-        if self.settings is None:
+    def __init__(self, settings: PipefySettings) -> None:
+        if settings is None:
             raise ValueError("Settings must be provided to create a GraphQL client.")
-
-        if self.settings.graphql_url is None:
+        if settings.graphql_url is None:
             raise ValueError("GraphQL URL must be provided in settings.")
-        if self.settings.oauth_url is None:
+        if settings.oauth_url is None:
             raise ValueError("OAuth URL must be provided in settings.")
-        if self.settings.oauth_client is None:
+        if settings.oauth_client is None:
             raise ValueError("OAuth client ID must be provided in settings.")
-        if self.settings.oauth_secret is None:
+        if settings.oauth_secret is None:
             raise ValueError("OAuth client secret must be provided in settings.")
 
-        transport = HTTPXAsyncTransport(
-            url=self.settings.graphql_url,
-            auth=OAuth2ClientCredentials(
-                token_url=self.settings.oauth_url,
-                client_id=self.settings.oauth_client,
-                client_secret=self.settings.oauth_secret,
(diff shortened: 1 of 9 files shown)


src/pipefy_mcp/services/pipefy/base_client.py
```diff
@@ -13,74 +13,44 @@
...
         """
-
-        async with self.client as session:
+        transport = HTTPXAsyncTransport(
+            url=self.settings.graphql_url,
+            auth=self._auth,
+            timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
+        )
+        async with Client(
+            transport=transport, fetch_schema_from_transport=False
```
adriannoes: The current code (`main`) defaults to `fetch_schema_from_transport=True`, which fetches the GraphQL schema on the first connection. The PR changes this to `False`, which is correct for performance (avoids an extra introspection round-trip per request). 

However, this is a silent behavioral change that should be called out explicitly in the PR description or as a code comment.

**Why it matters:** If any part of the stack relied on `gql` client-side validation via the downloaded schema (e.g., catching typos in queries before they hit the server), this would silently disable it. In practice, the project doesn't seem to use this — queries are pre-defined `gql()` constants — but it's worth confirming and noting.

**Suggestion:** Add a one-line comment on `fetch_schema_from_transport=False` explaining
the trade-off (no client-side schema validation, but no extra round-trip per call).

src/pipefy_mcp/services/pipefy/base_client.py
```diff
@@ -13,74 +13,44 @@
...
-        transport = HTTPXAsyncTransport(
-            url=self.settings.graphql_url,
-            auth=OAuth2ClientCredentials(
-                token_url=self.settings.oauth_url,
-                client_id=self.settings.oauth_client,
-                client_secret=self.settings.oauth_secret,
-            ),
-            timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
+        self.settings = settings
+        self._auth = OAuth2ClientCredentials(
```
adriannoes: The PR correctly shares `self._auth` to reuse the token cache. However, `httpx_auth.OAuth2ClientCredentials` uses `TokenMemoryCache` which does **not** document concurrency guarantees for async code. 

If two parallel `execute_query()` calls trigger a token refresh simultaneously, there's a potential race condition.

In practice, this may work fine because:
- Token refreshes are rare (tokens typically last 30+ minutes).
- `httpx_auth` might serialize auth calls internally via HTTPX's flow.

**Suggestion:**
1. Add a brief comment acknowledging this is safe under the current token lifetime assumptions.
2. Or consider wrapping the auth instance with an `asyncio.Lock` only for the refresh path if issues appear in production (not necessarily now — just flag it for future).

src/pipefy_mcp/services/pipefy/base_client.py
```diff
@@ -13,74 +13,44 @@
...
-
-    def _create_client(self, schema: str | None) -> Client:
-        """Create and configure a `gql.Client` using project settings.
-
-        Note: This preserves the current behavior from `PipefyClient._create_client`.
-        """
-
-        if self.settings is None:
+    def __init__(self, settings: PipefySettings) -> None:
+        if settings is None:
```
adriannoes: The PR changes the constructor to `def __init__(self, settings: PipefySettings)` (line 23), which is great. However, it then checks `if settings is None: raise ValueError(...)` (line 24). 

If the type hint says `PipefySettings` (not `PipefySettings | None`), this check becomes dead code for any type-checked caller.

**Suggestion:** Two options (both valid):
- Keep the check as defensive validation (useful since this is a public API).
- Or remove the check and let the type system enforce it — the `PipefySettings` fields themselves have their own validators.

Either way, the intent is clear; just worth a quick decision.

src/pipefy_mcp/services/pipefy/client.py
```diff
@@ -14,15 +11,9 @@
...
-
-        # Keep `client` as a public attribute for backward compatibility.
-        self.client: Client = graphql.client
-
-        # Service layer (domain logic lives here).
-        self._pipe_service = PipeService(self.client)
-        self._card_service = CardService(self.client)
+    def __init__(self, settings: PipefySettings):
+        self._pipe_service = PipeService(settings=settings)
+        self._card_service = CardService(settings=settings)
```
adriannoes: In the PR's design, `PipefyClient` creates `PipeService(settings)` and `CardService(settings)`. Each of these calls `BasePipefyClient.__init__`, which creates its own `OAuth2ClientCredentials` instance. This means:

- `PipeService._auth` and `CardService._auth` are **different** objects.
- Each has its own `TokenMemoryCache`, so the first request on each service will trigger
  a separate token refresh.

The PR description says *"The OAuth2ClientCredentials instance is shared across calls"*, which is true **within** a single service but not **across** services.

**Impact:** Two token refreshes on cold start instead of one. Not a correctness issue, but a minor efficiency gap compared to what the PR description implies.

**Suggestion:** If desired, create `_auth` once in `PipefyClient` and pass it to both services. Alternatively, document that token caching is per-service.

```python
# In PipefyClient.__init__:
auth = OAuth2ClientCredentials(
    token_url=settings.oauth_url,
    client_id=settings.oauth_client,
    client_secret=settings.oauth_secret,
)
self._pipe_service = PipeService(settings=settings, auth=auth)
self._card_service = CardService(settings=settings, auth=auth)
```

adriannoes (approved):
Solid delivery, Dan! The PR description explains the issue and the chosen approach well and the code is cleaner after the refactor.

I added **inline feedback on**:

- **#1** — Document the `fetch_schema_from_transport=False` trade-off (no client-side schema validation vs. no extra round-trip).
- **#2** — Briefly note the async-safety assumption for the shared `OAuth2ClientCredentials` / `TokenMemoryCache`.
- **#3** — Decide whether to keep or remove the defensive `if settings is None` check given the type hint.
- **#4** — Consider creating one `_auth` in `PipefyClient` and injecting it into both services so cold start does a single token refresh.

Nothing blocking — happy to merge and handle these in a follow-up if you prefer. Thanks again.

src/pipefy_mcp/services/pipefy/base_client.py
Danielmoraisg: I accidentally vibe changed this :hehe:, Just reverted it

src/pipefy_mcp/services/pipefy/base_client.py
Danielmoraisg: I'm now sharing the auth, so we should be good for now

src/pipefy_mcp/services/pipefy/base_client.py
Danielmoraisg: Removed the defensive clause and test on this

src/pipefy_mcp/services/pipefy/client.py
Danielmoraisg: we are now passing it for the services

PR #54, opened 2026-03-26
feat(tools): add search_tables for org-wide database discovery

MatheusFernandesDuarte:
## Summary

- Adds `search_tables` to `TableService`, mirroring the `search_pipes` pattern (rapidfuzz WRatio, 70% threshold)
- Adds `SEARCH_TABLES_QUERY` to `table_queries.py`
- Registers `search_tables` MCP tool in `TableTools` with `readOnlyHint=True`
- Adds `search_tables` delegation in `PipefyClient`
- 6 unit tests for the service layer (`tests/services/test_table_service.py`)

## Why

`pipe-full-toolset` exposes `get_table` and `get_tables` for fetching by known ID, but there is no way to discover tables when the ID is unknown. `search_tables` fills that gap — same UX as `search_pipes`.

## Test plan

- [ ] `pytest tests/services/test_table_service.py` — 6 tests (fuzzy match, no match, sort order, score field, empty orgs)
- [ ] `pytest tests/tools/test_table_tools.py` — all 48 existing table tool tests still pass (no regressions)
- [ ] Call `search_tables` without arguments → returns all tables across all orgs
- [ ] Call `search_tables(table_name="Clients")` → returns only fuzzy-matched tables with `match_score`

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@
 | Category | Tools | Description | Docs |
 |----------|:-----:|-------------|------|
 | **Pipes & cards** | 33 | Read, create, update, and delete pipes, phases, fields, labels, cards, and field conditions. | [Details](docs/tools/pipes-and-cards.md) |
-| **Database tables** | 15 | Tables, records (rows), and schema columns (table fields). | [Details](docs/tools/database-tables.md) |
+| **Database tables** | 16 | Tables, records (rows), schema columns (table fields), and org-wide table discovery. | [Details](docs/tools/database-tables.md) |
 | **Relations** | 6 | Link pipes, tables, and cards across workflows. | [Details](docs/tools/relations.md) |
 | **Reports** | 16 | Pipe and organization reports: discovery, CRUD, and async exports. | [Details](docs/tools/reports.md) |
 | **Automations & AI** | 15 | Traditional automations (rules engine) and AI-powered automations and agents. | [Details](docs/tools/automations-and-ai.md) |
--- a/src/pipefy_mcp/services/pipefy/client.py
+++ b/src/pipefy_mcp/services/pipefy/client.py
@@ -717,6 +717,10 @@ class PipefyClient:
         """Search for pipes across all organizations"""
         return await self._pipe_service.search_pipes(pipe_name)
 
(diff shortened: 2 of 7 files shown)


adriannoes (changes requested):
Nice work Matheus 👌🏻, the overall structure mirrors `search_pipes` well. Some things to address before merge:

**1. Missing substring match**

`search_pipes` has a substring check before falling back to fuzzy matching (`pipe_service.py:100-107`):

```python
if pipe_name.lower() in pipe_display_name.lower():
    matching_pipes.append((100.0, pipe))
else:
    score = fuzz.WRatio(...)
```

`search_tables` skips this and goes straight to fuzz.WRatio. 

Please add the same substring check so both tools behave consistently, obvious matches (e.g. searching "Client" against "Clients") should always score 100.

**2. Tool docstring needs more detail**

Compare search_pipes' docstring (mentions fuzzy threshold, sorting by score, return structure) with search_tables' (just "fuzzy match, 70% threshold"). 

Since LLMs use these docstrings for routing, please match the same level of detail: mention match_score field, sort order and response shape.

**3. Missing tool-layer test**

The PR only adds service-level tests (`tests/services/test_table_service.py`). The existing `search_pipes` tool has tests in `tests/tools/test_pipe_tools.py` that verify the tool is wired correctly end-to-end. 

Please add at least one happy-path test for `search_tables` in `tests/tools/test_table_tools.py`, this ensures the MCP registration, facade delegation, and response shape all work together.

MatheusFernandesDuarte:
Hey @adriannoes. Thanks for the detailed review!

All three points addressed:
1. **Substring check** — added, mirrors `search_pipes` exactly
2. **Docstring** — expanded with return shape, sort order and `match_score` detail
3. **Tool-layer tests** — added 3 tests to `test_table_tools.py` (no name, with name, response shape); also wired `search_tables` into the `mock_table_client` fixture

All 57 tests passing.

adriannoes:
LGTM, all three review items addressed. Substring match, enriched docstring, and tool-layer tests now align with the `search_pipes` pattern. 

Merging here, thanks!

PR #124, opened 2026-05-18
fix(sdk): skip malformed field definitions in create_form_model

adriannoes:
## Summary
- Introduce `MalformedFieldDefinitionError` and `ensure_valid_field_definitions()` in the SDK.
- `create_form_model` fails fast with a clear message when Pipefy returns incomplete field shapes (`id` / `type` missing, non-dict entries).
- MCP tools `create_card` and `fill_card_phase_fields` return `tool_error` on both elicitation and `skip_elicitation` paths.
- Fields without `label` still work (fallback to `field_id` as title).

Closes #30

Follow-up improvements tracked in a separate issue (field-filter gap, comment parsing, automation false success, etc.).

## Test plan
- [x] `uv run pytest packages/sdk/tests/models/test_form.py -v`
- [x] `uv run pytest packages/mcp/tests/tools/test_pipe_tool_helpers.py -k filter_fields`
- [x] `uv run pytest packages/mcp/tests/tools/test_pipe_tools.py -k malformed`
- [x] `uv run ruff check .`
- [x] CI green (lint + test)

### DIFF
--- a/packages/cli/src/pipefy_cli/skills/pipefy-introspection.md
+++ b/packages/cli/src/pipefy_cli/skills/pipefy-introspection.md
@@ -24,10 +24,10 @@ This is **Tier 2** in the resolution strategy: when a dedicated MCP tool fails o
 
 | Tool (MCP) | CLI | Read-only | Purpose |
 |------------|-----|-----------|---------|
-| `introspect_type` | `pipefy introspect type` | Yes | Discover a type's shape: `fields`, `inputFields`, `enumValues`. |
-| `introspect_query` | `pipefy introspect query` | Yes | Get a query's argument types and return shape. |
-| `introspect_mutation` | `pipefy introspect mutation` | Yes | Get a mutation's argument types and return shape. |
-| `search_schema` | `pipefy introspect schema search` | Yes | Keyword search across type / query / mutation names. |
+| `introspect_type` | `pipefy introspect type` | Yes | Type shape: `fields`, `inputFields`, `enumValues`. Optional `max_depth`. |
+| `introspect_query` | `pipefy introspect query` | Yes | Root query arguments and return type. Optional `max_depth`. |
+| `introspect_mutation` | `pipefy introspect mutation` | Yes | Root mutation arguments and return type. Optional `max_depth`. |
+| `search_schema` | `pipefy introspect schema search` | Yes | Keyword search on type names/descriptions. Optional `kind` filter. |
 | `execute_graphql` | `pipefy graphql exec` | No | Execute arbitrary GraphQL (`--yes` required for mutations). |
 | `get_organization` | `pipefy org get` | Yes | Load organization info (name, plan, UUID, member count, pipe count). |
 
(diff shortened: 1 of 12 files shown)


adriannoes:
Follow-up work (field-filter gap, comment parsing, automation false success, etc.) is tracked in #128 — intentionally out of scope for this PR.

**Merge blocked:** branch protection requires at least one approving review from someone with write access (cannot self-approve). CI is green.

gbrlcustodio:
**question (blocking):** What does the offending Pipefy response from #30 actually look like?

`field_definitions` is the raw GraphQL response from `GET_START_FORM_FIELDS_QUERY` / `GET_PHASE_FIELDS_QUERY`, surfaced unmodified by `PipeService.get_start_form_fields` / `get_phase_fields`. Given that:

- If `id` / `type` are non-null in Pipefy's schema, a missing value fails the whole query — `execute_query` raises `PipefyAPIError`, and no malformed dict reaches `create_form_model`.
- If they're nullable, they come back as `None`, not as missing keys. `field_def["id"]` returns `None`, not `KeyError`.

…so the "missing key" shape this PR defends against isn't reachable via normal GraphQL behavior. The actual cause of #30 is most likely one of:

1. Pipefy returned a `null` `id`/`type`.
2. A GraphQL error was in the response and `execute_query` didn't surface it.
3. Something upstream constructed `field_definitions` by hand.

Each points to a different layer; validating inside `create_form_model` treats the symptom rather than the cause.

**suggestion (non-blocking):** Once we know which scenario it is, a `FieldDefinition` Pydantic model in `pipe_service` (with non-`Optional` `id`/`type`) would replace both `ensure_valid_field_definitions` and the `.get()` defaults — validation lands at the SDK boundary where the data first enters Python, and downstream consumers get typed inputs for free.

adriannoes:
Thanks for the review — this clarifies what we should defend against.

**What #30 actually is:** There is no captured production GraphQL payload in the issue. It came from a Red Team pass on PR #28: a *hypothetical* list item missing the `id` key (see issue body). So the strict "missing key → KeyError" path is unlikely when data only flows through `execute_query` + non-null schema fields, as you noted.

**What we still want to handle (reachable without missing keys):**
1. **GraphQL `null` for `id` or `type`** — nullable fields or partial data; `field_def["id"]` does not KeyError, but `create_form_model` used to build a broken form. This commit validates at **`PipeService.get_start_form_fields` / `get_phase_fields`** via a `FieldDefinition` Pydantic model (`parse_field_definitions`).
2. **Defense in depth** — `create_form_model` / `_filter_fields_by_definitions` still call the same parser for mocks, hand-built lists, or future entry points.
3. **Non-dict list entries** — not expected from GraphQL lists of objects; kept as a cheap guard.

**Not in scope here:** GraphQL errors not surfaced by `execute_query` (scenario 2) — agree that belongs in the client layer; tracked for follow-up in #128.

Latest commit moves validation to the SDK boundary per your suggestion; MCP tools continue to map `MalformedFieldDefinitionError` → `tool_error`.

gbrlcustodio:
**issue (blocking):** `required` is `null: true` in the Pipefy schema (`field :required, Boolean, 'Whether the field is required', null: true`), so a real-world response can carry `required: null`. Pydantic v2 does **not** coerce `None` → `False` from the field default — `model_validate({"id": "...", "type": "...", "required": None})` raises `ValidationError`, which `parse_field_definitions` rolls up into `MalformedFieldDefinitionError`. Any pipe whose fields come back with a null `required` would now fail `get_start_form_fields` / `get_phase_fields` / `create_card` / `fill_card_phase_fields`.

**suggestion:** Match the pattern already used for the other nullable Booleans in `FieldDefinition` (`editable: bool | None = None`, `label: str | None = None`, etc.):

```python
required: bool | None = None
```

`model_dump(exclude_none=True)` drops the key when null, and `create_form_model`'s existing `field_def.get("required", False)` already returns `False` for absent keys — no custom validator needed.

Plus a regression test in `tests/models/test_field_definition.py`:

```python
def test_parse_field_definitions_accepts_null_required():
    parsed = parse_field_definitions(
        [{"id": "title", "type": "short_text", "required": None}],
        action="test",
    )
    assert "required" not in parsed[0]
```

gbrlcustodio:
**suggestion (non-blocking):** `ensure_valid_field_definitions` is a thin wrapper that calls `parse_field_definitions` and discards the parsed result. Its only caller — `_filter_fields_by_definitions` (`pipe_tool_helpers.py:433-444`) — then reads field IDs from the **unparsed** list:

```python
def _filter_fields_by_definitions(
    fields: dict[str, object] | None, field_definitions: list[dict]
) -> dict[str, object]:
    ensure_valid_field_definitions(field_definitions, action="filter field values")
    if not fields:
        return {}
    editable_ids = {field_def["id"] for field_def in field_definitions}
    ...
```

So we parse → discard → loop over raw data anyway. Two ways to tighten:

1. **Drop `ensure_valid_field_definitions`** and consume `parse_field_definitions`'s return value directly — single validation pass, parsed dicts everywhere.
2. **Drop the validation call here entirely** — `expected_fields` is already SDK-validated upstream at `PipeService.get_start_form_fields` / `get_phase_fields`, and the "future hand-built lists" justification is hypothetical.

Either is cleaner than the current shape; #2 is the smaller diff.

adriannoes:
@gbrlcustodio — all review feedback addressed :)

**Blocking (`required: null`):** `FieldDefinition.required` is now `bool | None = None`; `model_dump(exclude_none=True)` drops the key when null. Regression test added in `test_field_definition.py`.

**Non-blocking (`ensure_valid_field_definitions`):** removed the wrapper; validation stays at the SDK boundary only (`parse_field_definitions` in `PipeService`). `_filter_fields_by_definitions` no longer re-validates. `create_card` / `fill_card_phase_fields` catch `MalformedFieldDefinitionError` from `get_start_form_fields` / `get_phase_fields` and return `tool_error`.

Related tests passing; CI green. Ready for another look when you have a moment.

gbrlcustodio (approved):

PR #125, opened 2026-05-18
feat(cli): add `pipefy auth login` (PKCE-based user sign-in)

gbrlcustodio:
## Summary

- Adds `pipefy auth login`: OAuth 2.0 Authorization Code + PKCE (RFC 7636) over a loopback HTTP server (RFC 8252), with the resulting session persisted in the OS keychain via `keyring`.
- OIDC-neutral surface: one env var `PIPEFY_AUTH_URL` (issuer URL, e.g. `https://signin.pipefy.com/realms/pipefy`) + optional `PIPEFY_AUTH_CLIENT_ID` (default `pipefy-cli`). Endpoints come from `.well-known/openid-configuration` so the underlying IdP is not hardcoded.
- Validated end-to-end against the piporacle realm: login → token exchange → `{ me { email name } }` round-trip → keychain roundtrip via `storage.load_session()`.

## Scope

This is intentionally the first slice. It lands the **login + persist** path only. Downstream commands are **not** wired to read the keychain yet — that lands in a follow-up so this PR stays reviewable.

### In scope
- `pipefy auth login` Typer command (`commands/auth.py`)
- `oauth/` package: PKCE (S256), OIDC discovery, race-free `LoopbackCapture` (binds before opening the browser), token exchange, keychain `store_session`
- `_warn_if_masked` (`gh`-style notice when `PIPEFY_TOKEN` or a complete `PIPEFY_OAUTH_*` triple is set in the shell env)
- `--no-browser` flag for SSH / headless triage (prints the auth URL)
- `keyring>=24` added to CLI deps
- `resolve_cli_settings()` in `config.py` returns the full validated `CliSettings` so we don't construct it twice on `ctx.obj`
- Tests: PKCE invariants, OIDC discovery happy/error, loopback callback parsing + timeout + idempotent close, keychain roundtrip with in-memory backend, state-mismatch path, command-level happy/error/masking paths

### Out of scope (planned follow-ups)
- `get_authenticated_client` reading the keychain + silent refresh-on-401
- `auth status` / `auth whoami`
- `auth logout` (with `end_session_endpoint` revocation)
- OAuth 2.0 Device Authorization Grant for truly headless flows
- SSRF validation on `PIPEFY_AUTH_URL` via `pipefy_sdk.utils.url_ssrf` (flagged in review; defensible to land here, but kept out to bound this slice)
- Defense-in-depth: scrub `response.text` from token-exchange error messages (low risk; verifier shouldn't appear in IdP error bodies, but it's not guaranteed)
- Docs in `docs/cli/`

## Coexistence / precedence

Stacks below the existing client-credentials flow. Mental model: **explicit env > stored session**.

| Priority | Source |
|----------|--------|
| 1 | `--token` flag |
| 2 | `PIPEFY_TOKEN` (env or `.env`) |
| 3 | `PIPEFY_OAUTH_*` client-credentials (env or `.env`) |
| 4 | Stored user session (this PR) |

On `auth login`, a `gh`-style note prints to stderr when a higher-precedence source is set in the **shell** env, so users aren't confused why their session isn't used by subsequent commands.

## Commits

```
56a9e81 refactor(cli): make LoopbackCapture a context manager with explicit close()
76cd1ea refactor(cli): single CliSettings instance via resolve_cli_settings
af888a1 refactor(cli): clean up auth-login per /simplify review
27be35d feat(cli): add `auth login` for PKCE-based user sign-in
```

## Size note

15 files / ~775 production LOC + ~483 test LOC. Above AGENTS.md's ≤10-file / ≤300-LOC atomic-PR guideline. The 6 `oauth/` files are intentionally small and individually testable; collapsing them wouldn't reduce review surface. The `uv.lock` churn is from adding `keyring`. Happy to split into "lib + tests" vs "command + wiring" if reviewers prefer.

## Test plan

- [x] `uv run pytest -m "not integration"` — 2198 passed, 42 deselected
- [x] `uv run ruff check packages/cli` — clean
- [x] `uv run ruff format --check packages/cli` — clean
- [x] Manual smoke test against piporacle Keycloak using `bash`+`curl`+`openssl`+`jq` — discovery, code exchange, GraphQL `{ me { email name } }`, refresh rotation, logout/revoke all pass
- [x] `pipefy auth login` against piporacle end-to-end: browser flow → loopback capture → token exchange → keychain write → `storage.load_session()` roundtrips → stored access token authenticates against `piporacle.pipefy.com/graphql`
- [x] `_warn_if_masked` fires correctly when `PIPEFY_OAUTH_*` is set in the shell

## Known review items still open

From the in-branch `/review`, two security items defended as follow-ups:

1. SSRF validation on `PIPEFY_AUTH_URL` — would be a 4-line change inside `_auth_config_from_ctx` reusing `pipefy_sdk.utils.url_ssrf.validate_https_service_endpoint_url`. Punted to keep this PR bounded; happy to fold in if requested before merge.
2. Token-exchange error includes raw `response.text[:300]`. Verifier shouldn't appear in IdP error bodies per RFC 6749, but not guaranteed. Worth tightening to structured fields.


packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -0,0 +1,145 @@
...
+    if not auth_url:
+        typer.echo(
+            "PIPEFY_AUTH_URL is required for `pipefy auth login` (the OIDC issuer "
+            "URL for Pipefy authentication, e.g. "
+            "https://signin.pipefy.com/realms/pipefy). "
+            f"See {DOCS_SETUP_REF}.",
+            err=True,
+        )
+        raise typer.Exit(2)
+    return auth_url, obj["auth_client_id"]
```
adriannoes, line 46: **Follow-up (SSRF):** `auth_url` is passed straight to discovery without `validate_https_service_endpoint_url` (already used for GraphQL/OAuth in `PipefySettings`). Recommend validating in `_auth_config_from_ctx` or during `resolve_cli_settings`, reusing `pipefy_sdk.utils.url_ssrf`. ~4 lines; deferring is OK for this slice but should land before keychain-read is wired.

packages/cli/src/pipefy_cli/oauth/discovery.py
```diff
@@ -0,0 +1,86 @@
...
+    missing = [key for key in required if not data.get(key)]
+    if missing:
+        raise ValueError(
+            f"OIDC discovery at {url} is missing required fields: {', '.join(missing)}"
+        )
+
+    return ProviderMetadata(
+        issuer=str(data["issuer"]),
+        authorization_endpoint=str(data["authorization_endpoint"]),
+        token_endpoint=str(data["token_endpoint"]),
```
adriannoes, line 77: **Follow-up (SSRF):** Endpoints returned by discovery (`authorization_endpoint`, `token_endpoint`) are not validated. A malicious issuer could redirect token exchange to an internal host. Recommend running both through `validate_https_service_endpoint_url` after parse.

packages/cli/src/pipefy_cli/oauth/flow.py
```diff
@@ -0,0 +1,194 @@
...
+                "redirect_uri": redirect_uri,
+                "code_verifier": code_verifier,
+            },
+        )
+    except httpx.HTTPError as exc:
+        raise LoginError(f"Token exchange request failed: {exc}") from exc
+
+    if response.status_code != 200:
+        raise LoginError(
+            f"Token exchange failed ({response.status_code}): {response.text[:300]}"
```
adriannoes, line 82: **Follow-up (defense-in-depth):** Token exchange errors include `response.text[:300]`. Prefer parsing OAuth JSON fields (`error`, `error_description`) when present; avoid echoing raw body (low risk of verifier leakage, but not guaranteed by RFC 6749).

packages/cli/src/pipefy_cli/oauth/flow.py
```diff
@@ -0,0 +1,194 @@
...
+                metadata=metadata,
+                client_id=client_id,
+                redirect_uri=capture.redirect_uri,
+                code_challenge=challenge_from_verifier(verifier),
+                state=state,
+                scopes=scopes,
+            )
+            if on_url is not None:
+                on_url(auth_url)
+            open_browser(auth_url)
```
adriannoes: **Follow-up (UX):** `open_browser(auth_url)` return value is ignored. If `webbrowser.open` fails silently, the user waits until callback timeout. Consider checking the return value and printing a hint to use `--no-browser` / copy the URL manually.

packages/cli/src/pipefy_cli/oauth/flow.py
```diff
@@ -0,0 +1,194 @@
...
+            exchange so the same TLS connection can serve both requests.
+
+    Raises:
+        LoginError: For any user-visible failure (discovery, state mismatch,
+            token exchange).
+        TimeoutError: When no browser callback arrives in time.
+    """
+    with _http_client(http_client, timeout=_TOKEN_EXCHANGE_TIMEOUT_S) as http:
+        try:
+            metadata = fetch_provider_metadata(issuer_url, client=http)
```
adriannoes, line 141: **Follow-up (OIDC best practice):** No check that `metadata.issuer` matches the requested `issuer_url` (normalized). Worth adding before production use to catch misconfigured or hostile discovery documents.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -0,0 +1,145 @@
...
+            "On headless Linux, ensure a Secret Service daemon (gnome-keyring, "
+            "kwallet) is running, or use a static PIPEFY_TOKEN.",
+            err=True,
+        )
+        raise typer.Exit(1) from exc
+
+    typer.echo(
+        f"Signed in to Pipefy ({result.issuer}). Session stored in {keychain_backend_name()}."
+    )
+    _warn_if_masked()
```
adriannoes, line 155: **Follow-up (scope):** `_warn_if_masked` correctly reflects current behavior. When the keychain-read follow-up lands, revisit whether the warning should also mention `.env`-sourced tokens (today only `os.environ` is checked, which matches the documented precedence model).

packages/cli/tests/test_auth_login.py
```diff
@@ -0,0 +1,483 @@
...
+            raise flow.LoginError("Token exchange failed (400): invalid_grant")
+
+        from pipefy_cli.commands import auth as auth_module
+
+        monkeypatch.setattr(auth_module, "run_login", _boom)
+
+        result = cli_runner.invoke(cli_app, ["auth", "login"])
+        assert result.exit_code == 1
+        assert "Login failed" in result.stderr
+        assert "invalid_grant" in result.stderr
```
adriannoes, line 570: **Follow-up (tests):** Good command-level coverage. Missing a test for `--no-browser` asserting the authorization URL is printed and the browser stub is not invoked — golden help covers the flag but not behavior.

adriannoes (approved):
## Approved

Solid first slice: modular `oauth/` package, race-free `LoopbackCapture`, lazy `keyring` import, good unit-test coverage (PKCE, discovery, loopback, storage, state mismatch, command happy/error paths), and CI green.

The scope boundary (login + persist only; downstream commands not wired to keychain yet) is clear and reasonable for reviewability.

**Non-blocking follow-ups** (tracked for a later PR — inline notes below):

1. SSRF validation on `PIPEFY_AUTH_URL` via `pipefy_sdk.utils.url_ssrf.validate_https_service_endpoint_url`
2. SSRF validation on discovered `authorization_endpoint` / `token_endpoint`
3. Sanitize token-exchange error messages (prefer OAuth JSON `error` / `error_description` over raw `response.text`)
4. Wire keychain read + silent refresh-on-401 into `get_authenticated_client`
5. `auth status` / `auth whoami` / `auth logout`
6. Docs in `docs/cli/`
7. Unit test for `--no-browser`
8. Handle / surface `webbrowser.open()` returning `False`
9. OIDC issuer mismatch check (`issuer_url` vs `metadata.issuer`)
10. OIDC-neutral wording in discovery error messages

Happy to merge as-is; the security items should land before the keychain-read follow-up goes to production users.

packages/cli/tests/test_auth_login.py
gbrlcustodio: Addressed in dc6a20e — `test_no_browser_prints_url_and_skips_webbrowser` covers URL on stdout, `open_browser` returns `False`, and `webbrowser.open` is never invoked.

packages/cli/src/pipefy_cli/oauth/flow.py
gbrlcustodio: Addressed in e827335 — `run_login` now invokes `on_url` as a fallback when `open_browser` returns `False`. The auth command always wires `on_url` (not just under `--no-browser`) and prints a "Could not open a browser automatically. Open the URL below, or re-run with --no-browser." hint to stderr. New `test_browser_open_failure_falls_back_to_printing_url` covers it.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: Addressed in 2ef510c — `CliSettings` now runs `validate_https_service_endpoint_url` on `auth_url` via a `@model_validator(mode="after")`, honoring `pipefy.allow_insecure_urls`. Four new tests in `test_config.py` cover https accepted / http rejected / localhost-with-insecure-flag / private-IP rejected.

packages/cli/src/pipefy_cli/oauth/discovery.py
gbrlcustodio: **deferred:** discovery-returned endpoint validation is the second SSRF guard. Keeping it out of this PR to stay bounded; will land in the follow-up that also adds the `metadata.issuer == requested_url` check (your separate thread on `flow.py:137`). Both go together — they defend the same threat (hostile discovery doc).

packages/cli/src/pipefy_cli/oauth/flow.py
gbrlcustodio: **deferred:** tracked alongside the discovery-endpoint validation in #126. Both checks defend the same threat layer (hostile / misconfigured discovery doc) — landing them together gives a complete fix rather than half-defense.

packages/cli/src/pipefy_cli/oauth/flow.py
gbrlcustodio: **deferred:** tracked in #126. Will parse `error` / `error_description` from the JSON body and drop the raw `response.text[:300]` fallback so the verifier (or any submitted parameter) can never reach a bug report.

packages/cli/src/pipefy_cli/oauth/discovery.py
gbrlcustodio: Tracked in #126 (paired with the issuer-claim check on `flow.py:137`).

Issue #126, opened 2026-05-19
auth login: deferred hardening from PR #125 review


gbrlcustodio:
Three hardening items surfaced during the [PR #125](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/125) review and explicitly deferred to keep the initial slice (login + keychain persist) bounded. Tracking here so they don't get lost.

## 1. Validate OIDC discovery-returned endpoints

**Threat:** A hostile discovery response can return arbitrary ``authorization_endpoint`` / ``token_endpoint`` URLs, redirecting the browser dance or the code-for-token exchange to attacker-controlled hosts.

**Fix:** After ``fetch_provider_metadata`` returns, run ``metadata.authorization_endpoint`` and ``metadata.token_endpoint`` through ``pipefy_sdk.utils.url_ssrf.validate_https_service_endpoint_url`` (honoring ``allow_insecure_urls``).

**Files:** ``packages/cli/src/pipefy_cli/oauth/discovery.py``

**Reference:** [review comment r3262731297](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/125#discussion_r3262731297)

## 2. Verify ``metadata.issuer`` matches the requested issuer URL

**Threat:** OIDC Discovery 1.0 requires the ``issuer`` claim to match the URL the discovery doc was fetched from. Without the check we silently accept a discovery doc that claims a different issuer — useful for catching misconfigured realms (e.g. piporacle vs prod) and one layer of defense against tampered responses.

**Fix:** After ``fetch_provider_metadata``, compare ``normalize(metadata.issuer)`` to ``normalize(issuer_url)`` (rstrip ``/``); raise ``LoginError`` on mismatch.

**Files:** ``packages/cli/src/pipefy_cli/oauth/flow.py`` or ``oauth/discovery.py``

**Reference:** [review comment r3262731306](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/125#discussion_r3262731306)

**Note:** items 1 and 2 defend the same threat layer (hostile / misconfigured discovery doc) and should land together — each on its own is half-defense.

## 3. Scrub raw ``response.text`` from token-exchange errors

[...]

PR #131, opened 2026-05-19
feat(cli): wire stored auth-login session into `get_authenticated_client`

gbrlcustodio:
Closes the loop opened by #125: ``pipefy auth login`` writes a session to the keychain, but until this PR nothing downstream read it. After this lands, the four-tier credential precedence chain becomes fully operative:

| # | Source | Status before | Status after |
|---|---|---|---|
| 1 | ``--token`` flag | ✓ | unchanged |
| 2 | ``PIPEFY_TOKEN`` env | ✓ | unchanged |
| 3 | ``PIPEFY_OAUTH_*`` client-creds triple | ✓ | unchanged |
| 4 | Stored user session | ✗ dead | **operative** |

## Eager refresh

Before each ``pipefy <cmd>`` invocation, ``ensure_fresh_session`` checks the stored access token's age. If ≤ 60 s of leeway remains, the CLI POSTs ``grant_type=refresh_token`` to the issuer's token endpoint, persists rotated tokens back to the keychain, and builds the ``PipefyClient`` with the new access token. The existing cache key already includes ``bearer_token`` — refresh-induced rotation invalidates correctly with no key-shape change.

Reactive refresh-on-401 (via a custom httpx auth hook) is **deferred** to a separate slice. Eager refresh covers the common case (one-shot CLI invocations across the lifetime of a refresh token) and stays out of the SDK transport entirely.

## What's new

- ``oauth/refresh.py`` — ``RefreshError``, ``refresh_access_token``, ``ensure_fresh_session``.
- ``oauth/_http.py`` — shared httpx-client context manager extracted from ``flow.py`` so ``flow.py`` and ``refresh.py`` reuse the same borrow-or-create helper.
- ``oauth/__init__.py`` — re-exports for the new symbols.
- ``auth.py::get_authenticated_client`` — extended signature with ``auth_url`` / ``auth_client_id`` kwargs. New priority-4 branch runs ``ensure_fresh_session`` *before* the cache lookup so refresh-rotated tokens produce the right cache signature. Updated "no auth source" error mentions ``pipefy auth login``.
- ``commands/_common.py`` — new frozen dataclass ``AuthContext`` + helper ``auth_context_from_ctx(ctx)`` that threads the new ``auth_url`` / ``auth_client_id`` ctx values to the auth boundary. The 3 internal call sites (``run_pipefy_client_coroutine``, ``run_cli_command``, ``authenticated_client_from_ctx``) use it; ``settings_and_token`` stays unchanged for its 9 external command callers.
- ``tests/conftest.py`` — promoted ``InMemoryKeyring`` + ``fake_keyring`` from ``test_auth_login.py`` for cross-file reuse.

## Out of scope (tracked elsewhere or coming separately)

- Reactive refresh-on-401 (custom httpx auth hook).
- ``pipefy auth status`` / ``pipefy auth logout``.
- ``InternalApiClient`` over user-auth (AI-automation features remain unavailable on the bearer path — same limitation as ``PIPEFY_TOKEN`` today; lifting requires SDK changes).
- Distinguishing "keychain unreachable" from "no session" in the error message (both currently surface as ``None`` from ``load_session``).

## Test plan

- [x] ``test_auth_refresh.py`` — 12 new cases: ``ensure_fresh_session`` for no-session / fresh / refresh / rotation paths; ``RefreshError`` on ``invalid_grant`` / network error / discovery failure / non-object body; refresh failure preserves the stored session for retry.
- [x] ``test_auth.py`` — 5 new precedence cases: bearer wins over session, OAuth-triple wins over session, session used when no other source, cache invalidates when access token rotates, ``RefreshError`` → ``typer.Exit(2)`` with relogin hint.
- [x] ``test_common_helpers.py`` — 4 existing mocks updated to the new ``auth_context_from_ctx`` boundary.
- [x] ``uv run pytest -m "not integration"`` — **2222 passed**, 42 deselected.
- [x] ``uv run ruff check packages/cli`` — clean.
- [x] ``uv run ruff format --check packages/cli`` — clean.

### Manual E2E against piporacle (planned post-review)

```
unset PIPEFY_TOKEN PIPEFY_OAUTH_URL PIPEFY_OAUTH_CLIENT PIPEFY_OAUTH_SECRET
export PIPEFY_AUTH_URL=https://signin-piporacle.pipefy.com/realms/st-piporacle-pud1m
export PIPEFY_GRAPHQL_URL=https://piporacle.pipefy.com/graphql
uv run pipefy auth login         # browser flow → keychain
uv run pipefy pipe list           # should succeed via keychain session
PIPEFY_TOKEN=garbage uv run pipefy pipe list  # should 401 (proves --token wins)
```

## Size note

11 files / ~782 insertions / ~96 deletions. Above the AGENTS.md ≤300-line / ≤10-file guideline. Bulk is test scaffolding (~315 LOC in ``test_auth_refresh.py`` alone). Production code is ~295 LOC across 6 files. Happy to split into "refresh module + its tests" vs "auth.py wiring + its tests" if reviewers prefer.

### DIFF
--- a/packages/cli/src/pipefy_cli/auth.py
+++ b/packages/cli/src/pipefy_cli/auth.py
@@ -1,4 +1,15 @@
-"""Build an authenticated :class:`PipefyClient` from CLI configuration (OAuth or ``--token``)."""
+"""Build an authenticated :class:`PipefyClient` from CLI configuration.
+
+The precedence chain — most explicit wins:
+
+1. ``bearer_token`` arg (CLI ``--token`` flag or ``PIPEFY_TOKEN`` env).
+2. ``PIPEFY_OAUTH_*`` triple → client-credentials grant (service account).
+3. Stored user session from ``pipefy auth login`` (keychain) — refreshed
+   eagerly when the access token is within the leeway window.
+
+The cache key includes whichever bearer ultimately reaches the SDK, so a
+refresh-rotated access token naturally invalidates the cached client.
+"""
 
 from __future__ import annotations
 
@@ -15,6 +26,7 @@ from pipefy_cli.config import (
     describe_missing_oauth_vars,
     ensure_public_graphql_configured,
 )
+from pipefy_cli.oauth import RefreshError, ensure_fresh_session
 
 _cached_signature: tuple[object, ...] | None = None
 # One-shot CLIs reuse this; long-lived programmatic use should call
@@ -44,26 +56,39 @@ def _cache_key(
     )
 
 
+def _missing_auth_message(pipefy_settings: PipefySettings) -> str:
+    missing = describe_missing_oauth_vars(pipefy_settings)
+    return (
+        "Missing authentication. Use --token, set PIPEFY_TOKEN, configure "
+        f"PIPEFY_OAUTH_* ({missing}), or run `pipefy auth login`. "
+        f"See {DOCS_SETUP_REF}."
+    )
+
+
 def get_authenticated_client(
     pipefy_settings: PipefySettings,
     *,
     bearer_token: str | None = None,
+    auth_url: str | None = None,
+    auth_client_id: str | None = None,
 ) -> PipefyClient:
-    """Return a facade client using OAuth client-credentials or a static bearer token.
-
-    Uses the same OAuth wiring as ``pipefy-mcp-server`` via :class:`PipefyClient`
-    (``httpx_auth.OAuth2ClientCredentials`` when no bearer is supplied).
+    """Return a facade client using the highest-precedence available auth source.
 
     Args:
         pipefy_settings: Validated Pipefy endpoint settings (``PIPEFY_*``).
-        bearer_token: When set, skips OAuth and uses this bearer for GraphQL transports.
+        bearer_token: When set (``--token`` / ``PIPEFY_TOKEN``), skips OAuth and
+            uses this bearer directly. Highest precedence.
+        auth_url: Issuer URL for the stored user session lookup. From
+            ``PIPEFY_AUTH_URL``.
+        auth_client_id: Public client id for the stored session lookup.
 
     Returns:
-        Shared in-process instance when settings match a prior call; otherwise a new
-        client. Tokens are not written to disk.
+        Shared in-process instance when settings match a prior call; otherwise
+        a new client.
 
     Raises:
-        typer.Exit: Code 2 when configuration is incomplete for the chosen auth mode.
+        typer.Exit: Code 2 when no auth source resolves, or when a stored
+            session exists but refresh failed.
     """
     global _cached_signature, _cached_client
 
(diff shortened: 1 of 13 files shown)


packages/cli/src/pipefy_cli/oauth/discovery.py
```diff
@@ -6,6 +6,8 @@
 
 import httpx
 
+from pipefy_cli.oauth._http import http_client
```
Danielmoraisg: **nitpick:** On `flow.py` and on `refresh.py` we are importing it as _http_client, if the intent is for this to be a private module we should calll this _http_client. Or use the "public"  name of the helper inside the package.

packages/cli/src/pipefy_cli/auth.py
```diff
@@ -44,26 +60,39 @@ def _cache_key(
...
+        f"See {DOCS_SETUP_REF}."
+    )
+
+
 def get_authenticated_client(
     pipefy_settings: PipefySettings,
     *,
     bearer_token: str | None = None,
+    auth_url: str | None = None,
+    auth_client_id: str | None = None,
```
Danielmoraisg: **suggestion:** This might be too big of a refactor, but as far as i understood, we are creating the dataclass `AuthContext` only to immediatelly unpack it into each of the arguments of `get_authenticated_client` , I believe it would be better to just pass the dataclass directly to `get_authenticated_client`

packages/cli/src/pipefy_cli/commands/_common.py
```diff
@@ -17,6 +18,17 @@
...
 
+
+@dataclass(frozen=True)
+class AuthContext:
+    """All four credential-resolution inputs the auth boundary needs."""
+
+    settings: PipefySettings
+    bearer_token: str | None
+    auth_url: str | None
+    auth_client_id: str | None
```
Danielmoraisg: **suggestion:** since we can either have bearer_token or auth_url + client_id , we could have a separate dataclass for the auth attributes like this

@dataclass(frozen=True)
class AuthAttributes #or some other better name
  auth_url: str
  client_id: str

and then use the class in the `AuthContext`

packages/cli/src/pipefy_cli/oauth/_http.py
```diff
@@ -0,0 +1,24 @@
...
+@contextmanager
+def http_client(
+    provided: httpx.Client | None, *, timeout: float
+) -> Iterator[httpx.Client]:
+    """Yield ``provided`` (without closing it) or a fresh ``httpx.Client``."""
+    if provided is not None:
+        with nullcontext(provided) as client:
+            yield client
+        return
+    with httpx.Client(timeout=timeout) as client:
```
Danielmoraisg, line 20: **suggestion:** To make this a bit more robust we could retry in case the request fails on the server side. 

using tenacity package, it would be a decorator like this.

`@retry(retry=retry_if_exception_type(httpx.RequestError))`

Danielmoraisg (commented):
Great work my friend!

packages/cli/src/pipefy_cli/auth.py
gbrlcustodio: I was just working on this refactor, and it’s great to see that we’re on the same page. Thanks a lot, my padawan, for your review.

packages/cli/src/pipefy_cli/oauth/_http.py
gbrlcustodio: That’s something we had in mind, and it will be included in a separate PR that also handles refreshing tokens on 401.

packages/cli/src/pipefy_cli/commands/_common.py
gbrlcustodio: Done in a02fa7a — went a bit further than the original suggestion, lmk what you think:

- The nested dataclass is named `OidcClient` (issuer_url + client_id) to mirror the existing `ProviderMetadata` (the provider-side counterpart) and to use the term the codebase already uses internally ("OIDC issuer" appears across docstrings, comments, and `docs/cli/auth.md` in PR #141).
- Field on `AuthContext` is `oidc_client: OidcClient | None` — its presence/absence is exactly the "tier 4 configured" bit, replacing the old `auth_url`/`auth_client_id` independent fields.
- Took the chance to also drop `AuthContext.settings: PipefySettings` — the auth context now carries identity only (`bearer_token` + `oidc_client`). Settings stays a separate argument to `get_authenticated_client(pipefy_settings, auth)`. Cleaner separation between "where to call" and "who I am", and it preps the surface for #126 (which threads `allow_insecure_urls` to the OAuth layer via `pipefy_settings`, not via the auth context).
- New `settings_and_auth_from_ctx` helper in `_common.py` replaces the now-unused `auth_context_from_ctx`. `auth_login` uses it too, inlining the "PIPEFY_AUTH_URL is required" guard.

packages/cli/src/pipefy_cli/oauth/discovery.py
gbrlcustodio: Done in 3047014. Went with module-qualified imports — `from pipefy_cli.oauth import _http` and `_http.http_client(...)` in all three files (`discovery.py`, `flow.py`, `refresh.py`). The `_http` module's leading underscore carries the "package-private" signal, so the function name itself doesn't need it; this also drops the alias that flow/refresh had to dodge the name collision with their `http_client` parameter.

PR #142, opened 2026-05-20
feat(cli): pipefy auth status (#135)

### LINKED ISSUE #135: Add `pipefy auth status` / `whoami`
## Motivation

PR #125 added ``pipefy auth login`` and PR #131 made the stored session operative across all commands. There's still no way for a user to answer the obvious diagnostic questions:

- Am I signed in?
- Which Pipefy user am I authenticated as?
- Which auth source is the CLI actually using right now (``--token``, ``PIPEFY_TOKEN``, ``PIPEFY_OAUTH_*``, or my stored session)?
- When does my session expire?

This is the ``gh auth status`` / ``aws sts get-caller-identity`` analogue. Without it, users running into auth problems have no first step beyond re-logging in.

## Proposed command surface

```
pipefy auth status [--json]
```

Output (text mode):
```
Signed in to Pipefy (https://signin.pipefy.com/realms/pipefy)
  Identity:      gabriel.custodio@pipefy.com (Gabriel)
  Auth source:   stored session (`pipefy auth login`)
  Expires:       in 4h 12m
  Refresh token: valid (offline)
  Keychain:      macOS Keychain (Keyring)

Other auth sources detected in env:
  PIPEFY_OAUTH_* (would mask the stored session if no --token is passed)
```

Output (no session, no other auth):
```
Not signed in. Run `pipefy auth login`, set PIPEFY_TOKEN, or configure PIPEFY_OAUTH_*.
```

A ``--json`` flag emits the same data as JSON for scripting.

## Implementation notes

- Reuse ``oauth.ensure_fresh_session`` to load the stored session (already handles the refresh case).
- Reuse ``commands.auth._session_masking_env_vars`` to list "would-mask" env vars.
- Identity fetch: GraphQL ``{ me { email name } }`` via ``PipefyClient`` — same query the smoke test in PR #125 used, same auth path as any other command.
- Skip the GraphQL call when there's no auth source at all (print "Not signed in" and exit non-zero).

## Out of scope

- Multiple-session / profile support (single session per ``(issuer, client_id)`` is still the contract — see PR #125).
- ``auth list`` / ``auth show <name>`` for multiple stored sessions.

## Acceptance criteria

- [ ] ``pipefy auth status`` prints identity + auth source + expiry when signed in.
- [ ] ``pipefy auth status`` exits non-zero with a clear "not signed in" message when no source is configured.
- [ ] ``--json`` emits a stable schema (fields: ``signed_in``, ``identity``, ``auth_source``, ``issuer``, ``expires_at``, ``keychain_backend``, ``masking_env_vars``).
- [ ] Existing precedence is reflected — if ``--token`` is passed, status reports the static-bearer source, not the stored session.
- [ ] Tests cover: signed-in via session, signed-in via ``PIPEFY_TOKEN``, signed-in via ``PIPEFY_OAUTH_*``, not signed in.
- [ ] Documentation in ``docs/cli/auth.md`` lists the command (tracked separately if not yet present).

## References

- PR #125 (adds ``pipefy auth login``)
- PR #131 (makes the stored session operative)
- ``packages/cli/src/pipefy_cli/commands/auth.py``

gbrlcustodio:
## Summary

- New `pipefy auth status [--json|-j]` — reports which auth source is active, the authenticated identity, and (for stored sessions) access/refresh expiry. Locked JSON schema with `signed_in`, `identity`, `auth_source`, `detected_sources`, `issuer`, `state`, `access_expires_at`, `refresh_expires_at`, `token_rejected`, `keychain_backend`, `masking_env_vars`.
- Exit codes: **2** for no credentials / unrefreshable session, **1** for upstream rejection (401) or transport error, **0** for signed-in.
- SDK: typed `client.get_me() -> MePayload | None` mirroring `get_organization` (new `queries/me_queries.py`, `services/user_service.py`, `MePayload` TypedDict). `me` and `User.name` are both schema-nullable (verified via live introspection); types and renderers respect that.
- Refactor: `AuthContext.bearer_token` promoted to a `BearerToken(value, source)` dataclass so status can distinguish `--token` from `PIPEFY_TOKEN` without re-inspecting argv/env. New `detect_all_sources` / `detect_auth_source` helpers; `get_authenticated_client` now dispatches off the detection helper so runtime and status share one precedence definition.

Closes #135.

## Test plan

- [x] SDK unit tests (`packages/sdk/tests/services/pipefy/test_user_service.py`) — 3 cases (populated identity, `me: null`, `name: null`).
- [x] CLI command tests (`packages/cli/tests/test_auth_status.py`) — 16 cases covering all four sources, refresh-expired, 401, 5xx (distinct from 401), `me: null`, `name: null`, masked sources.
- [x] Golden CLI help refreshed (`tests/fixtures/cli_help_golden.txt`).
- [x] Full regression: SDK 729 passed; CLI 227 passed + 10 skipped (live-creds integration).
- [x] Live smoke: `pipefy auth status --json` against the service-account session returned the expected schema with the real synthetic identity.
- [x] Live smoke on a stored-session bearer (requires `pipefy auth login` first — reviewer's call whether to re-verify).

### DIFF
--- a/docs/cli/auth.md
+++ b/docs/cli/auth.md
@@ -107,7 +107,7 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 | Command | Status | Notes |
 |---------|--------|-------|
 | `pipefy auth login` | Available | Browser-based PKCE login; persists the session in the OS keychain. |
-| `pipefy auth status` | Forthcoming (#135) | Print which auth source is active, identity, and session expiry. |
+| `pipefy auth status` | Available | Print which auth source is active, identity, and session expiry. |
 | `pipefy auth logout` | Forthcoming (#136) | Delete the stored session and revoke the refresh token. |
 
 #### `pipefy auth login` flags
@@ -116,6 +116,43 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 |------|---------|--------|
 | `--no-browser` | _off_ | Print the authorization URL to stdout instead of trying to launch a browser. |
 | `--callback-timeout <s>` | `180.0` | Seconds to wait for the browser callback (minimum 5). |
+
+#### `pipefy auth status` flags
+
+| Flag | Default | Effect |
+|------|---------|--------|
+| `--json` / `-j` | _off_ | Emit a stable JSON schema instead of human-readable text. |
+
+The command answers four diagnostic questions: am I signed in, as whom, via which precedence tier, and (for stored sessions) when does the access/refresh token expire. It also reports which *other* credential sources are configured, so a CI failure where `PIPEFY_OAUTH_*` masks a stored login is one command away from being obvious.
+
+##### JSON schema
+
+```jsonc
+{
+  "signed_in": true,
+  "identity": {"email": "user@pipefy.com", "name": "Pipefy User"},
+  "auth_source": "stored-session",                // or "flag-token" | "env-token" | "service-account" | "none"
+  "detected_sources": ["stored-session"],         // every configured source (winner + masked)
+  "issuer": "https://signin.pipefy.com/realms/pipefy",
+  "state": "active",                              // or "refresh-expired" | "needs-login" | "n/a"
+  "access_expires_at": "2026-05-20T22:14:03Z",    // ISO 8601, null for static-bearer
+  "refresh_expires_at": "2026-06-19T18:02:00Z",   // ISO 8601, null when not stored-session
+  "token_rejected": false,                        // true only when the identity `me` query returned 401
+  "keychain_backend": "Keyring",                  // null for non-stored-session sources
+  "masking_env_vars": []                          // env vars masking a stored session, if any
+}
+```
+
+The shape is stable across all sources (fields you don't have are `null` rather than absent), so a script can `jq .auth_source` without branching on which tier is active.
+
+##### Exit codes
+
+| Case | Exit |
+|------|------|
+| `auth_source == "none"` | **2** — same code as a domain command that fails on missing credentials. |
+| Stored session present but refresh grant rejected (`state` ∈ {`refresh-expired`, `needs-login`}) | **2** |
+| Signed in, but the identity `me` query returned 401 or transport-failed | **1** |
+| Signed in, identity fetched successfully | **0** |
 
 ### Pipefy issuer URLs
 
--- a/packages/cli/src/pipefy_cli/auth.py
+++ b/packages/cli/src/pipefy_cli/auth.py
@@ -34,7 +34,6 @@ from pipefy_cli.config import (
 )
 from pipefy_cli.oauth import RefreshError, ensure_fresh_session, load_session
 
-
 AuthSource = Literal[
     "flag-token",
     "env-token",
@@ -126,10 +125,14 @@ def detect_all_sources(
         )
     if not describe_missing_oauth_vars(pipefy_settings):
         sources.append("service-account")
-    if auth.oidc_client is not None and load_session(
-        issuer=auth.oidc_client.issuer_url,
-        client_id=auth.oidc_client.client_id,
-    ) is not None:
+    if (
+        auth.oidc_client is not None
+        and load_session(
+            issuer=auth.oidc_client.issuer_url,
+            client_id=auth.oidc_client.client_id,
+        )
+        is not None
+    ):
         sources.append("stored-session")
     return sources
 
--- a/packages/cli/src/pipefy_cli/auth.py
+++ b/packages/cli/src/pipefy_cli/auth.py
@@ -17,6 +17,7 @@ SDK, so a refresh-rotated access token naturally invalidates the cached client.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from typing import Literal
 
 import typer
 from pipefy_sdk import (
@@ -31,7 +32,16 @@ from pipefy_cli.config import (
     describe_missing_oauth_vars,
     ensure_public_graphql_configured,
 )
-from pipefy_cli.oauth import RefreshError, ensure_fresh_session
+from pipefy_cli.oauth import RefreshError, ensure_fresh_session, load_session
+
+
+AuthSource = Literal[
+    "flag-token",
+    "env-token",
+    "service-account",
+    "stored-session",
+    "none",
+]
 
 
 @dataclass(frozen=True)
@@ -47,6 +57,18 @@ class OidcClient:
 
 
 @dataclass(frozen=True)
+class BearerToken:
+    """Static bearer token plus the surface that produced it (``--token`` or env).
+
+    Carrying the origin lets diagnostics (e.g. ``pipefy auth status``) report which
+    precedence tier won without having to re-inspect ``argv``/``os.environ``.
+    """
+
+    value: str
+    source: Literal["flag", "env"]
+
+
+@dataclass(frozen=True)
 class AuthContext:
     """User-auth identity for a single CLI invocation.
 
@@ -55,7 +77,7 @@ class AuthContext:
     passed alongside.
     """
 
-    bearer_token: str | None
+    bearer_token: BearerToken | None
     oidc_client: OidcClient | None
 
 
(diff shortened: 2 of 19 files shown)


packages/cli/src/pipefy_cli/auth.py
```diff
@@ -126,12 +185,20 @@ def get_authenticated_client(
...
+        typer.echo(_missing_auth_message(pipefy_settings), err=True)
+        raise typer.Exit(2)
 
     # Resolve the effective bearer BEFORE the cache lookup so a refresh-rotated
     # access token produces the right (new) cache signature.
-    effective_bearer = auth.bearer_token
-    if not effective_bearer and missing_oauth and auth.oidc_client is not None:
+    effective_bearer: str | None = None
+    if source in ("flag-token", "env-token"):
+        assert auth.bearer_token is not None  # narrowed by detect_auth_source
```
adriannoes: **Context:** `get_authenticated_client` narrows `auth.bearer_token` / `auth.oidc_client` after `detect_auth_source`.

**Issue:** Uses `assert` for type narrowing.

**Why:** Assertions are not a stable CLI error path and are removed under `python -O`.

**Suggestion:** Replace with explicit guards that emit the same missing-auth message and `typer.Exit(2)`.

**Done when:** Narrowing uses `if` / `raise typer.Exit(2)` only (no `assert` on this path).

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -148,4 +197,221 @@ def _open(url: str) -> bool:
...
+    report.issuer = oidc.issuer_url
+    report.keychain_backend = keychain_backend_name()
+    report.masking_env_vars = _session_masking_env_vars()
+    try:
+        fresh_session = ensure_fresh_session(
+            issuer=oidc.issuer_url, client_id=oidc.client_id
+        )
+    except RefreshError as exc:
+        report.state = (
+            "refresh-expired" if "invalid_grant" in str(exc) else "needs-login"
```
adriannoes: **Context:** Stored-session refresh failures set `state` to `refresh-expired` or `needs-login`.

**Issue:** The branch uses `"invalid_grant" in str(exc)`.

**Why:** Exit code 2 and `state` drive automation; a changed error string from the token endpoint would silently flip categories.

**Suggestion:** Add a structured field on `RefreshError` (or parse the OAuth `error` code from the token response) and branch on that instead of a substring.

**Done when:** Tests cover both `invalid_grant` and another refresh failure without relying on substring matching.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -148,4 +197,221 @@ def _open(url: str) -> bool:
...
+    """Print which auth source is active, the authenticated identity, and session expiry."""
+    settings, auth = settings_and_auth_from_ctx(ctx)
+    detected = detect_all_sources(settings, auth)
+    source: AuthSource = detected[0] if detected else "none"
+    report = AuthStatusReport(auth_source=source, detected_sources=detected)
+
+    try:
+        if source == "none":
+            raise _StatusExit(report=report, exit_code=2)
+        if source == "stored-session":
```
adriannoes, line 410: **Context:** `auth_status` only calls `_populate_stored_session` when `auth_source == "stored-session"`, and that helper sets `masking_env_vars`.

**Issue:** When `env-token` or `service-account` wins but a stored session is still in `detected_sources`, `masking_env_vars` stays `[]` and the human “Other auth sources detected in env” block never appears.

**Why:** The docs and locked JSON schema describe `masking_env_vars` for diagnosing env vars that mask a keychain login — a common CI failure called out in the PR. Relying only on `detected_sources` forces scripts to infer masking logic themselves.

**Suggestion:** Set `report.masking_env_vars = _session_masking_env_vars()` whenever `"stored-session" in detected` (not only when it wins). Add a test for `test_status_service_account_wins_over_stored_session` asserting non-empty `masking_env_vars`.

**Done when:** `--json` reports masking env vars when a stored session exists but a higher-precedence source is active.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -148,4 +197,221 @@ def _open(url: str) -> bool:
...
+    stderr: str | None = None
+
+
+def _populate_stored_session(report: AuthStatusReport, oidc: OidcClient) -> None:
+    """Populate stored-session fields on ``report``; raise ``_StatusExit`` on failure."""
+    report.issuer = oidc.issuer_url
+    report.keychain_backend = keychain_backend_name()
+    report.masking_env_vars = _session_masking_env_vars()
+    try:
+        fresh_session = ensure_fresh_session(
```
adriannoes, line 323: **Context:** `auth status` with a stored-session winner calls `ensure_fresh_session` in `_populate_stored_session`, then `_fetch_identity` → `get_authenticated_client` may refresh again.

**Issue:** Two refresh attempts per invocation on the stored-session path.

**Why:** Extra latency and token-endpoint load on a diagnostic command that users may run frequently in CI.

**Suggestion:** Optional follow-up — reuse the refreshed session for the `get_me` client (or skip the second refresh when the token is already fresh).

**Done when:** At most one `ensure_fresh_session` call per `auth status` run (observable via mock call counts in tests).

docs/cli/auth.md
```diff
@@ -107,7 +107,7 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 | Command | Status | Notes |
 |---------|--------|-------|
 | `pipefy auth login` | Available | Browser-based PKCE login; persists the session in the OS keychain. |
-| `pipefy auth status` | Forthcoming (#135) | Print which auth source is active, identity, and session expiry. |
+| `pipefy auth status` | Available | Print which auth source is active, identity, and session expiry. |
```
adriannoes, line 110: **Context:** This PR adds a new CLI subcommand documented here.

**Issue:** `CHANGELOG.md` `[Unreleased]` has no entry for `pipefy auth status`.

**Why:** We version the monorepo in lockstep and use the changelog for beta/release notes; skipping it makes the next release notes incomplete.

**Suggestion:** Add under `CHANGELOG.md` → `### Added`:
`- **CLI**: \`pipefy auth status [--json|-j]\` — reports auth source, identity, session expiry, and exit codes.`

**Done when:** A changelog bullet exists under `[Unreleased]`.

adriannoes (approved):
## Summary

Nice work — this is a thoughtful `gh auth status`-style diagnostic with a clear locked JSON contract, shared precedence via `detect_auth_source`, and solid test coverage (16 CLI + 3 SDK cases). CI is green. **Approved** with the follow-ups noted inline; none are merge blockers.

## Follow-ups (inline threads)

| ID | Topic | Why it matters |
|----|--------|----------------|
| **F1** | `masking_env_vars` only when stored-session wins | Docs/schema promise env vars that mask a keychain session; when `service-account`/`env-token` wins, `detected_sources` lists both but `masking_env_vars` stays `[]`. |
| **F2** | `CHANGELOG.md` entry | Release-notes hygiene for a user-visible CLI command. |
| **F3** | `"invalid_grant" in str(exc)` | Fragile `state` classification if OIDC error text changes. |
| **F4** | `assert` narrowing in `get_authenticated_client` | Prefer explicit `typer.Exit(2)` guards over `assert` on CLI paths. |
| **F5** | Double `ensure_fresh_session` | Optional perf polish on the stored-session path. |

## Using this review with AI

Full review record (IDs + done-when): `.cursor/dev-planning/code-review/pr-142-cli-auth-status.md` in the reviewer workspace — each thread is structured for agent fixes.

## Verification

```bash
uv run pytest packages/cli/tests/test_auth_status.py packages/cli/tests/test_auth.py -q
uv run pytest packages/sdk/tests/services/pipefy/test_user_service.py -q
uv run ruff check packages/cli packages/sdk
```

Manual smoke: `pipefy auth status --json` with stored session + `PIPEFY_OAUTH_*` set (validates F1 scenario).

packages/cli/src/pipefy_cli/auth.py
gbrlcustodio: **note:** Addressed in 608c5f3 with one tweak — guards emit an *internal-error* message ("`<source>` detected but no bearer/OIDC client configured. Please file an issue."), not `_missing_auth_message`. Reaching the branch means our classifier and the AuthContext disagree, which is internal state inconsistency, not a user-misconfiguration. Same exit code 2. Workspace-wide `S101` lint enforcement (preventing reintroduction) tracked in #147.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **note:** Addressed in b4a3fec. `RefreshError` gains an `error_code: str | None` attribute populated from the RFC 6749 `error` field; `auth status` branches on `exc.error_code == "invalid_grant"`. As a side-effect the refresh error path no longer echoes raw response bodies — closes the deferred follow-up from PR #126 (same bug from two angles). New tests cover `error_code` propagation, structured `error_description` rendering, and a sentinel regression guard ensuring `refresh_token=SENTINEL_REFRESH_LEAK` in a hostile body never reaches the rendered exception. There is also a complementary `auth status` test that feeds `unauthorized_client` with `"invalid_grant"` in prose and verifies `state == "needs-login"` — proves the branch is code-driven, not text-driven.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **note:** Addressed in 953d4bb. `report.masking_env_vars = _session_masking_env_vars()` lifted out of `_populate_stored_session` and into `auth_status` body, gated on `"stored-session" in detected`. When stored-session wins, the helper returns `[]` (no higher-precedence env var can be set or it would have won), so this is a no-op in the winning path and correct in the masking path. Scenarios 5 (`PIPEFY_OAUTH_*`) and 6 (`PIPEFY_TOKEN`) now assert the exact `masking_env_vars` payload — the existing setups covered the precondition but didnt check the field, which is how this slipped through.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **note:** Deferred to #146 per your "optional follow-up" framing. Worth flagging that tracing call #2 (`get_authenticated_client` → `ensure_fresh_session`) shows it short-circuits on freshness — the `time.time() < deadline` check passes against the just-stored `obtained_at`, so no second token-endpoint refresh happens. The actual cost is one extra `load_session` keychain round-trip (~10 ms on macOS, more on Linux Secret Service), not two refresh round-trips as the framing suggests. Still worth fixing; just smaller than implied.

docs/cli/auth.md
gbrlcustodio: **note:** Addressed in a6ab4d0. Added the suggested bullet verbatim under `[Unreleased] → ### Added`.

Issue #138, opened 2026-05-19
Add Device Authorization Grant (`pipefy auth login --device`) for headless flows


gbrlcustodio:
## Motivation

The current ``pipefy auth login`` (PR #125) requires a working browser on the same host as the CLI:

1. Bind a loopback HTTP server on ``127.0.0.1:<ephemeral>``.
2. Launch the user's browser at the auth URL.
3. The browser redirects to ``http://127.0.0.1:<port>/callback`` — only reachable from the same host.

This breaks for:

- SSH sessions to remote servers (no local browser; the redirect target is on the remote machine, not the user's laptop).
- CI runners doing one-time interactive setup.
- Container shells, locked-down corp laptops with restricted browser access, headless dev environments.

``--no-browser`` (added in PR #125) only changes the launch step — the loopback callback still has to land on the host running the CLI. So it doesn't help.

The OAuth 2.0 **Device Authorization Grant** (RFC 8628) is the canonical solution: the CLI prints a short code + a URL; the user opens that URL on *any* device with a browser (laptop, phone), enters the code, signs in; the CLI polls the IdP until authorization completes.

The piporacle Keycloak realm already advertises ``device_authorization_endpoint`` in its discovery doc (confirmed in PR #125 smoke test step 0). The ``pipefy-cli`` Keycloak client config also has device-flow enabled (per the handoff doc's "Keycloak client config" section).

## Proposed command surface

```
pipefy auth login --device
```

Behavior:

1. POST to ``device_authorization_endpoint`` (from discovery) with ``client_id=pipefy-cli`` + ``scope=openid profile email offline_access``.
2. Print to the user:

   ```
   To sign in:
     1. Open https://signin.pipefy.com/device on any browser.
     2. Enter code: WDJB-MJHT
   
   Waiting for sign-in... (this will time out in 5 minutes)
   ```

3. Poll the token endpoint with ``grant_type=urn:ietf:params:oauth:grant-type:device_code`` at the interval the IdP specifies (typically 5s).
4. On success, persist tokens to the keychain via existing ``storage.store_session`` — same shape as the loopback flow.
5. On user denial or timeout, surface a clear error.

## Implementation notes

- New module: ``oauth/device.py`` — analogous to ``oauth/flow.py`` but for the device grant.
- Reuses: ``fetch_provider_metadata``, ``store_session``, ``_http.http_client``.
- No loopback HTTP server, no browser launch — much simpler than the existing flow.
[...]

Issue #132, opened 2026-05-19
Cache `ProviderMetadata` to avoid re-fetching discovery on every refresh


gbrlcustodio:
## Motivation

``oauth/refresh.py::refresh_access_token`` re-fetches the OIDC discovery doc on every refresh (calls ``fetch_provider_metadata`` inside the function). For scripted ``pipefy`` use inside the refresh leeway window, every command incurs **2 HTTPS round-trips per refresh**: one for discovery (~50-150 ms) + one for the actual token POST.

Discovery responses for a given issuer are stable per RFC 8414. The CLI re-fetching them on every refresh is pure overhead — RFC 8414 §2.1 even says clients SHOULD cache the response.

Per the [/simplify review of e685210](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/131), this was deferred to keep the slice bounded since it requires a ``StoredSession`` schema change.

## Approach

Add ``ProviderMetadata`` (or its three URLs) to ``StoredSession``, populated at login time in ``flow.run_login``, consumed by ``refresh.refresh_access_token`` to skip discovery.

### Fields to add

```python
@dataclass(frozen=True)
class StoredSession:
    ...
    authorization_endpoint: str | None  # new
    token_endpoint: str | None          # new
    # `issuer` already present; no other discovery fields are
    # currently used past login, so we don't need to persist them.
```

### Migration

``load_session`` currently silently returns ``None`` on a ``TypeError`` from ``StoredSession(**data)`` — old keychain entries without the new fields would be treated as "no session", forcing a re-login. Two paths:

- **Hard migration (simplest):** accept the one-time re-login; ship a CHANGELOG note for the bump.
- **Soft migration:** make the new fields ``Optional`` and fall back to discovery when absent. After a release cycle, drop the fallback.

Recommend soft migration since the keychain is on each user's machine and there's no upgrade story for it.

### Refresh path change

```python
def refresh_access_token(*, session: StoredSession, http_client=None) -> dict:
    # Skip discovery; go straight to the persisted token_endpoint.
    ...
```

Signature becomes ``session``-shaped rather than the loose tuple it currently takes.

## Acceptance criteria

- [ ] ``StoredSession`` gains ``authorization_endpoint`` / ``token_endpoint`` (optional during migration).
- [ ] ``flow.run_login`` persists discovered endpoints alongside the tokens.
[...]

PR #144, opened 2026-05-20
feat(cli): harden `pipefy auth login` (#126)

### LINKED ISSUE #126: auth login: deferred hardening from PR #125 review

gbrlcustodio:
## Summary

Three security-grade fixes for `pipefy auth login`, all deferred from PR #125's review and tracked in #126. Rebased on top of merged #142 (`auth status`) and the merged refresh body-scrub side-effect; the originally deferred follow-up on `refresh.py:56` is now closed.

- **Issuer match (commit 1)** — Verify `metadata.issuer` from the discovery document matches the URL it was fetched from (OIDC Discovery 1.0 §4.3). Compared on the `rstrip("/")` form so trailing-slash variants behave the same; both URLs land in the error message. Without this, a misrouted realm (piporacle vs prod) or a tampered discovery doc lands silently.
- **SSRF on discovery-returned endpoints (commit 2)** — `authorization_endpoint` and `token_endpoint` are now run through the shared `pipefy_sdk.utils.url_ssrf.validate_https_service_endpoint_url` before reaching the caller. A tampered discovery response otherwise sends the browser dance / token exchange wherever the doc says. Plumbing uses a new `DiscoveryPolicy` frozen dataclass (one field today, `allow_insecure_urls`) so the next two follow-ups (#132 cached metadata, #138 device flow) don't have to re-thread kwargs. The validation lives **inside** `fetch_provider_metadata` because `ProviderMetadata` is a frozen dataclass other paths construct directly — pushing the invariant down to where the type is born means future callers inherit it.
- **Scrub raw response body from token-exchange errors (commit 3)** — `exchange_code` was rendering `response.text[:300]` into `LoginError` on any non-200. RFC 6749 doesn't forbid IdPs from echoing submitted params (like our `code_verifier`); a `[:N]` window is a guaranteed echo channel under a hostile IdP. Now surfaces only the OAuth-standard `error` / `error_description` fields, with status-only fallback on missing/non-JSON/non-dict bodies. Includes a regression-guard test that asserts a sentinel `code_verifier=...` substring outside `error`/`error_description` does not appear in the resulting message.
- **Test-helper cleanup (commit 4)** — `TestFlow._call_exchange` was a class method that never used `self`. Lifted to module-level `_exchange_error_for`, matching the file's existing convention (`_discovery_payload`, `_fire_callback`).
- **Discovery non-200 scrub + IdP trust-boundary doc (commit 6, from review)** — `discovery.py` non-200 path was still echoing `response.text[:200]`; status-only now. Documented in `_format_token_error` that `error_description` is free-form per RFC 6749 §5.2 and surfacing it is part of trusting the IdP (no length cap — attackers can truncate to fit, tight caps kill legitimate context).

Closes #126.

## Out of scope (filed as follow-ups)

- ~~`discovery.py` non-200 path also does `response.text[:200]`~~ — **addressed in commit 6** in response to PR review.
- ~~`refresh.py` does the same `response.text[:300]` on refresh-grant failures~~ — **closed by PR #142** as a side-effect of the `RefreshError.error_code` refactor (squash `b3b05b1` on `dev`, originating commit `b4a3fec`).

## Test plan

- [x] Targeted unit tests: `pytest packages/cli/tests/test_auth_login.py` — passing (covers issuer mismatch, trailing-slash variants, SSRF rejection on http/internal-IP, `allow_insecure_urls` bypass, structured token-error rendering, sentinel regression guard, and the new `test_fetch_non_200_raises` sentinel for the D3 discovery scrub).
- [x] Full CLI regression: `pytest packages/cli/` — 242 passed, 10 skipped (clean after the #142 rebase).
- [x] Live smoke against the piporacle realm:
  - [x] **Happy path** — `pipefy auth login` completes against `https://signin-piporacle.pipefy.com/realms/st-piporacle-pud1m`, session stored in macOS Keyring. `pipefy auth status --json` returns `auth_source: "stored-session"`, `state: "active"`, populated `identity.email`/`identity.name`.
  - [x] **`masking_env_vars` (PR #142 cross-check)** — with `PIPEFY_OAUTH_*` set, `pipefy auth status --json` reports `auth_source: "service-account"` while `masking_env_vars: ["PIPEFY_OAUTH_*"]` surfaces the masked stored session. Confirms the C3 fix from #142 fires in production.
  - [x] **Tampered-discovery negative path** — `PIPEFY_AUTH_URL=…/wrong-realm-does-not-exist` returns `Login failed: OIDC discovery failed (404) at …`. Browser does NOT open; **no HTML body echoed** (D3 scrub working live).
  - [x] **SSRF reject without flag** — `PIPEFY_AUTH_URL=http://localhost:9999/realms/test` rejected at the settings layer (Pydantic `auth_url: must use HTTPS`) before the command starts.
  - [x] **`PIPEFY_ALLOW_INSECURE_URLS=true` bypass** — same URL with the flag set bypasses the settings HTTPS check and reaches the discovery network call (`ECONNREFUSED`, expected). Confirms `DiscoveryPolicy(allow_insecure_urls=…)` plumbs end-to-end from `PipefySettings` through `run_login` to `fetch_provider_metadata`.
  - Note: the PR #144 SSRF gate on discovery-*returned* endpoints (vs the settings-layer check on the issuer URL itself) can't be exercised live without a synthetic IdP serving a malicious discovery doc. That gate is covered by unit tests in `test_auth_login.py` (`test_fetch_returns_invalid_authorization_endpoint`, `test_fetch_returns_localhost_token_endpoint`, plus the `allow_insecure_urls=True` bypass case).

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -27,6 +27,7 @@ from pipefy_cli.auth import (
 )
 from pipefy_cli.commands._common import settings_and_auth_from_ctx
 from pipefy_cli.oauth import (
+    DiscoveryPolicy,
     LoginError,
     RefreshError,
     ensure_fresh_session,
@@ -123,7 +124,7 @@ def auth_login(
     # Lazy to keep keyring's ~30-80ms backend-discovery cost off every CLI startup.
     from keyring.errors import KeyringError
 
-    _, auth = settings_and_auth_from_ctx(ctx)
+    settings, auth = settings_and_auth_from_ctx(ctx)
     if auth.oidc_client is None:
         typer.echo(
             "PIPEFY_AUTH_URL is required for `pipefy auth login` (the OIDC issuer "
@@ -159,6 +160,9 @@ def auth_login(
             callback_timeout_s=callback_timeout,
             open_browser=_open,
             on_url=_print_url,
+            discovery_policy=DiscoveryPolicy(
+                allow_insecure_urls=settings.allow_insecure_urls
+            ),
         )
     except LoginError as exc:
         typer.echo(f"Login failed: {exc}", err=True)
--- a/packages/cli/src/pipefy_cli/oauth/discovery.py
+++ b/packages/cli/src/pipefy_cli/oauth/discovery.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 import httpx
+from pipefy_sdk.utils.url_ssrf import validate_https_service_endpoint_url
 
 from pipefy_cli.oauth import _http
 
@@ -21,6 +22,17 @@ class ProviderMetadata:
     token_endpoint: str
 
 
+@dataclass(frozen=True)
+class DiscoveryPolicy:
+    """Knobs governing how ``fetch_provider_metadata`` validates its inputs.
+
+    Bundled so future flags (cached metadata, custom timeouts) can grow here
+    without re-threading kwargs through every caller.
+    """
+
+    allow_insecure_urls: bool = False
+
+
 def _normalize_issuer(issuer_url: str) -> str:
     return issuer_url.rstrip("/")
 
@@ -33,14 +45,22 @@ def discovery_url(issuer_url: str) -> str:
 def fetch_provider_metadata(
     issuer_url: str,
     *,
+    policy: DiscoveryPolicy = DiscoveryPolicy(),
     timeout: float = _DEFAULT_TIMEOUT,
     client: httpx.Client | None = None,
 ) -> ProviderMetadata:
     """Fetch and parse the issuer's OIDC discovery document.
 
+    Validates that ``metadata.issuer`` matches ``issuer_url`` (per OIDC
+    Discovery 1.0 §4.3) and that the returned ``authorization_endpoint`` and
+    ``token_endpoint`` aren't pointing at internal hosts or non-HTTPS URLs —
+    a tampered or misconfigured discovery doc must not be allowed to redirect
+    the browser dance or token exchange to an attacker-controlled target.
+
     Raises:
-        ValueError: When the discovery document is unreachable, malformed, or
-            missing required endpoints. Message is user-facing.
+        ValueError: When the discovery document is unreachable, malformed,
+            issuer-mismatched, or returns endpoint URLs that don't pass the
+            shared SSRF check. Message is user-facing.
     """
     url = discovery_url(issuer_url)
     with _http.http_client(client, timeout=timeout) as http:
@@ -80,15 +100,31 @@ def fetch_provider_metadata(
             f"document claims {claimed_issuer!r}"
         )
 
+    authorization_endpoint = str(data["authorization_endpoint"])
+    token_endpoint = str(data["token_endpoint"])
+    for field, value in (
+        ("authorization_endpoint", authorization_endpoint),
+        ("token_endpoint", token_endpoint),
+    ):
+        try:
+            validate_https_service_endpoint_url(
+                value, field, allow_insecure=policy.allow_insecure_urls
+            )
+        except ValueError as exc:
+            raise ValueError(
+                f"OIDC discovery returned invalid {field}: {exc}"
+            ) from exc
+
     return ProviderMetadata(
         issuer=claimed_issuer,
-        authorization_endpoint=str(data["authorization_endpoint"]),
-        token_endpoint=str(data["token_endpoint"]),
+        authorization_endpoint=authorization_endpoint,
+        token_endpoint=token_endpoint,
     )
 
 
 __all__ = [
     "DISCOVERY_PATH",
+    "DiscoveryPolicy",
     "ProviderMetadata",
     "discovery_url",
     "fetch_provider_metadata",
--- a/packages/cli/src/pipefy_cli/oauth/discovery.py
+++ b/packages/cli/src/pipefy_cli/oauth/discovery.py
@@ -111,9 +111,7 @@ def fetch_provider_metadata(
                 value, field, allow_insecure=policy.allow_insecure_urls
             )
         except ValueError as exc:
-            raise ValueError(
-                f"OIDC discovery returned invalid {field}: {exc}"
-            ) from exc
+            raise ValueError(f"OIDC discovery returned invalid {field}: {exc}") from exc
 
     return ProviderMetadata(
         issuer=claimed_issuer,
--- a/packages/cli/src/pipefy_cli/oauth/discovery.py
+++ b/packages/cli/src/pipefy_cli/oauth/discovery.py
@@ -71,8 +71,17 @@ def fetch_provider_metadata(
             f"OIDC discovery at {url} is missing required fields: {', '.join(missing)}"
         )
 
+    # OIDC Discovery 1.0 §4.3 — the ``issuer`` claim must match the URL the
+    # document was fetched from. Trailing slashes don't carry meaning here.
+    claimed_issuer = str(data["issuer"])
+    if _normalize_issuer(claimed_issuer) != _normalize_issuer(issuer_url):
+        raise ValueError(
+            f"OIDC discovery issuer mismatch: requested {issuer_url!r}, "
+            f"document claims {claimed_issuer!r}"
+        )
+
     return ProviderMetadata(
-        issuer=str(data["issuer"]),
+        issuer=claimed_issuer,
         authorization_endpoint=str(data["authorization_endpoint"]),
         token_endpoint=str(data["token_endpoint"]),
     )
--- a/packages/cli/src/pipefy_cli/oauth/discovery.py
+++ b/packages/cli/src/pipefy_cli/oauth/discovery.py
@@ -72,10 +72,10 @@ def fetch_provider_metadata(
             ) from exc
 
     if response.status_code != 200:
-        raise ValueError(
-            f"OIDC discovery failed ({response.status_code}) at {url}: "
-            f"{response.text[:200]}"
-        )
+        # Status-only; never echo the raw body. Discovery isn't OAuth so there's
+        # no RFC 6749 ``error`` field to surface, and a `[:N]` window of the
+        # body is the same echo-channel class scrubbed in ``_format_token_error``.
+        raise ValueError(f"OIDC discovery failed ({response.status_code}) at {url}")
 
     try:
         data = response.json()
--- a/packages/cli/src/pipefy_cli/oauth/flow.py
+++ b/packages/cli/src/pipefy_cli/oauth/flow.py
@@ -11,7 +11,11 @@ from urllib.parse import urlencode
 import httpx
 
 from pipefy_cli.oauth import _http
-from pipefy_cli.oauth.discovery import ProviderMetadata, fetch_provider_metadata
+from pipefy_cli.oauth.discovery import (
+    DiscoveryPolicy,
+    ProviderMetadata,
+    fetch_provider_metadata,
+)
 from pipefy_cli.oauth.loopback import CallbackResult, LoopbackCapture
 from pipefy_cli.oauth.pkce import challenge_from_verifier, generate_verifier
 
@@ -99,6 +103,7 @@ def run_login(
     open_browser: Callable[[str], bool] = webbrowser.open,
     on_url: Callable[[str], None] | None = None,
     http_client: httpx.Client | None = None,
+    discovery_policy: DiscoveryPolicy = DiscoveryPolicy(),
 ) -> LoginResult:
     """Run the full PKCE loopback login. Returns tokens; does **not** persist them.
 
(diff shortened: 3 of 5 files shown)


packages/cli/src/pipefy_cli/oauth/flow.py
```diff
@@ -90,6 +92,30 @@ def exchange_code(
...
+    try:
+        payload = response.json()
+    except ValueError:
+        return generic
+    if not isinstance(payload, dict):
+        return generic
+    error = payload.get("error")
+    if not error:
+        return generic
+    description = payload.get("error_description")
```
adriannoes, line 117: **Context:** `_format_token_error` surfaces OAuth `error` and `error_description` to the user.

**Issue:** The regression test guards extra JSON keys and raw HTML, but not a hostile `error_description` that echoes submitted parameters.

**Why:** RFC 6749 permits free-form `error_description`; that field is still user-visible CLI output. Worth documenting as residual risk unless you want a length cap.

**Suggestion:** Add a one-line comment in `_format_token_error` documenting the trust boundary, or add a length limit on `error_description` in user messages.

**Done when:** Residual risk is explicit in code or docs, or descriptions are bounded.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -110,6 +111,9 @@ def _open(url: str) -> bool:
             callback_timeout_s=callback_timeout,
             open_browser=_open,
             on_url=_print_url,
+            discovery_policy=DiscoveryPolicy(
```
adriannoes, line 163: **Context:** #144 touches `auth_login` here; #142 (`pipefy auth status`) adds a large block in the same file.

**Issue:** Parallel edits will conflict at merge time.

**Why:** Both are intended to land; order doesn't matter functionally but the combined file must compile and pass tests (including `test_auth_integration` fixed in #142).

**Suggestion:** Rebase this branch after #142 merges to `dev`, or merge #142 first, then re-run `pytest packages/cli/`.

**Done when:** Rebased branch is green on full CLI pytest.

packages/cli/src/pipefy_cli/oauth/discovery.py
```diff
@@ -71,15 +91,38 @@ def fetch_provider_metadata(
             f"OIDC discovery at {url} is missing required fields: {', '.join(missing)}"
         )
 
+    # OIDC Discovery 1.0 §4.3 — the ``issuer`` claim must match the URL the
+    # document was fetched from. Trailing slashes don't carry meaning here.
+    claimed_issuer = str(data["issuer"])
```
adriannoes, line 96: **Context:** Non-200 OIDC discovery responses (a few lines above this hunk) still raise `ValueError` with `response.text[:200]`.

**Issue:** That path was not updated in this PR; only token-exchange errors were scrubbed.

**Why:** Truncated discovery bodies are the same echo-channel class as commit 3 — inconsistent policy and riskier bug reports if an error page echoes sensitive content.

**Suggestion:** Align the discovery non-200 path with `_format_token_error` (status-only or OAuth `error` fields) or a shared helper — you already flagged this as follow-up in the PR body.

**Done when:** Discovery failure messages never include raw response text.

packages/cli/src/pipefy_cli/oauth/flow.py
```diff
@@ -90,6 +92,30 @@ def exchange_code(
     return payload
 
 
+def _format_token_error(response: httpx.Response) -> str:
```
adriannoes, line 95: **Context:** `_format_token_error` (this function) scrubs raw bodies on token-exchange failures; `refresh.py` still uses `response.text[:300]` on refresh-grant non-200.

**Issue:** Refresh path is out of scope for #126 but shares the same threat model.

**Why:** A hostile IdP could echo the submitted `refresh_token` into the body; users paste CLI errors into tickets, and #142 `auth status` surfaces refresh failures.

**Suggestion:** Small follow-up: reuse this formatter (or extract a shared helper) in `oauth/refresh.py`.

**Done when:** `RefreshError` messages never include truncated raw bodies.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -74,7 +75,7 @@ def auth_login(
     # Lazy to keep keyring's ~30-80ms backend-discovery cost off every CLI startup.
     from keyring.errors import KeyringError
 
-    _, auth = settings_and_auth_from_ctx(ctx)
+    settings, auth = settings_and_auth_from_ctx(ctx)
```
adriannoes, line 127: **Context:** PR is still in **draft**; live smoke against piporacle is unchecked in the test plan.

**Issue:** Unit tests are thorough but manual issuer-mismatch and `PIPEFY_ALLOW_INSECURE_URLS` scenarios are not yet confirmed.

**Why:** These changes guard production misconfiguration; one real realm run catches wiring mistakes mocks won't (especially `DiscoveryPolicy.allow_insecure_urls` passed to `run_login` below).

**Suggestion:** Run the three smoke steps in the PR test plan, tick the checkboxes, then mark ready for review.

**Done when:** Smoke items are checked and PR is no longer draft (or note that smoke was done locally).

adriannoes (approved):
## Summary

Strong, focused security hardening for OIDC login: issuer §4.3 validation, SSRF checks on discovery-returned endpoints via shared SDK `url_ssrf`, and safe token-exchange error formatting with a sentinel regression test. **Approved** with follow-ups below; none are merge blockers. Complements [#142](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/142) without overlapping diffs.

## Follow-ups (inline threads)

| ID | Topic | Why it matters |
|----|--------|----------------|
| **F1** | Discovery non-200 still echoes `response.text[:200]` | Same echo-channel class as commit 3; align when you consolidate helpers. |
| **F2** | `refresh.py` still uses `response.text[:300]` | Refresh token echo risk; reuse login scrub (noted in PR out-of-scope). |
| **F3** | Trust boundary on `error_description` | Residual risk if a hostile IdP echoes secrets in an OAuth-legal field. |
| **F4** | Draft + live smoke unchecked | Run piporacle smoke before marking ready. |
| **F6** | Merge coordination with #142 | `commands/auth.py` conflict + integration test fix in #142. |

*(F5 changelog entry intentionally omitted per reviewer.)*

## Using this review with AI

Full record: `.cursor/dev-planning/code-review/pr-144-cli-auth-hardening.md`

## Verification

```bash
uv run pytest packages/cli/tests/test_auth_login.py -q
uv run pytest packages/cli/ -m "not integration" -q
```

Manual: piporacle happy login; tampered issuer → mismatch error; `PIPEFY_ALLOW_INSECURE_URLS=true` + localhost IdP.

packages/cli/src/pipefy_cli/oauth/flow.py
gbrlcustodio: **note:** Addressed in b03369b — added a comment in `_format_token_error` documenting that `error_description` is free-form per RFC 6749 §5.2 and surfacing it is part of trusting the IdP. Skipped the length-cap option: an attacker can truncate to fit any cap, and a tight cap kills legitimate error context (e.g. `"PKCE failed: code verifier does not match the stored challenge"`). Making the trust assumption explicit in code beats pretending a number fixes it.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **note:** Resolved by the rebase — #142 merged to `dev` (squash `b3b05b1`); this branch was rebased on top with all 5 commits replaying clean, no conflicts. `pytest packages/cli/` reports 242 passed, 10 skipped (the 10 are pre-existing).

packages/cli/src/pipefy_cli/oauth/discovery.py
gbrlcustodio: **note:** Addressed in b03369b — non-200 discovery message now reads `OIDC discovery failed ({status}) at {url}` with no body. Discovery isnt OAuth so there is no RFC 6749 `error`/`error_description` field to surface; status-only is the right shape, same threat-model as `_format_token_error`. Added a sentinel assertion to `test_fetch_non_200_raises` so the property is enforced, not only commented.

packages/cli/src/pipefy_cli/oauth/flow.py
gbrlcustodio: **note:** Already addressed by PR #142 (squash `b3b05b1` on `dev`; originating commit `b4a3fec`). The C2 review thread there forced a `RefreshError.error_code` restructure that, as a side effect, also parses the OAuth `error`/`error_description` fields and never echoes the raw body. The symmetry between token-exchange and refresh-grant paths is now in place.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **note:** PR was flipped from draft to ready-for-review earlier this cycle (not currently draft). Smoke against the piporacle realm is still pending — the three checkboxes in the test plan have not been ticked. Will either run the smoke and tick them, or note explicitly that we are merging without, before the final ready-to-land call.

gbrlcustodio: **note:** Smoke complete against piporacle. All checkboxes ticked in the PR body. Highlights:
- Happy path: login + identity round-trip green; stored session active in macOS Keyring.
- `masking_env_vars` cross-check: with `PIPEFY_OAUTH_*` set, `auth status` surfaces the masked stored session correctly (#142 C3 fix confirmed live).
- D3 scrub confirmed in production: `wrong-realm-does-not-exist` returns `OIDC discovery failed (404) at …` with **no HTML body echoed**.
- `PIPEFY_ALLOW_INSECURE_URLS=true` plumbs end-to-end: same URL that gets settings-layer-rejected without the flag passes through to the discovery network call with the flag set.
- PR #144 SSRF gate on discovery-returned endpoints (vs the settings check on the issuer URL) requires a synthetic malicious IdP — covered by unit tests in `test_auth_login.py`, can not be exercised live.

PR #149, opened 2026-05-21
feat(cli): pipefy auth logout with token revocation

### LINKED ISSUE #136: Add `pipefy auth logout` with token revocation
## Motivation

PR #125 added ``pipefy auth login`` (PKCE, stores tokens in the OS keychain). Today the only ways to "log out" are:

- Manually delete the keychain entry via the OS UI (``Keychain Access.app`` on macOS, ``seahorse`` on Linux, etc.).
- Run a Python one-liner against ``storage.delete_session``.

Neither is discoverable; neither revokes the refresh token at the IdP. So after a "logout" the stored access token is gone but the **refresh token is still valid at Keycloak** — anyone who recovers it from a backup / shoulder-surfs the original session can mint new access tokens until natural expiry.

This is the ``gh auth logout`` / ``aws sso logout`` analogue.

## Proposed command surface

```
pipefy auth logout
```

Behavior:

1. Load the stored session via ``storage.load_session``.
2. POST to the OIDC ``end_session_endpoint`` (advertised in the discovery doc) with ``client_id`` + ``refresh_token`` to revoke at Keycloak. The smoke test in PR #125 confirmed Keycloak returns 204 here and subsequent refreshes fail with ``invalid_grant``.
3. On 204 (or any non-network success), call ``storage.delete_session`` to clear the keychain entry.
4. If the revocation request fails (network down, IdP unreachable), still clear the keychain entry but warn that the refresh token couldn't be revoked at the server — the user knows to follow up via the Pipefy UI if needed.
5. Print ``Signed out of Pipefy (https://...).`` and exit 0.

When no session is stored: print ``Not signed in. Nothing to do.`` and exit 0 (idempotent — matches ``gh auth logout`` and most other CLIs).

## Implementation notes

- ``oauth/discovery.py::ProviderMetadata`` would need to expose ``end_session_endpoint`` (currently only carries ``authorization_endpoint`` + ``token_endpoint``). Cheap one-field addition.
- New module ``oauth/revoke.py`` (or method on ``refresh.py``) to do the POST. ~30 LOC.
- Smoke test confirmed:

  ```
  POST https://signin.pipefy.com/realms/pipefy/protocol/openid-connect/logout
  client_id=pipefy-cli
  refresh_token=<token>
  → 204
  ```

## Out of scope

- Logging out of a specific session when multiple are stored (single-session contract still holds — see PR #125).
- Clearing tokens from any external process that may have cached them (the SDK's in-process cache is reset on the next CLI invocation anyway).

## Acceptance criteria

- [ ] ``pipefy auth logout`` revokes the refresh token at the IdP's ``end_session_endpoint``.
- [ ] On revocation success, the keychain entry is removed.
- [ ] On revocation failure (network/IdP), the keychain entry is **still** removed and the user is warned in stderr.
- [ ] ``pipefy auth logout`` with no stored session is idempotent (exit 0, no error).
- [ ] After logout, subsequent ``pipefy <cmd>`` falls back to whichever lower-precedence source is configured (or prints the "no auth source" message).
- [ ] Tests cover: happy path, network failure during revocation, idempotent no-session case.

## References

- PR #125 (``pipefy auth login`` smoke test step 6 — confirms Keycloak ``end_session_endpoint`` returns 204 and revokes)
- ``packages/cli/src/pipefy_cli/oauth/storage.py::delete_session`` (already implemented)
- ``packages/cli/src/pipefy_cli/oauth/discovery.py::ProviderMetadata`` (needs ``end_session_endpoint`` field)

gbrlcustodio:
## Summary

Adds `pipefy auth logout`. Closes #136.

The command revokes the refresh token at the IdP's `end_session_endpoint` and then clears the keychain entry. Without server-side revocation a "logout" today deletes the keychain entry locally but leaves the refresh token valid at Keycloak until natural expiry — recoverable from a backup, the `gh auth logout` / `aws sso logout` shape closes that gap.

### Behavior

| Path | stdout | stderr | exit |
|------|--------|--------|------|
| Happy: stored session + revoke 2xx | `Signed out of Pipefy (<issuer>).` | — | 0 |
| No session stored | `Not signed in. Nothing to do.` | — | 0 |
| Revoke network / non-2xx failure | `Signed out of Pipefy (<issuer>).` | `Could not revoke refresh token at the IdP: <reason>. Clearing local session anyway; …` | 0 |
| IdP doesn't advertise `end_session_endpoint` | `Signed out of Pipefy (<issuer>).` | `Pipefy auth server does not advertise a logout endpoint; …` | 0 |
| `PIPEFY_AUTH_URL` unset | — | `PIPEFY_AUTH_URL is required for \`pipefy auth logout\` …` | 2 |

Revoke runs before delete because once the keychain entry is cleared we no longer have the refresh_token to send. On revoke failure the local session is still cleared (a stranded keychain entry that no command can refresh is worse than an honest "we tried, may still be valid server-side" warning).

### Design choices

- `ProviderMetadata.end_session_endpoint: str | None = None` — optional per OIDC Discovery 1.0; not every IdP advertises it. The CLI soft-fails when it's absent.
- SSRF-validate `end_session_endpoint` through the same `validate_https_service_endpoint_url` gate as `authorization_endpoint` and `token_endpoint` (introduced in #144 / closes the same threat model — a tampered discovery doc shouldn't be able to redirect the revoke POST to an attacker-controlled host).
- New `oauth/revoke.py` module rather than a method on `refresh.py` — parallel shape (POST to OIDC endpoint, scrubbed error rendering), unrelated semantics, distinct error hierarchy (`RevocationError` / `RevocationUnsupportedError`).
- Body scrubbing matches the pattern in `flow._format_token_error` and `refresh._refresh_error_from`: status code only on non-2xx, never echo `response.text[:N]`. We just sent `refresh_token` in the POST body; a window into a hostile IdP's reply is a guaranteed leak channel. Regression-guarded with a sentinel-substring assertion (`test_non_2xx_raises_with_status_only`).

### Out of scope

- Multi-session logout (`--all`). Single-session contract from #125 still holds.
- RFC 7009 access-token `/revoke`. Access tokens expire in minutes; killing the refresh chain via `end_session_endpoint` is the meaningful step.
- DNS-resolution SSRF. Same posture as #126.
- `--local-only` flag. Could land later if users ask for it.

### Commit shape

- `feat(cli): expose end_session_endpoint in ProviderMetadata`
- `feat(cli): add oauth/revoke.py for OIDC end-session revocation`
- `feat(cli): add pipefy auth logout command (closes #136)`
- `docs: changelog + auth.md entry for pipefy auth logout`
- `test(cli): refresh --help golden fixture for auth logout`

## Test plan

- [x] Unit: `pytest packages/cli/tests/test_auth_logout.py` — 10 passed (5 `revoke_session`, 5 command).
- [x] Discovery field coverage: `pytest packages/cli/tests/test_auth_login.py::TestDiscovery` — 12 passed (4 new for `end_session_endpoint`).
- [x] Full CLI regression: `pytest packages/cli/` — 256 passed, 10 skipped.
- [x] `ruff check` + `ruff format --check` — green.
- [x] Live smoke against piporacle:
  - Happy path: stored session → `pipefy auth logout` → `Signed out of Pipefy (...)`, exit 0. `pipefy auth status` then reports "Not signed in."
  - Idempotent: re-run `pipefy auth logout` with no session → `Not signed in. Nothing to do.`, exit 0.
  - (Revoke-network-failure path is covered deterministically by `test_revoke_network_failure_still_deletes_keychain`; not retried live to avoid a second interactive re-login.)

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -30,9 +30,13 @@ from pipefy_cli.oauth import (
     DiscoveryPolicy,
     LoginError,
     RefreshError,
+    RevocationError,
+    RevocationUnsupportedError,
+    delete_session,
     ensure_fresh_session,
     keychain_backend_name,
     load_session,
+    revoke_session,
     run_login,
     store_session,
 )
@@ -430,6 +434,59 @@ def auth_status(
         raise typer.Exit(exit_.exit_code) from exit_
 
     _render(report, json_out=json_out)
+
+
+@auth_app.command("logout")
+def auth_logout(ctx: typer.Context) -> None:
+    """Revoke the stored refresh token at the IdP and clear the local session."""
+    settings, auth = settings_and_auth_from_ctx(ctx)
+    if auth.oidc_client is None:
+        typer.echo(
+            "PIPEFY_AUTH_URL is required for `pipefy auth logout` (the OIDC "
+            "issuer URL used at login). "
+            f"See {DOCS_CLI_AUTH_REF}.",
+            err=True,
+        )
+        raise typer.Exit(2)
+
+    issuer = auth.oidc_client.issuer_url
+    client_id = auth.oidc_client.client_id
+
+    session = load_session(issuer=issuer, client_id=client_id)
+    if session is None:
+        typer.echo("Not signed in. Nothing to do.")
+        return
+
+    # Revoke at the IdP first — once we delete the keychain entry we no longer
+    # have the refresh_token to send. On any revoke failure still proceed to
+    # delete (warning makes the difference honest: server-side credential may
+    # remain valid until natural expiry).
+    revoke_warning: str | None = None
+    try:
+        revoke_session(
+            issuer=issuer,
+            client_id=client_id,
+            refresh_token=session.refresh_token,
+            policy=DiscoveryPolicy(allow_insecure_urls=settings.allow_insecure_urls),
+        )
+    except RevocationUnsupportedError:
+        revoke_warning = (
+            "Pipefy auth server does not advertise a logout endpoint; the "
+            "refresh token could not be revoked server-side. Clearing local "
+            "session only."
+        )
+    except RevocationError as exc:
+        revoke_warning = (
+            f"Could not revoke refresh token at the IdP: {exc}. Clearing "
+            "local session anyway; the refresh token may remain valid at the "
+            "server until natural expiry."
+        )
+
+    delete_session(issuer=issuer, client_id=client_id)
+
+    if revoke_warning:
+        typer.echo(revoke_warning, err=True)
+    typer.echo(f"Signed out of Pipefy ({issuer}).")
 
 
 __all__ = ["auth_app"]
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -32,6 +32,7 @@ from pipefy_cli.oauth import (
     RefreshError,
     RevocationError,
     RevocationUnsupportedError,
+    SessionDeleteError,
     delete_session,
     ensure_fresh_session,
     keychain_backend_name,
@@ -482,7 +483,18 @@ def auth_logout(ctx: typer.Context) -> None:
             "server until natural expiry."
         )
 
-    delete_session(issuer=issuer, client_id=client_id)
+    try:
+        delete_session(issuer=issuer, client_id=client_id)
+    except SessionDeleteError as exc:
+        if revoke_warning:
+            typer.echo(revoke_warning, err=True)
+        typer.echo(
+            f"Could not delete local session from the keychain: {exc}. "
+            "The stored credential may still be present; remove it manually "
+            f"via your OS keychain (service: 'pipefy-cli'). See {DOCS_CLI_AUTH_REF}.",
+            err=True,
+        )
+        raise typer.Exit(1) from exc
 
     if revoke_warning:
         typer.echo(revoke_warning, err=True)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 - **CLI**: `pipefy auth status [--json|-j]` — reports auth source, identity, session expiry, and exit codes.
+- **CLI**: `pipefy auth logout` — revokes the refresh token at the IdP and clears the stored session.
 
 ### Changed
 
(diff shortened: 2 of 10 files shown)


packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -432,4 +436,57 @@ def auth_status(
...
+            "session only."
+        )
+    except RevocationError as exc:
+        revoke_warning = (
+            f"Could not revoke refresh token at the IdP: {exc}. Clearing "
+            "local session anyway; the refresh token may remain valid at the "
+            "server until natural expiry."
+        )
+
+    delete_session(issuer=issuer, client_id=client_id)
```
adriannoes: `delete_session` already returns `False` when the keyring backend fails, but `auth_logout` always prints `Signed out of Pipefy (...)` afterward. If revoke at the IdP succeeded and the local delete fails, the user gets a success line with no stderr warning while the refresh token may still be in the keychain — that's a worse failure mode than the documented revoke-failure path because nothing signals the problem.

I'd capture the return value and, on failure, emit an honest stderr message and avoid claiming a full sign-out (exit 1 is reasonable here). A small unit test that stubs `delete_session` to return `False` after a successful revoke would lock the behavior.

adriannoes (approved):
Solid, well-scoped logout command.. revoke-before-delete ordering, SSRF validation on `end_session_endpoint`, and error scrubbing all match the login hardening posture from #144. Docs, CHANGELOG, and tests (including the sentinel leak guard) look good. I ran the targeted logout/discovery tests locally (22 passed) on top of green CI.

## Also noted

- **(inline on `delete_session`)**: if keychain delete fails after a successful IdP revoke, stdout still claims a full sign-out with no stderr warning — worth handling the `delete_session` return value or documenting the edge case.

## Verification

```bash
uv run pytest packages/cli/tests/test_auth_logout.py packages/cli/tests/test_auth_login.py::TestDiscovery -m "not integration" -q
uv run pytest packages/cli/ -m "not integration" -q
```

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **resolution:** addressed in 1c21b20.

`delete_session` no longer conflates the two `False` paths — `PasswordDeleteError` ("entry absent") still returns `False`, backend-level `KeyringError` now raises `SessionDeleteError`. `auth_logout` catches it, emits the stderr warning ("stored credential may still be present; remove it manually …"), skips the "Signed out" line, and exits 1.

Mirrors the `RevocationError` / `RevocationUnsupportedError` split so the storage layer ends up with the same vocabulary as the OAuth layer.

Tests: `TestStorage::test_delete_raises_on_backend_failure` (storage contract) + `test_keychain_delete_failure_after_revoke_exits_1` (command shape — no "Signed out", stderr warning, exit 1).

PR #151, opened 2026-05-21
perf(cli): cache OIDC endpoints in stored session (#132)

### LINKED ISSUE #132: Cache `ProviderMetadata` to avoid re-fetching discovery on every refresh

adriannoes:
## Summary
- Persist `authorization_endpoint` and `token_endpoint` in `StoredSession` at login time (soft migration: optional fields default to `None`).
- `refresh_access_token` uses the cached `token_endpoint` when present, skipping OIDC discovery on refresh; legacy keychain entries without endpoints still re-discover.
- Tests cover cached refresh (no discovery HTTP call) and legacy session fallback.

Closes #132

## Test plan
- [x] `uv run pytest packages/cli/tests/test_auth_refresh.py packages/cli/tests/test_auth_login.py -m "not integration"`
- [x] `uv run ruff check` / `uv run ruff format` on touched files

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -179,6 +179,8 @@ def auth_login(
             issuer=result.issuer,
             client_id=client_id,
             token_response=result.token_response,
+            authorization_endpoint=result.authorization_endpoint,
+            token_endpoint=result.token_endpoint,
         )
     except ValueError as exc:
         typer.echo(
--- a/packages/cli/src/pipefy_cli/oauth/flow.py
+++ b/packages/cli/src/pipefy_cli/oauth/flow.py
@@ -33,6 +33,8 @@ class LoginResult:
 
     issuer: str
     token_response: dict[str, object]
+    authorization_endpoint: str
+    token_endpoint: str
 
 
 def build_authorization_url(
@@ -196,7 +198,12 @@ def run_login(
                 code_verifier=verifier,
                 client=http,
             )
-    return LoginResult(issuer=metadata.issuer, token_response=token_response)
+    return LoginResult(
+        issuer=metadata.issuer,
+        token_response=token_response,
+        authorization_endpoint=metadata.authorization_endpoint,
+        token_endpoint=metadata.token_endpoint,
+    )
 
 
 def _ensure_callback_ok(callback: CallbackResult, *, expected_state: str) -> None:
--- a/packages/cli/src/pipefy_cli/oauth/refresh.py
+++ b/packages/cli/src/pipefy_cli/oauth/refresh.py
@@ -37,16 +37,23 @@ def refresh_access_token(
     *,
     session: StoredSession,
     http_client: httpx.Client | None = None,
-) -> dict[str, object]:
+) -> tuple[dict[str, object], str | None, str | None]:
     """POST ``grant_type=refresh_token`` to the issuer's token endpoint.
 
     Uses the persisted ``token_endpoint`` when available; otherwise falls back
     to OIDC discovery (legacy keychain entries from before endpoint caching).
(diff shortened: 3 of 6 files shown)


gbrlcustodio:
note (blocking): closing — the design pins `authorization_endpoint` / `token_endpoint` in the keychain `StoredSession`, which is the antipattern OIDC discovery exists to prevent.

A keychain entry survives across many refresh cycles, and if the IdP rotates a path every existing entry is silently wrong until the user re-logs in. The soft-migration fallback only handles the *schema* case (missing fields → re-discover); when the fields are present but stale there's no signal to fall back.

If we want to cut the refresh-time round-trip later, the cache has to be TTL-bounded (e.g., a file under `XDG_CACHE_HOME` honoring `Cache-Control` / `max-age` from the discovery response), not pinned to the session. See close note on #132.

PR #152, opened 2026-05-21
chore: enable ruff S101 workspace-wide (#147)

### LINKED ISSUE #147: chore: enable ruff `S101` (no `assert` in production code) workspace-wide
## Motivation

PEP 8 says: *"Don't use `assert` for checking the validity of input or for any other purpose where the check needs to run in production."* Asserts are stripped under `python -O` / `PYTHONOPTIMIZE=1`, silently turning a narrowing guard into an `AttributeError` on the next line.

PR #142 removed three asserts from the auth path (`packages/cli/src/pipefy_cli/auth.py:198,201`, `packages/cli/src/pipefy_cli/commands/auth.py:405`) by replacing them with explicit `if x is None: raise typer.Exit(2)` guards. One assert still exists in production code:

- `packages/cli/src/pipefy_cli/oauth/flow.py:153` — `assert callback.code is not None` after `_ensure_callback_ok(callback, expected_state=state)` has already raised `LoginError` for the None/empty cases.

Without a lint rule, anyone can reintroduce the same pattern in a future PR.

## Approach

Enable ruff's `S101` (from `flake8-bandit`) workspace-wide via the **"each package extends root"** config shape (the per-package configs already exist for `TID251` banned-imports — they currently don't inherit from the root config because ruff replaces rather than merges when subdirectories define `[tool.ruff]`).

### Root `pyproject.toml`

```toml
[tool.ruff]
lint.extend-select = ["I", "S101"]
lint.per-file-ignores = { "**/tests/**" = ["S101"] }
```

### Each package `pyproject.toml` (`packages/{cli,sdk,mcp}/pyproject.toml`)

```toml
[tool.ruff]
extend = "../../pyproject.toml"
lint.extend-select = ["TID251"]   # per-package banned-imports stay local

[tool.ruff.lint.flake8-tidy-imports.banned-api]
# ... unchanged ...
```

### `packages/cli/src/pipefy_cli/oauth/flow.py:153`

Replace `assert callback.code is not None` with `typing.cast(str, callback.code)` — the runtime invariant is already enforced by `_ensure_callback_ok` raising `LoginError` for both `None` and empty-string cases, so no defensive double-check is warranted; `cast` purely satisfies the type narrower.

```python
from typing import cast
# ...
_ensure_callback_ok(callback, expected_state=state)
code = cast(str, callback.code)
```

## Acceptance criteria

- [ ] `uv run ruff check packages/` passes on `flow.py` after the cast change.
- [ ] Removing the cast (or reintroducing `assert callback.code is not None`) makes ruff fail with `S101 use of `assert` detected`.
- [ ] `uv run ruff check packages/cli/tests/` continues to pass (tests are exempt via per-file-ignores).
- [ ] Per-package `TID251` banned-imports still fire as before (regression check: introduce a banned import in `packages/cli/src/`, confirm it's still flagged).

## Scope

- Root `pyproject.toml`: +2 lines (extend-select + per-file-ignores).
- Each of `packages/{cli,sdk,mcp}/pyproject.toml`: rewrap with `extend = "../../pyproject.toml"`, keep `TID251` block.
- `packages/cli/src/pipefy_cli/oauth/flow.py`: 1 import, 2-line change.

## References

- PR #142 review thread #1 (adriannoes): https://github.com/gbrlcustodio/pipefy-mcp-server/pull/142#discussion_r3277256847 — argument for replacing asserts in production.
- PEP 8 on `assert` usage: https://peps.python.org/pep-0008/

adriannoes:
## Summary

- Enable ruff rule **S101** (`assert` banned in production code) at the workspace root, with `**/tests/**` exempt via `per-file-ignores`.
- Point each package `pyproject.toml` at the root config via `extend = "../../pyproject.toml"`, keeping local `TID251` and banned-api rules.
- Replace the lone production `assert` in OAuth flow with `typing.cast` after `_ensure_callback_ok` validates the callback code.

Closes #147

## Test plan

- [x] `uv run ruff check packages/` — all checks passed
- [x] `uv run pytest -m "not integration"` — 2257 passed
- [x] `uv run ruff format .` — no changes needed

### DIFF
--- a/packages/cli/pyproject.toml
+++ b/packages/cli/pyproject.toml
@@ -46,7 +46,8 @@ markers = [
 ]
 
 [tool.ruff]
-lint.extend-select = ["I", "TID251"]
+extend = "../../pyproject.toml"
+lint.extend-select = ["TID251"]
 
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "pipefy_mcp".msg = "pipefy_cli must not import the MCP package; use pipefy_sdk instead."
--- a/packages/cli/src/pipefy_cli/oauth/flow.py
+++ b/packages/cli/src/pipefy_cli/oauth/flow.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import secrets
 import webbrowser
 from dataclasses import dataclass
-from typing import Callable
+from typing import Callable, cast
 from urllib.parse import urlencode
 
 import httpx
@@ -186,12 +186,12 @@ def run_login(
 
             callback = capture.await_callback(timeout=callback_timeout_s)
             _ensure_callback_ok(callback, expected_state=state)
-            assert callback.code is not None
+            code = cast(str, callback.code)
 
             token_response = exchange_code(
                 metadata=metadata,
                 client_id=client_id,
-                code=callback.code,
+                code=code,
                 redirect_uri=capture.redirect_uri,
                 code_verifier=verifier,
                 client=http,
--- a/packages/mcp/pyproject.toml
+++ b/packages/mcp/pyproject.toml
@@ -56,7 +56,8 @@ markers = [
 ]
 
 [tool.ruff]
-lint.extend-select = ["I", "TID251"]
+extend = "../../pyproject.toml"
+lint.extend-select = ["TID251"]
 
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "pipefy_cli".msg = "pipefy_mcp must not import the CLI package; keep MCP independent of pipefy-cli."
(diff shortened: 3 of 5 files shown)


gbrlcustodio (approved):
LGTM. `cast` is safe — `_ensure_callback_ok` already raises on falsy `callback.code`, so the invariant holds by the time we reach line 189. `extend` plumbing is correct and the `**/tests/**` ignore matches the actual layout.

**suggestion (non-blocking, if-minor):** have `_ensure_callback_ok` return the validated `code: str` so the cast goes away entirely. Defer if uninteresting.

**thought:** the remaining `# noqa: S101` in `packages/sdk/src/pipefy_sdk/client.py` (lines 798/805/942) are service-injection sentinels — worth a follow-up to tighten the types or convert to explicit `if … is None: raise`. Out of scope here.

PR #153, opened 2026-05-21
perf(cli): avoid duplicate load_session in auth status (#146)

### LINKED ISSUE #146: perf(cli): `auth status` does two `load_session` round-trips on the stored-session path
## Motivation

`pipefy auth status` on the stored-session path invokes `ensure_fresh_session` twice per run:

1. `_populate_stored_session` (commands/auth.py:324) — does the actual refresh when the token is near expiry.
2. `_fetch_identity` → `get_authenticated_client` (auth.py:203) — re-enters `ensure_fresh_session` to obtain a bearer for the `get_me` call.

Tracing call #2 against the freshly-stored session:

- `load_session(...)` reads the keychain again (~10 ms on macOS, more on Linux/Secret Service).
- The freshness check (`time.time() < session.obtained_at + expires_in - leeway_s`) passes because we just stored a brand-new `obtained_at`.
- Returns without touching the token endpoint.

So the actual cost is one extra keychain read plus one extra `load_session` deserialization per `auth status`, **not** two network refreshes. The PR review framed this as "extra token-endpoint load," which is true only on the first call; call #2 short-circuits on freshness.

Raised by adriannoes on PR #142 (https://github.com/gbrlcustodio/pipefy-mcp-server/pull/142) as an explicit "Optional follow-up."

## Approach

Two equally viable shapes:

1. **Thread the refreshed `StoredSession` down**: `_populate_stored_session` returns the (possibly rotated) session; `_fetch_identity` accepts it and passes the bearer directly into the GraphQL client, skipping `get_authenticated_client`'s stored-session branch entirely.
2. **`get_authenticated_client` accepts an optional pre-fetched session**: same effect, opt-in at the call site, no changes to non-`auth status` callers.

Option 2 is less invasive (one new kwarg, default `None`) and keeps the `get_authenticated_client` API as the single entry point.

## Change

1. Add `prefetched_session: StoredSession | None = None` kwarg to `get_authenticated_client`.
2. When supplied, bypass the `load_session` + `ensure_fresh_session` round-trip for the stored-session branch; use its `access_token` as the bearer.
3. In `auth_status`, capture the refreshed session from `_populate_stored_session` and pass it into the `_fetch_identity` → `get_authenticated_client` call chain.

## Acceptance criteria

- [ ] `auth status` on the stored-session path calls `load_session` exactly once per invocation (observable via mock call counts in a new test).
- [ ] No behavior change on other auth paths (`flag-token`, `env-token`, `service-account`).
- [ ] No behavior change for other CLI commands consuming `get_authenticated_client`.

## Scope

- 1 source file (`packages/cli/src/pipefy_cli/auth.py`) — add kwarg, conditional branch.
- 1 source file (`packages/cli/src/pipefy_cli/commands/auth.py`) — capture + thread session.
- 1 test file (`packages/cli/tests/test_auth_status.py`) — call-count assertion.

## References

- PR #142 review thread: https://github.com/gbrlcustodio/pipefy-mcp-server/pull/142#discussion_r3277256916
- adriannoes labeled this "Optional follow-up" in the review.

adriannoes:
## Summary
- Add optional `prefetched_session` to `get_authenticated_client` so the stored-session branch skips a second `ensure_fresh_session` / keychain read when the caller already refreshed the session.
- `pipefy auth status` preloads the session once, reuses it for source detection and population, and passes the refreshed session into identity fetch.
- Add regression tests asserting a single `load_session` per `auth status` invocation on the stored-session path.

## Test plan
- [x] `uv run pytest packages/cli/tests/test_auth_status.py packages/cli/tests/test_auth.py -m "not integration"`
- [x] `uv run ruff check` / `ruff format` on touched files

Closes #146

### DIFF
--- a/packages/cli/src/pipefy_cli/auth.py
+++ b/packages/cli/src/pipefy_cli/auth.py
@@ -16,6 +16,7 @@ SDK, so a refresh-rotated access token naturally invalidates the cached client.
 
 from __future__ import annotations
 
+import time
 from dataclasses import dataclass
 from typing import Literal, NoReturn
 
@@ -32,7 +33,14 @@ from pipefy_cli.config import (
     describe_missing_oauth_vars,
     ensure_public_graphql_configured,
 )
-from pipefy_cli.oauth import RefreshError, ensure_fresh_session, load_session
+from pipefy_cli.oauth import (
+    RefreshError,
+    StoredSession,
+    ensure_fresh_session,
+    load_session,
+)
+
+_SESSION_FRESHNESS_LEEWAY_S = 60
 
 AuthSource = Literal[
     "flag-token",
@@ -108,9 +116,19 @@ def _cache_key(
     )
 
 
+def _session_is_fresh(
+    session: StoredSession, *, leeway_s: int = _SESSION_FRESHNESS_LEEWAY_S
+) -> bool:
+    """Return whether the access token is outside the refresh leeway window."""
+    expires_in = session.expires_in or 0
+    return time.time() < session.obtained_at + expires_in - leeway_s
+
+
 def detect_all_sources(
     pipefy_settings: PipefySettings,
     auth: AuthContext,
+    *,
+    cached_stored_session: bool | None = None,
 ) -> list[AuthSource]:
     """Return every configured credential source, highest-precedence first.
 
@@ -125,15 +143,18 @@ def detect_all_sources(
         )
     if not describe_missing_oauth_vars(pipefy_settings):
         sources.append("service-account")
-    if (
-        auth.oidc_client is not None
-        and load_session(
-            issuer=auth.oidc_client.issuer_url,
-            client_id=auth.oidc_client.client_id,
+    if auth.oidc_client is not None:
+        has_stored = (
+            cached_stored_session
+            if cached_stored_session is not None
+            else load_session(
+                issuer=auth.oidc_client.issuer_url,
+                client_id=auth.oidc_client.client_id,
+            )
+            is not None
         )
-        is not None
-    ):
-        sources.append("stored-session")
+        if has_stored:
+            sources.append("stored-session")
     return sources
 
 
@@ -169,6 +190,8 @@ def _raise_internal_error(detail: str) -> NoReturn:
 def get_authenticated_client(
     pipefy_settings: PipefySettings,
     auth: AuthContext,
+    *,
+    prefetched_session: StoredSession | None = None,
 ) -> PipefyClient:
     """Return a facade client using the highest-precedence available auth source.
 
@@ -179,6 +202,9 @@ def get_authenticated_client(
             tier 4 (stored user session keyed by issuer + client id). Tier 3
             (service-account client credentials) is resolved from
             ``pipefy_settings`` alone.
+        prefetched_session: When set (e.g. by ``pipefy auth status`` after a
+            single keychain read + refresh), skips a second
+            ``ensure_fresh_session`` round-trip for the stored-session branch.
 
     Returns:
         Shared in-process instance when settings match a prior call; otherwise
@@ -196,7 +222,10 @@ def get_authenticated_client(
         typer.echo(str(exc), err=True)
         raise typer.Exit(2) from exc
 
-    source = detect_auth_source(pipefy_settings, auth)
+    if prefetched_session is not None:
+        source: AuthSource = "stored-session"
+    else:
+        source = detect_auth_source(pipefy_settings, auth)
 
     if source == "none":
         typer.echo(_missing_auth_message(pipefy_settings), err=True)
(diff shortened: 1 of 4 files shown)


packages/cli/src/pipefy_cli/auth.py
```diff
@@ -32,7 +33,14 @@
...
 )
-from pipefy_cli.oauth import RefreshError, ensure_fresh_session, load_session
+from pipefy_cli.oauth import (
+    RefreshError,
+    StoredSession,
+    ensure_fresh_session,
+    load_session,
+)
+
+_SESSION_FRESHNESS_LEEWAY_S = 60
```
gbrlcustodio, line 43: **issue (non-blocking):** duplicates `_LEEWAY_S = 60` from `oauth/refresh.py:13`. If either drifts, the fresh-path predicate here and the stale-check inside `ensure_fresh_session` will silently disagree.

**suggestion:** have `ensure_fresh_session` accept an optional `session: StoredSession | None = None` and skip the inner `load_session` when supplied. Then `_populate_stored_session` just calls `ensure_fresh_session(session=loaded_session, ...)` — freshness predicate + leeway stay in one place, `_session_is_fresh` goes away, and the stale path collapses to 1 load too (currently still 2).

packages/cli/src/pipefy_cli/auth.py
```diff
@@ -125,24 +143,31 @@ def detect_all_sources(
...
-        and load_session(
-            issuer=auth.oidc_client.issuer_url,
-            client_id=auth.oidc_client.client_id,
+    if auth.oidc_client is not None:
+        has_stored = (
+            cached_stored_session
+            if cached_stored_session is not None
+            else load_session(
+                issuer=auth.oidc_client.issuer_url,
+                client_id=auth.oidc_client.client_id,
```
gbrlcustodio, line 152: **nitpick:** correct (the `is not None` binds to `load_session(...)`, not the whole ternary) but a re-read. Splitting helps:

```py
if cached_stored_session is None:
    cached_stored_session = (
        load_session(issuer=auth.oidc_client.issuer_url, client_id=auth.oidc_client.client_id)
        is not None
    )
if cached_stored_session:
    sources.append("stored-session")
```

packages/cli/src/pipefy_cli/auth.py
```diff
@@ -196,7 +226,11 @@ def get_authenticated_client(
         typer.echo(str(exc), err=True)
         raise typer.Exit(2) from exc
 
-    source = detect_auth_source(pipefy_settings, auth)
+    source = detect_auth_source(
+        pipefy_settings,
+        auth,
+        cached_stored_session=True if prefetched_session is not None else None,
```
gbrlcustodio, line 232: **nitpick (if-minor):** `cached_stored_session=prefetched_session is not None or None` says the same thing without the ternary.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -22,6 +22,7 @@
     AuthContext,
     AuthSource,
     OidcClient,
+    _session_is_fresh,
```
gbrlcustodio, line 25: **nitpick:** importing `_session_is_fresh` across module boundaries breaks the leading-underscore "module-private" convention. Folds away if `ensure_fresh_session(session=...)` absorbs the optimization — see the comment on `auth.py:43`.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -399,7 +423,19 @@ def auth_status(
...
+        loaded_session = load_session(
+            issuer=auth.oidc_client.issuer_url,
+            client_id=auth.oidc_client.client_id,
+        )
+    detected = detect_all_sources(
+        settings,
+        auth,
+        cached_stored_session=loaded_session is not None
+        if auth.oidc_client is not None
+        else None,
```
gbrlcustodio, line 437: **suggestion (if-minor):** the `if auth.oidc_client is not None else None` arm never fires usefully — `detect_all_sources` skips the `cached_stored_session` consultation entirely when `oidc_client is None` (`auth.py:146`). Just:

```py
cached_stored_session=loaded_session is not None,
```

is equivalent and one line shorter.

gbrlcustodio (commented):
Optimization is sound and the second commit correctly preserves bearer/service-account precedence (locked in by `test_prefetched_session_does_not_override_bearer_precedence`). Main feedback inline.

**chore:** PR is `CONFLICTING` against `dev` — needs a rebase before merge (likely the `describe_missing_oauth_vars` → `describe_missing_service_account_vars` rename landing on a parallel branch).

packages/cli/tests/test_auth_status.py
```diff
@@ -77,6 +77,42 @@ def _invoke_status(runner: Any, args: list[str] | None = None) -> Any:
     return runner.invoke(app, ["auth", "status", *(args or [])])
 
 
+_LOAD_SESSION_PATCH_TARGETS = (
+    "pipefy_cli.oauth.storage.load_session",
+    "pipefy_cli.oauth.refresh.load_session",
+    "pipefy_cli.commands.auth.load_session",
+    "pipefy_cli.auth.load_session",
+)
```
gbrlcustodio, line 85: **note (non-blocking):** silently under-counts if a 5th importer is added later — the new site bypasses the counter and `load_calls == 1` still holds. A docstring calling that out would be enough; not a blocker.

packages/cli/tests/test_auth_status.py
```diff
@@ -161,11 +197,14 @@ def test_status_refresh_expired_exits_2(
...
-        side_effect=RefreshError(
-            "Refresh failed (HTTP 400): invalid_grant", error_code="invalid_grant"
+    with (
+        patch(
+            "pipefy_cli.commands.auth.ensure_fresh_session",
+            side_effect=RefreshError(
+                "Refresh failed (HTTP 400): invalid_grant", error_code="invalid_grant"
+            ),
         ),
+        patch("pipefy_cli.commands.auth._session_is_fresh", return_value=False),
```
gbrlcustodio, line 207: **suggestion:** patching `_session_is_fresh` couples the test to a private helper. Seeding a stale session exercises the real freshness path instead:

```py
_seed_session(monkeypatch, obtained_at=int(time.time()) - 7200)
```

Same applies to lines 231 and 258. Folds away entirely if `ensure_fresh_session(session=...)` lands.

PR #155, opened 2026-05-21
refactor: rename PIPEFY_OAUTH_* env vars to PIPEFY_SERVICE_ACCOUNT_*

### LINKED ISSUE #127: Rename `PIPEFY_OAUTH_*` env vars to `PIPEFY_SERVICE_ACCOUNT_*`
## Motivation

PR #125 introduced ``PIPEFY_AUTH_URL`` for the new ``pipefy auth login`` (OIDC user auth via Authorization Code + PKCE). It sits one letter away from the existing ``PIPEFY_OAUTH_URL`` (service-account client-credentials), which is a real ergonomic footgun — users will mix them up in ``.env`` files and shell configs.

Beyond the visual similarity, ``PIPEFY_OAUTH_*`` is also imprecise: both flows are technically OAuth, and the name doesn't say *which grant* or *who the credentials represent*. The CLI and MCP server already use the domain term "service account" everywhere else:

- ``PipefySettings.service_account_ids``
- ``PIPEFY_SERVICE_ACCOUNT_IDS`` env var
- ``_SERVICE_ACCOUNT_ENV_KEYS`` constant in ``packages/cli/src/pipefy_cli/commands/auth.py``

The three credential vars are the only holdouts.

## Proposed rename

```
PIPEFY_OAUTH_URL    → PIPEFY_SERVICE_ACCOUNT_URL
PIPEFY_OAUTH_CLIENT → PIPEFY_SERVICE_ACCOUNT_CLIENT
PIPEFY_OAUTH_SECRET → PIPEFY_SERVICE_ACCOUNT_SECRET
```

The new ``PIPEFY_AUTH_URL`` stays as-is — it's correctly scoped already (user login has one auth URL, no qualifier needed).

## Backward compatibility — `AliasChoices` + deprecation warning

pydantic-settings supports clean migration via ``AliasChoices``:

```python
oauth_url: str | None = Field(
    default=None,
    validation_alias=AliasChoices(
        "PIPEFY_SERVICE_ACCOUNT_URL",  # new — preferred
        "PIPEFY_OAUTH_URL",            # old — accept for one release cycle, then drop
    ),
)
```

Plus a one-time ``stderr`` warning at startup when an old-name env var is used. Old configs keep working; users see a clear nudge.

After one beta release cycle, drop the legacy names. ``0.2.0-beta.x`` is the right window — beta semver still permits breaking changes, and the alias shim makes the deprecation painless.

## Scope — files this likely touches

- ``packages/sdk/src/pipefy_sdk/settings.py`` — rename fields on ``PipefySettings`` (``oauth_url`` → ``service_account_url``, etc.); update the URL validators to use the new field names in their error labels.
- ``packages/sdk/`` — internal callers using ``settings.oauth_*`` attributes.
- ``packages/mcp/`` — any MCP wiring that reads the renamed fields.
- ``packages/cli/src/pipefy_cli/config.py`` — ``describe_missing_oauth_vars`` becomes ``describe_missing_service_account_vars``; warning messages update.
- ``packages/cli/src/pipefy_cli/commands/auth.py`` — ``_SERVICE_ACCOUNT_ENV_KEYS`` tuple values update; the warning string ``"PIPEFY_OAUTH_*"`` becomes ``"PIPEFY_SERVICE_ACCOUNT_*"``.
- ``.env.example`` — update both the variable names and the inline help comments.
- ``docs/setup.md``, ``docs/MIGRATION.md`` — add migration guide; update setup walkthroughs.
- ``skills/**/SKILL.md`` — grep for env-var references.
- Tests across SDK / MCP / CLI — fixtures, ``conftest.py`` env-key tuples, assertions.
- ``packages/cli/tests/fixtures/cli_help_golden.txt`` — regen after rename touches help text.

## Acceptance criteria

- [ ] All three env vars renamed to ``PIPEFY_SERVICE_ACCOUNT_*`` across SDK / MCP / CLI source.
- [ ] ``PipefySettings`` field names renamed; old names not present.
- [ ] ``AliasChoices`` accepts legacy ``PIPEFY_OAUTH_*`` for backward compat.
- [ ] Stderr deprecation warning printed when any legacy var is detected at settings load.
- [ ] ``MIGRATION.md`` updated with a "0.2.0 → next" section covering the rename.
- [ ] ``.env.example`` shows new names; old names mentioned only in MIGRATION.
- [ ] Skills audited for env-var references (CI ``skills-lint.yml`` should catch most).
- [ ] Tests updated; full suite green.
- [ ] No CHANGELOG-mentioned "breaking" callout for this beta release cycle (the alias shim covers it); the eventual drop-of-aliases PR carries the breaking flag.

## Out of scope

- ``PIPEFY_TOKEN`` stays — it's a static bearer override, not a service-account credential.
- ``PIPEFY_AUTH_URL`` / ``PIPEFY_AUTH_CLIENT_ID`` (user-auth flow) — already correctly scoped.

## References

- Discussion in PR #125 (see [conversation](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/125#discussion))
- Existing precedent: ``service_account_ids`` field, ``PIPEFY_SERVICE_ACCOUNT_IDS`` env var

gbrlcustodio:
## Summary

Renames the three service-account credential env vars across SDK / CLI / MCP. Closes #127.

| Old | New |
|---|---|
| `PIPEFY_OAUTH_URL` | `PIPEFY_SERVICE_ACCOUNT_URL` |
| `PIPEFY_OAUTH_CLIENT` | `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID` |
| `PIPEFY_OAUTH_SECRET` | `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET` |

`PIPEFY_OAUTH_URL` sat one letter from `PIPEFY_AUTH_URL` (the OIDC issuer for the interactive user-login flow in #125 / #142 / #149) — a real ergonomic footgun, and `PIPEFY_OAUTH_*` was imprecise on its own (both flows are OAuth). The rest of the codebase already uses the domain term **service account** (`PIPEFY_SERVICE_ACCOUNT_IDS`, `_SERVICE_ACCOUNT_ENV_KEYS`).

## Naming choices

- **`SERVICE_ACCOUNT`** prefix matches existing `PIPEFY_SERVICE_ACCOUNT_IDS` precedent.
- **`_CLIENT_ID` / `_CLIENT_SECRET`** match OAuth 2.0 RFC 6749 (`client_id` / `client_secret`) and the existing internal `PIPEFY_AUTH_CLIENT_ID` from the user-login flow. `_CLIENT` / `_SECRET` would diverge from both.
- **`_URL`** (not `_TOKEN_URL`) mirrors `PIPEFY_AUTH_URL`. The prefix change alone fully resolves the footgun; the service-account endpoint is a Doorkeeper token URL with no issuer or `.well-known` discovery (confirmed against `pipefy-core`), so a forward-looking `_TOKEN_URL` would carry no future benefit.

`PIPEFY_TOKEN` (static bearer) and `PIPEFY_AUTH_URL` / `PIPEFY_AUTH_CLIENT_ID` (user-login) are **not** renamed — already correctly scoped.

## Back-compat mechanics

- **`AliasChoices(new, old)`** on each `PipefySettings` field accepts both forms; new wins when both are set. Honors `PIPEFY_OAUTH_*` env vars and `.env` keys without code changes on the user side.
- **TOML key alias**: `apply_toml_fallback` normalizes legacy `oauth_*` keys in `~/.config/pipefy/config.toml` to the new names; new key wins when both present.
- **One-shot stderr deprecation warning** per legacy key still in use, fired via a `model_validator(mode="before")` so it surfaces even when downstream validation fails. Process-global dedup; short-circuits once all legacy keys have warned.

The shim ships in the Unreleased section without a breaking-change callout. A follow-up PR drops the aliases and carries the breaking flag.

## Out of scope

- `PIPEFY_TOKEN` — static bearer, not a service-account credential.
- `PIPEFY_AUTH_URL` / `PIPEFY_AUTH_CLIENT_ID` — user-login flow, already correctly scoped.
- Dropping the `AliasChoices` shim — separate PR after one beta release cycle.

## Commit shape

- ` refactor: rename PIPEFY_OAUTH_* to PIPEFY_SERVICE_ACCOUNT_* (#127)` — source-file rename across SDK/CLI/MCP + AliasChoices + stderr warning + TOML alias + `InternalApiClient` kwargs + golden fixture regen.
- ` test: cover AliasChoices, env+TOML deprecation warnings for #127` — 13 new back-compat tests pinning the legacy paths and the new-wins precedence.
- ` docs: rename PIPEFY_OAUTH_* env vars to PIPEFY_SERVICE_ACCOUNT_* (#127)` — `.env.example`, `bootstrap.sh`, `docs/setup.md`, `docs/MIGRATION.md` (new section), READMEs, `SKILL.md`, `CHANGELOG.md` Unreleased entry.
- ` refactor: polish per /simplify review of the #127 rename` — early-out guards in the warning emitter and TOML normalizer, dedup the legacy-key map between settings and test files.

## Test plan

- [x] `uv run pytest packages/` — 2048 passed, 42 skipped.
- [x] `uv run ruff check packages/` + `uv run ruff format --check packages/` — clean.
- [x] Live smoke with legacy env vars set: all three stderr warnings fire naming the correct replacements; values populate the new fields.
- [x] Live smoke with new env vars only: clean stderr; values populate the new fields.
- [x] Source-file grep audit: only intentional legacy refs remain (`AliasChoices` shim, `_LEGACY_TOML_KEYS_TO_NEW` map, deprecation-warning message strings, docstrings naming both forms).

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -72,9 +72,9 @@ auth_app = typer.Typer(
 )
 
 _SERVICE_ACCOUNT_ENV_KEYS = (
-    "PIPEFY_OAUTH_URL",
-    "PIPEFY_OAUTH_CLIENT",
-    "PIPEFY_OAUTH_SECRET",
+    "PIPEFY_SERVICE_ACCOUNT_URL",
+    "PIPEFY_SERVICE_ACCOUNT_CLIENT_ID",
+    "PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET",
 )
 
 
@@ -82,15 +82,15 @@ def _session_masking_env_vars() -> list[str]:
     """Env vars that outrank a stored session in the credential precedence chain.
 
     Only ``os.environ`` is consulted — by the precedence model, ``.env`` defaults
-    sit below the stored session. ``PIPEFY_OAUTH_*`` is listed only when the
-    *complete* triple is configured (otherwise the client-credentials path
+    sit below the stored session. ``PIPEFY_SERVICE_ACCOUNT_*`` is listed only when
+    the *complete* triple is configured (otherwise the client-credentials path
     wouldn't activate and the warning would be misleading).
     """
     env_vars: list[str] = []
     if os.environ.get("PIPEFY_TOKEN"):
         env_vars.append("PIPEFY_TOKEN")
     if all(os.environ.get(k) for k in _SERVICE_ACCOUNT_ENV_KEYS):
-        env_vars.append("PIPEFY_OAUTH_*")
+        env_vars.append("PIPEFY_SERVICE_ACCOUNT_*")
     return env_vars
 
 
@@ -204,7 +204,7 @@ def auth_login(
 _AUTH_SOURCE_LABELS: dict[AuthSource, str] = {
     "flag-token": "--token flag",
     "env-token": "PIPEFY_TOKEN environment variable",
-    "service-account": "PIPEFY_OAUTH_* (client credentials)",
+    "service-account": "PIPEFY_SERVICE_ACCOUNT_* (client credentials)",
     "stored-session": "stored session (`pipefy auth login`)",
     "none": "none",
 }
@@ -260,7 +260,7 @@ def _render_status_text(report: AuthStatusReport) -> None:
     if not report.signed_in:
         typer.echo(
             "Not signed in. Run `pipefy auth login`, set PIPEFY_TOKEN, or "
-            f"configure PIPEFY_OAUTH_*. See {DOCS_CLI_AUTH_REF}."
+            f"configure PIPEFY_SERVICE_ACCOUNT_*. See {DOCS_CLI_AUTH_REF}."
         )
         return
 
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -16,6 +16,7 @@ from gql.transport.exceptions import (
     TransportServerError,
 )
 from pipefy_sdk import MePayload, PipefySettings
+from pipefy_sdk.settings import _LEGACY_ENV_KEYS_TO_NEW
 
 from pipefy_cli._docs import DOCS_CLI_AUTH_REF
 from pipefy_cli.auth import (
@@ -71,30 +72,18 @@ auth_app = typer.Typer(
     no_args_is_help=True,
 )
 
-_SERVICE_ACCOUNT_ENV_KEYS = (
-    "PIPEFY_SERVICE_ACCOUNT_URL",
-    "PIPEFY_SERVICE_ACCOUNT_CLIENT_ID",
-    "PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET",
-)
-_LEGACY_SERVICE_ACCOUNT_ENV_KEYS = (
-    "PIPEFY_OAUTH_URL",
-    "PIPEFY_OAUTH_CLIENT",
-    "PIPEFY_OAUTH_SECRET",
-)
+_SERVICE_ACCOUNT_ENV_KEYS = tuple(_LEGACY_ENV_KEYS_TO_NEW.values())
+_LEGACY_SERVICE_ACCOUNT_ENV_KEYS = tuple(_LEGACY_ENV_KEYS_TO_NEW.keys())
 
 
 def _session_masking_env_vars() -> list[str]:
-    """Env vars that outrank a stored session in the credential precedence chain.
+    """Env vars in ``os.environ`` that outrank a stored session.
 
-    Only ``os.environ`` is consulted — by the precedence model, ``.env`` defaults
-    sit below the stored session. A service-account triple is listed only when
-    *complete* (otherwise the client-credentials path wouldn't activate and the
-    warning would be misleading). Both the canonical
-    ``PIPEFY_SERVICE_ACCOUNT_*`` and the legacy ``PIPEFY_OAUTH_*`` triples are
-    recognised during the deprecation window — ``PipefySettings.AliasChoices``
-    resolves either, so the masking diagnostic has to match or it would
-    silently regress on the legacy form. The label echoes the form the user
-    actually set so they know which keys to unset.
+    A service-account triple counts only when *complete* (otherwise the
+    client-credentials path wouldn't activate and the warning would mislead).
+    Both the canonical and legacy forms are accepted during the deprecation
+    window, and the label echoes whichever the user actually set so they know
+    which keys to unset.
     """
     env_vars: list[str] = []
     if os.environ.get("PIPEFY_TOKEN"):
--- a/packages/cli/src/pipefy_cli/commands/auth.py
+++ b/packages/cli/src/pipefy_cli/commands/auth.py
@@ -76,21 +76,33 @@ _SERVICE_ACCOUNT_ENV_KEYS = (
     "PIPEFY_SERVICE_ACCOUNT_CLIENT_ID",
     "PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET",
 )
+_LEGACY_SERVICE_ACCOUNT_ENV_KEYS = (
+    "PIPEFY_OAUTH_URL",
+    "PIPEFY_OAUTH_CLIENT",
+    "PIPEFY_OAUTH_SECRET",
+)
 
 
 def _session_masking_env_vars() -> list[str]:
     """Env vars that outrank a stored session in the credential precedence chain.
 
     Only ``os.environ`` is consulted — by the precedence model, ``.env`` defaults
-    sit below the stored session. ``PIPEFY_SERVICE_ACCOUNT_*`` is listed only when
-    the *complete* triple is configured (otherwise the client-credentials path
-    wouldn't activate and the warning would be misleading).
+    sit below the stored session. A service-account triple is listed only when
+    *complete* (otherwise the client-credentials path wouldn't activate and the
+    warning would be misleading). Both the canonical
+    ``PIPEFY_SERVICE_ACCOUNT_*`` and the legacy ``PIPEFY_OAUTH_*`` triples are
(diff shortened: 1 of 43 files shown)


adriannoes (changes requested):
Solid rename with thoughtful back-compat (`AliasChoices`, TOML normalizer, deduped stderr warnings) and good test coverage — SDK/CLI/MCP boundaries stay clean, CI is green, and I ran the full unit suite locally on `f69efe7` (2131 tests passed across sdk/cli/mcp).

Two doc/diagnostic gaps should land before merge.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -260,7 +260,7 @@ def _render_status_text(report: AuthStatusReport) -> None:
     if not report.signed_in:
         typer.echo(
             "Not signed in. Run `pipefy auth login`, set PIPEFY_TOKEN, or "
-            f"configure PIPEFY_OAUTH_*. See {DOCS_CLI_AUTH_REF}."
+            f"configure PIPEFY_SERVICE_ACCOUNT_*. See {DOCS_CLI_AUTH_REF}."
```
adriannoes, line 264: This PR renames the service-account env vars everywhere except `docs/cli/auth.md`, which is still the doc linked from CLI errors via `DOCS_CLI_AUTH_REF`. The precedence table, quick-start snippet, env reference, headless guidance, and troubleshooting section there all still teach `PIPEFY_OAUTH_*` and quote error text that no longer matches `pipefy_cli/auth.py` / `commands/auth.py`.

I'd update `docs/cli/auth.md` in this PR — new names as primary, legacy names in a short migration note like `setup.md`, and troubleshooting strings that mirror the live CLI output.

packages/cli/src/pipefy_cli/commands/auth.py
```diff
@@ -72,25 +72,25 @@ def signed_in(self) -> bool:
...
+    sit below the stored session. ``PIPEFY_SERVICE_ACCOUNT_*`` is listed only when
+    the *complete* triple is configured (otherwise the client-credentials path
     wouldn't activate and the warning would be misleading).
     """
     env_vars: list[str] = []
     if os.environ.get("PIPEFY_TOKEN"):
         env_vars.append("PIPEFY_TOKEN")
     if all(os.environ.get(k) for k in _SERVICE_ACCOUNT_ENV_KEYS):
-        env_vars.append("PIPEFY_OAUTH_*")
+        env_vars.append("PIPEFY_SERVICE_ACCOUNT_*")
```
adriannoes: `_session_masking_env_vars()` now only looks for `PIPEFY_SERVICE_ACCOUNT_*` in `os.environ`, but during the alias window a full `PIPEFY_OAUTH_*` triple still wins over a stored session via `PipefySettings` + `AliasChoices`. Auth works, yet `masking_env_vars` stays empty and `_warn_if_masked()` never fires — the CI "why isn't my login being used?" case that `auth.md` still describes for `PIPEFY_OAUTH_*`.

Worth treating a resolved legacy triple (or the old env key names) the same as the new ones until the shim is removed. A test mirroring scenario 5 in `test_auth_status.py` with `PIPEFY_OAUTH_*` only would lock it in.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **resolution:** addressed in 5642d1a.

`docs/cli/auth.md` rewritten — new names primary in the precedence table, quick-start `.env` snippet, env reference, headless guide, status-command commentary, and the four troubleshooting subheadings (which now copy the live `pipefy_cli/auth.py` / `commands/auth.py` strings verbatim).

Legacy-names admonition mirrors the one in `docs/setup.md` (one-shot stderr warning, removal in a future beta, links to `docs/MIGRATION.md`). The disambiguation paragraph keeps a parenthetical noting the previous name and *why* the rename happened (`PIPEFY_OAUTH_URL` was one letter from `PIPEFY_AUTH_URL`) for anyone landing here from the deprecation warning. The masking-warning section also notes the warning fires identically for the legacy triple during the deprecation window, naming whichever form is set — coupled with the fix on the other thread.

packages/cli/src/pipefy_cli/commands/auth.py
gbrlcustodio: **resolution:** addressed in bed106e.

`_session_masking_env_vars()` now detects either complete triple (canonical `PIPEFY_SERVICE_ACCOUNT_*` or legacy `PIPEFY_OAUTH_*`) and labels the masking source by the form the user actually set — so the diagnostic names keys they can identify and unset, both in the post-login stderr warning and in the `masking_env_vars` JSON field of `auth status`.

New-wins precedence still lives in `PipefySettings` via `AliasChoices`; this is purely additive detection for the legacy form during the shim window.

Test: new scenario 5b in `test_auth_status.py` (`test_status_legacy_service_account_triple_masks_stored_session`) mirrors scenario 5 with only the `PIPEFY_OAUTH_*` triple set and pins `masking_env_vars == ["PIPEFY_OAUTH_*"]`.

gbrlcustodio:
**resolution:** both gaps addressed — bed106e (masking diagnostic) and 5642d1a (docs).

- `_session_masking_env_vars()` now detects either complete triple and labels by the form actually set, so `_warn_if_masked()` and the `masking_env_vars` JSON field stop silently regressing on the legacy form during the shim window. Scenario 5b in `test_auth_status.py` pins it.
- `docs/cli/auth.md` rewritten — precedence table, quick-start, env reference, headless guide, status commentary, and the four troubleshooting subheadings (verbatim from the live CLI strings). Legacy-names admonition mirrors `docs/setup.md`; disambiguation paragraph keeps a parenthetical explaining the rename's motivation (`PIPEFY_OAUTH_URL` was one letter from `PIPEFY_AUTH_URL`).

Local: `uv run ruff check packages/` + `--check` formatting clean; `uv run pytest packages/cli/` 247 passed, 13 deselected. Both inline threads resolved. Ready for a re-stamp.

adriannoes (approved):
Image

Issue #207, opened 2026-05-21
Portal CRUD 2.0 — Portal read (list_portals, get_portal)


adriannoes:
## Goal

Ship read-only portal discovery: **`list_portals`** and **`get_portal`** across SDK, MCP, and CLI.

## Context

- **Depends on:** #206 (Task 1.0) merged into `dev`
- **Enables:** all write flows (need org UUID / portal UUID)
- **Tools:** `list_portals`, `get_portal` | `pipefy portal list`, `pipefy portal get`
- **Branch:** `feat/portal-crud/2-portal-read` from `dev`
- **Suggested PR:** 2 → merges into `dev`

## Acceptance criteria

- [ ] One main portal per org documented in tool docstrings (mention `mainPortal` query option)
- [ ] `get_portal` selects `uuid, name, visibility, published, pages { … elements { uuid type metadata } }, subPortals`
- [ ] MCP tools use `readOnlyHint=True`
- [ ] CLI supports `--json`
- [ ] `docs/parity.md` rows for both tools
- [ ] Full TDD cycle: SDK → MCP → CLI (red before green at each layer)

## Sub-tasks (TDD order)

- [ ] **2.1 RED** — SDK tests: `list_portals(org_uuid, search_term=None)`, `get_portal(uuid)` (mock GraphQL)
- [ ] **2.2 GREEN** — SDK queries + service methods + facade
- [ ] **2.3 RED** — MCP tests for both read tools
- [ ] **2.4 GREEN** — Register MCP tools + helpers + `PIPEFY_TOOL_NAMES`
- [ ] **2.5 RED** — CLI tests for `portal list` and `portal get`
- [ ] **2.6 GREEN** — Implement CLI commands
- [ ] **2.7 REFACTOR** — Parity rows; optional integration read smoke

## Verify

```bash
uv run pytest packages/sdk/tests packages/mcp/tests packages/cli/tests -m "not integration" -k portal -v
```

adriannoes:
Closed on https://github.com/gbrlcustodio/pipefy-mcp-server/pull/214.

Issue #206, opened 2026-05-21
Portal CRUD 1.0 — SDK infrastructure (multi-endpoint + PortalService)


adriannoes:
## Goal

Add SDK foundation for portal GraphQL: `/graphql/interfaces` + reuse of `internal_api` for sub-portal wiring. Wire `PortalService` into `PipefyClient`.

## Context

- **Depends on:** nothing (first slice)
- **Enables:** tasks 2.0–6.0
- **Branch:** `feat/portal-crud/1-sdk-infrastructure` from `dev`
- **Suggested PR:** 1 → merges into `dev`

## Acceptance criteria

- [ ] `PIPEFY_INTERFACES_GRAPHQL_URL` / `interfaces_graphql_url` in settings (default `https://app.pipefy.com/graphql/interfaces`)
- [ ] `_validate_pipefy_endpoint_urls` extended to SSRF-validate the new URL (same path as `internal_api_url`)
- [ ] `PortalService` builds a `BasePipefyClient` against the Interfaces URL via `settings.model_copy(update={"graphql_url": settings.interfaces_graphql_url})` and accepts an injected `InternalApiClient`
- [ ] `PipefyClient` instantiates `PortalService` (no business methods yet)
- [ ] TDD: failing tests written before implementation; all pass after refactor

## Sub-tasks (TDD order)

- [ ] **1.1 RED** — `packages/sdk/tests/test_settings.py`: tests for `interfaces_graphql_url` default + env override + HTTPS rejection
- [ ] **1.2 RED** — `packages/sdk/tests/services/pipefy/test_portal_service.py`: routing tests (Interfaces base client vs injected `InternalApiClient`)
- [ ] **1.3 GREEN** — Implement settings field, extend SSRF validator, create `PortalService` shell, wire into `PipefyClient`
- [ ] **1.4 REFACTOR** — Exports + ruff; SDK portal tests green

## Verify

```bash
uv run pytest packages/sdk/tests -k "portal or interfaces_graphql_url" -v
uv run ruff check packages/sdk && uv run ruff format .
```

adriannoes:
Closed on https://github.com/gbrlcustodio/pipefy-mcp-server/pull/214.

PR #214, opened 2026-05-21
feat: portal read (list/get) with Interfaces schema and org id auto-resolve

adriannoes:
## Issues

This 1st PR of portal's CRUD [milestone](https://github.com/gbrlcustodio/pipefy-mcp-server/milestone/9; and closes:

- **#206** — Task **1.0** SDK infrastructure (`interfaces_graphql_url`, `PortalService` multi-endpoint foundation)
- **#207** — Task **2.0** Portal read (`list_portals`, `get_portal`) including sub-task **2.8** org identifier auto-resolve (numeric org id → UUID)

## Summary

- **SDK:** `PortalService` on `PIPEFY_INTERFACES_GRAPHQL_URL`; GraphQL aligned with live Interfaces schema (`InterfaceSubTypeFilter`, `id` fields, `pages[]` array); normalizes `uuid` alias from `id`; `list_portals` accepts org UUID or numeric org id (auto-resolve via public GraphQL, shared `looks_like_uuid` util).
- **MCP:** `list_portals`, `get_portal` read tools with structured `{ success, data }` / error envelopes (GraphQL transport errors included).
- **CLI:** `pipefy portal list --org-uuid`, `pipefy portal get <uuid>`; help golden updated.
- **Docs / CI:** `docs/parity.md` rows; `portal` registered in skills-lint CLI allowlist.

## Test plan

- [x] `uv run pytest -m "not integration"` — 2290 passed
- [x] `uv run ruff check packages/sdk/src packages/mcp/src packages/cli/src`
- [x] `uv run ruff format --check packages/sdk/src packages/mcp/src packages/cli/src`
- [x] `uv run pytest tests/test_parity.py`
- [x] `uv build packages/mcp`
- [x] Live smoke: MCP `list_portals(302398434)` + `get_portal`; CLI `portal list --org-uuid 302398434 --json`


packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -0,0 +1,75 @@
...
+    @staticmethod
+    def register(mcp: FastMCP, client: PipefyClient) -> None:
+        """Register portal-related tools on the MCP server."""
+
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=True),
+        )
+        async def list_portals(
+            ctx: Context[ServerSession, None],
+            org_uuid: str,
```
Danielmoraisg: **`org_uuid: str` diverges from `PipefyId` convention and the name contradicts the contract.**

Two issues land on this same line:

**(1a) Missing `int → str` coercion.** Every other MCP tool that accepts an organization identifier that may be a UUID *or* a numeric id types it as `PipefyId` and runs `validate_tool_id` at the boundary — see `observability_tools.get_agents_usage`, `get_automations_usage`, `get_ai_credit_usage`, and `get_organization`. `PipefyId` exists specifically because *"mcporter CLI infers unquoted numeric values as int, but Pipefy IDs are always strings"* (`packages/sdk/src/pipefy_sdk/models/validators.py`). If a client sends `list_portals(org_uuid=302398434)` unquoted — which is exactly the live-smoke example in the PR description — the SDK calls `.strip()` on an `int` inside `_resolve_organization_identifier_for_interfaces` and raises `AttributeError`. None of the new tests exercise int input.

**Suggested fix:**
```python
async def list_portals(
    ctx: Context[ServerSession, None],
    org_uuid: PipefyId,
    search_term: str | None = None,
) -> dict[str, Any]:
    org_uuid, err = validate_tool_id(org_uuid, "org_uuid")
    if err is not None:
        return err
    ...
```
Please add a unit test that calls the MCP tool with `{"org_uuid": 302398434}` (int, not str).

**(2e) Parameter naming.** The name `org_uuid` propagates through MCP → CLI (`--org-uuid`) → SDK while the docstring says *"Organization UUID, or numeric organization id (string)."* The name contradicts the contract and will mislead callers reading the tool schema. Observability tools settled on `organization_uuid` (still slightly imperfect, but at least matches the GraphQL variable spelling). Preferred names, in order:

1. `organization_identifier` — most truthful given the dual-acceptance behavior.
2. `organization_uuid` — matches observability's convention.

Worth fixing now, before this gets referenced in skills — the same-PR rule means a later rename is a coupled change across `skills/**`.

packages/sdk/src/pipefy_sdk/services/portal_service.py
```diff
@@ -0,0 +1,180 @@
...
+        Args:
+            organization_identifier: Organization UUID or numeric organization id (string).
+
+        Returns:
+            Organization UUID for Interfaces GraphQL variables.
+
+        Raises:
+            ValueError: When the identifier is empty, invalid, or resolution yields no uuid.
+        """
+        trimmed = organization_identifier.strip()
```
Danielmoraisg: **`.strip()` on a non-`str` caller raises `AttributeError`, not the documented `ValueError`.**

The signature is typed `organization_identifier: str` but there is no Pydantic coercion at the SDK boundary (unlike MCP tools, which can use `PipefyId`). A library caller that passes `int` here — plausible from notebook/scripting users — hits a confusing `AttributeError: 'int' object has no attribute 'strip'` instead of the documented `ValueError`. This is the same risk class flagged in comment 1a on the MCP layer.

Two options:

1. Defensive coerce at the top of the method:
   ```python
   if isinstance(organization_identifier, int):
       organization_identifier = str(organization_identifier)
   trimmed = organization_identifier.strip()
   ```
2. If you adopt comment 2a (shared `resolve_organization_uuid` helper in `pipefy_sdk/utils/organization_identifiers.py`), do the coercion there once and let both `ObservabilityService` and `PortalService` benefit — `ObservabilityService._resolve_organization_uuid` has the same issue today.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -0,0 +1,75 @@
...
+            except Exception as exc:  # noqa: BLE001
+                return build_error_payload(str(exc))
+            return build_success_payload({"portals": portals}, include_parsed=True)
+
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=True),
+        )
+        async def get_portal(
+            ctx: Context[ServerSession, None],
+            uuid: str,
```
Danielmoraisg: **Parameter `uuid` shadows the stdlib `uuid` module.**

The parameter name `uuid` shadows Python's stdlib `uuid` module within this function scope. There is no runtime bug today (none of `portal_tools.py`, `portal_service.py`, or `cli/commands/portal.py` import the `uuid` module), but the same `uuid: str` name leaks into:

- `PortalService.get_portal(self, uuid: str)` — `packages/sdk/src/pipefy_sdk/services/portal_service.py:166`
- `portal_get(... uuid: str = resource_id_argument(...))` — `packages/cli/src/pipefy_cli/commands/portal.py:48`

And `_resolve_organization_identifier_for_interfaces` in the same SDK file *does* go near `uuid.UUID` semantics via `looks_like_uuid`, so this is one careless edit away from a name collision. It is also out of step with the rest of the codebase, which uniformly names IDs `<resource>_id` / `<resource>_uuid`.

Suggest renaming all three (MCP arg, SDK arg, CLI positional) to `portal_uuid`. As a bonus, the GraphQL variable mapping reads more clearly:

```python
data = await self.execute_interfaces_query(GET_PORTAL_QUERY, {"uuid": portal_uuid})
```

packages/sdk/src/pipefy_sdk/services/portal_service.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Raises:
+            ValueError: When no internal API client was injected.
+        """
+        if self._internal_api_client is None:
+            msg = "Internal API client is not configured."
+            raise ValueError(msg)
+        return await self._internal_api_client.execute_query(query, variables)
+
+    async def _resolve_organization_identifier_for_interfaces(
```
Danielmoraisg: **`_resolve_organization_identifier_for_interfaces` is a near-verbatim copy of `ObservabilityService._resolve_organization_uuid`.**

The PR rightly extracted `looks_like_uuid` into `pipefy_sdk/utils/organization_identifiers.py` (✓), but the surrounding *resolve* helper is now copy-pasted in two places:

- `packages/sdk/src/pipefy_sdk/services/observability_service.py:98-132` (`_resolve_organization_uuid`)
- `packages/sdk/src/pipefy_sdk/services/portal_service.py:104-140` (`_resolve_organization_identifier_for_interfaces`)

Same algorithm, same error strings, same `RESOLVE_ORGANIZATION_UUID_QUERY`. A future change (e.g., a new org-id format, a different error message convention, or the int-coercion fix from comment 1b) will need to be applied twice and the next reviewer has to spot the drift.

**Suggested extraction** (in `pipefy_sdk/utils/organization_identifiers.py`):

```python
from collections.abc import Awaitable, Callable
from typing import Any

ExecuteQuery = Callable[[Any, dict[str, Any]], Awaitable[dict[str, Any]]]

async def resolve_organization_uuid(
    execute_query: ExecuteQuery,
    organization_identifier: str,
) -> str:
    """Resolve UUID-or-numeric organization identifier to a UUID via the public GraphQL endpoint."""
    ...
```

Then both services delegate. `PortalService` passes `self._graphql_client.execute_query`; `ObservabilityService` passes `self.execute_query` (since it inherits `BasePipefyClient`). Optional: bake the int-coercion fix from comment 1b into the helper.

packages/sdk/src/pipefy_sdk/services/portal_service.py
```diff
@@ -0,0 +1,180 @@
...
+        _with_uuid_alias(sub_portal) for sub_portal in portal.get("subPortals") or []
+    ]
+    return {
+        **_with_uuid_alias(portal),
+        "pages": normalized_pages,
+        "subPortals": sub_portals,
+    }
+
+
+class PortalService:
```
Danielmoraisg, line 41: **Question: was the composition-over-inheritance choice deliberate?**

Every other service under `packages/sdk/src/pipefy_sdk/services/` does `class FooService(BasePipefyClient):` and calls `self.execute_query`. `PortalService` is the only one that *composes* two `BasePipefyClient` instances (`_interfaces_client` + `_graphql_client`) instead of inheriting. The two-endpoints motivation makes sense — Interfaces lives on `/graphql/interfaces`, the public schema on `/graphql` — but the trade-offs aren't currently documented:

- The `gql_reuse_fetched_graphql_schema` cache is now *per composed client*, not *per logical service*. Each transport does its own introspection cycle when that flag is enabled.
- If anyone ever constructs `PortalService(settings=..., auth=None)` outside the `PipefyClient` facade (e.g., a future test, a one-off script), each `BasePipefyClient` will build its **own** `OAuth2ClientCredentials` independently from `settings.oauth_*`, resulting in two token caches per service instance. In production this never happens because `PipefyClient` always passes a shared `auth=` (✓ `client.py:133`), but it's a sharp edge for future direct users.

**Was this an explicit design call, or just the simplest path?** Two ways forward:

1. *(Light)* Add a sentence to the class docstring — something like "Composes two `BasePipefyClient` transports rather than inheriting because the Interfaces schema lives on a different host than the public GraphQL endpoint. When constructed outside `PipefyClient`, pass a pre-built `auth` to share one token cache between the transports."
2. *(Heavier, separate refactor)* Generalize `BasePipefyClient` to optionally hold multiple named endpoints, so all services can keep inheriting. Probably overkill if Interfaces is the only multi-endpoint case for the foreseeable future.

(1) is sufficient for this PR; (2) is a separate ticket if the milestone adds more cross-endpoint services.

packages/sdk/src/pipefy_sdk/utils/field_tokens.py
```diff
@@ -2,16 +2,9 @@
...
-def looks_like_uuid_token(token: str) -> bool:
-    """True when ``token`` parses as a UUID (Pipefy often uses uuid for field ``id``)."""
-    try:
-        uuid_mod.UUID(str(token).strip())
-    except ValueError:
-        return False
-    return True
+from pipefy_sdk.utils.organization_identifiers import (
+    looks_like_uuid as looks_like_uuid_token,
+)
```
Danielmoraisg: ** `looks_like_uuid_token` is now *literally* an alias for `looks_like_uuid` — two public names for one function.**

After this refactor:

```python
from pipefy_sdk.utils.organization_identifiers import (
    looks_like_uuid as looks_like_uuid_token,
)
```

…and both names are re-exported from `pipefy_sdk/utils/__init__.py` (lines 14 + 27). Two public names for the same function is needless API surface and confuses future maintainers about which is canonical.

Two cleaner options:

1. **Drop the alias.** Replace the two existing call sites of `looks_like_uuid_token` (`slug_like_field_token` here on line 13, plus any others) with `looks_like_uuid` and remove the alias + the `looks_like_uuid_token` export from `utils/__init__.py`. Cleanest.
2. **Keep but deprecate.** Add `# Deprecated: use looks_like_uuid` next to the alias and the `__all__` entry. Useful only if there are external consumers via `pipefy_sdk.utils.looks_like_uuid_token` we need to grace-period.

Given the SDK is still on `0.2.0-beta.1`, option 1 is preferable and a 2-line change.

packages/sdk/src/pipefy_sdk/queries/portal_queries.py
```diff
@@ -0,0 +1,62 @@
...
+            name
+            visibility
+            published
+            pages {
+                id
+                title
+                elements {
+                    id
+                    type
+                    metadata
```
Danielmoraisg, line 46: **`pages[].elements[].metadata` is returned as a raw `JSON` scalar with no documented shape.**

Agents calling `get_portal` blind will not know whether to expect `{ formId }`, `{ pipeId }`, `{ url }`, or something else — the GraphQL selection just says `metadata` with no sub-fields, and the MCP tool docstring in `portal_tools.py` doesn't enumerate the variants either. The MCP description is what the LLM steers off; that's where the contract lives.

The PR's test fixture only exercises `{"formId": "123"}`, which is not enough surface for an agent to reason about real responses.

Two options (either, or both):

1. **Narrow the GraphQL selection.** If the Interfaces schema has typed metadata variants per element `type` (e.g., a union or interface), select the concrete fields. Best for typed clients.
2. **Document the observed shapes in the `get_portal` MCP tool docstring** in `packages/mcp/src/pipefy_mcp/tools/portal_tools.py`, e.g.:
   ```
   Element metadata shapes by `type`:
     - 'forms'    → { formId: str }
     - 'pipe'     → { pipeId: str }
     - 'link'     → { url: str, label?: str }
     - ...
   ```
   Even a non-exhaustive list with a "see Pipefy docs for additional types" hint is much more actionable than a bare `JSON` blob.

packages/sdk/src/pipefy_sdk/settings.py
```diff
@@ -27,6 +27,11 @@ class PipefySettings(BaseModel):
         description="Internal API URL for AI Automation endpoints",
     )
 
+    interfaces_graphql_url: str = Field(
+        default="https://app.pipefy.com/graphql/interfaces",
+        description="GraphQL URL for Pipefy Interfaces schema (portals, pages, elements)",
+    )
```
Danielmoraisg, line 68: **The new `PIPEFY_INTERFACES_GRAPHQL_URL` env var is undocumented outside of code.**

This setting is functionally a sibling of `PIPEFY_INTERNAL_API_URL`, but it's missing from every place operators look:

1. **`docs/setup.md`** — the env-table around line 63-66 documents `PIPEFY_GRAPHQL_URL`, `PIPEFY_INTERNAL_API_URL`, `PIPEFY_OAUTH_URL` but not `PIPEFY_INTERFACES_GRAPHQL_URL`.
2. **MCP `env` shape examples** in `docs/setup.md` (lines ~92-95, ~119-122, ~155-158) list `PIPEFY_INTERNAL_API_URL` but not this one.
3. **`.env.example`** — has commented entries for every other optional knob but not this one.

Operators running against a self-hosted or proxied Pipefy will hit this on their first portal tool call with no breadcrumb.

**Suggested changes (in this PR or a quick follow-up):**

1. Add a row to the env table in `docs/setup.md`:
   ```md
   | `PIPEFY_INTERFACES_GRAPHQL_URL` | GraphQL endpoint for the Interfaces schema (portals, pages, elements). Optional; defaults to `https://app.pipefy.com/graphql/interfaces`. |
   ```
2. Add `PIPEFY_INTERFACES_GRAPHQL_URL` to the three MCP `env` shape examples in `docs/setup.md`, next to `PIPEFY_INTERNAL_API_URL`.
3. Append to `.env.example`:
   ```bash
   # Optional: GraphQL endpoint for the Interfaces schema (portals, pages, elements). Default below.
   # PIPEFY_INTERFACES_GRAPHQL_URL=https://app.pipefy.com/graphql/interfaces
   ```

Per `AGENTS.md`: *"PRs must include: summary, testing performed (commands + results), docs updates if tool behavior or config changed."* A new env var is a config change.

docs/parity.md
```diff
@@ -109,6 +109,7 @@ For **database records**, `find_records` result nodes may use **`fields`** while
 | `get_pipe_report_export` | (poll via `pipefy report-pipe export --format json`) | shipped | Reports + exports. |
 | `get_pipe_report_filterable_fields` | `pipefy report-pipe filterable-fields` | shipped | Reports domain. |
 | `get_pipe_reports` | `pipefy report-pipe list` | shipped | Reports domain. |
+| `get_portal` | `pipefy portal get` | shipped | Portal CRUD task **2.0**; Interfaces schema; read-only; includes `published`. |
```
Danielmoraisg, line 112: **No per-area MCP tool reference exists for portals under `docs/mcp/tools/`.**

`AGENTS.md` documentation map lists `docs/mcp/tools/` as the canonical *"per-area MCP tool reference (parameters, edge cases, cross-cutting behavior)."* Other read-only domains (organization, observability, etc.) already have entries there. This is where durable contract docs belong:

- "At most one main portal per org" semantics.
- That `list_portals` returns a flat list (not Relay edges) and *does not* include `published` (the parity row notes the latter).
- That `get_portal` *does* include `published`, plus the page/element/sub-portal shape.
- The Interfaces schema endpoint distinction — explains why `PIPEFY_INTERFACES_GRAPHQL_URL` exists.
- Element metadata shapes (see comment 3b on `portal_queries.py`).
- Numeric-org-id auto-resolve behavior (`list_portals` only; `get_portal` does not need it).

Not strictly required by parity.md itself, but it is the documented pattern in the repo guide. Suggest a `docs/mcp/tools/portal.md` either in this PR or as the next sub-task of the Portal CRUD milestone — ideally before more CRUD verbs (create/update/delete) ship, so the contract page grows once instead of being retrofitted later.

.github/workflows/scripts/lint_skill_refs.py
```diff
@@ -30,6 +30,7 @@
         "org",
         "phase",
         "pipe",
+        "portal",
```
Danielmoraisg, line 33: **Skills update deferred; flagging for the milestone closeout.**

`AGENTS.md` same-PR rule: *"breaking command renames must update affected skills in the same PR (or a paired PR opened in the same review window). CI (`skills-lint.yml`) validates starter-pack bundle drift, MCP tool names, and `pipefy` CLI subcommands referenced in `skills/**/SKILL.md`."*

A quick scan of `skills/**` finds zero mentions of `portal`, `list_portals`, or `get_portal` (only `skills/api-troubleshoot/pipefy-api-fallback/SKILL.md` references the developer-portal URL, which is unrelated). So this PR is acceptable as a first-of-its-kind capability — there are no affected skills *yet* to update, and allowlisting `portal` here (✓) unblocks the CLI sub-app surface for future skill authors.

**Please track for the milestone:** the Portal CRUD milestone should land a `skills/portal-management/SKILL.md` (or similar) before "shipped" status in the parity table is meaningful to agents — today an agent has no way to discover `list_portals` / `get_portal` without trawling the parity matrix. Ideal scope for that skill: when to call `list_portals` vs `get_portal`, the org-uuid-or-numeric input contract, the "at most one main portal" guarantee, and the sub-portal vs main-portal distinction. Worth scheduling alongside or immediately after task 2.x of the milestone.

Danielmoraisg (approved):

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Good catch — you're right that `str` without `PipefyId` breaks on unquoted numeric MCP input (exactly our smoke case). Fixed in this push: `organization_uuid: PipefyId` + `validate_tool_id` at the tool boundary, plus a unit test with `{"organization_uuid": 302398434}` (int).

On naming: we renamed to `organization_uuid` to stay consistent with the observability tools (`get_agents_usage`, `get_ai_credit_usage`, etc.) rather than `organization_identifier`. The docstring and CLI flag (`--organization-uuid`) now match that dual UUID/numeric contract. Happy to revisit if you'd prefer the more explicit name across the board in a follow-up.

packages/sdk/src/pipefy_sdk/services/portal_service.py
adriannoes: Agreed — library callers passing `int` would hit `AttributeError` instead of the documented `ValueError`. We extracted a shared `resolve_organization_uuid` helper in `pipefy_sdk/utils/organization_identifiers.py` that coerces `int → str` once before `.strip()`. Both `PortalService` and `ObservabilityService` now delegate to it.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Fair point. Renamed the public surface to `portal_uuid` across MCP, SDK, and CLI (`pipefy portal get <portal_uuid>`). The GraphQL variable is still `uuid` internally — only the caller-facing names changed.

packages/sdk/src/pipefy_sdk/services/portal_service.py
adriannoes: Done — `resolve_organization_uuid` now lives in `organization_identifiers.py` with the int-coercion fix baked in. `PortalService` passes `self._graphql_client.execute_query`; `ObservabilityService` passes `self.execute_query`. Removed both private copies.

packages/sdk/src/pipefy_sdk/services/portal_service.py
adriannoes: Deliberate choice: Interfaces and public GraphQL are different endpoints, so we compose two `BasePipefyClient` transports rather than inherit one URL. You're right the trade-offs weren't documented — added a class docstring note about the two-transport layout and passing shared `auth` when constructing outside `PipefyClient`. A generalized multi-endpoint `BasePipefyClient` feels like overkill while Interfaces is the only cross-endpoint case; happy to ticket that if more services appear in the milestone.

packages/sdk/src/pipefy_sdk/utils/field_tokens.py
adriannoes: Agreed — redundant public surface. Dropped the alias; `field_tokens.py` and `utils/__init__.py` now use/export only `looks_like_uuid`.

packages/sdk/src/pipefy_sdk/queries/portal_queries.py
adriannoes: Added observed element metadata shapes to the `get_portal` MCP docstring (forms → `{formId}`, pipe → `{pipeId}`, link → `{url, label?}`, plus a note that additional types may appear). Same summary is in the new `docs/mcp/tools/portal.md`. We kept the GraphQL selection as the raw `JSON` scalar since the Interfaces schema doesn't expose typed sub-fields per element type in our current introspection.

packages/sdk/src/pipefy_sdk/settings.py
adriannoes: Added to `docs/setup.md` (env table + MCP client JSON examples), `.env.example`, and cross-referenced from the new portal tool doc. Thanks for flagging — config change should have shipped with the setting.

docs/parity.md
adriannoes: Added `docs/mcp/tools/portal.md` covering Interfaces endpoint, org UUID/numeric auto-resolve on `list_portals`, main-vs-sub-portal semantics, list vs get field differences (`published`), and element metadata notes. Aiming to grow this page as CRUD verbs land rather than retrofit later.

.github/workflows/scripts/lint_skill_refs.py
adriannoes: Acknowledged — no skills reference portal tools yet, so nothing to update in this PR. We'll track `skills/portal-management/SKILL.md` in the Portal CRUD milestone before the write verbs ship. The `portal` CLI allowlist entry stays as the hook for future skill authors.

adriannoes:
## Review feedback addressed

Pushed 4 commits addressing Daniel's review threads (all resolved):

1. **`refactor(sdk)`** — shared `resolve_organization_uuid` helper (int coercion), `organization_uuid` / `portal_uuid` renames, PortalService composition docstring, dropped `looks_like_uuid_token` alias
2. **`fix(mcp)`** — `PipefyId` + `validate_tool_id` on `list_portals`, metadata docstring on `get_portal`, int org-id test
3. **`feat(cli)`** — `--organization-uuid` flag rename + help golden
4. **`docs`** — `docs/mcp/tools/portal.md`, `PIPEFY_INTERFACES_GRAPHQL_URL` in setup/`.env.example`

**Verification:** `uv run pytest -m "not integration"` (2299 passed), `test_parity.py`, `lint_skill_refs.py`, ruff check/format.

Skills update (thread 10) deferred to a later Portal CRUD milestone task as agreed.

Ready for another look when you have a moment — thanks!

Issue #213, opened 2026-05-21
Extend stored OAuth user session to the MCP server runtime


gbrlcustodio:
## Problem

The CLI's user-auth flow (PRs #125, #131, #142, #144) lets a user run `pipefy auth login` once and have every subsequent CLI command pick up the session from the OS keychain. The MCP server has no equivalent path — it still requires either `PIPEFY_TOKEN` or a `PIPEFY_SERVICE_ACCOUNT_*` triple at startup (`packages/mcp/src/pipefy_mcp/core/container.py:23-50`).

For local stdio deployments — the MCP server launched per-user by an MCP client (Claude Desktop, etc.) — this is a missed reuse: the same user with the same keychain entry should not also need to export service-account env vars to use the MCP tools.

## Non-goals

Hosted / remote MCP server with its own OAuth gate (`mcp.server.auth.*`, MCP-spec PRM/OASM discovery, DCR for arbitrary clients). The MCP authorization spec ([docs](https://modelcontextprotocol.io/docs/tutorials/security/authorization)) explicitly scopes OAuth to HTTP transports — stdio servers are directed to use "credentials provided by third-party libraries embedded directly in the MCP server" (i.e. exactly the keychain-reuse path this issue proposes).

## Implementation plan (4 sequential PRs)

Each PR is implemented, reviewed, and merged before the next starts. No parallel branches off `dev`.

### PR-A — Extract `oauth/` into a shared package
Move `packages/cli/src/pipefy_cli/oauth/` into a new workspace member `packages/auth/` (`name = pipefy-auth`, module `pipefy_auth`). Mechanical refactor only — CLI behavior unchanged. Also moves `OidcClient` and `DEFAULT_AUTH_CLIENT_ID` into `pipefy_auth.identity`, and renames the keychain service from `pipefy-cli` to `pipefy` (clean cut; early adopters re-run `pipefy auth login`).

### PR-B — Auth-layer consolidation + resolver + #134 wire-up
Make `pipefy-auth` the single source of truth for authentication:
- New `pipefy_auth.bearer`: hosts `StaticBearerAuth` (moved from SDK) + new `CallableBearerAuth` (sync + async, with `asyncio.Lock` to dedup in-process refresh contention).
[...]

PR #216, opened 2026-05-22
refactor: extract pipefy_cli.oauth into shared pipefy-auth package

### LINKED ISSUE #133: Concurrent `pipefy` invocations race the refresh-token grant
## Problem

When two ``pipefy <cmd>`` processes start at roughly the same time and both decide the stored access token is within the leeway window, they both POST the **same** ``refresh_token`` to the IdP. Keycloak (and most IdPs) rotate the refresh token on each successful grant — so one process gets fresh tokens, and the other gets ``invalid_grant`` because the refresh token it sent is no longer valid by the time the request lands.

Today the second process surfaces ``RefreshError("Refresh failed (400): invalid_grant")`` and exits 2 with **"Run \`pipefy auth login\` to sign in again."** — but the user's session is actually fine; they just hit a race.

## When this matters

Sequential CLI use: **never**. The window is ≤ leeway (60 s by default) and only opens when the access token is expiring.

Real usage that triggers this:

- ``xargs -P 4 pipefy card get`` over a list of card ids
- ``parallel pipefy ...``
- shell scripts that `&` multiple ``pipefy`` invocations
- IDE integrations / Cursor MCP that spawn multiple commands during a workflow

### Option A — Filesystem mutex during refresh

A short-lived ``flock`` (Unix) / Windows-equivalent over a file in the user's config dir (``~/.config/pipefy/refresh.lock`` or similar). First refresher wins; others block briefly and then re-load the keychain (now containing fresh tokens) instead of refreshing themselves.

**Pros:** the simplest correct fix; matches how ``aws sso login`` and ``gh auth`` handle the same race.
**Cons:** new filesystem dependency (must work on macOS, Linux, Windows); needs careful timeout handling so a stuck process doesn't block the whole user's CLI.

### Option B — Retry-once on ``invalid_grant``

When ``refresh_access_token`` returns ``invalid_grant``, sleep briefly and re-load the keychain. If the access token has changed (because the other process refreshed), use it. If unchanged, the user really is logged out.

**Pros:** no new infra; one extra keychain read on failure.
**Cons:** doesn't prevent the wasted refresh round-trip; relies on the IdP using ``invalid_grant`` consistently for this case (it does in Keycloak, but not universally guaranteed).

### Option C — Reactive refresh via custom httpx auth class

The deferred reactive-refresh-on-401 follow-up subsumes this: both processes would see the access token, one would 401, and the auth class would refresh-and-retry. The other process would 401 next and lose the race the same way as today.

**Pros:** no extra work if we're already going to build this.
**Cons:** doesn't actually fix the race — just moves it from the refresh path to the API call path.

## Recommendation

**Option A in a follow-up PR**, after the reactive-refresh slice. The filesystem mutex is the right shape for the threat: it's the same problem AWS SSO and gh CLI solved this way.

Until then, **Option B as a thin retry layer** would noticeably improve the UX for the (rare but real) scripted-parallel case at minimal cost.

## Acceptance criteria

- [ ] Two concurrent ``pipefy <cmd>`` invocations within the leeway window do not both surface ``RefreshError``; at most one performs the refresh and the others adopt its result.
- [ ] The mutex / retry path is observable in tests (e.g., a test that fires two ``ensure_fresh_session`` calls in parallel and asserts only one network round-trip).
- [ ] No regression in single-process CLI use — refresh latency unchanged in the common case.

## References

- PR #131 (introduces the eager-refresh path that races)

### LINKED ISSUE #134: Wire `InternalApiClient` over bearer-token credentials (unlock AI-automation on user auth)
## Problem

``InternalApiClient`` requires the OAuth client-credentials triple (``oauth_url`` + ``oauth_client`` + ``oauth_secret``) at construction:

```python
# packages/sdk/src/pipefy_sdk/services/internal_api_client.py
class InternalApiClient:
    def __init__(self, url, oauth_url, oauth_client, oauth_secret, *, allow_insecure_urls=False):
        ...
        self._auth = OAuth2ClientCredentials(token_url=oauth_url, client_id=oauth_client, client_secret=oauth_secret)
```

So the CLI only builds it on the **client-credentials path** in ``auth.py`` — the bearer branches (``--token``, ``PIPEFY_TOKEN``, the new stored user session from #131) get a ``PipefyClient`` without an attached ``InternalApiClient``. Any feature routed through ``AiAutomationService(client=internal_client)`` is silently unavailable when the CLI authenticates via a bearer.

Users hit this as:

- "I ran ``pipefy auth login`` and now AI automation commands return errors."
- "``PIPEFY_TOKEN`` works for everything except AI automation."

The inconsistency was inherited from the pre-#125 era (when ``PIPEFY_TOKEN`` was the only bearer path); PR #131 widened the user-auth slot to the same shape but didn't lift the limitation.

## Proposed change

**SDK** (``pipefy-sdk``):

- Add a bearer-token alternative for ``InternalApiClient``'s auth surface. Two reasonable shapes:
  - **Option A:** a second classmethod constructor, e.g. ``InternalApiClient.from_bearer_token(url, bearer_token, *, allow_insecure_urls=False)`` that swaps the ``OAuth2ClientCredentials`` for a ``StaticBearerAuth`` (mirroring ``PipefyClient``'s existing pattern at ``base_client.py:15-23``).
  - **Option B:** factor the auth surface into a protocol (e.g. ``InternalApiAuth``) and have ``InternalApiClient`` accept an instance. Two implementations: ``OAuth2ClientCredentialsAuth`` (today's path) and ``StaticBearerAuth`` (new).
- Mark the old ``__init__(url, oauth_url, oauth_client, oauth_secret, …)`` shape as deprecated only if Option B is taken; Option A keeps it as the primary constructor.

**CLI** (``pipefy-cli``):

- In ``auth.py::get_authenticated_client``, build an ``InternalApiClient`` over the bearer in the priorities 1 / 2 / 4 branches the same way the client-credentials branch already does.
- For the stored-session path specifically (priority 4 from PR #131), the bearer is ``session.access_token``. The refresh path already invalidates the in-process cache when the access token rotates, so the new ``InternalApiClient`` rebuilds correctly.

## Refresh consideration (user-auth path)

The user-auth flow uses **eager refresh**: ``ensure_fresh_session`` rotates the access token before each command. As long as ``InternalApiClient`` reads the bearer at the same point in the flow, this works without changes.

If the SDK later grows reactive refresh-on-401 (deferred follow-up), ``InternalApiClient``'s bearer would need to be re-readable from the keychain rather than captured by value at construction — but that's the same problem ``PipefyClient`` will face, and the solution can be shared.

## Out of scope

- Reactive 401 refresh in ``InternalApiClient`` (deferred with the analogous ``PipefyClient`` work).
- Reworking the ``oauth_*`` env-var naming (tracked in #127).

## Acceptance criteria

- [ ] ``InternalApiClient`` accepts a bearer token (Option A or B), without breaking existing client-credentials callers.
- [ ] ``pipefy_cli.auth.get_authenticated_client`` builds an ``InternalApiClient`` in the bearer-token branches (``--token``, ``PIPEFY_TOKEN``, stored session).
- [ ] ``AiAutomationService(client=internal_client)`` is wired in all four precedence tiers.
- [ ] AI-automation CLI commands succeed end-to-end against piporacle when authenticated via ``pipefy auth login`` (no ``PIPEFY_OAUTH_*`` configured).
- [ ] SDK tests cover the bearer constructor; CLI tests cover that the bearer branches build a fully-wired ``PipefyClient``.

## References

- PR #131 (introduces the user-auth bearer path that surfaces this limitation)
- PR #125 (introduces ``pipefy auth login`` — the first user-facing way to authenticate without ``PIPEFY_OAUTH_*``)
- ``packages/sdk/src/pipefy_sdk/services/internal_api_client.py:26-54`` (the constructor that needs the bearer alternative)
- ``packages/cli/src/pipefy_cli/auth.py:123-148`` (where the bearer branches skip ``InternalApiClient`` construction today)

gbrlcustodio:
First of four PRs implementing #213. The sequence and rationale are tracked in that issue.

## What

Lifts `packages/cli/src/pipefy_cli/oauth/` into a new workspace member `packages/auth/` (importable as `pipefy_auth`). Mechanical refactor — CLI behavior is unchanged.

## Why

The MCP server needs to read the same stored OAuth session the CLI populates. Today the keychain helpers live inside `pipefy_cli`, which the MCP package cannot import. This PR turns them into a shared workspace member so the remaining work for #213 can land on top.

## Notable choices

- **Leaf package.** `pipefy-auth` depends only on `httpx` + `keyring`. A ruff ban (`TID251`) enforces "no imports from `pipefy_sdk` / `pipefy_cli` / `pipefy_mcp`".
- **One inline-copied utility.** `pipefy_auth.discovery` needs `validate_https_service_endpoint_url`, which lives in `pipefy_sdk.utils.url_ssrf`. To keep `pipefy-auth` a leaf, the function (and its small helpers) is inline-copied into `pipefy_auth._url_ssrf`. The duplication is documented in the file header as a candidate for a future shared-utils package.
- **Keychain service rename.** `_SERVICE = "pipefy-cli"` → `"pipefy"` so the entry is consumer-neutral (the MCP server will read the same one). Clean cut, no migration code: anyone who'd logged in pre-rename (early adopters of #125 / #131) just re-runs `pipefy auth login`. The user-auth feature is recent enough that the install base is small.
- **`OidcClient` and `DEFAULT_AUTH_CLIENT_ID`** move into `pipefy_auth.identity`. The value of `DEFAULT_AUTH_CLIENT_ID` stays `"pipefy-cli"` because it's the registered Keycloak public client_id — refresh tokens are scoped to the issuing client_id, so renaming it would invalidate every existing session.
- **Tests.** Only `test_auth_refresh.py` was purely oauth-internal — moved to `packages/auth/tests/test_refresh.py` with a minimal `conftest.py` (just `InMemoryKeyring` + `fake_keyring`). Other CLI auth tests (`test_auth.py`, `test_auth_login.py`, `test_auth_logout.py`, `test_auth_status.py`) test Typer wiring and stay in `packages/cli/tests/` with rewritten imports.
- **Version lockstep + release wiring.** `pipefy-auth` is added to the CI lockstep check (`.github/workflows/ci.yml`), to `scripts/bump_version.py`, and to the release wheel verification + PyPI staging step (`.github/workflows/release.yml`). Without the last one, `pip install pipefy-cli` from PyPI would fail to resolve the new `pipefy-auth` runtime dep.

## Verification

- `uv run pytest`: 2315 passed, 24 skipped (unchanged from `dev`).
- `uv run ruff check packages/`: clean.
- Manual smoke: `pipefy auth login` → `pipefy auth status` → `pipefy auth logout` round-trips against the new `pipefy`-named keychain entry.

## What's NOT in this PR

The remaining work for #213 — single-`auth` SDK constructors, the shared precedence-chain resolver (closes #134 as a side effect), the cross-process refresh lock (closes #133), and the MCP container wiring — will land in follow-up PRs against the same issue.


.github/workflows/ci.yml
```diff
@@ -67,6 +67,7 @@ jobs:
             "sdk": Path("packages/sdk/src/pipefy_sdk/__init__.py"),
             "mcp": Path("packages/mcp/src/pipefy_mcp/__init__.py"),
             "cli": Path("packages/cli/src/pipefy_cli/__init__.py"),
+            "auth": Path("packages/auth/src/pipefy_auth/__init__.py"),
```
adriannoes, line 70: Good addition for version lockstep. While you're touching this workflow, the "Lint (full tree)" / format steps a few lines below still only cover sdk/mcp/cli — not `packages/auth/src`. This PR adds the leaf-package `TID251` ban in `packages/auth/pyproject.toml`, but that rule only runs when someone lints auth locally; a `pipefy_sdk` import under auth wouldn't fail CI today.

I'd add `packages/auth/src` to those ruff steps (or a dedicated `cd packages/auth && uv run ruff check src` job, mirroring SDK). You'll know it's covered when CI fails on a deliberate banned import in auth.

adriannoes (approved):
## Summary

Solid mechanical extraction — `pipefy-auth` as a leaf package, CLI rewired cleanly, release/lockstep includes the fourth wheel, and the full unit suite passes. Review included a local checkout + pytest run on the PR head.

## Before merge (non-blocking)

**F1 — CI lint coverage for auth:** The new `TID251` banned-import rules in `packages/auth/pyproject.toml` aren't exercised by the full-tree lint step (still sdk/mcp/cli only). Please extend the workflow while you're already editing `ci.yml` for auth lockstep — inline comment on the workflow file.

## Using this review with AI

Structured findings live in `.cursor/dev-planning/code-review/pr-216-extract-auth-package.md` (repo-local review artifact).

## Verification

```bash
uv run ruff check packages/auth/src
uv run pytest packages/auth/tests packages/cli/tests/test_auth*.py -m "not integration" -q
```

.github/workflows/ci.yml
gbrlcustodio: **resolution:** addressed in #217 via 52b2444 — `packages/auth/src` is now included in the workspace-wide `ruff check`/`ruff format --check` steps, plus a dedicated "Lint auth (banned upward imports)" job mirroring the SDK one so the leaf-package `TID251` ban actually runs in CI.

PR #217, opened 2026-05-22
refactor: consolidate auth precedence chain in pipefy_auth

### LINKED ISSUE #134: Wire `InternalApiClient` over bearer-token credentials (unlock AI-automation on user auth)

gbrlcustodio:
Stacked on #216. Targets that branch directly so the diff stays focused on this step; will be retargeted to `dev` once #216 merges. Part of #213's resolver/settings work; closes #134 as a side effect.

## Summary

- `pipefy_auth.resolver` owns a fixed three-tier precedence chain: `static-token` → `service-account` → `stored-session`. `resolve_pipefy_auth(...)` short-circuits at the first tier that resolves and returns `httpx.Auth | None` directly — the tier identity is carried by the concrete subclass (`StaticBearerAuth`, `OAuth2ClientCredentials`, `CallableBearerAuth`). Use `tier_for(auth)` to recover the tier name when needed (e.g. for the `pipefy auth status` wire schema). `detect_pipefy_tiers(...)` is the diagnostics-only enumerator. No open Protocol, no policy class hierarchy.
- `pipefy_auth.bearer` exports `StaticBearerAuth` (moved out of the SDK) and a new `CallableBearerAuth` that resolves a fresh token per request, with an `asyncio.Lock` to dedup in-process refresh contention.
- SDK auth is now explicit. `PipefyClient(settings, *, auth)`, `InternalApiClient(url, *, auth, allow_insecure_urls)`, and `BasePipefyClient(settings, *, auth)` all require an `httpx.Auth`. `httpx-auth` moves from `pipefy-sdk`'s dependencies to `pipefy-auth`'s.
- CLI's `get_authenticated_client` becomes a thin wrapper around `pipefy_auth.resolve_pipefy_auth`. The bearer-token path now wires `InternalApiClient` whenever `PIPEFY_INTERNAL_API_URL` is set, closing #134 (AI-automation tools were silently unavailable on the bearer path).
- MCP container uses the same resolver. `PIPEFY_TOKEN` is wired as the static-token tier ahead of the service-account triple, matching CLI precedence by construction. The flag-vs-env distinction lives entirely in CLI display code.
- CLI in-process client cache: the cache key is now a SHA-256 digest over `model_dump(mode="json")` of the settings + `asdict(auth)` + tier name. Adding a new field anywhere in the inputs participates in the key by construction (no more hand-picked tuple to maintain), and the bearer/secret never linger in module state past the digest computation.
- `docs/cli/auth.md` refreshed: the "Limitations on the bearer path" section is gone (every tier wires `InternalApiClient` now), the precedence-table notes are updated, and the troubleshooting quote matches the new `missing_auth_message()` output.

## Naming

The resolver uses *tier* (a precedence slot) — `STATIC_TOKEN_TIER`, `SERVICE_ACCOUNT_TIER`, `STORED_SESSION_TIER`. *Source* was the wrong word at that layer: a source is the *surface* a value came from (flag, env var, config file). The CLI keeps `FLAG_TOKEN_SOURCE`/`ENV_TOKEN_SOURCE` for its display layer, where "source" is correct, and internal variables that hold tier values are named `tier` (not `source`). The wire-schema string values (`"static-token"`, `"service-account"`, `"stored-session"`) are unchanged.

## Test plan

- [x] `uv run pytest -m "not integration"` — 2312 passing locally.
- [x] `uv run ruff check` + `uv run ruff format --check` — clean.
- [x] CI green.
- [x] `pipefy auth status` reports the correct `flag-token` vs `env-token` source on a sample invocation. Manual smoke confirmed flag wins over env when both are set.
- [x] Bearer path wires `InternalApiClient` (regression for #134). Manual smoke confirmed `internal_api_available = True` and `ai_automation_available = True` when `PIPEFY_TOKEN` + `PIPEFY_INTERNAL_API_URL` are set.
- [x] MCP container boots on both auth paths (`PIPEFY_SERVICE_ACCOUNT_*` and `PIPEFY_TOKEN`); raises with the canonical `missing_auth_message()` when neither is configured.

## Related work

- #218 — settings split (`PipefySettings` endpoint-only + new `pipefy_auth.AuthSettings`). Stacked on this branch; review the two together when ready.

## Follow-ups against #213

- Picking up the stored OAuth user session in the MCP container (with eager refresh warmup) — separate PR once #218 lands.
- Cross-process refresh lock (#133) — separate PR.

### DIFF
--- a/packages/cli/tests/test_auth.py
+++ b/packages/cli/tests/test_auth.py
@@ -1,4 +1,4 @@
-"""Tests for ``pipefy_cli.auth`` (service-account / bearer factory and CLI exits)."""
+"""Tests for ``pipefy_cli.auth`` (resolver wiring, eager warmup, CLI exits)."""
 
 from __future__ import annotations
 
@@ -7,12 +7,18 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 import typer
-from pipefy_auth import OidcClient, StoredSession
+from pipefy_auth import (
+    CallableBearerAuth,
+    OidcClient,
+    StaticBearerAuth,
+    StoredSession,
+)
 from pipefy_sdk import PipefySettings
 
 from pipefy_cli.auth import (
     AuthContext,
     BearerToken,
+    detect_cli_sources,
     get_authenticated_client,
 )
 from pipefy_cli.main import app
@@ -49,28 +55,39 @@ def _auth(
     return AuthContext(bearer_token=bearer, oidc_client=oidc_client)
 
 
-def test_get_authenticated_client_passes_bearer_to_pipefy_client(clean_pipefy_env):
+def test_get_authenticated_client_passes_auth_to_pipefy_client(clean_pipefy_env):
     settings = _minimal_service_account_settings()
     with patch("pipefy_cli.auth.PipefyClient") as mock_pc:
         mock_pc.return_value = MagicMock()
         client = get_authenticated_client(settings, _auth(bearer_token="tok"))
-        mock_pc.assert_called_once_with(settings, bearer_token="tok")
+        kwargs = mock_pc.call_args.kwargs
+        assert mock_pc.call_args.args == (settings,)
+        assert isinstance(kwargs["auth"], StaticBearerAuth)
         assert client is mock_pc.return_value
 
 
-def test_get_authenticated_client_service_account_mode_no_bearer(clean_pipefy_env):
+def test_get_authenticated_client_service_account_wires_internal_api(clean_pipefy_env):
     settings = _minimal_service_account_settings()
-    with patch("pipefy_cli.auth.PipefyClient") as mock_pc:
+    with (
+        patch("pipefy_cli.auth.PipefyClient") as mock_pc,
+        patch("pipefy_cli.auth.InternalApiClient") as mock_internal,
+        patch("pipefy_cli.auth.AiAutomationService"),
+    ):
         mock_pc.return_value = MagicMock()
         get_authenticated_client(settings, _auth())
-        mock_pc.assert_called_once_with(settings)
+        mock_pc.assert_called_once()
+        mock_internal.assert_called_once()
 
 
 def test_cache_returns_same_instance_for_identical_service_account_settings(
     clean_pipefy_env,
 ):
     settings = _minimal_service_account_settings()
-    with patch("pipefy_cli.auth.PipefyClient") as mock_pc:
+    with (
+        patch("pipefy_cli.auth.PipefyClient") as mock_pc,
+        patch("pipefy_cli.auth.InternalApiClient"),
+        patch("pipefy_cli.auth.AiAutomationService"),
+    ):
         mock_pc.return_value = MagicMock()
         first = get_authenticated_client(settings, _auth())
         second = get_authenticated_client(settings, _auth())
@@ -118,7 +135,28 @@ def test_cli_uses_pipefy_token_env_when_no_flag(
 
 
 # --------------------------------------------------------------------------- #
-# Priority 4: stored user session                                             #
+# Display source mapping                                                      #
+# --------------------------------------------------------------------------- #
+
+
+def test_detect_cli_sources_maps_static_token_to_flag_label(clean_pipefy_env):
+    """``--token`` source surfaces as the locked ``flag-token`` wire value."""
+    settings = _minimal_service_account_settings()
+    sources = detect_cli_sources(
+        settings, _auth(bearer_token="t", bearer_source="flag")
+    )
+    assert sources[0] == "flag-token"
+
+
+def test_detect_cli_sources_maps_env_token_to_env_label(clean_pipefy_env):
+    """``PIPEFY_TOKEN`` source surfaces as the locked ``env-token`` wire value."""
+    settings = _minimal_service_account_settings()
+    sources = detect_cli_sources(settings, _auth(bearer_token="t", bearer_source="env"))
+    assert sources[0] == "env-token"
+
+
+# --------------------------------------------------------------------------- #
+# Stored user session                                                         #
 # --------------------------------------------------------------------------- #
 
 
@@ -142,15 +180,18 @@ def _fresh_stored_session(*, access_token: str = "SESSION_ACCESS") -> StoredSess
 
 
 def _public_only_settings() -> PipefySettings:
-    """``PIPEFY_SERVICE_ACCOUNT_*`` triple absent → priority 3 unavailable, falls through to 4."""
+    """``PIPEFY_SERVICE_ACCOUNT_*`` triple absent → service-account tier unavailable, stored session wins."""
     return PipefySettings(graphql_url="https://unit.example.com/graphql")
 
 
 def test_bearer_token_wins_over_stored_session(clean_pipefy_env):
-    """Priority 1/2 (bearer) MUST short-circuit before the keychain is even consulted."""
+    """The static-token tier MUST short-circuit before the keychain is even consulted."""
     settings = _minimal_service_account_settings()
     with (
         patch("pipefy_cli.auth.PipefyClient") as mock_pc,
+        patch("pipefy_cli.auth.InternalApiClient"),
+        patch("pipefy_cli.auth.AiAutomationService"),
+        patch("pipefy_auth.resolver.load_session") as mock_load,
         patch("pipefy_cli.auth.ensure_fresh_session") as mock_ensure,
     ):
         mock_pc.return_value = MagicMock()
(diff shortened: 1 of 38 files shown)


packages/mcp/src/pipefy_mcp/core/container.py
```diff
@@ -26,23 +32,31 @@ def initialize_services(self, settings: Settings) -> None:
...
+            and pipefy.service_account_client_secret
         ):
+            service_account = ServiceAccount(
+                token_url=pipefy.service_account_url,
+                client_id=pipefy.service_account_client_id,
+                client_secret=pipefy.service_account_client_secret,
+            )
+        # The stored-session tier is wired in a follow-up against #213.
+        resolved = resolve_pipefy_auth(
+            static_token=os.environ.get("PIPEFY_TOKEN"),
```
adriannoes, line 49: The service-account tier is built from `settings.pipefy` (so `.env` works), but `static_token` comes from `os.environ.get("PIPEFY_TOKEN")`. Pydantic-settings loads `.env` into model fields without populating `os.environ`, so a repo-root `.env` with only `PIPEFY_TOKEN` + `PIPEFY_GRAPHQL_URL` loads GraphQL yet fails auth — I reproduced `resolve_pipefy_auth(static_token=os.environ.get(...))` returning `None` while `settings.pipefy.graphql_url` is set.

I'd thread the bearer through the same settings path as the SA triple (or a small shared helper both CLI and MCP call) and add a container test that uses a `.env` file without `monkeypatch.setenv`. You'll know it's fixed when `.env`-only `PIPEFY_TOKEN` boots the MCP container the same way injected env vars do.

packages/cli/tests/test_auth.py
```diff
@@ -50,28 +55,39 @@ def _auth(
...
+    with (
+        patch("pipefy_cli.auth.PipefyClient") as mock_pc,
+        patch("pipefy_cli.auth.InternalApiClient") as mock_internal,
+        patch("pipefy_cli.auth.AiAutomationService"),
+    ):
         mock_pc.return_value = MagicMock()
         get_authenticated_client(settings, _auth())
-        mock_pc.assert_called_once_with(settings)
+        mock_pc.assert_called_once()
+        mock_internal.assert_called_once()
```
adriannoes, line 79: #134 is fixed in code — `get_authenticated_client` wires `InternalApiClient` for all tiers — but the only internal-API test covers the service-account path and doesn't assert the `auth` instance. A small bearer-path test mirroring that wiring (`mock_internal.call_args.kwargs["auth"] is mock_pc.call_args.kwargs["auth"]`) would lock the regression this PR is meant to close.

packages/mcp/tests/core/test_container.py
```diff
@@ -100,3 +104,48 @@ def test_initialize_services_creates_ai_services(
...
+            pipefy=PipefySettings(
+                graphql_url="https://api.pipefy.com/graphql",
+                service_account_url="https://auth.pipefy.com/oauth/token",
+                service_account_client_id="client_id",
+                service_account_client_secret="client_secret",
+            )
+        )
+        ServicesContainer().initialize_services(settings)
+        auth = mock_pipefy_client_class.call_args.kwargs["auth"]
+        assert isinstance(auth, StaticBearerAuth)
```
adriannoes: `test_initialize_services_picks_up_pipefy_token_over_service_account` confirms `StaticBearerAuth` wins precedence, but unlike the SA AI test it never asserts `InternalApiClient` or `set_internal_api_client`. Extending it with the same assertions as `test_initialize_services_creates_ai_services` would catch a bearer-path regression where GraphQL auth works but AI automation stays unwired.

adriannoes (approved):
## Summary

Clean consolidation of the auth precedence chain into `pipefy_auth` — the resolver is readable, SDK clients correctly require explicit `auth=`, and the bearer path now wires `InternalApiClient` (closing #134). I checked out the head locally: `ruff check` clean, **2312** unit tests passing.

**Approved with reservations** — nothing here blocks merge in the #216 stack, but I'd address F1 before this reaches users who keep `PIPEFY_TOKEN` in `.env` only.

## Also noted

- **F1 (inline):** MCP reads `PIPEFY_TOKEN` from `os.environ` while the service-account triple loads via pydantic-settings — `.env`-only bearer setups fail auth even when GraphQL loads. Worth fixing or documenting before the stack lands on `dev`.
- **F2 (inline):** Add a bearer-path test that asserts `InternalApiClient` receives the same `auth` instance as `PipefyClient` — guards the #134 fix against future drift.
- **F3 (inline):** Extend the MCP token-precedence test with `InternalApiClient` / AI wiring assertions, mirroring the service-account path.

## Using this review with AI

Full structured findings (IDs, evidence, done-when) live in `.cursor/dev-planning/code-review/pr-217-auth-resolver-consolidation.md` in the reviewer's clone.

## Verification

```bash
uv run ruff check .
```

packages/cli/tests/test_auth.py
gbrlcustodio: **resolution:** addressed in bed542d — `test_get_authenticated_client_bearer_path_shares_auth_with_internal_api` now asserts both `PipefyClient` and `InternalApiClient` receive the same `StaticBearerAuth` instance.

packages/mcp/tests/core/test_container.py
gbrlcustodio: **resolution:** addressed in bed542d — `test_initialize_services_picks_up_pipefy_token_over_service_account` now also asserts `InternalApiClient` is wired, the `auth` instance is shared with `PipefyClient`, and `set_ai_automation_service` is called, mirroring the SA path.

packages/mcp/src/pipefy_mcp/core/container.py
gbrlcustodio: **acknowledged, deferring to #218:** the fix routes `PIPEFY_TOKEN` through `AuthSettings` (the new `pipefy_auth.AuthSettings` introduced in #218), which is the architecturally correct home for the bearer field — no point landing a stop-gap on `PipefySettings` here just to move it one PR later. Picking this up next on #218.

PR #218, opened 2026-05-22
refactor(auth): own auth-related settings inside pipefy_auth

gbrlcustodio:
Stacked on #217. Targets that branch directly so the diff is just the settings split; will be retargeted to `dev` once #217 merges. Second of two PRs that together close out #213's resolver/settings work.

## Summary

Splits the fat `PipefySettings` into endpoint config (stays in `pipefy-sdk`) and auth config (new `pipefy_auth.AuthSettings`). Both consumers (`pipefy-cli`, `pipefy-mcp`) compose the two models side by side in their own settings type.

- `pipefy_auth.AuthSettings` (new) is a `BaseSettings` with `env_prefix="PIPEFY_"`. It owns the service-account triple, the legacy `PIPEFY_OAUTH_*` aliases + one-shot deprecation warning, and the OIDC fields (`auth_url`, `auth_client_id`). Exposes `to_service_account()` / `to_oidc_client()` projections and `validate_urls(allow_insecure=...)`.
- `pipefy_sdk.PipefySettings` shrinks to endpoint + runtime concerns (`graphql_url`, `internal_api_url`, `allow_insecure_urls`, etc.). Drops the SA fields and the `_LEGACY_ENV_KEYS_TO_NEW` machinery (moves with the fields).
- `pipefy_cli.config.CliSettings` and `pipefy_mcp.settings.Settings` embed `auth: AuthSettings` alongside `pipefy: PipefySettings`. The CLI's `AuthContext` gains a `service_account` slot so the resolver receives its three inputs in one shape; the CLI projects `AuthSettings` into `ServiceAccount` / `OidcClient` at the boundary.
- The `describe_missing_service_account_vars` helper moves to `pipefy_auth.settings`. The TOML loader routes auth keys to `AuthSettings` (derived from `AuthSettings.model_fields` so renames stay in sync).
- Auth-URL SSRF validation runs once per `resolve_cli_settings` call against the final `allow_insecure_urls` value.

Drive-by `/simplify` cleanups also included:
- `_revalidate_pipefy` + `_revalidate_auth` unified into one generic helper.
- `_detect_stored_session` trivial wrapper inlined in `detect_pipefy_sources`.
- `_minimal_service_account_settings` test alias dropped (call sites now use `_minimal_settings` directly).

## Test plan

- [x] `uv run pytest -m "not integration"` — 2310 passing locally.
- [x] `uv run ruff check` + `uv run ruff format --check` — clean.
- [x] CI green.
- [x] Smoke: legacy `PIPEFY_OAUTH_*` env vars still populate `AuthSettings.service_account_*` and emit the deprecation warning once per process.
- [x] Smoke: `pipefy auth status` reports the correct source after the split.
- [x] Smoke: MCP server boots with both flavours of env (`PIPEFY_SERVICE_ACCOUNT_*` and `PIPEFY_TOKEN`).

## Migration

`PipefySettings` no longer accepts `service_account_url` / `service_account_client_id` / `service_account_client_secret`. In-repo callers are updated; out-of-repo consumers should migrate to `AuthSettings`.

The back-compat suite for the `PIPEFY_OAUTH_*` rename moves from `packages/sdk/tests/test_settings_back_compat.py` to `packages/auth/tests/test_settings_back_compat.py` and asserts against `AuthSettings`.

## Follow-ups

- #220 — tighten ``auth: Auth | None = None`` on 14 SDK service constructors (loose end from #217 caught while reviewing this PR).
- #221 — collapse ``InternalApiClient`` into ``BasePipefyClient`` (surfaced during this review; design discussion in the issue).

### DIFF
--- a/packages/auth/pyproject.toml
+++ b/packages/auth/pyproject.toml
@@ -9,6 +9,7 @@ dependencies = [
     "httpx-auth>=0.23.1",
     "keyring>=24",
     "pydantic>=2.11.3",
+    "pydantic-settings>=2.8.1",
 ]
 
 [dependency-groups]
--- a/packages/auth/pyproject.toml
+++ b/packages/auth/pyproject.toml
@@ -8,6 +8,7 @@ dependencies = [
     "httpx>=0.28",
     "httpx-auth>=0.23.1",
     "keyring>=24",
+    "pydantic>=2.11.3",
 ]
 
 [dependency-groups]
--- a/packages/auth/src/pipefy_auth/settings.py
+++ b/packages/auth/src/pipefy_auth/settings.py
@@ -1,0 +1,204 @@
+"""Pydantic model for auth-related settings (env vars, config file fields).
+
+Owns every value that describes *how* to authenticate against Pipefy:
+
+* ``PIPEFY_SERVICE_ACCOUNT_*`` — OAuth2 client-credentials grant inputs
+  (legacy ``PIPEFY_OAUTH_*`` names still honoured via :class:`AliasChoices`).
+* ``PIPEFY_AUTH_URL`` — OIDC issuer URL for the stored-session tier.
+* ``PIPEFY_AUTH_CLIENT_ID`` — OIDC public client id (defaults to
+  :data:`pipefy_auth.identity.DEFAULT_AUTH_CLIENT_ID`).
+
+Endpoint settings (``PIPEFY_GRAPHQL_URL``, ``PIPEFY_INTERNAL_API_URL``) live
+on :class:`pipefy_sdk.PipefySettings` — they are SDK concerns, not auth.
+Consumers compose both models side by side in their own settings type.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+
+from pydantic import AliasChoices, Field, model_validator
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from pipefy_auth.identity import DEFAULT_AUTH_CLIENT_ID, OidcClient
+from pipefy_auth.resolver import ServiceAccount
+
+# Legacy ``PIPEFY_OAUTH_*`` env vars still resolve to the new
+# ``PIPEFY_SERVICE_ACCOUNT_*`` fields. The mapping is exported for
+# diagnostics (e.g. CLI's ``pipefy auth status`` lists which legacy keys
+# would still mask a stored session).
+_LEGACY_ENV_KEYS_TO_NEW: dict[str, str] = {
+    "PIPEFY_OAUTH_URL": "PIPEFY_SERVICE_ACCOUNT_URL",
+    "PIPEFY_OAUTH_CLIENT": "PIPEFY_SERVICE_ACCOUNT_CLIENT_ID",
+    "PIPEFY_OAUTH_SECRET": "PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET",
+}
+
+_warned_legacy_env_keys: set[str] = set()
+
+
+def _warn_once_for_legacy_oauth_env_keys() -> None:
+    """Emit a one-shot stderr deprecation warning for each legacy env var still set."""
+    if len(_warned_legacy_env_keys) == len(_LEGACY_ENV_KEYS_TO_NEW):
+        return
+    for legacy, new in _LEGACY_ENV_KEYS_TO_NEW.items():
+        if legacy in _warned_legacy_env_keys:
+            continue
+        if legacy in os.environ:
+            sys.stderr.write(
+                f"warning: {legacy} is deprecated; rename to {new}. "
+                "The legacy name will be removed in a future beta.\n"
+            )
+            _warned_legacy_env_keys.add(legacy)
+
+
+def _reset_legacy_oauth_warning_state() -> None:
+    """Test helper: clear the one-shot dedup so a fixture can re-trigger the warning."""
+    _warned_legacy_env_keys.clear()
+
+
+class AuthSettings(BaseSettings):
+    """Auth-related configuration loaded from env / config files.
+
+    Reads its own ``PIPEFY_*`` env vars directly (with the ``PIPEFY_`` prefix
+    folded into the field names). Pure data — SSRF validation happens via
+    :meth:`validate_urls` once the surrounding settings know whether insecure
+    URLs are allowed.
+    """
+
+    model_config = SettingsConfigDict(
+        env_prefix="PIPEFY_",
+        env_file=".env",
+        env_file_encoding="utf-8",
+        extra="ignore",
+        # Required so ``env_prefix`` is applied to the field name on top of the
+        # ``AliasChoices`` lookups (otherwise env lookups would only consider
+        # the aliases, and ``PIPEFY_SERVICE_ACCOUNT_URL`` would be ignored).
+        populate_by_name=True,
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _emit_legacy_oauth_env_var_warning(cls, data: object) -> object:
+        # Mirrors the pre-split behaviour: warn once per legacy env key still set.
+        _warn_once_for_legacy_oauth_env_keys()
+        return data
+
+    # ``AliasChoices`` precedence is left-to-right. The fully-prefixed
+    # canonical env var comes first to outrank the legacy ``PIPEFY_OAUTH_*``
+    # name. The unprefixed form (e.g. ``oauth_url``) keeps direct kwarg
+    # construction working (``AuthSettings(oauth_url=...)``).
+    service_account_url: str | None = Field(
+        default=None,
+        validation_alias=AliasChoices(
+            "PIPEFY_SERVICE_ACCOUNT_URL",
+            "service_account_url",
+            "oauth_url",
+            "PIPEFY_OAUTH_URL",
+        ),
+        description=(
+            "Service-account token endpoint (OAuth 2.0 client-credentials grant) "
+            "(env: PIPEFY_SERVICE_ACCOUNT_URL; legacy PIPEFY_OAUTH_URL still honored)."
+        ),
+    )
+
+    service_account_client_id: str | None = Field(
+        default=None,
+        validation_alias=AliasChoices(
+            "PIPEFY_SERVICE_ACCOUNT_CLIENT_ID",
+            "service_account_client_id",
+            "oauth_client",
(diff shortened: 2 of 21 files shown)


packages/mcp/src/pipefy_mcp/settings.py
```diff
@@ -23,8 +26,9 @@ class Settings(BaseSettings):
     )
 
     pipefy: PipefySettings = Field(default_factory=PipefySettings)
+    auth: AuthSettings = Field(default_factory=AuthSettings)
```
adriannoes, line 31: After the split, auth URL SSRF checks only run when something calls `AuthSettings.validate_urls()`. CLI does that in `CliSettings` and again after flag overrides, but MCP `Settings()` never does — I can import the module with `PIPEFY_SERVICE_ACCOUNT_URL=http://169.254.169.254/...` and nothing fails until token fetch time. That's a regression from the old automatic check on `PipefySettings.service_account_url`.

I'd add an `mode="after"` validator on MCP `Settings` that calls `self.auth.validate_urls(allow_insecure=self.pipefy.allow_insecure_urls)`, mirroring what `CliSettings` already does. You'll know it's fixed when a unit test sets that env var and `Settings()` (or server init) raises before any network I/O.

packages/auth/pyproject.toml
```diff
@@ -8,6 +8,7 @@ dependencies = [
     "httpx>=0.28",
     "httpx-auth>=0.23.1",
     "keyring>=24",
+    "pydantic>=2.11.3",
```
adriannoes, line 11: `AuthSettings` is a `BaseSettings` subclass, but `pipefy-auth` only adds `pydantic` in this PR — `pydantic_settings` isn't declared. It works in the monorepo because CLI/MCP pull it in transitively, but the auth package should stand alone.

Please add `pydantic-settings` here with the same lower bound as the rest of the workspace (e.g. `>=2.8.1`) and refresh `uv.lock`.

packages/mcp/tests/core/test_container.py
```diff
@@ -140,12 +140,20 @@ def test_initialize_services_raises_when_no_auth_source_configured(
...
+            "PIPEFY_OAUTH_CLIENT",
+            "PIPEFY_OAUTH_SECRET",
+        ):
+            monkeypatch.delenv(key, raising=False)
         settings = Settings(
-            pipefy=PipefySettings(
-                graphql_url="https://api.pipefy.com/graphql",
-            )
+            pipefy=PipefySettings(graphql_url="https://api.pipefy.com/graphql"),
+            auth=AuthSettings(),
```
adriannoes, line 174: Good fix here for the "no auth" case — `auth=AuthSettings()` is the right shape after the split. The success-path tests above (`test_initialize_services_creates_pipefy_client`, `test_initialize_services_creates_ai_services`, and the token-precedence test) still pass the service-account triple on `PipefySettings`, which no longer has those fields. Pydantic drops the kwargs silently; `initialize_services` reads `settings.auth.to_service_account()` instead.

Those tests fail locally for me (`Missing Pipefy authentication`) and only pass in CI because the workflow injects `PIPEFY_OAUTH_*`, which `AuthSettings()` picks up from the environment. Please update those fixtures the same way: `Settings(pipefy=..., auth=AuthSettings(service_account_url=..., ...))` with SA env vars cleared in the test.

adriannoes (changes requested):
## Summary

Solid direction on the settings split — composing `AuthSettings` alongside `PipefySettings` in CLI/MCP is the right boundary, and the CLI path (nested env binding, single canonical SSRF pass after flag override) looks correct. Review included a local checkout of `2de3848`: ruff clean; unit suite **2307 passed / 3 failed** locally (`test_container` ×2, `test_register_tools`) — failures align with fixtures still passing SA creds via removed `PipefySettings` fields while CI masks them via workflow `PIPEFY_OAUTH_*` env.

## Required before merge

- **F1 (MCP SSRF):** Wire `AuthSettings.validate_urls()` on MCP settings load — unsafe auth URLs currently import without error.
- **F2 (container tests):** Success-path fixtures must pass creds via `auth=AuthSettings(...)`, not dropped kwargs on `PipefySettings`.
- **F4 (`pydantic-settings`):** Declare the dependency on `pipefy-auth` so the package stands alone.

## Also noted

- **F3:** `packages/mcp/tests/test_server.py` — `_MINIMAL_PIPEFY_SETTINGS` still embeds SA creds on `PipefySettings` (file not in this diff, but `test_register_tools` fails locally for the same reason). Move the triple to `auth=AuthSettings(...)` when fixing F2.
- **F5:** `docs/MIGRATION.md` covers the env-var rename but not the model move — library/script users constructing `PipefySettings(service_account_url=...)` will silently drop creds. Worth a short note pointing them at `pipefy_auth.AuthSettings` composed alongside `PipefySettings`.

## Using this review with AI

Structured findings (F1–F8) are in `.cursor/dev-planning/code-review/pr-218-split-auth-settings.md` in the repo — address by ID.

## Verification

```bash
uv run pytest packages/mcp/tests/core/test_container.py packages/mcp/tests/test_server.py -m "not integration"
```

packages/mcp/src/pipefy_mcp/settings.py
gbrlcustodio: **resolution:** addressed in b671a2a — MCP `Settings` now runs `self.auth.validate_urls(allow_insecure=self.pipefy.allow_insecure_urls)` in a `model_validator(mode="after")`, mirroring `CliSettings`. Pinned by `test_settings_rejects_link_local_service_account_url`: setting `PIPEFY_SERVICE_ACCOUNT_URL=https://169.254.169.254/...` now raises on `Settings()` load before any I/O.

packages/mcp/tests/core/test_container.py
gbrlcustodio: **resolution:** addressed in b671a2a — `test_initialize_services_creates_pipefy_client`, `test_initialize_services_creates_ai_services`, and the token-precedence test now pass the SA triple via `auth=AuthSettings(...)`. Added an `autouse clear_auth_env` fixture on `TestServicesContainer` to strip `PIPEFY_*` from `os.environ` before each test, so the suite no longer relies on the CI workflow injecting `PIPEFY_OAUTH_*`. Same fix applied to `_MINIMAL_PIPEFY_SETTINGS` in `test_server.py`.

packages/auth/pyproject.toml
gbrlcustodio: **resolution:** addressed in b671a2a — added `pydantic-settings>=2.8.1` to `packages/auth/pyproject.toml` (same lower bound used by SDK/CLI/MCP); `uv.lock` refreshed.

packages/sdk/src/pipefy_sdk/settings.py
```diff
@@ -1,53 +1,23 @@
...
-def _warn_once_for_legacy_oauth_env_keys() -> None:
-    """Print a one-shot stderr deprecation warning for each ``PIPEFY_OAUTH_*`` env var still set.
+class PipefySettings(BaseModel):
+    """Pipefy API connection and shared runtime knobs (MCP, CLI, scripts).
 
-    Pydantic accepts the legacy names via ``AliasChoices`` so old configs keep working;
-    this nudge tells users to rename. Dedup state is process-global; tests reset via
-    :func:`_reset_legacy_oauth_warning_state`.
+    Endpoint configuration only — credentials live on
+    :class:`pipefy_auth.AuthSettings`. Consumers compose both side by side in
```
adriannoes, line 13: The split looks good end-to-end — this docstring is the right pointer for library users (`AuthSettings` composed alongside `PipefySettings`). One small docs gap: `docs/MIGRATION.md` still covers only the env-var rename, not the model move. Anyone still doing `PipefySettings(service_account_url=...)` will silently drop creds (`extra=ignore`). A short subsection there mirroring this docstring would close the loop for out-of-repo consumers. Happy to defer to a follow-up PR if you prefer.

packages/auth/src/pipefy_auth/settings.py
```diff
@@ -0,0 +1,228 @@
...
+            validate_https_service_endpoint_url(
+                u, "service_account_url", allow_insecure=allow_insecure
+            )
+        if self.auth_url and (u := self.auth_url.strip()):
+            validate_https_service_endpoint_url(
+                u, "auth_url", allow_insecure=allow_insecure
+            )
+
+
+def describe_missing_service_account_vars(auth: AuthSettings) -> str:
```
adriannoes: This helper was moved here from `pipefy_cli.config`, but its CLI callers were already removed in #217 — on this branch it has zero in-repo callers and no direct test. Re-exporting it in `__all__` reads like public diagnostic API, yet nothing exercises it. I'd either drop it (our no-dead-code rule) or keep it with a small unit test that pins the "missing vars" message shape. Not blocking merge; just flagging so we don't carry orphan surface area.

adriannoes (approved):
## Summary

Re-reviewed at `0c6da7a` after the follow-up commits — all four blockers from the first pass are fixed and verified locally (2356 unit tests green, ruff clean, MCP SSRF regression test in place, hermetic container fixtures, `pydantic-settings` declared on `pipefy-auth`). The settings split is clean: CLI and MCP compose `AuthSettings` alongside `PipefySettings`, bearer loads from `.env` via `static_token`, and the cache key is hashed.

## Also noted (non-blocking)

- **G1:** `docs/MIGRATION.md` could mention the model move to `AuthSettings` — inline on `PipefySettings` docstring.
- **G3:** `describe_missing_service_account_vars` has no callers post-#217 — inline on `settings.py`; drop or pin with a test in a follow-up.

## Verification

```bash
uv run ruff check .
```

packages/auth/src/pipefy_auth/settings.py
gbrlcustodio: **resolution:** addressed in 9936a69 — dropped per the no-dead-code rule. The helper was an orphan after #217 removed the CLI callers; carrying it forward as advertised public surface in ``__all__`` without a test or caller was misleading.

packages/sdk/src/pipefy_sdk/settings.py
gbrlcustodio: **resolution:** addressed in 9936a69 — added a "Settings model split" subsection to ``docs/MIGRATION.md`` covering the ``PipefySettings`` → ``AuthSettings`` field moves, the ``extra="ignore"`` silent-drop hazard, and a worked example of composing both models side by side. End-user wording at the top of the section makes clear that env / `.env` / TOML consumers are unaffected.

PR #220, opened 2026-05-22
refactor(sdk): require explicit auth on service constructors

gbrlcustodio:
Follow-up to #217 (which tightened ``BasePipefyClient`` / ``PipefyClient`` / ``InternalApiClient`` to require ``auth: httpx.Auth``).

## Summary

14 SDK service-layer constructors still declared ``auth: Auth | None = None`` and forwarded the value to the now-required base. That was type-system rot rather than a runtime bug — every actual call site already passed a real ``Auth`` — but the stale signature undermined the post-#217 invariant that ``auth`` is required. This sweep tightens them all to the same ``*, auth: Auth`` shape as the base classes.

Touched: ``attachment``, ``ai_agent``, ``card``, ``member``, ``organization``, ``pipe``, ``pipe_config``, ``portal``, ``relation``, ``report``, ``schema_introspection``, ``table``, ``user``, ``webhook``.

## Test plan

- [x] ``uv run pytest -m \"not integration\"`` — 2355 passing.
- [x] ``uv run ruff check`` + ``uv run ruff format --check`` — clean.
- [x] Seven test sites that relied on the ``None`` default (4 ``AiAgentService`` constructions and 3 in ``pipefy/*`` suites focused on auxiliary kwargs like ``pipe_service`` / ``card_service``) now pass the standard ``StaticBearerAuth`` stub.

## Background

When reviewing #218, I noticed ``PortalService`` (the Interfaces client) still accepted ``auth: Auth | None`` even though it passes the value straight to ``BasePipefyClient`` which now rejects ``None``. Grepping showed 14 services with the same stale shape — every one a forwarder. This PR is the mechanical fix.

### DIFF
--- a/packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
+++ b/packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
@@ -65,7 +65,8 @@ class AiAgentService(BasePipefyClient):
     def __init__(
         self,
         settings: PipefySettings,
-        auth: Auth | None = None,
+        *,
+        auth: Auth,
     ) -> None:
         """Initialize the GraphQL client.
 
--- a/packages/sdk/src/pipefy_sdk/services/attachment_service.py
+++ b/packages/sdk/src/pipefy_sdk/services/attachment_service.py
@@ -27,7 +27,8 @@ class AttachmentService(BasePipefyClient):
     def __init__(
         self,
         settings: PipefySettings,
-        auth: Auth | None = None,
+        *,
+        auth: Auth,
     ) -> None:
         super().__init__(settings=settings, auth=auth)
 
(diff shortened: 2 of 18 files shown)


adriannoes (approved):
Clean mechanical sweep — LGTM. 

Reviewed locally: 747 unit tests passing, ruff clean, CI green.

**One follow-up for later (not blocking merge):** The `live_svc` fixture in `packages/sdk/tests/services/pipefy/test_schema_introspection_integration.py:26` still builds `SchemaIntrospectionService(settings=live_pipefy_settings())` without `auth`, so it will raise `TypeError` on any `-m integration` run after this merges. 

The fix is a one-liner mirroring `test_portal_service_integration.py:34` — resolve live OAuth auth and pass `auth=` to the constructor. Default CI isn't affected, but worth catching before the next local integration run.

PR #222, opened 2026-05-22
refactor(sdk): collapse InternalApiClient onto BasePipefyClient

### LINKED ISSUE #221: refactor(sdk): collapse InternalApiClient into BasePipefyClient
## Context

Surfaced during review of #218: ``InternalApiClient`` exists as a separate transport class even though the internal_api endpoint is a regular GraphQL endpoint. The asymmetry is best explained as historical drift, not a real protocol difference.

Looking at the two transports:

| | ``BasePipefyClient`` (public + Interfaces) | ``InternalApiClient`` |
|---|---|---|
| Transport | ``gql.Client`` over ``HTTPXAsyncTransport`` | raw ``httpx.AsyncClient.post`` |
| Schema fetch | opt-in via ``gql_reuse_fetched_graphql_schema`` | n/a |
| Body shape | ``{query, variables}`` | ``{query, variables}`` (same) |
| Surface | ``async def execute_query(query, variables) -> dict`` | identical signature |
| Error envelope | propagates ``gql`` exceptions | parses ``extensions.code`` / ``extensions.correlation_id`` into ``[code=…] [correlation_id=…]`` suffix on ``ValueError`` |
| ``data`` unwrapping | done by ``gql`` | ``data.get(\"data\", data)`` |

So the only real differentiator is the **custom error suffix** the AI-automation MCP tools then strip with ``strip_internal_api_diagnostic_markers``. Everything else is duplicated machinery.

## Proposal

Fold internal_api transport into ``BasePipefyClient`` and replace ``InternalApiClient`` with either a thin subclass or a configuration option:

**Option A — subclass with hook (recommended).** Add an ``on_graphql_error: Callable[[list[dict]], str] | None = None`` keyword to ``BasePipefyClient.__init__`` that lets a subclass post-process the error list into the raised message. ``InternalApiClient`` becomes a 10-line subclass that passes ``_format_internal_api_error`` as that hook and accepts ``url`` (not ``settings``) by synthesising ``PipefySettings(graphql_url=url, allow_insecure_urls=...)`` internally. Keeps the typed distinction at the call sites (``set_internal_api_client`` still takes ``InternalApiClient``).

**Option B — flatten.** Delete the class entirely; callers do ``BasePipefyClient(PipefySettings(graphql_url=internal_api_url, ...), auth=auth)`` and use a free function for error formatting. Smaller API surface but loses the type tag at the call sites — ``client.py`` would have to track which ``BasePipefyClient`` is which.

Option A keeps the existing type contracts in ``PipefyClient.set_internal_api_client`` and ``PortalService.__init__`` unchanged.

## Scope

~18 files reference ``InternalApiClient`` — most are unchanged (just import path / type annotation). The real work is in:

- ``packages/sdk/src/pipefy_sdk/base_client.py`` — add ``on_graphql_error`` hook
- ``packages/sdk/src/pipefy_sdk/services/internal_api_client.py`` — rewrite as subclass
- ``packages/sdk/tests/services/test_internal_api_client.py`` — already tests the public behaviour (error envelope, HTTP errors, timeouts); should mostly survive intact

## Acceptance criteria

- [ ] ``InternalApiClient`` no longer maintains its own ``httpx.AsyncClient`` POST loop.
- [ ] The ``[code=…] [correlation_id=…]`` error suffix is preserved verbatim (existing ``test_internal_api_client.py`` and AI-automation tests expect this exact format).
- [ ] ``data.get(\"data\", data)`` fallback behaviour is preserved (or documented as removed if ``BasePipefyClient``'s unwrap is equivalent).
- [ ] All call sites (CLI ``auth.py``, MCP ``container.py``, ``PipefyClient.set_internal_api_client``, ``PortalService``, AI-automation service) keep their current shapes — this is an internal refactor.
- [ ] ``uv run pytest -m \"not integration\"`` passes.
- [ ] ``uv run ruff check`` + ``uv run ruff format --check`` clean.

## Out of scope

- Replacing the MCP-tool ``strip_internal_api_diagnostic_markers`` step (the suffix-stripping that hides ``[code=…]`` from end users) — that stays.
- Promoting an ``InterfacesClient`` class as a sibling to make all three transports look uniform. Worth a separate discussion if anyone wants it; this issue is just about removing the unjustified duplication on the internal_api side.

## Background

Caught during the auth-refactor review chain (#217 / #218 / #220). I claimed in conversation that internal_api was structurally different because "no schema to fetch" — that was wrong; ``gql.Client`` doesn't fetch schemas by default either. The transport divergence has no protocol-level justification.

gbrlcustodio:
## Summary

- `InternalApiClient` now subclasses `BasePipefyClient` and runs over the same `gql.Client` transport as the rest of the SDK, removing the duplicated raw-httpx POST loop / `raise_for_status` / `data`-envelope / `errors` extraction.
- The `[code=…] [correlation_id=…]` error envelope AI-automation MCP tools depend on is preserved verbatim via a new `on_graphql_error: Callable[[list[dict]], str] | None` hook on `BasePipefyClient.__init__` (passed by `InternalApiClient` only — `PipefyClient` is unchanged).
- All call-site type contracts (`set_internal_api_client`, `PortalService`, `AiAutomationService`) keep their current shapes — internal refactor.

## Behaviour change worth flagging

5xx responses from `internal_api` now raise `gql.transport.exceptions.TransportServerError` instead of `httpx.HTTPStatusError`. This brings internal_api into line with how the public-API client surfaces server errors; no source-level caller catches `HTTPStatusError` from this path. The 500-response test was updated accordingly.

## Acceptance criteria

- [x] ``InternalApiClient`` no longer maintains its own ``httpx.AsyncClient`` POST loop.
- [x] The ``[code=…] [correlation_id=…]`` error suffix is preserved verbatim.
- [x] ``data.get("data", data)`` fallback dropped — `gql.Client` unwraps the `data` envelope natively; the no-`data` edge case is dead and not exercised.
- [x] All call sites keep their current shapes.
- [x] `uv run pytest -m "not integration"` — 2356 passed.
- [x] `uv run ruff check && uv run ruff format --check` — clean.

Closes #221.

## Test plan

- [x] `uv run pytest packages/sdk/tests/services/test_internal_api_client.py` — 15/15 green
- [x] `uv run pytest -m "not integration"` — full suite green (2356 passed)
- [x] `uv run ruff check packages && uv run ruff format --check packages` — clean
- [ ] Smoke: exercise an AI-automation CLI command against real Pipefy and confirm the success path
- [ ] Smoke: trigger a `PERMISSION_DENIED` on `internal_api` and confirm the raised `ValueError` still carries `[code=PERMISSION_DENIED] [correlation_id=…]`
- [ ] Smoke: boot the MCP server, dispatch an AI-automation tool, confirm `strip_internal_api_diagnostic_markers` still matches and removes the suffix from end-user output

### DIFF
--- a/packages/sdk/src/pipefy_sdk/base_client.py
+++ b/packages/sdk/src/pipefy_sdk/base_client.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Callable
 from typing import Any, ClassVar
 
 from gql import Client
+from gql.transport.exceptions import TransportQueryError
 from gql.transport.httpx import HTTPXAsyncTransport
 from graphql import GraphQLSchema
 from httpx import Auth, Timeout
@@ -45,12 +47,18 @@ class BasePipefyClient:
         settings: PipefySettings,
         *,
         auth: Auth,
+        on_graphql_error: Callable[[list[dict]], str] | None = None,
     ) -> None:
         if settings.graphql_url is None:
             raise ValueError("GraphQL URL must be provided in settings.")
 
         self.settings = settings
         self._auth = auth
+        # When set, ``TransportQueryError`` is converted to ``ValueError`` using the
+        # formatter's output. Used by ``InternalApiClient`` to surface its
+        # ``[code=…] [correlation_id=…]`` envelope; ``None`` leaves gql exceptions
+        # untouched (PipefyClient's behaviour).
+        self._on_graphql_error = on_graphql_error
         # Populated when gql_reuse_fetched_graphql_schema is True; avoids repeating
         # introspection on every new Client (see Cons5 code review).
         self._fetched_gql_schema: GraphQLSchema | None = None
@@ -70,30 +78,35 @@ class BasePipefyClient:
             auth=self._auth,
             timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
         )
-        if self.settings.gql_reuse_fetched_graphql_schema:
-            if self._fetched_gql_schema is None:
-                async with self._fetched_gql_schema_lock:
-                    if self._fetched_gql_schema is None:
-                        client = Client(
-                            transport=transport,
-                            fetch_schema_from_transport=True,
-                        )
-                        async with client as session:
-                            result = await session.execute(
-                                query, variable_values=variables
+        try:
+            if self.settings.gql_reuse_fetched_graphql_schema:
+                if self._fetched_gql_schema is None:
+                    async with self._fetched_gql_schema_lock:
+                        if self._fetched_gql_schema is None:
+                            client = Client(
+                                transport=transport,
+                                fetch_schema_from_transport=True,
                             )
-                        if client.schema is not None:
-                            self._fetched_gql_schema = client.schema
-                        return result
-            reuse_client = Client(
-                transport=transport,
-                schema=self._fetched_gql_schema,
-                fetch_schema_from_transport=False,
-            )
-            async with reuse_client as session:
-                return await session.execute(query, variable_values=variables)
+                            async with client as session:
+                                result = await session.execute(
+                                    query, variable_values=variables
+                                )
+                            if client.schema is not None:
+                                self._fetched_gql_schema = client.schema
+                            return result
+                reuse_client = Client(
+                    transport=transport,
+                    schema=self._fetched_gql_schema,
+                    fetch_schema_from_transport=False,
+                )
+                async with reuse_client as session:
+                    return await session.execute(query, variable_values=variables)
 
-        async with Client(
-            transport=transport, fetch_schema_from_transport=False
-        ) as session:
-            return await session.execute(query, variable_values=variables)
+            async with Client(
+                transport=transport, fetch_schema_from_transport=False
+            ) as session:
+                return await session.execute(query, variable_values=variables)
+        except TransportQueryError as exc:
+            if self._on_graphql_error is None:
+                raise
+            raise ValueError(self._on_graphql_error(exc.errors or [])) from exc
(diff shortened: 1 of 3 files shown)


adriannoes (approved):
## Summary

Clean refactor — `InternalApiClient` correctly collapses onto `BasePipefyClient` without changing call-site contracts. The `[code=…] [correlation_id=…]` error envelope is preserved verbatim, the duplicate httpx POST loop is gone, and 5xx handling now aligns with the public client via `TransportServerError`. Review included a local checkout (`31cb829`) with ruff clean and 747 SDK unit tests passing; CI is green.

No blockers. Approved.

## Follow-ups (non-blocking — worth a quick follow-up PR)

- **F1 — Pin schema reuse off for internal_api:** Pass `gql_reuse_fetched_graphql_schema=False` explicitly in the synthetic `PipefySettings` inside `InternalApiClient.__init__`, so a process-wide `PIPEFY_GQL_REUSE_FETCHED_GRAPHQL_SCHEMA=true` cannot trigger introspection against `internal_api`.
- **F2 — Unit-test the new `on_graphql_error` hook:** Add a pair of mocked tests in `test_pipefy_base_client.py` — one with `on_graphql_error=None` asserting `TransportQueryError` propagates unchanged, and one with a formatter asserting `ValueError` plus `__cause__`. The internal-api respx tests cover the formatter path indirectly, but the shared base hook is new infrastructure.
- **F3 — Restore docstring on the `execute_query` override:** The pre-refactor override documented the `query: str` contract and expected exceptions. Without it, `help()` falls back to the base docstring (`query: Any`, schema-reuse notes) which does not match this subclass's surface.

## Using this review with AI

Structured notes live at `.cursor/dev-planning/code-review/pr-222-collapse-internal-api-client.md` in the repo — each follow-up is tagged F1–F3 for easy pickup.

## Verification

```bash
uv run pytest packages/sdk/tests/services/test_internal_api_client.py -m "not integration" -q
uv run pytest packages/sdk/tests -m "not integration" -q
uv run ruff check packages/sdk
```

PR #223, opened 2026-05-22
fix(sdk-tests): repair live-integration fixtures post-#218

gbrlcustodio:
Follow-up to [#220's review comment](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/220#pullrequestreview-4349172218). Investigating the flagged `SchemaIntrospectionService(settings=...)` call surfaced two adjacent breakages with the same root cause: nothing under `-m integration` had been exercised since #218 moved credentials onto `pipefy_auth.AuthSettings` and #220 made `auth` required.

## Summary

- `_shared/live_settings.py` accessed `PipefySettings.service_account_url` (AttributeError post-#218); the helper now loads `AuthSettings` and exposes a `live_resolved_auth()` that runs through `pipefy_auth.resolve_pipefy_auth` — same precedence chain the CLI / MCP container use in production.
- `test_portal_service_integration.py` built `OAuth2ClientCredentials` by hand from `settings.oauth_url` / `oauth_client` / `oauth_secret` (also gone post-#218) — now calls `live_resolved_auth()`.
- `test_schema_introspection_integration.py` was missing the `auth=` kwarg the reviewer flagged; same shape as the portal fixture.

`pipefy_live_configured()` is also broadened to accept any single auth tier (static-token, service-account, or stored-session) — previously it required the full service-account triple.

## Test plan

- [x] `uv run pytest -m "not integration" packages/sdk` — 747 passed
- [x] `uv run pytest --collect-only` on both integration files — 12 tests collected without error
- [x] `uv run ruff check packages && uv run ruff format --check packages` — clean
- [ ] Local `uv run pytest -m integration` with credentials — to confirm at the next live run

### DIFF
--- a/packages/sdk/tests/_shared/live_settings.py
+++ b/packages/sdk/tests/_shared/live_settings.py
@@ -4,11 +4,16 @@ from __future__ import annotations
 
 import pytest
 from httpx import Auth
-from pipefy_auth import AuthSettings, resolve_pipefy_auth
+from pipefy_auth import AuthSettings, missing_auth_message, resolve_pipefy_auth
 from pydantic import Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from pipefy_sdk.settings import PipefySettings
+
+_MISSING_CREDS_MESSAGE = (
+    "Pipefy credentials not configured: PIPEFY_GRAPHQL_URL required; "
+    + missing_auth_message()
+)
 
 
 class _LiveEnvSettings(BaseSettings):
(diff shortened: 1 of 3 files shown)


adriannoes (approved):
## Summary

Focused fix — SDK live-integration fixtures now wire auth through `resolve_pipefy_auth`, matching the production precedence chain post-#218/#220. The flagged `SchemaIntrospectionService(settings=...)` gap is closed, and the portal fixture no longer hand-rolls `OAuth2ClientCredentials` from removed `PipefySettings` OAuth fields. Review included a local checkout (`6b6fd9e`) with ruff clean, 747 SDK unit tests passing, and 12 integration tests collecting without error; CI is green.

No blockers. Approved.

## Follow-ups (non-blocking — worth a quick follow-up PR)

- **F1 — Align live cred detection with resolver:** `pipefy_live_configured()` counts `to_oidc_client()` as configured auth, but `resolve_pipefy_auth` only picks up stored-session when there's actually a keychain session. That can make `require_live_creds()` pass and `live_resolved_auth()` skip right after with a different message. Derive `has_auth` from `resolve_pipefy_auth(...) is not None` so detection matches what the fixtures use.
- **F2 — Canonical skip messages:** Skip strings still only mention `PIPEFY_TOKEN` / `PIPEFY_SERVICE_ACCOUNT_*`, but this helper now supports stored-session too. Reusing `missing_auth_message()` (prefixed with the GraphQL URL requirement) would keep SDK integration skips aligned with CLI/MCP copy.

**Also noted (out of this PR's scope):** MCP live fixtures (`test_introspection_tools_live.py`, `test_attachment_tools_live.py`) and CLI `test_auth_integration.py` still construct clients without `auth=` — same root cause, separate follow-up.

PR #225, opened 2026-05-23
feat(auth): serialize concurrent refresh-token grants (#133)

### LINKED ISSUE #133: Concurrent `pipefy` invocations race the refresh-token grant

gbrlcustodio:
PR-C of [#213](https://github.com/gbrlcustodio/pipefy-mcp-server/issues/213). Closes #133.

## Problem

Two ``pipefy`` processes near token expiry both POST the same ``refresh_token`` to the IdP. Keycloak rotates the refresh token on each successful grant, so one process gets a fresh session and the other surfaces ``RefreshError("Refresh failed (HTTP 400): invalid_grant")`` even though the user's session is fine. Triggered by ``xargs -P``, ``parallel``, shell ``&`` fan-out, IDE / MCP integrations spawning concurrent commands.

## Fix

Cross-process filesystem mutex around the ``refresh + store`` window in ``ensure_fresh_session``, with a re-load + re-check inside the critical section so the loser of the race observes the winner's rotated session and skips its own POST.

- POSIX: ``fcntl.flock(LOCK_EX | LOCK_NB)`` polled with a 30s timeout
- Windows: ``msvcrt.locking(LK_NBLCK)`` polled with the same shape
- Lock file: ``~/.config/pipefy/refresh.lock`` (POSIX) / ``%APPDATA%/pipefy/refresh.lock`` (Windows)
- Lock window: ~200-500ms (one HTTPS POST + one keychain write); the 30s timeout is the deadlock backstop, not the expected wait

## Crash safety

| Scenario | Behaviour |
|---|---|
| POSIX ``SIGKILL`` / ``SIGTERM`` / segfault during hold | kernel releases ``flock`` on FD close — no deadlock; covered by ``test_file_lock_released_after_sigkill`` |
| ``KeyboardInterrupt`` during hold | ``@contextmanager``'s ``finally`` runs ``os.close(fd)``, releasing cleanly |
| Windows ``taskkill /F`` | mandatory locking can rarely leave a region locked; the 30s timeout surfaces ``RefreshError`` instead of hanging |
| Hung-but-alive holder | bounded by ``timeout_s=30``; clean ``RefreshError`` |
| Stale lock file present, no kernel holder | next acquirer's ``flock(LOCK_EX)`` succeeds immediately (normal state) |

## Test plan

- [x] ``uv run pytest -m "not integration"`` — 2365 passing
- [x] ``uv run ruff check packages && uv run ruff format --check packages`` — clean
- [x] ``test_locks.py`` covers happy path, contention timeout, reentrancy, POSIX crash safety via subprocess + ``SIGKILL``, and parent-dir creation
- [x] ``test_refresh_concurrent.py`` pins the double-checked-locking shape inside ``ensure_fresh_session``: lock entered before the POST and exited after the store; second invocation observes the fresh session and skips its POST; ``RefreshLockTimeout`` → ``RefreshError``
- [x] **Manual end-to-end smoke against piporacle** — 4 workers gated on a ``multiprocessing.Barrier`` so they hit ``ensure_fresh_session`` at the same instant, after force-staling the keychain entry to ``expires_in=0``:

  | Signal | Reading |
  |---|---|
  | All 4 workers returned ``ok`` | No ``invalid_grant``; pre-PR three of four would have errored |
  | Refresh token rotated exactly once | Only one IdP round-trip, not four |
  | Worker elapsed times (535, 576, 585, 627 ms) | ~50 ms increments between successive acquirers = the ``_POLL_INTERVAL_S`` signature — unmistakably the lock serializing them |
  | Post-test ``expires_in=259200`` | Keycloak restored a normal ~3-day refresh-token lifetime |

## Notes

- ``RefreshLockTimeout`` is intentionally not exported from ``pipefy_auth`` — callers see ``RefreshError`` like today, which preserves the existing CLI / MCP exception-handling contract.
- An autouse fixture in ``packages/auth/tests/conftest.py`` redirects the lock file path to ``tmp_path`` for all auth tests; without it, every ``ensure_fresh_session`` test would create ``~/.config/pipefy/refresh.lock`` on the developer's machine.
- A two-process IdP-mock test would require ``pytest-httpserver`` (not currently a dep). The kernel's ``flock`` correctness isn't our code under test; our code is the *use* of it, which the shape tests cover.

## Risks

- **NFS-home users** silently lose the cross-process guarantee (``fcntl.flock`` is unreliable over NFS). The refresh still works; just no serialisation. Documented in ``locks.py``'s module docstring.
- **Windows mandatory locking** quirks under ungraceful termination — documented and backstopped by the 30s timeout.
- **PR-D (MCP container) callers from async contexts** will block the event loop on the sync ``flock``. The umbrella plan wraps ``ensure_fresh_session`` in ``asyncio.to_thread``; out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **CLI**: concurrent `pipefy` invocations near token expiry no longer surface `invalid_grant` errors; the refresh-token grant is now serialized across processes via a filesystem lock at `~/.config/pipefy/refresh.lock` (`%APPDATA%/pipefy/refresh.lock` on Windows). Closes #133.
+
 ### Removed
 
 ## [0.2.0-beta.1] - 2026-05-18
--- a/packages/auth/src/pipefy_auth/locks.py
+++ b/packages/auth/src/pipefy_auth/locks.py
@@ -1,0 +1,115 @@
+"""Cross-process serialization for the refresh-token grant.
+
+POSIX uses ``fcntl.flock(LOCK_EX | LOCK_NB)`` polled until acquired (kernel-
+managed, FD-bound, released on process exit including SIGKILL). Windows uses
+``msvcrt.locking(LK_NBLCK)`` with the same polled shape. Lock window is
+~200-500ms (one POST + one keychain write); 30s is the deadlock backstop.
+
+The advisory `fcntl.flock` guarantee is cross-process on local filesystems
+but historically unreliable over NFS; the lock file lives under
+``~/.config/pipefy`` which is typically local. NFS-home users silently lose
+the cross-process serialisation but the refresh still works.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+import time
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
+_POLL_INTERVAL_S = 0.05
+
+
+class RefreshLockTimeout(RuntimeError):
+    """Could not acquire the cross-process refresh lock within timeout."""
+
+
+def refresh_lock_path() -> Path:
+    """Filesystem path used to coordinate concurrent refreshes."""
+    if sys.platform == "win32":
+        appdata = os.environ.get("APPDATA")
+        base = Path(appdata) if appdata else Path.home() / "AppData" / "Roaming"
+        directory = base / "pipefy"
+    else:
+        directory = Path.home() / ".config" / "pipefy"
+    directory.mkdir(parents=True, exist_ok=True)
+    return directory / "refresh.lock"
+
+
+@contextmanager
+def file_lock(path: Path, *, timeout_s: float = 30.0) -> Iterator[None]:
+    """Acquire an exclusive cross-process lock backed by ``path``.
+
+    Open with ``O_CREAT | O_RDWR`` (never ``O_TRUNC`` — concurrent acquirers
+    would race on truncation). The lock file persists across runs; that is
+    fine: an existing inode with no active kernel lock is the normal state.
+
+    Raises:
+        RefreshLockTimeout: When the lock cannot be acquired within ``timeout_s``.
+    """
+    fd = os.open(path, os.O_CREAT | os.O_RDWR, 0o600)
+    try:
+        if sys.platform == "win32":
+            _acquire_windows(fd, timeout_s)
+        else:
+            _acquire_posix(fd, timeout_s)
+        try:
+            yield
+        finally:
+            _release(fd)
+    finally:
+        os.close(fd)
+
+
+def _acquire_posix(fd: int, timeout_s: float) -> None:
+    import fcntl
+
+    deadline = time.monotonic() + timeout_s
+    while True:
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            return
+        except BlockingIOError:
+            if time.monotonic() >= deadline:
+                raise RefreshLockTimeout(
+                    f"refresh lock at fd {fd} not acquired within {timeout_s:.1f}s"
+                ) from None
+            time.sleep(_POLL_INTERVAL_S)
+
+
+def _acquire_windows(fd: int, timeout_s: float) -> None:
+    import msvcrt
+
+    deadline = time.monotonic() + timeout_s
+    while True:
+        try:
+            msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+            return
+        except OSError:
+            if time.monotonic() >= deadline:
+                raise RefreshLockTimeout(
+                    f"refresh lock at fd {fd} not acquired within {timeout_s:.1f}s"
+                ) from None
+            time.sleep(_POLL_INTERVAL_S)
+
+
+def _release(fd: int) -> None:
+    # Swallow release errors so they never mask an exception raised inside the
+    # ``with`` block; the kernel releases the lock when the FD closes anyway.
+    try:
+        if sys.platform == "win32":
+            import msvcrt
+
+            msvcrt.locking(fd, msvcrt.LK_UNLCK, 1)
+        else:
+            import fcntl
+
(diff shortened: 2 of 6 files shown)


adriannoes (approved):
## Summary

Solid concurrency fix — cross-process filesystem lock with double-checked re-load inside the critical section, clean `RefreshLockTimeout` → `RefreshError` boundary, and strong POSIX test coverage (including SIGKILL subprocess recovery). Review included a local checkout (`db44017`): `ruff check` clean, 25/25 targeted tests passed; CI `lint` + `test` green. Architecture boundaries look good (`RefreshLockTimeout` intentionally not exported; no leak into CLI/MCP/SDK). **Approved** — merge when ready.

## Optional follow-ups (non-blocking)

These are suggestions for a follow-up PR or whenever convenient — no need to block this merge:

- **F1 — Windows path coverage:** The Windows branches in `locks.py` (`refresh_lock_path` APPDATA path + `_acquire_windows`) never run in CI and `test_file_lock_released_after_sigkill` is skipped on win32 without a compensating test. A small monkeypatch test (`sys.platform = "win32"`, set `APPDATA`, assert the resolved path) would lock in the path-resolution branch without needing a Windows CI runner. Defer the full Windows matrix job if you prefer.

- **F2 — CHANGELOG bullet:** Worth a line under `[Unreleased] → Fixed` so release notes reflect the race fix for #133. The pre-PR behavior was a real `invalid_grant` storm under `xargs -P` / parallel `pipefy` invocations, and your piporacle smoke proves the fix end-to-end.

- **F3 — Global mutex trade-off:** `refresh_lock_path()` is a single global mutex across all issuer/client pairs, while the keychain is keyed per `(issuer, client_id)`. For one IdP + `pipefy-cli` that's a reasonable simplification, but the docstring doesn't say so explicitly. Either a one-liner ("single lock per host; multi-account isn't a goal today") or hashing `(issuer, client_id)` into the filename would make the trade-off legible. No action needed if the current scope is permanent.

- **F4 — Autouse lock isolation scope:** The autouse `_isolate_refresh_lock` only protects auth tests. Once PR #226 lands, the MCP container starts calling `ensure_fresh_session` from its own test suite. A future test that doesn't fully mock the call path would touch `~/.config/pipefy/refresh.lock` on the developer's box. Cheap fix: lift the fixture to a root `conftest.py` (or a shared helper) so the isolation travels with `ensure_fresh_session`. Defer if you want to keep this PR narrow.

## Verification

```bash
uv run pytest packages/auth -m "not integration"
uv run ruff check packages/auth
```

gbrlcustodio:
Thanks for the review.

**F1**: addressed in d3c6770 — two monkeypatched tests cover the Windows `APPDATA`-set and `APPDATA`-unset path branches.
**F2**: addressed in d3c6770 — `[Unreleased] → Fixed` bullet.
**F3**: addressed in d3c6770 — docstring line on `refresh_lock_path` noting the per-host (not per-`(issuer, client_id)`) scope.
**F4**: deferred — PR #226 already mocks `ensure_fresh_session` so there's no immediate exposure. Worth lifting the fixture once a non-mocked caller actually lands.

adriannoes:
LGTM

PR #224, opened 2026-05-23
fix(mcp/cli-tests): wire live fixtures through resolve_pipefy_auth

gbrlcustodio:
Companion to #223. Same root cause (live-integration fixtures drifted off post-#218 / post-#220 invariants), surfaced by reviewer's out-of-scope note on [#223's review](https://github.com/gbrlcustodio/pipefy-mcp-server/pull/223#pullrequestreview-4349207940).

**Stacked on #223** — uses the ``live_resolved_auth`` / ``live_auth_settings`` helpers it adds to ``packages/sdk/tests/_shared/live_settings.py``. Will rebase cleanly once #223 merges.

## Summary

- ``packages/mcp/tests/tools/test_introspection_tools_live.py`` — fixture passes ``auth=live_resolved_auth()`` to ``PipefyClient``.
- ``packages/mcp/tests/tools/test_attachment_tools_live.py`` — extract a module-level ``_live_pipefy_client`` helper (one source of truth for the five ``PipefyClient(settings=settings.pipefy)`` sites inside the S3 / upload tests) so each call goes through ``live_resolved_auth``.
- ``packages/cli/tests/test_auth_integration.py`` — ``AuthContext`` was missing the ``service_account=`` field (required, no default post-#218); now built from ``live_auth_settings().to_service_account()`` + ``to_oidc_client()``.

## Test plan

- [x] ``uv run pytest --collect-only`` on the three files — 11 tests collect without error
- [x] ``uv run pytest -m "not integration"`` — 2356 passing
- [x] ``uv run ruff check packages && uv run ruff format --check packages`` — clean
- [ ] Local ``uv run pytest -m integration`` with credentials — confirm at next live run

PR #226, opened 2026-05-23
feat(mcp): wire stored-session tier into ServicesContainer

gbrlcustodio:
## Summary

Final step of #213 (umbrella). The MCP server now reaches the keychain-backed user session that PR-A (#216), PR-B (#217), the auth-settings split (#218), and PR-C (#225) put in place.

- ``ServicesContainer.initialize_services`` passes ``oidc_client=settings.auth.to_oidc_client()`` to ``resolve_pipefy_auth``, so the stored-session tier participates in the precedence chain.
- When the resolved tier is the stored session, ``ensure_fresh_session`` runs eagerly at lifespan init (wrapped in ``asyncio.to_thread``). Stale or revoked sessions surface as a startup error rather than as a confusing failure on the first MCP tool call.
- Higher-precedence tiers (``PIPEFY_TOKEN`` static-token, ``PIPEFY_SERVICE_ACCOUNT_*``) skip the warm-up; nothing about their wiring changes.
- ``initialize_services`` becomes ``async`` to host the awaited warm-up; the only caller (``lifespan`` in ``server.py``) is already async. ``test_server.py`` mocks swap ``MagicMock().initialize_services`` for ``AsyncMock``.

No new env vars: ``PIPEFY_AUTH_URL`` and ``PIPEFY_AUTH_CLIENT_ID`` already exist on ``AuthSettings`` (introduced by #218); this PR is the consumer.

## Stacking

Stacked on ``#224`` so the live-fixture rewrites in ``packages/mcp/tests/`` don't conflict. Once ``#224`` lands this rebases trivially onto ``dev``. ``#225`` (cross-process refresh lock) is independent — the lock is internal to ``ensure_fresh_session`` and the API surface used here is unchanged.

## Test plan

- [x] ``uv run pytest packages/mcp packages/auth -m "not integration"`` — green
- [x] ``uv run pytest -m "not integration"`` — full repo green (2358 passed)
- [x] ``uv run ruff check packages && uv run ruff format --check packages`` — clean
- [ ] Manual: run the MCP server with a stored session active (no ``PIPEFY_TOKEN`` / ``PIPEFY_SERVICE_ACCOUNT_*``), confirm a tool call succeeds and the startup log emits ``Pipefy stored session warmed up at startup``.
- [ ] Manual: run the MCP server with an expired/revoked stored session, confirm the lifespan raises a clear ``RefreshError`` instead of failing on the first tool call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - **CLI**: `pipefy auth status [--json|-j]` — reports auth source, identity, session expiry, and exit codes.
 - **CLI**: `pipefy auth logout` — revokes the refresh token at the IdP and clears the stored session.
+- **MCP**: stored-session tier wired into `ServicesContainer`; setting `PIPEFY_AUTH_URL` after `pipefy auth login` now lets the MCP server reuse the keychain-backed session, with the refresh pre-warmed at startup so a stale or revoked session surfaces before the first tool call.
 
 ### Changed
 
--- a/packages/mcp/src/pipefy_mcp/core/container.py
+++ b/packages/mcp/src/pipefy_mcp/core/container.py
@@ -58,10 +58,11 @@ class ServicesContainer:
                     issuer=oidc_client.issuer_url,
                     client_id=oidc_client.client_id,
                 )
-            except RefreshError:
+            except RefreshError as exc:
                 logger.error(
-                    "Stored Pipefy session could not be refreshed at startup. "
-                    "Run `pipefy auth login` to sign in again."
+                    "Stored Pipefy session could not be refreshed at startup: %s. "
+                    "Run `pipefy auth login` to sign in again.",
+                    exc,
                 )
                 raise
             logger.info("Pipefy stored session warmed up at startup")
--- a/packages/mcp/src/pipefy_mcp/core/container.py
+++ b/packages/mcp/src/pipefy_mcp/core/container.py
@@ -6,6 +6,7 @@ from typing import Self
 
 from pipefy_auth import (
     STORED_SESSION_TIER,
+    RefreshError,
     ensure_fresh_session,
     missing_auth_message,
     resolve_pipefy_auth,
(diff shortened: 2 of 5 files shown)


adriannoes (approved):
## Summary

Clean, focused wiring PR — passes `oidc_client` into `resolve_pipefy_auth`, eagerly warms stored sessions at lifespan init via `asyncio.to_thread`, and migrates `initialize_services` to async with matching test updates. Review included a local checkout (`bea0e12`): `ruff check` clean, 17/17 targeted tests passed. Architecture boundaries look good (MCP → `pipefy_auth` + `pipefy_sdk`, no CLI import). **Approved** — merge when the stack is ready.

## Optional follow-ups (non-blocking)

These are suggestions for a follow-up PR or whenever convenient — no need to block this merge:

- **F1 — RefreshError unit test:** The warm-up path is the main behavioral change, and the manual test plan already calls out a revoked session. Worth locking that in with a unit test: if `ensure_fresh_session` raises `RefreshError`, init should abort before `PipefyClient` is constructed (same contract as the CLI warm-up in `pipefy_cli/auth.py`). A small test with `side_effect=RefreshError(...)` plus `mock_pipefy_client_class.assert_not_called()` would make regressions on the fail-fast story obvious in CI.

- **F2 — RefreshError UX hint:** On a stale keychain session the warm-up will raise `RefreshError` and the lifespan will abort — that's the right fail-fast behavior. The CLI wraps this with a "run `pipefy auth login`" hint before exiting; here the operator mostly sees the generic lifespan fatal log plus the raw exception. If you want parity for MCP-only users, a thin catch that logs or re-raises with that hint would save debugging time. Fine to defer.

- **F3 — CHANGELOG bullet:** Worth one line under `[Unreleased] → Added` so release notes match the CLI bullets you already have for stored-session work — e.g. *MCP: stored-session tier wired into `ServicesContainer`; setting `PIPEFY_AUTH_URL` after `pipefy auth login` now lets the MCP server reuse the keychain-backed session, with refresh pre-warmed at startup.*

## Verification

```bash
uv run pytest packages/mcp packages/auth -m "not integration"
uv run ruff check packages/mcp packages/auth
```

gbrlcustodio:
Thanks for the review.

**F1**: addressed in 68c9174 — `test_initialize_services_aborts_before_client_when_refresh_fails` pins the fail-fast contract (`RefreshError` propagates; `PipefyClient` not constructed).
**F2**: addressed in 68c9174 — `initialize_services` wraps the warm-up in a `RefreshError` catch that logs the "run `pipefy auth login`" hint before re-raising. Mirrors the CLI's contract in `pipefy_cli.auth.get_pipefy_client_async`.
**F3**: addressed in 68c9174 — `[Unreleased] → Added` bullet for the MCP stored-session wiring.

PR #227, opened 2026-05-25
feat(auth): reactive refresh-on-401 for stored-session tier

### LINKED ISSUE #137: Add reactive refresh-on-401 (complement to eager refresh)
## Motivation

PR #131 wired **eager** refresh: before each ``pipefy <cmd>``, ``ensure_fresh_session`` checks the stored access token's age and refreshes if it's within the 60s leeway window. This covers the common case (one-shot CLI invocations across the lifetime of a refresh token) with zero SDK contact and minimal complexity.

It does **not** cover:

- **IdP-side revocation mid-session**: an admin force-logs the user out; the access token still has 4 hours left by our clock but the next API call 401s. Eager check passes; API call fails; user sees a generic error instead of a "session invalidated; please re-login" prompt.
- **Long-running processes** (e.g. MCP server consuming user-auth in a future world): the access token genuinely expires while the process is running. Eager check ran 30 minutes ago and passed; current API call now 401s.

Reactive refresh would fill these gaps: a custom ``httpx.Auth`` class that intercepts 401 responses, attempts a refresh, retries the original request once.

## Proposed approach

New module: ``oauth/auth_class.py`` (or extend ``refresh.py``).

```python
class RefreshableBearerAuth(httpx.Auth):
    \"\"\"Reads access token from a callable; on 401, refreshes and retries once.\"\"\"

    def __init__(self, get_session: Callable[[], StoredSession]):
        self._get_session = get_session

    def sync_auth_flow(self, request):
        session = self._get_session()
        request.headers["Authorization"] = f"Bearer {session.access_token}"
        response = yield request
        if response.status_code != 401:
            return
        # Try refresh once, retry.
        refreshed = ensure_fresh_session(
            issuer=session.issuer,
            client_id=session.client_id,
            leeway_s=session.expires_in,  # force refresh
        )
        if refreshed is None or refreshed.access_token == session.access_token:
            return  # refresh failed or didn't change anything; let the 401 propagate
        request.headers["Authorization"] = f"Bearer {refreshed.access_token}"
        yield request
```

This auth class would plug into ``PipefyClient`` instead of ``StaticBearerAuth`` for the user-session path.

## Tradeoffs

**Pros:**
- Catches IdP-side revocation gracefully.
- Recovers from "the eager check happened, but the IdP rotated tokens between then and now" race conditions.
- Required for long-lived consumers of user-auth (future MCP-over-user-auth).

**Cons:**
- New SDK touch surface (``PipefyClient`` needs to accept a refresh-aware auth class; today it only takes a static bearer or builds ``OAuth2ClientCredentials`` itself).
- Reactive + eager together mean two refresh paths to reason about; needs careful interaction with the in-process cache.
- ``InternalApiClient`` would need the same treatment (#134 also depends on this question).

## Approach decision

Two paths to consider:

1. **Build it ourselves** — full control; one extra class.
2. **Adopt ``httpx_auth.OAuth2AuthorizationCodeRefreshToken``** — already in the dependency tree (the SDK uses ``httpx_auth.OAuth2ClientCredentials``). Need to verify its API fits our "load refresh token from keychain at startup, persist rotated tokens" lifecycle. Discussed during PR #131 planning; deferred pending an API check.

Option 2 is preferred if the library fits.

## Out of scope

- Replacing eager refresh entirely. Both can coexist: eager covers the predictable case at near-zero cost; reactive handles the edge cases.
- ``InternalApiClient`` over user-auth (tracked in #134; the auth class designed here should plug into both clients).

## Acceptance criteria

- [ ] User-auth path uses a refresh-aware ``httpx.Auth`` so that a 401 mid-call triggers refresh + retry.
- [ ] Eager refresh remains the primary path; reactive is the safety net.
- [ ] On reactive refresh, rotated tokens are persisted back to the keychain (so future commands see them).
- [ ] On reactive refresh failure, the 401 propagates with a clear "session expired" message — not a silent retry loop.
- [ ] Tests cover: happy 401-refresh-retry, refresh fails on the reactive path, the auth class doesn't loop forever.

## References

- PR #131 (introduces eager refresh; explicitly defers reactive to this issue)
- #133 (concurrent-refresh race — orthogonal; both this and #133 touch the refresh path but solve different problems)
- #134 (``InternalApiClient`` over bearer — likely consumes the same auth class)

Danielmoraisg:
## Summary

PR #131 wired **eager** refresh: `ensure_fresh_session` checks the access token's clock-side age before each request through `CallableBearerAuth._stored_session_provider`. It does not cover **IdP-side revocation**: an admin force-logs the user out, the token still looks fresh by our clock, the next API call 401s and surfaces as a generic transport error instead of a "session expired" prompt.

This adds the reactive safety net: when an API call returns 401, force a refresh, retry the request once if the bearer changed, persist any rotated tokens. Eager refresh stays the primary path.

## What's new

- **`ensure_fresh_session(force=True)`** — bypasses the freshness deadline; reuses the carry-forward + keychain persistence path. Failure semantics unchanged: `RefreshError` on IdP failure, `None` when no session is stored.
- **`RefreshableBearerAuth(httpx.Auth)`** — per-request bearer + 401-refresh-retry hook. Sync (`auth_flow`) and async (`async_auth_flow`). Refreshes serialized via `asyncio.Lock`. The class trusts its `force_refresh: Callable[[], str | None]` contract — translating an IdP `RefreshError` into `None` is the resolver's job (the imperative shell that does the keychain + IdP I/O), keeping the auth flow free of error handling.
- **Resolver wiring** — `_stored_session_provider` returns `RefreshableBearerAuth`; its `force_refresh` callable runs `ensure_fresh_session(force=True)` and maps `RefreshError`/no-session to `None`. `CallableBearerAuth` stays a public primitive and `tier_for(CallableBearerAuth())` still resolves to `STORED_SESSION_TIER` (regression test included).

## Concurrency model (serialize, don't coalesce)

Three concurrent 401s run **three** refresh POSTs back-to-back under the lock, not one shared refresh. Coalescing belongs to the cross-process mutex tracked in #133; the in-class lock only guarantees refreshes are not interleaved. `test_concurrent_401s_serialize_under_async_lock` probes the lock with `threading.Barrier(parties=3)` — deterministic, no sleep-based race-widening.

## Out of scope

- Concurrent-refresh race / filesystem mutex (#133).
- Replacing eager refresh — both coexist; reactive is the safety net.
- `StaticBearerAuth` / `OAuth2ClientCredentials` paths — untouched.

## Test plan

- [x] `uv run pytest packages/auth/tests -m "not integration"` — auth package green (65 tests; sync + async behavior matrix, deterministic barrier-based concurrency probe)
- [x] `uv run pytest packages/cli/tests/test_auth.py` + `packages/mcp/tests/core/test_container.py` — stored-session warm-up asserts `RefreshableBearerAuth`
- [x] `uv run pytest -m "not integration"` (whole repo) — green except `test_initialize_services_raises_when_no_auth_source_configured`, a pre-existing failure on `dev` too (repo `.env` leaks `PIPEFY_OAUTH_*` into `AuthSettings()`; passes hermetically and in CI, which has no `.env`)
- [x] `uv run ruff check` + `ruff format --check` on touched packages — clean
- [x] Manual E2E against piporacle — corrupted the stored access token, ran `pipefy pipe list`; wire log showed `POST /graphql 401 → GET /.well-known → POST /token 200 → POST /graphql 200`, keychain access_token rotated, real data returned

Closes #137.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/auth/src/pipefy_auth/resolver.py
+++ b/packages/auth/src/pipefy_auth/resolver.py
@@ -18,9 +18,13 @@ from dataclasses import dataclass
 from httpx import Auth
 from httpx_auth import OAuth2ClientCredentials
 
-from pipefy_auth.bearer import CallableBearerAuth, StaticBearerAuth
+from pipefy_auth.bearer import (
+    CallableBearerAuth,
+    RefreshableBearerAuth,
+    StaticBearerAuth,
+)
 from pipefy_auth.identity import OidcClient
-from pipefy_auth.refresh import ensure_fresh_session
+from pipefy_auth.refresh import RefreshError, ensure_fresh_session
 from pipefy_auth.storage import load_session
 
 STATIC_TOKEN_TIER = "static-token"
@@ -40,10 +44,13 @@ class ServiceAccount:
 # Maps each ``httpx.Auth`` implementation back to the resolver-tier name that
 # produced it. ``tier_for`` is the public lookup; the mapping itself is the
 # single source of truth so the wire-schema strings (e.g. ``"stored-session"``)
-# stay in one place.
+# stay in one place. ``CallableBearerAuth`` is listed alongside
+# ``RefreshableBearerAuth`` so consumers who build one directly still resolve
+# to the stored-session tier.
 _TIER_BY_AUTH_TYPE: dict[type[Auth], str] = {
     StaticBearerAuth: STATIC_TOKEN_TIER,
     OAuth2ClientCredentials: SERVICE_ACCOUNT_TIER,
+    RefreshableBearerAuth: STORED_SESSION_TIER,
     CallableBearerAuth: STORED_SESSION_TIER,
 }
 
@@ -64,8 +71,8 @@ def tier_for(auth: Auth) -> str:
     )
 
 
-def _stored_session_provider(oidc_client: OidcClient) -> CallableBearerAuth:
-    def _provider() -> str:
+def _stored_session_provider(oidc_client: OidcClient) -> RefreshableBearerAuth:
+    def _token() -> str:
         session = ensure_fresh_session(
             issuer=oidc_client.issuer_url, client_id=oidc_client.client_id
         )
@@ -75,7 +82,18 @@ def _stored_session_provider(oidc_client: OidcClient) -> CallableBearerAuth:
             )
         return session.access_token
 
-    return CallableBearerAuth(_provider)
+    def _force_refresh() -> str | None:
+        try:
+            session = ensure_fresh_session(
+                issuer=oidc_client.issuer_url,
+                client_id=oidc_client.client_id,
+                force=True,
+            )
+        except RefreshError:
+            return None
+        return session.access_token if session is not None else None
+
+    return RefreshableBearerAuth(token_provider=_token, force_refresh=_force_refresh)
 
 
 def _has_stored_session(oidc_client: OidcClient) -> bool:
@@ -108,7 +126,9 @@ def resolve_pipefy_auth(
         service_account: Service-account client-credentials inputs.
         oidc_client: OIDC client identity for the stored-session tier; the
             session is loaded from the keychain at detection time, and a fresh
-            access token is fetched per request via :class:`CallableBearerAuth`.
+            access token is fetched per request via
+            :class:`pipefy_auth.bearer.RefreshableBearerAuth`, which also
+            forces a refresh + retry on a 401 response.
     """
     if static_token and static_token.strip():
         return StaticBearerAuth(static_token.strip())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **CLI**: `pipefy auth status [--json|-j]` — reports auth source, identity, session expiry, and exit codes.
 - **CLI**: `pipefy auth logout` — revokes the refresh token at the IdP and clears the stored session.
 - **MCP**: stored-session tier wired into `ServicesContainer`; setting `PIPEFY_AUTH_URL` after `pipefy auth login` now lets the MCP server reuse the keychain-backed session, with the refresh pre-warmed at startup so a stale or revoked session surfaces before the first tool call.
+- **Auth**: reactive refresh-on-401 for the stored-session tier (`RefreshableBearerAuth`). When an API call returns 401, the SDK transport forces a refresh via `ensure_fresh_session(force=True)`, persists any rotated tokens, and retries the request once before surfacing the error. Complements the eager refresh path for IdP-side revocation and mid-process token expiry. Closes #137.
 
 ### Changed
 
--- a/packages/auth/src/pipefy_auth/__init__.py
+++ b/packages/auth/src/pipefy_auth/__init__.py
@@ -11,7 +11,11 @@ from __future__ import annotations
 
 __version__ = "0.2.0-beta.1"
 
-from pipefy_auth.bearer import CallableBearerAuth, StaticBearerAuth
+from pipefy_auth.bearer import (
+    CallableBearerAuth,
+    RefreshableBearerAuth,
+    StaticBearerAuth,
+)
 from pipefy_auth.discovery import (
     DiscoveryPolicy,
     ProviderMetadata,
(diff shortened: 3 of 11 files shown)


packages/auth/src/pipefy_auth/resolver.py
```diff
@@ -75,7 +82,18 @@ def _provider() -> str:
...
+            session = ensure_fresh_session(
+                issuer=oidc_client.issuer_url,
+                client_id=oidc_client.client_id,
+                force=True,
+            )
+        except RefreshError:
+            return None
+        return session.access_token if session is not None else None
+
+    return RefreshableBearerAuth(token_provider=_token, force_refresh=_force_refresh)
```
adriannoes, line 96: Wiring looks good here — one follow-up outside this hunk: `packages/mcp/tests/core/test_container.py` still asserts `CallableBearerAuth` in `test_initialize_services_warms_up_stored_session` (line 221), but the resolver now returns `RefreshableBearerAuth` (you already updated `packages/cli/tests/test_auth.py`). 

Please align the MCP warm-up test so CI doesn’t fail when the stored-session tier is exercised.

adriannoes (changes requested):
Solid work on the reactive 401 safety net — `RefreshableBearerAuth`, `ensure_fresh_session(force=True)`, and the auth package tests are thorough and match the serialize-not-coalesce contract. 

I checked out `9286db6` locally: `packages/auth/tests` and `packages/cli/tests/test_auth.py` pass cleanly; ruff is clean on the touched packages.

One regression guard needs a small follow-up before merge.

## Required before merge

**MCP container test type:** `_stored_session_provider` now returns `RefreshableBearerAuth`, but `packages/mcp/tests/core/test_container.py` still `isinstance(..., CallableBearerAuth)` in `test_initialize_services_warms_up_stored_session`. CLI was already updated; please mirror that in MCP so warm-up tests match what `ServicesContainer` passes to `PipefyClient`.

## Also noted

**Follow-up, not blocking:** With `force=True`, the in-lock freshness skip is bypassed, so back-to-back reactive 401s can each POST to the IdP even after a peer refreshed under `file_lock`. That fits your documented model until coalescing lands; worth keeping in mind if refresh-token rotation gets stricter.

packages/auth/src/pipefy_auth/resolver.py
Danielmoraisg: Fixed in `0f89712` — `test_initialize_services_warms_up_stored_session` now asserts `RefreshableBearerAuth`, mirroring the CLI test. Touched-package sweep is green (88 passed).

On the `force=True` follow-up: agreed, that's the documented serialize-not-coalesce behavior — each reactive 401 forces its own grant even after a peer refreshed under `file_lock`. If refresh-token rotation gets stricter, the in-class lock could re-check the keychain after acquiring it (use the peer's rotated token instead of POSTing again)

gbrlcustodio (approved):
LGTM. Reactive 401 path complements eager refresh cleanly. Force-refresh contract (re-check inside FS lock, RefreshError surfaces, None on no-session) consistent with existing semantics. Serialize-not-coalesce documented and tested. Tier mapping keeps CallableBearerAuth as public primitive. Test coverage thorough — sync + async behavior matrix + deterministic barrier-based concurrency probe.

Non-blocking nits:
- bearer.py: AsyncGenerator imported from typing; rest of module uses collections.abc. Pre-existing inconsistency, worth normalizing.
- _safe_force_refresh catches RefreshError that resolver's _force_refresh already swallows. Defense-in-depth for direct constructor users; either drop one layer or note in docstring.
- PR body says _safe_force_refresh narrows to (RefreshError, RuntimeError); code catches only RefreshError. Update body wording.

gbrlcustodio (approved):

Danielmoraisg:
Thanks for the review — all three nits addressed in `9ed8127`:

1. **`AsyncGenerator` import** — moved to `collections.abc` alongside `Callable`/`Generator`; dropped the `typing` import.
2. **Redundant `RefreshError` catch** — dropped the layer rather than documenting it. The auth class now trusts its `force_refresh: Callable[[], str | None]` contract and calls it directly; translating an IdP `RefreshError` into `None` stays solely in the resolver (the imperative shell doing the keychain + IdP I/O). `_safe_force_refresh` is gone. The two bearer tests that pinned the old swallow-behavior now assert the contract directly: a raising `force_refresh` surfaces rather than being silently turned into a 401.
3. **PR body wording** — updated; the stale `(RefreshError, RuntimeError)` line is removed (the method it referenced no longer exists).

Touched-package tests green (auth 65, cli_auth 14, mcp container warm-up passes).

PR #228, opened 2026-05-25
feat(portal): portal page CRUD — SDK, MCP, and CLI (task 4.0)

### LINKED ISSUE #209: Portal CRUD 4.0 — Portal pages (CRUD, sort, layout)
## Goal

Page management: **create, update, delete, sort, and layout update** for portal pages.

## Context

- **Depends on:** #208 (Task 3.0) merged into `dev`
- **Enables:** #210 (Task 5.0 needs `page_id`)
- **Tools:** 5 MCP tools | `pipefy portal page …`
- **Branch:** `feat/portal-crud/4-portal-pages` from `dev`
- **Suggested PR:** 4 → merges into `dev`

## Acceptance criteria

- [ ] Five operations shipped SDK + MCP + CLI
- [ ] Input shapes follow `InterfacePage*MutationInput` exactly (snake_case `interface_uuid`, `page_id`, `page_ids`)
- [ ] `update_portal_page_layout` only requires `page_id` + `layout: Json!` (no `interface_uuid`)
- [ ] CLI page delete requires `--yes`
- [ ] Five parity matrix rows

## Sub-tasks (TDD order)

- [ ] **4.1 RED** — SDK tests for five page operations
- [ ] **4.2 GREEN** — SDK implementation
- [ ] **4.3 RED** — MCP tests for five page tools
- [ ] **4.4 GREEN** — MCP page tools
- [ ] **4.5 RED** — CLI tests for `portal page` subcommands
- [ ] **4.6 GREEN** — CLI page subcommands
- [ ] **4.7 REFACTOR** — Parity rows + ruff

## Verify

```bash
uv run pytest -m "not integration" -k portal_page -v
```

adriannoes:
## Summary

- Ship five portal page capabilities on the Interfaces schema: create, update, delete, sort, and layout update (SDK → MCP → CLI).
- Add parity rows and `docs/mcp/tools/portal.md` for the 10 portal tools (read + CRUD + five page tools).
- Harden sort/layout MCP envelopes (success only when `sortPages` / `updatePageLayout` return `success: true`) and validate `page_ids` / CLI sort ID inputs (code review F1/F2).

Closes #209

## Test plan

- [x] `uv run ruff check packages/sdk/src packages/mcp/src packages/cli/src packages/auth/src`
- [x] `uv run ruff format --check` (same paths)
- [x] `uv run pytest packages/sdk/tests/services/pipefy/test_portal_service.py packages/mcp/tests/tools/test_portal_tools.py packages/cli/tests/test_portal_commands.py tests/test_parity.py -m "not integration"` — **243 passed**
- [x] `uv run python .github/workflows/scripts/lint_skill_refs.py` — passed
- [x] `uv build packages/mcp` — wheel OK
- [x] Manual MCP smoke (PipeClaw org `302398434`, main portal `51831a06-…`): create → update → sort → delete preview/execute; layout deferred (no real layout blob in `get_portal`)

**Note:** Full `uv run pytest -m "not integration"` may fail locally if `.env` provides auth (`test_initialize_services_raises_when_no_auth_source_configured`); CI runs without that env and should stay green.


packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -238,3 +238,242 @@ async def delete_portal(
...
+                layout: Layout JSON (full layout object for the page).
+            """
+            page_id, err = validate_tool_id(page_id, "page_id")
+            if err is not None:
+                return err
+            await ctx.debug(f"update_portal_page_layout: page_id={page_id}")
+            try:
+                result = await client.update_portal_page_layout(page_id, layout)
+            except Exception as exc:  # noqa: BLE001
+                return build_error_payload(str(exc))
```
Danielmoraisg: **Important — inconsistent portal-error mapping.**

All four sibling page tools (`create_portal_page`, `update_portal_page`, `delete_portal_page`, `sort_portal_pages`) wrap exceptions with `build_error_payload(map_portal_error_to_message(exc))`, but here it's `build_error_payload(str(exc))`. The SDK's `_execute_interfaces_query_with_portal_errors` still raises `PortalPermissionError` on `PERMISSION_DENIED`, so a permission failure here will leak the raw `TransportQueryError` message instead of the actionable "Request organization permissions such as `create_portal` or `manage_portals`…" guidance the other tools surface.

Suggest:
```python
    return build_error_payload(map_portal_error_to_message(exc))
```
…and add a `test_update_portal_page_layout_permission_denied_surfaces_actionable_message` analogous to the existing `test_create_portal_page_permission_denied_…` to lock the behavior in.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -238,3 +238,242 @@ async def delete_portal(
...
+                return tool_error(
+                    "Invalid 'page_ids': must be a non-empty list of page UUIDs.",
+                    code="INVALID_ARGUMENTS",
+                )
+            cleaned_page_ids: list[str] = []
+            for page_id in page_ids:
+                cleaned_id, id_err = validate_tool_id(page_id, "page_ids")
+                if id_err is not None:
+                    return id_err
+                cleaned_page_ids.append(cleaned_id)
```
Danielmoraisg, line 442: **Important — `page_ids` does not reject duplicates.**

The PR description's input-validation contract calls for "no duplicates" on `page_ids`. Each item is validated individually, but `["a", "a", "b"]` will pass straight through to `sortPages`, where the server's behavior on duplicate IDs is unspecified (likely either an opaque GraphQL error or a silent reorder using last-seen). Cheap to defend in depth at the boundary:

```python
if len(set(cleaned_page_ids)) != len(cleaned_page_ids):
    return tool_error(
        "Invalid 'page_ids': must not contain duplicate page UUIDs.",
        code="INVALID_ARGUMENTS",
    )
```

Same fix is needed on the CLI side in `_parse_page_ids_for_sort` (`packages/cli/src/pipefy_cli/commands/portal.py:61`) — raise `typer.BadParameter` if the deduped list is shorter. Extend `test_sort_portal_pages_rejects_invalid_page_id_items` with a duplicate-ID case.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -238,3 +238,242 @@ async def delete_portal(
...
+                return build_success_payload(result, include_parsed=True)
+            return build_error_payload(
+                f"Failed to delete portal page '{page_id}'. "
+                "Please try again or contact support."
+            )
+
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=False),
+        )
+        async def sort_portal_pages(
```
Danielmoraisg, line 418: **Important — plan/implementation divergence on `page_ids` shape validation.**

The PR plan describes `page_ids` validation as "every entry a UUID", but `validate_tool_id` (`packages/mcp/src/pipefy_mcp/tools/validation_helpers.py`) only enforces "non-empty string or positive integer" — `UUID_RE` exists in that module but is never invoked here. This is **consistent** with the rest of the portal CRUD (`portal_uuid` also goes through `validate_tool_id`, not `UUID_RE`), so the implementation isn't internally inconsistent — but it doesn't match the stricter spec in the PR description.

Two options, no strong preference:
1. Leave the validation as-is and tweak the PR body to say "non-empty string or positive integer" rather than "UUID" — matches the existing portal tooling and avoids tightening a contract Pipefy IDs don't actually require.
2. Tighten to `UUID_RE.fullmatch(...)` for the page-level identifiers (`page_id`, `page_ids[*]`) on the grounds that Interfaces-schema IDs really are UUID-shaped.

Flagging so the choice is intentional, not accidental.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
```diff
@@ -238,3 +238,242 @@ async def delete_portal(
...
+
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=False),
+        )
+        async def create_portal_page(
+            ctx: Context[ServerSession, None],
+            portal_uuid: str,
+            title: str,
+            description: str | None = None,
+            index: int | None = None,
```
Danielmoraisg, line 252: **Important — `index` is unvalidated at both MCP and CLI tiers.**

`create_portal_page` (here, line 250) and `update_portal_page` (line 301) accept `index: int | None = None` and forward it straight to the SDK. The CLI equivalents (`packages/cli/src/pipefy_cli/commands/portal.py:252` and `:292`) use `typer.Option(None, "--index", ...)` with no bounds. Negative or zero values will reach the GraphQL server and likely come back as an opaque error.

Defensive fix at the MCP boundary (mirrors the existing `INVALID_ARGUMENTS` style in this file):

```python
if index is not None and not isinstance(index, int) or (
    isinstance(index, int) and index < 0
):
    return tool_error(
        "Invalid 'index': must be a non-negative integer.",
        code="INVALID_ARGUMENTS",
    )
```

And on the CLI, switch the option to a `typer.Option(None, "--index", min=0, ...)`. Apply to both `create_portal_page` and `update_portal_page`.

(Whether `0` should be the lower bound or `1` depends on whether `sortPages` indexes are 0- or 1-based — worth a one-line confirm from someone with a live portal.)

Danielmoraisg (commented):
Solid end-to-end wiring across SDK → MCP → CLI; envelope-success semantics for `sortPages` / `updatePageLayout` / `deletePage` are correctly hardened. Leaving four "Important" inline comments — none are blocking on their own, but worth resolving before merge:

1. `update_portal_page_layout` skips `map_portal_error_to_message` (other four page tools use it).
2. `sort_portal_pages` doesn't reject duplicate `page_ids` (MCP + CLI).
3. `page_ids` validation is shape-loose (not `UUID_RE`) despite the PR plan wording — pick one and align.
4. `index` is unvalidated at both MCP and CLI tiers — fail fast at the boundary instead of relying on the server.

CI is green and the destructive-guard / parity / docs lockstep all look good. Leaving as `COMMENT` rather than `REQUEST_CHANGES` — none of these are critical, but please ack 1, 2, 4 before merge.

Danielmoraisg (approved):

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Addressed in 22ca891.

`update_portal_page_layout` now wraps exceptions with `map_portal_error_to_message(exc)` like the other four page tools, so `PERMISSION_DENIED` surfaces the same `create_portal` / `manage_portals` guidance.

Added `test_update_portal_page_layout_permission_denied_returns_actionable_error` (TransportQueryError with `PERMISSION_DENIED`).

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Addressed in 22ca891.

**MCP:** after per-item `validate_tool_id`, `validate_sort_page_ids_no_duplicates()` returns `INVALID_ARGUMENTS` when `len(set(ids)) != len(ids)`.

**CLI:** `_reject_duplicate_sort_page_ids()` in `_parse_page_ids_for_sort` for both `--page-ids` and `--ids-json` (`typer.BadParameter`).

**Tests:** `test_sort_portal_pages_rejects_duplicate_page_ids`, `test_portal_page_sort_page_ids_csv_rejects_duplicate_ids`, `test_portal_page_sort_ids_json_rejects_duplicate_ids`.

adriannoes:
Thanks @Danielmoraisg — all four inline **Important** items are addressed in **22ca891** (`fix(portal): address PR review on page tool validation`).

| Item | What we did |
|------|-------------|
| 1. Layout error mapping | `map_portal_error_to_message` on `update_portal_page_layout` + permission-denied test |
| 2. Duplicate `page_ids` | MCP + CLI boundary checks + tests (CSV and JSON paths) |
| 3. UUID vs `validate_tool_id` | Option 1: keep existing ID rules; document in `docs/mcp/tools/portal.md` |
| 4. `index` validation | `validate_portal_page_index` (MCP), `min=0` (CLI), tests |

`uv run pytest packages/mcp/tests/tools/test_portal_tool_helpers.py packages/mcp/tests/tools/test_portal_tools.py packages/cli/tests/test_portal_commands.py -m "not integration"` — 81 passed.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Addressed in 22ca891 — went with **option 1** (intentional).

Kept `validate_tool_id` for `page_id` / `page_ids[*]` to stay consistent with the rest of portal CRUD (`portal_uuid`, etc.). Updated `docs/mcp/tools/portal.md` **Input validation** table to document the contract explicitly: non-empty string or positive integer, not a strict UUID regex.

packages/mcp/src/pipefy_mcp/tools/portal_tools.py
adriannoes: Addressed in 22ca891 (items **1**, **2**, **4** ack).

**MCP:** `validate_portal_page_index()` on `create_portal_page` and `update_portal_page` — rejects negative values and `bool` (subclass of `int`).

**CLI:** `--index` uses `min=0` on both `page create` and `page update`.

**Tests:** MCP `test_create_portal_page_rejects_negative_index`, `test_update_portal_page_rejects_negative_index`; CLI `test_portal_page_create_rejects_negative_index`. Lower bound is `0` (non-negative); happy to tighten to `1` if we confirm sort indexes are 1-based on a live portal.

Issue #229, opened 2026-05-25
plugin: ship Claude Code plugin marketplace entry with /pipefy-login slash command


gbrlcustodio:
## Outcome

Users running Claude Code can install a `pipefy` plugin that ships a `/pipefy-login` slash command and registers the MCP server, regardless of whether the CLI is permanently installed.

```sh
/plugin marketplace add gbrlcustodio/pipefy-mcp-server
/plugin install pipefy
/pipefy-login
```

`/pipefy-login` prompts the user to confirm running `uvx --from git+https://github.com/gbrlcustodio/pipefy-mcp-server pipefy auth login` (one-shot, no permanent install). On approval, the existing OAuth flow in `pipefy-auth` writes the keychain entry that the MCP server's stored-session tier consumes.

## Scope

**New files at repo root**

- `.claude-plugin/plugin.json` — plugin manifest (`name`, `version`, `description`, `repository`, inline `mcpServers` entry, reference to `commands/`).
- `.claude-plugin/marketplace.json` — marketplace entry exposing the `pipefy` plugin from this repo.
- `commands/pipefy-login.md` — slash command markdown. Frontmatter:
  ```yaml
  ---
  name: pipefy-login
  description: Authenticate with Pipefy via OAuth (browser login) and store the session in the OS keychain.
  disable-model-invocation: true
  argument-hint: "[--no-browser] [--callback-timeout <seconds>]"
  ---
  ```
  Body instructs the model to run the `uvx --from git+https://github.com/gbrlcustodio/pipefy-mcp-server pipefy auth login $ARGUMENTS` command via the Bash tool. `allowed-tools` is intentionally omitted so the user confirms each invocation.

**Inline MCP server registration in `plugin.json`**

```json
"mcpServers": {
  "pipefy": {
    "command": "uvx",
    "args": ["--from", "git+https://github.com/gbrlcustodio/pipefy-mcp-server", "pipefy-mcp-server"]
  }
}
```

**MCP error-hint updates**

- `packages/auth/src/pipefy_auth/resolver.py` — extend `missing_auth_message()` with a `slash_command` parameter (default `\"/pipefy-login\"`); the message lists both surfaces. Update `_stored_session_provider()` (line ~73) RuntimeError text the same way.
[...]

PR #232, opened 2026-05-25
feat(plugin): ship Claude Code plugin with /pipefy:login

gbrlcustodio:
## Summary

- Adds a Claude Code plugin manifest (`.claude-plugin/plugin.json` + `.claude-plugin/marketplace.json`) plus external `.mcp.json` (auto-discovered at plugin root, matching the convention used by `linear` / `terraform` / `discord` in `claude-plugins-official`). Installing the plugin registers the Pipefy MCP server and the new `/pipefy:login` slash command in one step.
- `commands/login.md` runs the existing `pipefy auth login` flow through `uvx --from "git+...#subdirectory=packages/cli"`, so plugin users do not need a permanent `pipefy-cli` install to seed the keychain entry the MCP stored-session tier reads. The `#subdirectory=` pin sidesteps a build failure where `uvx`'s default builder tries to compile the workspace root `pyproject.toml` and trips over the flat-layout `skills/` + `packages/` directories.
- MCP error text in `packages/mcp/src/pipefy_mcp/core/container.py` no longer names a specific host; both the missing-auth `RuntimeError` and the startup-refresh-failure log point users at `docs/setup.md` via a new `DOCS_SETUP_REF` constant (mirrors the CLI's `_docs.py` pattern).
- README + `docs/setup.md` gain a "Plugin install (Claude Code)" block under Installation. Each block flags the restart-on-fresh-install case so users whose MCP server crashed at startup know to restart it after `/pipefy:login`.
- `.gitignore` carve-out (`!/.mcp.json`) so the plugin's tracked `.mcp.json` at repo root is committed while nested machine-specific Cursor configs stay ignored.

`pipefy-auth` and `pipefy-cli` are intentionally untouched — both stay host-agnostic. `/pipefy:login` is the Claude Code surface only; terminal users keep running `pipefy auth login` directly.

Refs #229 (milestone "Simplify install + skills + login UX"; workstream C of 3).

## Test plan

- [x] `uv run pytest packages/mcp/tests/core/test_container.py --no-cov` — 10/10 pass.
- [x] `uv run ruff check .` and `uv run ruff format --check .` — clean.
- [x] `uv run pipefy-mcp-server` with no `PIPEFY_*` env vars — stderr surfaces both `pipefy auth login` and `https://github.com/gbrlcustodio/pipefy-mcp-server/blob/main/docs/setup.md`.
- [x] Local plugin install via `claude plugin marketplace add <repo>` + `claude plugin install pipefy@pipefy` — `claude plugin details pipefy` reports `Skills (1) login` and `MCP servers (1) pipefy`. (Two bugs caught at this step and fixed: inline `mcpServers` in `plugin.json` was not loaded — moved to external `.mcp.json`; bare `uvx --from git+<repo>` failed to build the workspace root — switched to `#subdirectory=packages/<pkg>` pin.)
- [x] `/pipefy:login` invocation — slash command body runs `uvx --from "git+...#subdirectory=packages/cli" pipefy auth login $ARGUMENTS`; verified the underlying `uvx` command builds and surfaces `pipefy --help` / `pipefy auth login --help`. Full OAuth flow against a live IdP is user-side (browser approval).
- [ ] Remote install (post-merge) from any other directory: `/plugin marketplace add gbrlcustodio/pipefy-mcp-server` → `/plugin install pipefy` → MCP server registers and `/pipefy:login` works as above.

## Known follow-up (out of scope)

The existing `README.md` / `docs/setup.md` install commands of the form `uvx --from git+<repo> pipefy-cli` / `uv tool install "pipefy-cli @ git+<repo>"` hit the same workspace-root build failure surfaced here. They need the same `#subdirectory=packages/<pkg>` treatment in a follow-up PR; not part of this workstream's scope.


README.md
```diff
@@ -49,6 +49,16 @@ Feedback and issues: [GitHub Issues](https://github.com/gbrlcustodio/pipefy-mcp-
...
 
+### Plugin install (Claude Code)
+
+```text
+/plugin marketplace add gbrlcustodio/pipefy-mcp-server
+/plugin install pipefy
+/pipefy:login
+```
+
+`/plugin install pipefy` registers the MCP server and the `/pipefy:login` slash command. `/pipefy:login` runs the OAuth browser flow and stores the session in the OS keychain; on first run it persists the Pipefy CLI as a `uv tool install` so subsequent commands have a stable binary on PATH. Environment variables (`PIPEFY_GRAPHQL_URL`, `PIPEFY_AUTH_URL`, service-account credentials, etc.) are configured per **[`docs/setup.md`](docs/setup.md#claude-code)** — that file is the single source of truth so the names do not drift. Claude Code only; other hosts use the terminal flow below.
```
adriannoes: The README and `docs/setup.md` plugin blocks still say `/pipefy:login` installs the CLI on first use, but `commands/login.md` bails out when `pipefy` is not on PATH and points people at `/pipefy:install` instead — only `install.md` runs `uv tool install`. Someone following the three-line quickstart (marketplace → install pipefy → login) will hit that mismatch. Add `/pipefy:install` to the quickstart in both doc files, or rewrite the prose so it does not promise install-on-login. Fixed when documented steps match `login.md` and `install.md`.

packages/mcp/src/pipefy_mcp/_docs.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+DOCS_SETUP_REF = (
+    "https://github.com/gbrlcustodio/pipefy-mcp-server/blob/main/docs/setup.md"
```
adriannoes: `DOCS_SETUP_REF` points at `blob/main/docs/setup.md`, but this PR’s plugin docs and `.mcp.json` are on `@dev` — and `main` doesn’t have the plugin install section yet. Folks who see the new RuntimeError or refresh log will land on setup instructions that skip the flow they’re actually using. Align the URL with `@dev` (or the same ref you pin in `.mcp.json`) until those docs are on `main`.

.mcp.json
```diff
@@ -0,0 +1,16 @@
+{
+  "mcpServers": {
+    "pipefy": {
+      "command": "uvx",
+      "args": [
+        "--with",
+        "pipefy-sdk @ git+https://github.com/gbrlcustodio/pipefy-mcp-server@dev#subdirectory=packages/sdk",
```
adriannoes, line 7: The three `git+…@dev` pins are fine for this beta workstream, but they’re mutable — every `uvx` refresh can pull different code. When #234 lands, consider pinning the same tag as `plugin.json` `version` (or a commit SHA) so plugin installs stay reproducible across machines.

commands/login.md
```diff
@@ -0,0 +1,16 @@
...
+argument-hint: "[--no-browser] [--callback-timeout <seconds>]"
+---
+
+If `command -v pipefy` fails, stop and tell the user to run `/pipefy:install` first.
+
+Otherwise prompt the user to confirm running:
+
+```
+pipefy auth login $ARGUMENTS
+```
```
adriannoes, line 14: If Claude Code ever wraps this block in a shell (`sh -c`), unquoted `$ARGUMENTS` could mis-parse metacharacters. Low priority while `disable-model-invocation` is on, but if you keep `$ARGUMENTS`, documenting only safe flags (`--no-browser`, etc.) or an argv-safe pattern would harden the command.

adriannoes (approved):
Nice work shipping the Claude Code plugin end-to-end — external `.mcp.json`, `#subdirectory=` pins, slash commands, and thin MCP doc hints in `container.py` all look sound. I checked out head `3b67606` locally: `ruff` clean and `packages/mcp/tests/core/test_container.py` 10/10; CI lint/test are green on your branch.

Approving with a few doc-alignment notes inline (F1/F2 are the ones I’d fix in this PR or immediately after merge; the rest are beta follow-ups).

## Also noted (inline)

- **F1** — README/setup quickstart vs `commands/login.md` / `install.md` on when the CLI gets installed.
- **F2** — `DOCS_SETUP_REF` uses `main` while plugin content ships on `@dev`.
- **F3** — `@dev` git+ pins in `.mcp.json` (track with #234).
- **F4** — `$ARGUMENTS` hardening on `/pipefy:login` (optional).

## Using this review with AI

A fuller structured report (IDs, evidence, verification commands) lives at `.cursor/dev-planning/code-review/pr-232-claude-plugin-pipefy-login.md` in the repo for agents fixing items by ID.

## Verification

```bash
uv run ruff check .
uv run pytest packages/mcp/tests/core/test_container.py -m "not integration" -q
```

README.md
gbrlcustodio: **resolved** (6ff193b): added `/pipefy:install` between `/plugin install pipefy` and `/pipefy:login` in both README and docs/setup.md, and rewrote the prose to describe the install as a separate idempotent step instead of an on-first-run side-effect of login. Docs now match `commands/login.md`'s `command -v pipefy` bail-out.

packages/mcp/src/pipefy_mcp/_docs.py
gbrlcustodio: **resolved** (6ff193b): flipped `DOCS_SETUP_REF` to `blob/dev/docs/setup.md`. This PR's base is `dev`, so once it merges the link lands users on the file that actually documents the plugin install. Will flip back to `blob/main` when `dev` → `main` ships the plugin section.

.mcp.json
gbrlcustodio: **deferred**: agreed. Tracked under #234 — the release-wheel URL migration will swap `@dev` for the released tag in both `.mcp.json` and `plugin.json` in lockstep so plugin version and pulled code stay aligned. No action this PR.

commands/login.md
gbrlcustodio: **won't fix**: with `disable-model-invocation: true` the model can't pass `$ARGUMENTS`; the only producer is the human, who sees the rendered command (per slash-command body) and confirms the Bash permission prompt before it runs. `argument-hint` already advertises just `--no-browser` and `--callback-timeout <seconds>`. The threat model (a shell wrap rewriting characters the human literally typed and confirmed) doesn't seem worth a workaround that would also block legitimate future flags. Happy to revisit if the slash-command surface area grows.

adriannoes:
LGTM

Danielmoraisg (approved):

PR #236, opened 2026-05-25
feat(auth): default PIPEFY_AUTH_URL to Pipefy production IdP

### LINKED ISSUE #233: cli: default PIPEFY_AUTH_URL to Pipefy prod IdP
## Problem

`pipefy auth login` exits 2 when `PIPEFY_AUTH_URL` is not set, with the error:

```
PIPEFY_AUTH_URL is required for `pipefy auth login` (the OIDC issuer URL for Pipefy authentication, e.g. https://signin.pipefy.com/realms/pipefy). See docs/cli/auth.md.
```

(`packages/cli/src/pipefy_cli/commands/auth.py:146-153`.)

Pipefy's production IdP URL — `https://signin.pipefy.com/realms/pipefy` — is public and the only value 99% of users would set. Surfacing it as a required env var defeats the one-command UX that the `/pipefy:login` slash command (PR #232) and the curl installer (#231) are trying to deliver.

Surfaced during the workstream-C smoke test: with `pipefy-cli` + `pipefy-auth` + `pipefy-sdk` correctly installed from `dev`, the slash command body still bails on a missing env var — a config requirement plugin users cannot reasonably be expected to set ahead of running `/pipefy:login`.

## Proposal

Default `PIPEFY_AUTH_URL` to `https://signin.pipefy.com/realms/pipefy` inside `pipefy-cli` so the env var becomes an override, not a precondition.

Likely landing spot: `packages/cli/src/pipefy_cli/auth.py` (the `BearerToken` / `AuthContext` wiring), or wherever `OidcClient` is constructed for the CLI. `pipefy-auth` already exposes a `DEFAULT_AUTH_CLIENT_ID = "pipefy-cli"` constant (`packages/auth/src/pipefy_auth/identity.py`); pairing it with a `DEFAULT_AUTH_URL` keeps the defaults colocated.

### Acceptance

- [ ] `pipefy auth login` with no `PIPEFY_AUTH_URL` set drives OAuth against `https://signin.pipefy.com/realms/pipefy`.
- [ ] `PIPEFY_AUTH_URL=https://my-custom-issuer.example` overrides the default (regression test).
- [ ] Self-hosted / non-prod users keep their override path; default does not break service-account, static-token, or any of the resolver tiers.
- [ ] Documentation (`docs/cli/auth.md`, `docs/setup.md`) updated to state the default and the override knob.

## Why this is separate from #229

`/pipefy:login` (#229) is host-wiring (Claude Code plugin manifest + slash command body). Baking a default URL touches `pipefy-cli` resolver semantics — host-agnostic CLI behavior. Keeping the two PRs separate matches the host-agnostic boundary established in #229 and avoids mixing scopes.

Refs #229, #231.

gbrlcustodio:
## Summary

- Adds `DEFAULT_AUTH_URL = \"https://signin.pipefy.com/realms/pipefy\"` to `pipefy_auth.identity` (sibling to the existing `DEFAULT_AUTH_CLIENT_ID`).
- `AuthSettings.auth_url` now defaults to that constant — `pipefy auth login` no longer exits 2 with `PIPEFY_AUTH_URL is required` when the env var is unset.
- Override path preserved: `PIPEFY_AUTH_URL=<custom>` still wins over the default; `PIPEFY_AUTH_URL=""` explicitly disables the stored-session tier (regression-tested).
- `docs/cli/auth.md` updated to document the new default + override shape.
- Existing `test_missing_auth_url_exits_2` tests in `packages/cli/tests/{test_auth_login,test_auth_logout}.py` renamed to `test_empty_auth_url_exits_2` and re-pointed at the explicit-empty case.

Surfaced during the workstream-C smoke test (PR #232): plugin / one-shot users following the documented `/pipefy:login` flow hit the missing-env exit because the canonical Pipefy production OIDC issuer URL was not baked in as a default. The URL is public, the only realm 99% of users would set, so requiring it as an env var introduced friction every install path had to pay through.

Closes #233.

## Test plan

- [x] `uv run pytest packages/auth/tests --no-cov` — 50/50 pass.
- [x] `uv run pytest packages/cli/tests -m \"not integration\" --no-cov` — 260/260 pass.
- [x] `uv run pytest packages/mcp/tests --no-cov` — 1203/1203 pass + 14 skipped (no regressions in MCP container's auth resolution).
- [x] `uv run ruff check .` + `uv run ruff format --check .` — clean.
- [x] `AuthSettings()` with no `PIPEFY_AUTH_URL` env returns `auth_url == DEFAULT_AUTH_URL` and a non-None `to_oidc_client()`.
- [x] `PIPEFY_AUTH_URL=https://other.example/realms/x` overrides the default.
- [x] `PIPEFY_AUTH_URL=""` opts out → `to_oidc_client()` returns None and `pipefy auth login` / `pipefy auth logout` exit 2 with the existing "PIPEFY_AUTH_URL is required" hint.

### DIFF
--- a/docs/cli/auth.md
+++ b/docs/cli/auth.md
@@ -91,7 +91,7 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID` | Tier 3 | Service-account client id. |
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET` | Tier 3 | Service-account client secret. |
 | `PIPEFY_INTERNAL_API_URL` | Tier 3 | Internal GraphQL endpoint for AI automations / some relation flows. Required for those tools only. |
-| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. Required for `pipefy auth login`. |
+| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. **Defaults to `https://signin.pipefy.com/realms/pipefy` (Pipefy production IdP)**; set to an empty string (`PIPEFY_AUTH_URL=`) to disable the stored-session tier, or override to point at a non-prod IdP. |
 | `PIPEFY_AUTH_CLIENT_ID` | Tier 4 | Public client id registered for the CLI. Defaults to `pipefy-cli`. |
 
 `PIPEFY_SERVICE_ACCOUNT_URL` and `PIPEFY_AUTH_URL` are **not** interchangeable: the first is a token URL for client-credentials, the second is an OIDC issuer URL for the user-login flow.
--- a/docs/cli/auth.md
+++ b/docs/cli/auth.md
@@ -91,7 +91,7 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID` | Tier 3 | Service-account client id. |
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET` | Tier 3 | Service-account client secret. |
 | `PIPEFY_INTERNAL_API_URL` | Tier 3 | Internal GraphQL endpoint for AI automations / some relation flows. Required for those tools only. |
-| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. **Defaults to `https://signin.pipefy.com/realms/pipefy` (Pipefy production IdP)**; set to an empty string (`PIPEFY_AUTH_URL=`) to disable the stored-session tier, or override to point at a non-prod IdP. |
+| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. **`pipefy auth login` and `pipefy auth logout` fall back to `https://signin.pipefy.com/realms/pipefy` (Pipefy production IdP) when this variable is unset**; set it explicitly to point at a non-prod IdP or to opt into stored-session tier resolution in `pipefy auth status` / the MCP server / the SDK (which leave the tier inactive when the variable is unset). |
 | `PIPEFY_AUTH_CLIENT_ID` | Tier 4 | Public client id registered for the CLI. Defaults to `pipefy-cli`. |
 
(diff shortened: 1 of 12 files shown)


docs/cli/auth.md
```diff
@@ -91,7 +91,7 @@ PIPEFY_TOKEN="$MY_BEARER" uv run pipefy pipe list
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID` | Tier 3 | Service-account client id. |
 | `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET` | Tier 3 | Service-account client secret. |
 | `PIPEFY_INTERNAL_API_URL` | Tier 3 | Internal GraphQL endpoint for AI automations / some relation flows. Required for those tools only. |
-| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. Required for `pipefy auth login`. |
+| `PIPEFY_AUTH_URL` | Tier 4 | **OIDC issuer URL** for interactive login. The CLI appends `/.well-known/openid-configuration` to discover the authorization and token endpoints. **Defaults to `https://signin.pipefy.com/realms/pipefy` (Pipefy production IdP)**; set to an empty string (`PIPEFY_AUTH_URL=`) to disable the stored-session tier, or override to point at a non-prod IdP. |
```
adriannoes, line 94: **Logout exit codes (~line 174):** That table still says exit 2 when `PIPEFY_AUTH_URL` is *unset*, but with this default an omitted var resolves to the production issuer — unset is no longer the same as "cannot run logout/login." I'd align row 174 with this env table: exit 2 when tier 4 is explicitly disabled (`PIPEFY_AUTH_URL=` / blank), normal logout when the var is simply omitted.

adriannoes, line 94: **Troubleshooting (~lines 209–211):** The "PIPEFY_AUTH_URL is required" section still reads like forgetting to export the issuer, but after this PR that message is mainly empty-string opt-out or a bad override — not "skipped prod export." Worth retitling/rewording to match the default + opt-out story here on line 94.

adriannoes (approved):
## Summary

Solid, focused change: baking in the production OIDC issuer removes real install friction while keeping override and empty-string opt-out behavior correct (`_blank_to_none` + new unit tests). I ran the auth and auth login/logout CLI suites locally on the PR head — all green.

## Also noted (inline on env table)

Two doc sections below this hunk still describe pre-default behavior (logout exit table ~L174, troubleshooting ~L209). Worth a quick follow-up edit in the same file; does not block merge.

## Verification

```bash
uv run pytest packages/auth/tests -m "not integration" -q
uv run pytest packages/cli/tests/test_auth_login.py packages/cli/tests/test_auth_logout.py -m "not integration" -q
```

docs/cli/auth.md
gbrlcustodio: **resolved** (e680436): exit table now reads `PIPEFY_AUTH_URL explicitly disabled (`PIPEFY_AUTH_URL=`)` with a note that unset falls back to the prod default.

gbrlcustodio: **resolved** (e680436): troubleshooting entry rewritten to attribute the message to the explicit-empty opt-out and to point at unset as the fastest restore path.

adriannoes:
LGTM

PR #240, opened 2026-05-26
feat: reduce Pipefy URL configuration surface area to PIPEFY_BASE_URL + PIPEFY_AUTH_URL

### LINKED ISSUE #238: Reduce Pipefy URL configuration surface area
## Problem

Pipefy MCP / CLI / SDK consumers need to configure five URLs to use a non-prod environment:

- `PIPEFY_GRAPHQL_URL` — public GraphQL endpoint.
- `PIPEFY_INTERNAL_API_URL` — internal GraphQL for AI Automation flows.
- `PIPEFY_INTERFACES_GRAPHQL_URL` — Interfaces schema endpoint (portals, pages, elements).
- `PIPEFY_SERVICE_ACCOUNT_URL` — OAuth 2.0 token endpoint for the client-credentials grant.
- `PIPEFY_AUTH_URL` — OIDC issuer URL for the stored-session tier.

For prod, only `PIPEFY_GRAPHQL_URL` was historically required; PR #236 defaulted `PIPEFY_AUTH_URL`. Operators still pay the friction of setting redundant URLs whenever they target a non-prod environment, even though all five URLs come from the same host family for any given environment.

## Goals

- Reduce the operator-facing URL configuration to the minimum number of env vars needed to disambiguate "which Pipefy environment".
- Keep an escape hatch for exotic deployments (regional CNAMEs, proxies, gateways, local-dev mocks).
- No internal-environment names baked into source code or `__all__` exports (public repo).
- No package release required when an internal environment is added or removed.
- Reject malformed values at settings construction; no overload of empty-string semantics (kill-switch tracked separately in #237).
- Preserve env-as-ephemeral-override convention (env > config file > built-in default).

## Out of scope

- Disabling the stored-session tier (tracked in #237).
- Validation / non-prod environment URL changes themselves.
- A public list of internal environment names anywhere in the package.

## Related

- PR #236 — `PIPEFY_AUTH_URL` default precedent (closes #233).
- PR #239 — closed unmerged in favor of this issue.
- Issue #237 — kill-switch / disable-tier mechanism.

gbrlcustodio:
Closes #238.

## Summary

Collapses the five per-URL env vars (`PIPEFY_GRAPHQL_URL`, `PIPEFY_INTERNAL_API_URL`, `PIPEFY_INTERFACES_GRAPHQL_URL`, `PIPEFY_SERVICE_ACCOUNT_URL`, `PIPEFY_OAUTH_URL`) into two:

- `PIPEFY_BASE_URL` (default `https://app.pipefy.com`) — API host root; drives the four API endpoints via pydantic `@computed_field` properties.
- `PIPEFY_AUTH_URL` (default `https://signin.pipefy.com/realms/pipefy`) — full OIDC issuer URL.

Hard break — dropped env vars carry no aliases. Pre-1.0 beta surface; consumers update their config to the two new vars.

## Design

- `PipefySettings.base_url` and `AuthSettings.base_url` each read `PIPEFY_BASE_URL` independently, no cross-package dependency at runtime.
- `@computed_field` properties (`graphql_url`, `internal_api_url`, `interfaces_graphql_url`, `service_account_url`) derive from `base_url`; round-trip through `model_dump(exclude_unset=True)` for `_revalidate`.
- `AuthSettings` self-validates: inline `model_validator(mode="after")` runs SSRF + scheme checks against `base_url` and `auth_url`, so direct construction outside the CLI/MCP composition is safe.
- `AuthSettings.to_oidc_client()` now returns a non-Optional `OidcClient` (the issuer URL field always carries a value).
- `BasePipefyClient` gains an `url_override` param so callers that need a non-default endpoint (e.g. `InternalApiClient`) pass it explicitly instead of mutating settings with `model_copy`.
- Legacy `PIPEFY_OAUTH_CLIENT` / `PIPEFY_OAUTH_SECRET` env names still resolve to `service_account_client_id` / `_secret` via `AliasChoices` plus a one-shot stderr deprecation warning. `PIPEFY_OAUTH_URL` is gone — no alias, no warning.

## Security

`AliasChoices` lists only fully-prefixed env-var names. Unprefixed entries would let pydantic-settings (which is case-insensitive on env loads) pick up bare-name env vars like `BASE_URL`, `STATIC_TOKEN`, or `OAUTH_CLIENT` — an auth-redirect / credential-leak primitive on any host whose environment happens to carry those common names. Coverage in `packages/auth/tests/test_settings_back_compat.py::test_bare_name_env_vars_do_not_leak_into_auth_settings`. Kwarg construction by field name still works via `populate_by_name=True`.

## CLI flag rename

`--graphql-url` → `--base-url` on all CLI subcommands. The `--allow-insecure-urls` flag is implemented as a temporary env-var override held for the whole `resolve_cli_settings` flow (including `_revalidate` calls) so an `PIPEFY_ALLOW_INSECURE_URLS=false` env doesn't bleed back in over the flag patch.

## Test plan
- [x] `uv run pytest packages/cli/tests packages/auth/tests packages/sdk/tests -m "not integration"` — 1115 passed.
- [x] `uv run pytest packages/mcp/tests -m "not integration"` — 1227 passed.
- [x] `uv run ruff check .` — clean.
- [ ] Smoke: `PIPEFY_BASE_URL=https://staging.example.com pipefy --version` (env recognized, no crash).
- [ ] Smoke: `pipefy auth login` against prod (default `PIPEFY_AUTH_URL`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,22 +12,37 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - **CLI**: `pipefy auth status [--json|-j]` — reports auth source, identity, session expiry, and exit codes.
 - **CLI**: `pipefy auth logout` — revokes the refresh token at the IdP and clears the stored session.
-- **Auth**: `AuthSettings.auth_url` now defaults to `https://signin.pipefy.com/realms/pipefy` (Pipefy production IdP) when `PIPEFY_AUTH_URL` is unset — removes the previous `PIPEFY_AUTH_URL is required` exit-2 friction on `pipefy auth login` / `pipefy auth logout` and wires the MCP stored-session tier automatically after `pipefy auth login`. Override by setting `PIPEFY_AUTH_URL=<url>` to a non-prod IdP; explicit-empty `PIPEFY_AUTH_URL=` disables the stored-session tier on hosts that need it. User-level `~/.config/pipefy/config.toml` `auth_url` keys still override the default via the existing TOML fallback. Closes #233.
+- **Auth**: `AuthSettings.auth_url` now defaults to the Pipefy production IdP when `PIPEFY_AUTH_URL` is unset — removes the previous `PIPEFY_AUTH_URL is required` exit-2 friction on `pipefy auth login` / `pipefy auth logout` and wires the MCP stored-session tier automatically after `pipefy auth login`. Override by setting `PIPEFY_AUTH_URL=<url>` to a non-prod IdP. Closes #233.
+- **SDK / Auth**: introduced `PIPEFY_BASE_URL` (default `https://app.pipefy.com`) that drives the four API endpoints (`graphql_url`, `internal_api_url`, `interfaces_graphql_url`, `service_account_url`) as pydantic `@computed_field` properties. Operators on non-prod environments set `PIPEFY_BASE_URL=<host>` once and all four endpoints follow. The OIDC issuer (`PIPEFY_AUTH_URL`, default `https://signin.pipefy.com/realms/pipefy`) remains a separate full-URL field because non-prod realm names don't follow a derivable convention. Closes #238.
 - **MCP**: stored-session tier wired into `ServicesContainer`; setting `PIPEFY_AUTH_URL` after `pipefy auth login` now lets the MCP server reuse the keychain-backed session, with the refresh pre-warmed at startup so a stale or revoked session surfaces before the first tool call.
 
 ### Changed
 
-- **SDK / MCP / CLI**: renamed the three service-account credential env vars for clarity and to remove the one-letter footgun against `PIPEFY_AUTH_URL` (interactive user-login issuer): `PIPEFY_OAUTH_URL` → `PIPEFY_SERVICE_ACCOUNT_URL`, `PIPEFY_OAUTH_CLIENT` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID`, `PIPEFY_OAUTH_SECRET` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET`. Closes #127.
+- **SDK / MCP / CLI**: renamed the two service-account credential env vars for clarity and to remove the one-letter footgun against `PIPEFY_AUTH_URL` (interactive user-login issuer): `PIPEFY_OAUTH_CLIENT` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID`, `PIPEFY_OAUTH_SECRET` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET`. `PIPEFY_OAUTH_URL` is dropped without a renamed counterpart — the OAuth token endpoint now derives from `PIPEFY_BASE_URL` (see the **Removed** section). Closes #127.
+- **Auth / SDK**: every `PIPEFY_*` env var is validated against a semantically meaningful pattern at settings construction. URL env vars (`PIPEFY_BASE_URL`, `PIPEFY_AUTH_URL`) require `https?://` plus non-whitespace; credential fields (`PIPEFY_TOKEN`, `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID`, `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET`, `PIPEFY_AUTH_CLIENT_ID`) reject leading / trailing whitespace; `PIPEFY_ORG_ID` must be an ASCII numeric string. There is no longer an empty-string opt-out for any tier — unset the variable to fall back to the default (kill-switch tracked separately in #237).
+- **SDK / MCP / CLI**: renamed `--graphql-url` CLI flag to `--base-url` to match the new env-var shape.
 
 ### Deprecated
 
-- **SDK**: legacy `PIPEFY_OAUTH_*` env vars and `oauth_*` keys in `~/.config/pipefy/config.toml` still resolve to the new `service_account_*` fields via an alias shim, with a one-shot stderr deprecation warning per legacy key. The aliases will be removed in a later `0.2.0-beta.x` release (carrying an explicit breaking-change callout). See [`docs/MIGRATION.md`](docs/MIGRATION.md#service-account-env-var-rename).
+- **SDK**: legacy `PIPEFY_OAUTH_*` env vars still resolve to the new `service_account_*` fields via an alias shim, with a one-shot stderr deprecation warning per legacy key. The aliases will be removed in a later `0.2.0-beta.x` release (carrying an explicit breaking-change callout). See [`docs/MIGRATION.md`](docs/MIGRATION.md#service-account-env-var-rename).
 
 ### Fixed
 
 - **CLI**: concurrent `pipefy` invocations near token expiry no longer surface `invalid_grant` errors; the refresh-token grant is now serialized across processes via a filesystem lock at `~/.config/pipefy/refresh.lock` (`%APPDATA%/pipefy/refresh.lock` on Windows). Closes #133.
 
 ### Removed
+
+- **HARD BREAK** — per-URL env vars dropped in favor of `PIPEFY_BASE_URL` + `PIPEFY_AUTH_URL`. The following env vars are no longer recognized; `PipefySettings` / `AuthSettings` are configured with `extra="ignore"`, so settings construction silently drops them with no exception or warning — stale `.env` keys look configured but the prod defaults still apply. Same wording as `docs/MIGRATION.md`. Audit `.env`, MCP client JSON, and CI secrets before upgrading:
+  - `PIPEFY_GRAPHQL_URL` — set `PIPEFY_BASE_URL` to your API host (graphql_url derives as `<base>/graphql`).
+  - `PIPEFY_INTERNAL_API_URL` — derives as `<base>/internal_api`.
+  - `PIPEFY_INTERFACES_GRAPHQL_URL` — derives as `<base>/graphql/interfaces`.
+  - `PIPEFY_SERVICE_ACCOUNT_URL` — derives as `<base>/oauth/token`.
+  - `PIPEFY_OAUTH_URL` (legacy alias for `PIPEFY_SERVICE_ACCOUNT_URL`) — no replacement; same derivation path.
+
+  Migration: replace the five per-URL env vars with a single `PIPEFY_BASE_URL` (default `https://app.pipefy.com`). If you need a non-prod OIDC issuer, also set `PIPEFY_AUTH_URL` to the full issuer URL.
+- **HARD BREAK** — user TOML config file `~/.config/pipefy/config.toml` is no longer read by either the CLI or the MCP server. Operators who relied on it must move credentials and `base_url` into shell environment variables, a `.env` file at the working directory, or their MCP client's `env` block. Two config surfaces (env + `.env`) instead of three; persistent global config is shell-rc territory.
+- `PIPEFY_TENANT` / `PIPEFY_AUTH_REALM` env vars (never shipped in a release; existed only on intermediate commits of this branch).
+- **Auth**: `pipefy_auth.DEFAULT_AUTH_URL` constant (and its re-export from the package root). Callers should consult the resolved `AuthSettings.auth_url` instead.
 
 ## [0.2.0-beta.1] - 2026-05-18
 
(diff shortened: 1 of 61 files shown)


packages/cli/src/pipefy_cli/config.py
```diff
@@ -78,7 +76,7 @@ def _revalidate(model: _ModelT, patch: dict[str, Any]) -> _ModelT:
     """Merge ``patch`` and re-validate (``model_copy`` would skip URL validators)."""
     if not patch:
         return model
-    merged = {**model.model_dump(), **patch}
+    merged = {**model.model_dump(exclude_unset=True), **patch}
     return type(model).model_validate(merged)
```
adriannoes: When `PIPEFY_BASE_URL` is already exported, `--base-url` updates `pipefy.base_url` but not `auth.base_url` — GraphQL follows the flag while `service_account_url` stays on the env host.

`_revalidate` calls `AuthSettings.model_validate(merged)`. Because `base_url` is aliased to `PIPEFY_BASE_URL`, the env source wins over the value in `merged`.

Hold `PIPEFY_BASE_URL` in `os.environ` for the whole `resolve_cli_settings` flow when the flag is set (same approach as `PIPEFY_ALLOW_INSECURE_URLS` today). Extend `test_base_url_flag_overrides_env` to assert `auth.base_url` and `auth.service_account_url`, not only `.pipefy`.

packages/cli/src/pipefy_cli/config.py
```diff
@@ -160,72 +159,58 @@ def apply_toml_fallback(
...
 
     Raises:
         ValueError: When validation fails (e.g. SSRF guard); message is user-facing.
     """
+    # Env override held through the whole flow (not just ``CliSettings()``).
+    # ``_revalidate`` invokes ``BaseSettings.model_validate`` on ``AuthSettings``,
+    # which re-reads env vars — a ``PIPEFY_ALLOW_INSECURE_URLS=false`` env would
+    # otherwise override the flag patch.
     prev_allow = os.environ.get(_ALLOW_INSECURE_ENV_KEY)
     if allow_insecure_urls_flag is True:
```
adriannoes: The env workaround only sets `PIPEFY_ALLOW_INSECURE_URLS` when the flag is `True`. If a caller passes `allow_insecure_urls_flag=False` while `PIPEFY_ALLOW_INSECURE_URLS=true` is exported, `PipefySettings` gets `False` but `AuthSettings` keeps `True`.

The CLI never passes `False` today, but `resolve_cli_settings` is public API. Holding both `PIPEFY_BASE_URL` and `PIPEFY_ALLOW_INSECURE_URLS` in the environment for the full resolve flow (for every flag value) fixes this together with the `--base-url` case above.

CHANGELOG.md
```diff
@@ -12,12 +12,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 - **MCP**: stored-session tier wired into `ServicesContainer`; setting `PIPEFY_AUTH_URL` after `pipefy auth login` now lets the MCP server reuse the keychain-backed session, with the refresh pre-warmed at startup so a stale or revoked session surfaces before the first tool call.
 
 ### Changed
 
 - **SDK / MCP / CLI**: renamed the three service-account credential env vars for clarity and to remove the one-letter footgun against `PIPEFY_AUTH_URL` (interactive user-login issuer): `PIPEFY_OAUTH_URL` → `PIPEFY_SERVICE_ACCOUNT_URL`, `PIPEFY_OAUTH_CLIENT` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID`, `PIPEFY_OAUTH_SECRET` → `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET`. Closes #127.
```
adriannoes: Unreleased still describes `PIPEFY_OAUTH_URL` → `PIPEFY_SERVICE_ACCOUNT_URL` and `PIPEFY_OAUTH_*` resolving. This PR drops URL env vars; only `PIPEFY_OAUTH_CLIENT` / `PIPEFY_OAUTH_SECRET` remain as legacy aliases.

Please split client/secret deprecation from URL removal (`PIPEFY_BASE_URL`).

CHANGELOG.md
```diff
@@ -29,6 +32,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Removed
 
+- **HARD BREAK** — per-URL env vars dropped in favor of `PIPEFY_BASE_URL` + `PIPEFY_AUTH_URL`. The following env vars are no longer recognized; settings construction does not silently ignore them but they have no effect:
```
adriannoes: Dropped per-URL env vars are `extra="ignore"` no-ops with no warning — stale `.env` keys look configured but prod defaults still apply.

Either match MIGRATION wording (“ignored, no effect”) or emit a one-shot stderr when legacy URL env keys are still set.

adriannoes (changes requested):
**Changes requested.** Feedback is in the inline comments on the Files changed tab.

docs/cli/auth.md
```diff
@@ -171,18 +170,13 @@ When no session is stored, `pipefy auth logout` prints `Not signed in. Nothing t
...
 
 ### Pipefy issuer URLs
 
-| Environment | `PIPEFY_AUTH_URL` |
-|-------------|--------------------|
-| Production | `https://signin.pipefy.com/realms/pipefy` |
-| Validation (piporacle) | `https://signin-piporacle.pipefy.com/realms/st-piporacle-pud1m` |
-
-Pair each issuer with the matching `PIPEFY_GRAPHQL_URL` (`https://app.pipefy.com/graphql` for prod, `https://piporacle.pipefy.com/graphql` for piporacle).
+Production: `https://signin.pipefy.com/realms/pipefy`. Set `PIPEFY_TENANT=<name>` (and `PIPEFY_AUTH_REALM=<realm>` if the realm name doesn't match the default `pipefy`) to point at a non-prod environment; non-prod hosts follow the convention `signin-<tenant>.pipefy.com` and the per-environment realm name comes from your environment's documentation. Operators on environments that don't follow the convention can still set `PIPEFY_AUTH_URL` to the full issuer URL directly.
```
adriannoes: This file still documents env vars this PR removes (all are `extra="ignore"` no-ops):

- **L29:** `PIPEFY_INTERNAL_API_URL` — derives from `PIPEFY_BASE_URL`; client is always wired
- **L65:** `PIPEFY_OAUTH_URL` is not honored — only `PIPEFY_OAUTH_CLIENT` / `PIPEFY_OAUTH_SECRET`
- **L179, L205:** `PIPEFY_TENANT` / `PIPEFY_AUTH_REALM` — use `PIPEFY_BASE_URL` and full `PIPEFY_AUTH_URL`

packages/sdk/tests/test_settings.py
```diff
@@ -5,34 +5,125 @@
...
+    monkeypatch.setenv("PIPEFY_BASE_URL", "https://staging.example.com")
     settings = live_pipefy_settings()
-    assert settings.interfaces_graphql_url == custom_url
+    assert settings.base_url == "https://staging.example.com"
+    assert settings.graphql_url == "https://staging.example.com/graphql"
+    assert settings.internal_api_url == "https://staging.example.com/internal_api"
+    assert (
+        settings.interfaces_graphql_url
+        == "https://staging.example.com/graphql/interfaces"
+    )
```
adriannoes, line 44: Worth a test that sets `PIPEFY_GRAPHQL_URL`, `PIPEFY_INTERNAL_API_URL`, `PIPEFY_INTERFACES_GRAPHQL_URL`, `PIPEFY_SERVICE_ACCOUNT_URL`, and `PIPEFY_OAUTH_URL` to non-prod hosts (without `PIPEFY_BASE_URL`) and asserts computed URLs still use prod defaults — locks the hard break in CI.

packages/mcp/src/pipefy_mcp/core/container.py
```diff
@@ -55,7 +55,9 @@ async def initialize_services(self, settings: Settings) -> None:
                 f"{missing_auth_message()} "
                 f"See {DOCS_SETUP_REF} for host-specific install steps."
             )
-        if oidc_client is not None and tier_for(resolved) == STORED_SESSION_TIER:
+        # ``oidc_client`` is always non-None (``auth_url`` defaults to prod IdP);
+        # tier check gates warmup — only stored-session needs it.
+        if tier_for(resolved) == STORED_SESSION_TIER:
```
adriannoes, line 60: Nearby (~L78): `if pipefy.internal_api_url:` is always true now that `internal_api_url` is a `@computed_field`. Drop the guard and construct `InternalApiClient` unconditionally.

packages/cli/src/pipefy_cli/auth.py
```diff
@@ -60,11 +59,15 @@ class AuthContext:
     stored-session). Built once at startup from the loaded
     :class:`pipefy_auth.AuthSettings` plus the per-invocation ``--token`` /
     ``PIPEFY_TOKEN`` resolution.
+
+    ``oidc_client`` is non-Optional: ``AuthSettings.to_oidc_client()`` always
```
adriannoes, line 63: Nearby (~L181): same dead `if pipefy_settings.internal_api_url:` guard — `internal_api_url` is always set. Wire the internal client unconditionally.

packages/cli/src/pipefy_cli/config.py
gbrlcustodio: **resolved**: settings init rewritten — no more `_revalidate` / env-mutation dance. Flag value now lands in pydantic-settings init source as an alias-keyed kwarg (`{"auth": {"PIPEFY_BASE_URL": ...}}`), which beats nested env. `test_base_url_flag_overrides_env` extended to assert both `auth.base_url` and `auth.service_account_url`. See 1656809.

packages/cli/src/pipefy_cli/config.py
gbrlcustodio: **resolved**: flag value (any branch — `True` / `False`) now lands in init source as alias-keyed kwarg `{"auth": {"PIPEFY_ALLOW_INSECURE_URLS": ...}}`, beating nested env in both directions. Added regression test `test_allow_insecure_urls_flag_false_overrides_env_true`. See 1656809.

CHANGELOG.md
gbrlcustodio: **resolved**: rename bullet now lists only `PIPEFY_OAUTH_CLIENT` / `PIPEFY_OAUTH_SECRET` (deprecation, alias shim). `PIPEFY_OAUTH_URL` moved to a dedicated **Removed / HARD BREAK** bullet alongside the four per-URL env vars, with explicit derivation notes. See 1656809.

CHANGELOG.md
gbrlcustodio: **resolved**: HARD BREAK bullet now spells out the silent-drop behaviour verbatim — _"`PipefySettings` / `AuthSettings` are configured with `extra=\"ignore\"`, so settings construction silently drops them with no exception or warning — stale `.env` keys look configured but the prod defaults still apply. Same wording as `docs/MIGRATION.md`."_ Going with the docs-match option rather than a stderr warning to keep the removed-vars list a clean no-op surface. See 1656809.

docs/cli/auth.md
gbrlcustodio: **resolved**: dropped `PIPEFY_INTERNAL_API_URL` (now derived), `PIPEFY_OAUTH_URL` (now legacy alias note only), and `PIPEFY_TENANT` / `PIPEFY_AUTH_REALM` (never shipped). Fixed a stray "tenant default" → "prod default" while there. See 1656809.

packages/sdk/tests/test_settings.py
gbrlcustodio: **resolved**: added parametrized `test_pipefy_settings_ignores_removed_per_url_env_vars` in `packages/sdk/tests/test_settings.py` — sets each removed var to a non-prod host with `PIPEFY_BASE_URL` unset, asserts the computed URL stays on prod default. Locks the hard break. See 1656809.

packages/mcp/src/pipefy_mcp/core/container.py
gbrlcustodio: **resolved**: dropped the `if pipefy.internal_api_url:` guard; `InternalApiClient` is now wired unconditionally. See 1656809.

packages/cli/src/pipefy_cli/auth.py
gbrlcustodio: **resolved**: dropped the `if pipefy_settings.internal_api_url:` guard; internal client wired unconditionally. See 1656809.

Issue #237, opened 2026-05-25
auth: kill-switch + disk-fallback for the stored-session keychain probe


gbrlcustodio:
## Context

#233 (PR #236) made `AuthSettings.auth_url` default to the Pipefy production IdP so that `pipefy auth login` and the MCP stored-session tier work without explicit env config. Side effect: every default-constructed `AuthSettings` now carries a non-`None` `OidcClient`, which means `resolve_pipefy_auth` / `detect_pipefy_tiers` reach the stored-session detection branch on hosts that don't actually want it (service-account-only deployments, headless Linux without Secret Service, CI runners without a working keyring backend).

The probe is bounded (`_has_stored_session` lazy-imports `keyring`, returns `None` on `KeyringError`), but it does carry a measurable cost:

- ~30-80 ms of `keyring` backend discovery on the first call per process (see `packages/auth/src/pipefy_auth/storage.py:8-11`).
- Log noise on headless Linux without a Secret Service daemon (the silenced `KeyringError`).
- On macOS in non-Aqua subprocess contexts, the documented-but-not-yet-reproducible `errSecParam (-25244)` failure mode (#235) becomes more reachable because more code paths trip the keychain probe.

#233's PR review surfaced this as a real concern but accepted it as a fair trade for the UX win, with the explicit follow-up that an opt-out / kill-switch should land here.

## Proposal

Two complementary mechanisms — either is useful on its own; together they cover all the leakage cases:

### 1. Kill-switch env var

Honor a `PIPEFY_DISABLE_STORED_SESSION=1` env (or equivalent boolean parsed via `pydantic-settings`) on `AuthSettings`. When set, `to_oidc_client()` returns `None` even if `auth_url` is populated, so `resolve_pipefy_auth` / `detect_pipefy_tiers` never reach the keychain probe. Equivalent to `PIPEFY_AUTH_URL=` (the current opt-out) but a positive signal that does not require remembering the `_blank_to_none` validator interaction documented at `settings.py:91` (review finding #10).

### 2. Disk-fallback storage backend

When `PIPEFY_KEYCHAIN_BACKEND=file` (or similar), swap the `keyring` backend for `keyrings.alt.file.PlaintextKeyring` (or an encrypted alternative scoped via a separate env var). This unblocks:

- Headless Linux hosts without a Secret Service daemon.
[...]

PR #244, opened 2026-05-27
feat(infra): shared filesystem-backed TOML config foundation

gbrlcustodio:
## Summary

- New leaf-layer workspace package **`pipefy-infra`** owns three primitives shared by `pipefy-sdk` and `pipefy-auth`: `config_dir()` (XDG_CONFIG_HOME / %APPDATA% aware), `config_file_path()` (with a `PIPEFY_CONFIG_FILE=<path>` env override), and `PipefyTomlConfigSource` (a schema-agnostic `PydanticBaseSettingsSource` subclass).
- Both `AuthSettings` and `PipefySettings` wire the new source into their `settings_customise_sources()` hook. Precedence: `init kwargs > env > .env > config.toml > defaults`. Shared TOML keys (`base_url`, `allow_insecure_urls`) populate both models from a single entry.
- `pipefy_auth.locks.refresh_lock_path()` now delegates to `config_dir()` so the refresh lock and `config.toml` share one discovery path.
- The package holds **no field schema** — `pipefy_infra` knows TOML, paths, and the source protocol; it does not know about `base_url`, `auth_url`, or any other field. Adding a new auth/SDK field never touches this package.
- Dep graph stays symmetric:

  ```
  pipefy_infra         (leaf; pydantic + pydantic-settings only)
       ↑
       ├── pipefy_sdk     (no auth dep)
       └── pipefy_auth    (no sdk dep)
                ↑
                ├── pipefy_cli
                └── pipefy_mcp
  ```

## Context

Foundation for **#237** (kill-switch + file-backed keyring). The two new `AuthSettings` fields (`disable_stored_session`, `keychain_backend`) added in that follow-up will read from this loader for free; that PR can stay surgical and focus on the keychain bypass / `keyrings.alt.file.PlaintextKeyring` wiring.

This PR drops the prior `HARD BREAK` CHANGELOG entry that removed `~/.config/pipefy/config.toml` and replaces it with an `Added` entry for the new schema-agnostic loader. The new schema is **not** the old schema (flat field names, schema-agnostic source, `env > file` precedence) — but functionally a `config.toml` is again read.

## Naming

Package name maps to DDD/clean-architecture's "infrastructure layer" — adapters between Pipefy's application code (`pipefy-sdk`, `pipefy-auth`) and external concerns (filesystem, OS, network boundary). The boundary the package defends: stdlib + `pydantic` / `pydantic-settings` only. No httpx, no keyring, no gql, no business logic.

## Notable design decisions

- **Schema-agnostic helper, not a `PipefyConfig` model.** Field schemas stay where they belong (`AuthSettings`, `PipefySettings`). Each consumer wires the source itself (~3-line classmethod). No junk-drawer coupling.
- **Flat TOML schema, not per-package tables.** `AuthSettings` and `PipefySettings` intentionally share `base_url` + `allow_insecure_urls`; an operator pinning `base_url = "..."` once and having both models pick it up is the desired ergonomic.
- **TOML uses bare pydantic field names, not env-style keys.** The env-only `AliasChoices` (e.g. `PIPEFY_TOKEN`) exist to refuse unprefixed env leakage; the source filters TOML lookups through model field names so the alias-as-TOML-key shadow path is closed.
- **XDG_CONFIG_HOME is honoured** (the previous `locks.py:41` implementation ignored it — small idiomatic fix bundled into the lift).
- **`platformdirs` deliberately not adopted.** Would land config under `~/Library/Application Support/pipefy` on macOS, breaking parity with `gh`, `uv`, `gcloud`, the existing `locks.py` location, and operator muscle memory.
- **Stdlib `tomllib`** (Python 3.11+; repo `requires-python = ">=3.11"`). Malformed TOML re-raises chained as `ValueError(f"invalid TOML in {path}") from exc`.

## Test plan

- [x] `uv run pytest` — 2527 passed, 45 skipped (full workspace).
- [x] `uv run ruff check packages` — clean.
- [x] Manual: with a valid TOML file → fields populated on both `PipefySettings` and `AuthSettings`.
- [x] Manual: env overrides file.
- [x] Manual: malformed TOML → `ValueError` quoting the file path.
- [x] Manual: missing file → defaults preserved.

## Follow-ups (not this PR)

- **#237**: `disable_stored_session: bool` + `keychain_backend: Literal["auto","file"]` on `AuthSettings`. Both will read from this loader automatically.
- **#245** (stacked): consolidate the duplicated URL SSRF helpers into `pipefy_infra`.
- A Typer-level `pipefy --config /path/to/foo.toml` flag (the idiomatic CLI override). `PIPEFY_CONFIG_FILE` env covers the same need for this PR.
- `pipefy config show` / `pipefy config init` CLI commands (separate issue once the file format settles).
- TOML profiles (`[prod]` / `[staging]`) once we have two real users juggling environments.


adriannoes (approved):
Solid foundation, leaf `pipefy-infra`, symmetric sdk/auth wiring, schema-agnostic TOML source, and tests/CI/release all line up. 

I checked out `c1ed243` locally: targeted pytest on the new infra + auth/sdk TOML suites and ruff on the touched packages are green (matches your CI).

**Approved.** This is ready to merge from my side.

## Adoption — two small doc hooks (optional follow-up)

`docs/config.md` is clear and the link from `docs/cli/auth.md` helps CLI users. For broader product adoption, two discoverability gaps are worth a quick pass (same PR or a tiny docs-only follow-up):

1. **`docs/setup.md` — “How configuration is loaded”** still reads as *env overrides `.env`* only. Most first-time installers start here, not `docs/config.md`, so they may never learn that `~/.config/pipefy/config.toml` exists again or how it fits the stack. A short pointer to `docs/config.md` plus the full precedence chain (`init kwargs > env > .env > config.toml > defaults`) would align onboarding with what `AuthSettings` / `PipefySettings` actually do after this change — especially for MCP users who pin `base_url` once for both CLI and server.

2. **`docs/README.md` — Shared docs table** does not list `config.md` yet. Adding a row next to `setup.md` makes the global config guide discoverable from the docs index, the same way operators already find auth and parity docs.

Neither blocks merge; they just reduce “I didn’t know I could use config.toml” friction now that the feature is back.

Danielmoraisg:
The precedence tests cover `init > toml` and `env > toml`, but not `.env > config.toml` — even though both models set `env_file=".env"` and `docs/config.md` promises that tier. The `env > toml` test doesn't cover it: env is position 2 and dotenv is position 3 in the source tuple, so a reorder that slid the TOML source between them (`init, env, toml, dotenv, ...`) would keep the env test green while silently flipping `.env > toml`. Low severity, but it's a live documented tier and one test pins it:

```python
def test_dotenv_wins_over_toml(monkeypatch, tmp_path):
    monkeypatch.chdir(tmp_path)  # env_file=".env" resolves relative to cwd
    (tmp_path / ".env").write_text("PIPEFY_BASE_URL=https://from-dotenv.example\n")
    monkeypatch.setenv("PIPEFY_CONFIG_FILE", str(tmp_path / "config.toml"))
    (tmp_path / "config.toml").write_text('base_url = "https://from-toml.example"\n')
    assert PipefySettings().base_url == "https://from-dotenv.example"
```

Danielmoraisg (approved):

gbrlcustodio:
**chore (non-blocking):** suggestion 2 (`docs/README.md` shared-docs table) landed as d28f143 on this branch. Suggestion 1 (broaden `docs/setup.md` to match its index slot) turned out larger than a precedence-line patch once scope-checked; tracked separately as #246 with a release-gated audit checklist, separate pass.

gbrlcustodio:
Good catch. Pinned dotenv > toml in both models at 49d227c. Sibling `test_env_wins_over_toml` would have stayed green under your reorder hypothetical (`init, env, toml, dotenv, ...`); the new test fails loudly there. Adapted your snippet to lean on the existing `_isolate_env` fixture (already clears `PIPEFY_*` and sets `PIPEFY_CONFIG_FILE` in tmp).

PR #245, opened 2026-05-27
refactor(infra): consolidate URL SSRF gates into pipefy_infra

gbrlcustodio:
## Summary

Eliminates the `pipefy_auth._url_ssrf` ↔ `pipefy_sdk.utils.url_ssrf` duplication. The auth module's own docstring spelled out the migration path: *"can be deduplicated later by extracting a shared utilities package if one is needed for other reasons."* #244 introduced `pipefy-infra` (renamed from `pipefy-config` in the same PR); this one cashes the IOU.

- `packages/sdk/src/pipefy_sdk/utils/url_ssrf.py` → `packages/infra/src/pipefy_infra/url_ssrf.py` (file rename; SDK copy was the superset).
- `packages/auth/src/pipefy_auth/_url_ssrf.py` deleted (subset).
- All call sites (auth `discovery` + `settings`, SDK `settings` + 3 services + test monkeypatch) import from `pipefy_infra` directly.
- `pipefy_sdk.utils.__init__` drops the SSRF re-export — no shim. Verified zero remaining consumers.

## Context

`pipefy_infra` was scoped narrowly in #244 (TOML loader + path discovery). This PR broadens its module docstring + `__all__` to advertise the second concern (URL SSRF gates). Both concerns stay schema-agnostic and depend only on stdlib + `pydantic` / `pydantic-settings`.

Net diff: **−63 LoC** (one whole `_url_ssrf.py` removed; the rest is import line rewrites).

## Test plan

- [x] `uv run pytest` — 2527 passed, 45 skipped (full workspace).
- [x] `uv run ruff check packages` — clean.
- [x] Verified no remaining consumers of `pipefy_sdk.utils.url_ssrf` or `pipefy_auth._url_ssrf` via grep.

## Follow-ups (out of scope here)

- `_URL_SHAPE_PATTERN = r"^(?i:https?)://\\S+$"` regex still duplicated between `packages/auth/src/pipefy_auth/settings.py` and `packages/sdk/src/pipefy_sdk/settings.py`. ~1 LoC each.
- `_strip_str` field_validator likewise duplicated. ~5 LoC each.
- Field definitions (`base_url`, `allow_insecure_urls`) remain duplicated by design — documented "no `env_nested_delimiter` → each model loads independently" stance.

Issue #235, opened 2026-05-25
cli: improve macOS keychain write-failure messaging and first-run UX


gbrlcustodio:
## Problem

`pipefy auth login` from a fresh install on macOS surfaces this when the keychain write fails:

```
Login succeeded but the session could not be stored in your OS keychain (Keyring):
Can't store password on keychain: (-25244, 'Unknown Error').
On headless Linux, ensure a Secret Service daemon (gnome-keyring, kwallet) is running,
or use a static PIPEFY_TOKEN.
```

(`packages/cli/src/pipefy_cli/commands/auth.py:205-213`.)

Two issues:

1. **The hint is platform-wrong.** The user is on macOS; the hint talks about headless Linux Secret Service daemons. Confusing.
2. **The actual cause isn't named.** On macOS, `-25244` (`errSecParam`) from `SecItemAdd` typically fires when `Security.framework` cannot surface the new-item ACL dialog to the user — e.g. when invoked from a non-TTY subprocess of an editor / IDE / agent host that doesn't share an interactive Aqua session. **The failure is independent of how `pipefy` was installed.** It hits both `uvx --from <git+url>` (ephemeral binary) and `uv tool install`-installed persistent binaries when the calling subprocess can't render the keychain prompt. The OAuth handshake itself succeeds; only the keychain persistence step rejects.

The fix users have to discover today: run `pipefy auth login` once from their own Terminal.app session, click **Always Allow** when macOS prompts to grant the Python binary access to the `pipefy` keychain item. From that point on, the keychain ACL remembers and the slash command (and refresh writes) work from subprocesses.

Surfaced during the workstream-C smoke test (#229): `/pipefy:login` via the Claude Code plugin completed OAuth but failed to persist, leaving users with a non-actionable error.

## Proposal

Branch the error message on `platform.system()` in the `KeyringError` handler at `packages/cli/src/pipefy_cli/commands/auth.py:205-213`:

\`\`\`python
except KeyringError as exc:
    hint = _keychain_hint_for_platform()
    typer.echo(
        f\"Login succeeded but the session could not be stored in your OS \"
        f\"keychain ({keychain_backend_name()}): {exc}. {hint}\",
        err=True,
    )
    raise typer.Exit(1) from exc
\`\`\`

`_keychain_hint_for_platform()` returns:

[...]

PR #247, opened 2026-05-27
feat(auth): kill-switch + file-backed keyring for stored sessions

### LINKED ISSUE #237: auth: kill-switch + disk-fallback for the stored-session keychain probe

gbrlcustodio:
Closes #237.

## Summary

Two new `AuthSettings` fields, both env- and TOML-driven, both reading through the `pipefy_infra` loader from #244 for free:

- **`PIPEFY_DISABLE_STORED_SESSION=1`** (or `disable_stored_session = true`) makes `AuthSettings.to_oidc_client()` return `None`. The resolver short-circuits the stored-session tier: no backend discovery, no `load_session` probe. `pipefy auth login` and `pipefy auth logout` refuse with exit 2 and an explanatory stderr. Avoids the keyring backend-discovery cost on cold start (~30-80 ms on macOS, more on headless Linux) and the macOS `errSecParam (-25244)` exposure from #235.

- **`PIPEFY_KEYCHAIN_BACKEND=file`** (or `keychain_backend = "file"`) swaps the active backend to `keyrings.alt.file.PlaintextKeyring` writing under `pipefy_infra.config_dir() / "keyring.cfg"`. Unblocks headless Linux without Secret Service and CI runners. The file is plaintext on disk, not OS-secured; opt-in only.

`#233`'s review accepted defaulting `PIPEFY_AUTH_URL` to the prod IdP with an explicit follow-up to ship this opt-out — every default-constructed `AuthSettings` now reaches `_has_stored_session()` on every cold start, paying the discovery cost. This PR delivers the opt-out.

## Context

PR #244 (shared TOML loader) and #245 (URL SSRF dedupe) built the `pipefy_infra` foundation. The new fields plug into `PipefyTomlConfigSource` and `config_dir()` with no new infra. Branch is cut from `origin/dev` (post-#244, no dependency on #245).

## Design

- **No new abstraction at the resolver.** The existing `resolve_pipefy_auth(oidc_client=None)` path already gated `STORED_SESSION_TIER` on a non-None client; `to_oidc_client()` returning `None` is the natural seam. `AuthContext.oidc_client` widens to `Optional`; the three sites that dereference it (`cli/auth.py`, `commands/auth.py`, `mcp/core/container.py`) narrow with type asserts inside their existing tier checks — the assert is unreachable when the kill-switch is on because the resolver can't return `STORED_SESSION_TIER` without a client.

- **`configure_keychain_backend()` lives on `pipefy_auth.storage`.** Idempotent; no-op for the default `"auto"`. Called once from the CLI typer callback and once from MCP `initialize_services` before any `load_session` / `set_password`. Lazy imports inside the function so `"auto"` doesn't pull `keyrings.alt.file` or `pipefy_infra` into the cold-start path.

- **Login `store_session` widens `except KeyringError` to `except (KeyringError, OSError)`.** `PlaintextKeyring.set_password` calls `os.makedirs` internally; a read-only `config_dir()` surfaces as `PermissionError`, not a `KeyringError`. The user-facing "could not be stored in your keychain" message stays the same.

- **`_strip_str` covers the two new fields.** A stray space in `PIPEFY_KEYCHAIN_BACKEND=" file "` or `PIPEFY_DISABLE_STORED_SESSION=" 1 "` strips before `Literal` / bool parsing instead of raising `ValidationError`. Matches the existing copy-paste-friendly contract on the URL and credential fields.

- **CLI conftest gets the two new env vars.** `clean_pipefy_env` was leaking them between tests when set in the dev shell.

## Files

- `packages/auth/src/pipefy_auth/settings.py` — two fields + `to_oidc_client()` gate + `_strip_str` coverage
- `packages/auth/src/pipefy_auth/storage.py` — `configure_keychain_backend()`
- `packages/auth/pyproject.toml` — `keyrings.alt>=5`
- `packages/cli/src/pipefy_cli/{main,auth}.py`, `packages/cli/src/pipefy_cli/commands/auth.py` — entrypoint wiring, login/logout refuse branches, login `OSError` catch
- `packages/mcp/src/pipefy_mcp/core/container.py` — entrypoint wiring, narrowed `Optional` access
- `docs/cli/auth.md`, `CHANGELOG.md`

## Test plan

- [x] `uv run pytest` — 2559 passed, 45 skipped (full workspace).
- [x] `uv run ruff check packages` — clean.
- [x] New: `test_storage_backend.py` (file backend swap, idempotency, round-trip under `config_dir()`).
- [x] New: kill-switch + backend tests in `test_settings_back_compat.py` (env parses, Literal rejects unknown / empty / whitespace-only values, `_strip_str` covers padded valid values).
- [x] New: TOML coverage in `test_settings_toml_source.py` (both fields load from TOML; env > TOML for the kill-switch).
- [x] New: resolver/detect skip + load_session never called when `oidc_client is None`.
- [x] New: CLI refuse paths for `pipefy auth login` and `pipefy auth logout` (exit 2 + explanatory stderr; `revoke_session` / `run_login` poisoned to prove no work is done).
- [x] New: `pipefy auth status` with `PIPEFY_DISABLE_STORED_SESSION=1` + seeded session + service-account env → stored-session tier omitted from `detected_sources`, service-account wins.
- [ ] Smoke (macOS): `PIPEFY_DISABLE_STORED_SESSION=1 pipefy auth status` → no `stored-session` in detected; no keychain prompt.
- [ ] Smoke (macOS): `PIPEFY_KEYCHAIN_BACKEND=file pipefy auth login` → completes OAuth, writes `~/.config/pipefy/keyring.cfg`, `auth status` shows `Keychain: PlaintextKeyring`. Verify via Keychain Access that no `pipefy` items are created.

## Out of scope

- Encrypted file backend (`keyrings.alt.file.EncryptedKeyring`) — separate follow-up.
- CLI flag (`pipefy --keychain-backend file`) — env/TOML only for now, matches the existing config surface.
- Auto-detect failure mode (switch to `file` automatically on Linux without Secret Service) — explicit opt-in only, by design.
- Migrating existing OS-keychain sessions into the file backend — users re-login.


adriannoes (approved):
Solid, minimal design: the kill-switch reuses `to_oidc_client() → None` and the resolver’s existing `oidc_client` gate, so there’s no new abstraction and cold start skips keychain discovery as intended. `configure_keychain_backend` is wired at the right entrypoints, tests poison `load_session` / OAuth paths when disabled, and the follow-up commit replacing `assert` with `RuntimeError` on tier invariants is the right call.

Review included a local checkout (`fcd2c7b`) — targeted auth/CLI/MCP tests and ruff are clean; CI lint + test are green on the PR head.

**Approved** — merge when you’re ready.

## Also noted (future polish, non-blocking)

- **MCP test env parity:** `packages/cli/tests/conftest.py` now scrubs `PIPEFY_DISABLE_STORED_SESSION` and `PIPEFY_KEYCHAIN_BACKEND`, but `packages/mcp/tests/core/test_container.py`’s `_AUTH_ENV_KEYS` does not yet. A dev shell with the kill-switch set can make MCP container tests flaky locally even though CI stays hermetic. Worth adding both keys to `_AUTH_ENV_KEYS` when convenient.

- **Login store error copy for file backend:** The widened `except (KeyringError, OSError)` path still tells users to fix “OS keychain” / Secret Service. With `PIPEFY_KEYCHAIN_BACKEND=file`, a `PermissionError` on `config_dir()/keyring.cfg` is better served by a hint about config-dir writability than gnome-keyring — small UX tweak for a follow-up.

gbrlcustodio:
chore: addressed both notes in f480ea4.

- `_AUTH_ENV_KEYS` in `packages/mcp/tests/core/test_container.py` now scrubs `PIPEFY_DISABLE_STORED_SESSION` and `PIPEFY_KEYCHAIN_BACKEND` (parity with the CLI conftest).
- Login's `(KeyringError, OSError)` branch now switches on `keychain_backend_name()`: `PlaintextKeyring` → config-dir writability hint; default → existing Secret Service guidance plus `PIPEFY_KEYCHAIN_BACKEND=file` as an explicit escape hatch. The misleading "OS keychain" qualifier is dropped. Two new tests pin both branches; `docs/cli/auth.md` updated to match.

PR #251, opened 2026-05-28
chore(test): migrate AI agent fixtures off production ids (#250)

### LINKED ISSUE #250: chore(test): migrate AI agent and skills fixtures off production pipe/org ids
## Context

Portal, observability, org CLI, and `organization_identifiers` unit tests now use fictional ids from `packages/sdk/tests/_shared/fixture_ids.py` (`EXAMPLE_NUMERIC_ORG_ID`, `EXAMPLE_ORG_UUID`, etc.). Live tests use `.env` only.

## Still using production-looking ids

- `packages/sdk/tests/_shared/ai_agent_test_payloads.py` — `306996636`, reference UUIDs
- `packages/sdk/tests/models/test_ai_agent.py`
- `packages/mcp/tests/tools/test_ai_agent_tools.py`
- `packages/mcp/tests/tools/test_ai_tool_helpers.py`
- `packages/mcp/tests/tools/test_ai_tool_helpers_slug_resolution.py`
- `skills/ai-agents/pipefy-ai-agents/SKILL.md` — example JSON with real pipe/field ids

## Done when

- [ ] Centralize AI agent test pipe/field placeholders in `fixture_ids.py` (or a sibling `ai_agent_fixture_ids.py`) with obviously fictional values
- [ ] Update skills examples to use placeholders documented in the skill (no customer pipe ids)
- [ ] Grep clean for `306996636` / customer-specific field ids in committed test/skill sources
- [ ] Unit tests green: `uv run pytest packages/sdk/tests packages/mcp/tests -m "not integration" -k ai_agent`

## Out of scope

- Changing Pipefy product docs that describe URL shape (org id in path is generic)

adriannoes:
## Summary

Centralizes fictional Pipefy identifiers for AI agent unit tests and skill examples in `packages/sdk/tests/_shared/fixture_ids.py`, replacing production-like IDs (`306996636`, `425829426`, `427911728`, `342688850`, `307077065`) across SDK/MCP tests and skills.

Docstrings and skills now state that `%{internal_id}` **syntax** is required while example digits (e.g. `900000101`) are illustrative — agents must discover real field IDs per pipe.

Closes #250


gbrlcustodio:
## Code review findings

### Validation correctness

**issue (blocking): silent `get_phase_fields` failure produces false-positive "fieldId not found"**

`packages/sdk/src/pipefy_sdk/ai_pipe_validation.py:445`

`collect_field_ids_for_pipe` catches every `get_phase_fields` exception with a debug log only. A transient timeout, rate-limit, or 5xx silently drops that phase's field IDs from `field_ids`; `validate_behaviors_against_pipe` then emits `fieldId X not found in pipe fields` with no warning explaining the partial fetch. Asymmetric with `ai_preflight.py:365`, which already surfaces a `tool_warning` when a target pipe's `get_pipe` fails.

Fix: return `(field_ids, failed_phase_ids)` from the helper, surface a `tool_warning` in `fetch_pipe_validation_context` and the cross-pipe loop when any phase failed.

---

**suggestion: handle `{"pipe": null}` defensively**

`packages/sdk/src/pipefy_sdk/ai_pipe_validation.py:480`, `packages/sdk/src/pipefy_sdk/ai_preflight.py:371`

`pipe_data.get("pipe", {})` returns `None` (not `{}`) when GraphQL responds with `{"pipe": null}` (stale id, no permission). `pipe_info.get(...)` then raises `AttributeError`. Pre-existing pattern, but this PR widens it by passing the value into `collect_field_ids_for_pipe`.

Fix: `pipe_data.get("pipe") or {}` at both sites.

---

**issue: empty `check_fields` set silently passes all validation**

`packages/sdk/src/pipefy_sdk/ai_pipe_validation.py:129`

`if fid and check_fields and fid not in check_fields:` — empty set is falsy in Python, so the membership check short-circuits. If every phase fetch fails and the pipe has no start-form fields, `field_ids` is empty and every behavior is declared valid regardless of whether its fieldIds are real.

Fix: drop `check_fields and`. Pair with the warning surfaced above.

---

### Performance + query shape

**suggestion: extend `GET_PIPE_QUERY` to include phase fields**

`packages/sdk/src/pipefy_sdk/queries/pipe_queries.py:13`

Today the query selects `phases { id name }`. `collect_field_ids_for_pipe` then issues N additional `get_phase_fields` round-trips per pipe. Selecting `phases { id fields { id internal_id } }` collapses `1 + N` to `1` and makes the existing embedded-fields branch in `collect_field_ids_for_pipe` actually live in production.

---

**suggestion: bucket the parallel field fetches**

`packages/sdk/src/pipefy_sdk/ai_preflight.py:363`, `packages/sdk/src/pipefy_sdk/ai_pipe_validation.py:432`

Both the cross-pipe target loop and the per-phase loop inside `collect_field_ids_for_pipe` await sequentially. With 5 cross-pipe targets × 10 phases at ~150ms per call, that's ~7.5s of pure serial wait. Gather both under a shared `asyncio.Semaphore` to cap concurrency below Pipefy's rate limit.

Mostly redundant once the `GET_PIPE_QUERY` change above lands; keep as defense for callers that pass stale-shaped `target_info`.

---

### Tests + fixtures

**issue: `test_target_pipe_fetch_for_cross_pipe_fields` passes by accident**

`packages/mcp/tests/tools/test_ai_agent_tools.py:1729`

The test hand-stuffs `fields` into `phases[].fields` (a code path production `GET_PIPE_QUERY` never returns) AND `mock_pipefy_client.get_phase_fields` is a bare MagicMock whose await raises `TypeError` that the helper's broad `except Exception` swallows.

Fix: reshape the mocks to match production (`phases` with only `{id, name}`) and add an explicit `get_phase_fields = AsyncMock(return_value=...)` setup.

---

**suggestion: add `get_phase_fields` to the shared mock fixture default**

`packages/mcp/tests/tools/test_ai_agent_tools.py:35`

`mock_pipefy_client` doesn't define `client.get_phase_fields = AsyncMock(...)`. Most tests under `TestValidateAiAgentBehaviors` and `TestEnrichWithValidation` pass only because the helper's broad `except` swallows the `TypeError` from awaiting a bare MagicMock. Any future tightening of that `except` breaks ~10 tests at once.

Fix: `client.get_phase_fields = AsyncMock(return_value={"fields": []})` in the fixture default; individual tests already override when they need specific behavior.

---

**suggestion: collapse `fixture_ids.py` to category prefixes**

`packages/sdk/tests/_shared/fixture_ids.py`

Current file mixes role-baked names (`EXAMPLE_DESTINATION_PHASE_ID`) and pairing-specific names (`EXAMPLE_MOCK_FIELD_INTERNAL_ID_COMPANY_NAME`). Side-effects: `EXAMPLE_DESTINATION_PHASE_ID` is used as a field `internal_id` in `test_ai_pipe_validation.py:45`, the `MOCK_` prefix is noise (everything in a fixture file is for mocks), and four constants are unused.

Suggested shape (subsumes the unused-constants cleanup and the namespace confusion):

```python
EXAMPLE_PIPE_ID          = "987654321"
EXAMPLE_PIPE_ID_2        = "900000301"

EXAMPLE_PHASE_ID         = "900000201"

EXAMPLE_FIELD_INTERNAL_ID    = "900000101"
EXAMPLE_FIELD_INTERNAL_ID_2  = "900000102"
EXAMPLE_FIELD_INTERNAL_ID_3  = "900000103"
EXAMPLE_FIELD_INTERNAL_ID_4  = "900000110"
EXAMPLE_FIELD_INTERNAL_ID_5  = "900000111"
EXAMPLE_FIELD_INTERNAL_ID_6  = "900000112"

EXAMPLE_FIELD_SLUG    = "resumo_de_briefing_ia"
EXAMPLE_FIELD_SLUG_2  = "objetivo_da_demanda"
EXAMPLE_FIELD_SLUG_3  = "phase_field_slug"

EXAMPLE_UUID    = "00000000-0000-4000-8000-000000000001"
EXAMPLE_UUID_2  = "00000000-0000-4000-8000-000000000002"

EXAMPLE_AI_ACTION_REFERENCE_UUID    = "00000000-0000-4000-8000-000000000010"
EXAMPLE_AI_ACTION_REFERENCE_UUID_2  = "00000000-0000-4000-8000-000000000011"
```

---

**nitpick: rename `test_validate_behaviors_rejects_numeric_field_id_when_set_has_slug_only`**

`packages/sdk/tests/test_ai_pipe_validation.py:113`

The name implies a slug-vs-numeric discrimination check, but the body asserts only "non-member produces a problem" — would pass for any non-member value. Rename to `test_validate_behaviors_rejects_non_member_field_id` or similar.

adriannoes:
Thanks for the thorough review, @gbrlcustodio : addressed everything in the findings, plus a few follow-ups we spotted while applying the same patterns elsewhere.

### Validation correctness (blocking / issues)

- **`collect_field_ids_for_pipe`** now returns `(field_ids, failed_phase_ids)`. Failures no longer produce silent false positives; we append `tool_warning`s from `fetch_pipe_validation_context` and the cross-pipe loop (same spirit as the existing target-pipe `get_pipe` warning).
- **Empty `check_fields`**: dropped the `check_fields and` guard so an empty set still rejects unknown `fieldId`s (with tests).
- **`{"pipe": null}`**: `pipe_data.get("pipe") or {}` in validation fetch, cross-pipe target load, `build_field_slug_map`, and **`validate_ai_automation_prompt_sdk`** (we hit the same `AttributeError` risk there).

### Performance / query shape (suggestions)

- **`GET_PIPE_QUERY`**: `phases { fields { id internal_id } }` so production uses the embedded-fields path and avoids N extra round-trips when possible.
- **Parallel phase fetches**: `asyncio.gather` + `Semaphore(8)` inside `collect_field_ids_for_pipe` and `build_field_slug_map` for phases that still need `get_phase_fields`.
- **Cross-pipe field collection**: after parallel `get_pipe`, we `gather` `collect_field_ids_for_pipe` across target pipes (was serial per target).

### Tests / fixtures (issues / suggestions / nit)

- **`test_target_pipe_fetch_for_cross_pipe_fields`**: mocks match production shape (`phases` without hand-stuffed `fields`) + explicit `get_phase_fields` `AsyncMock`.
- **Shared `mock_pipefy_client`**: default `get_phase_fields = AsyncMock(return_value={"fields": []})`.
- **`fixture_ids.py`**: collapsed to category prefixes (`EXAMPLE_PIPE_ID`, `EXAMPLE_FIELD_INTERNAL_ID_2`, …); `EXAMPLE_PHASE_ID` wired into a `move_card` phase test.
- Renamed **`test_validate_behaviors_rejects_non_member_field_id`**.

### Extra improvements (same patterns, not in the original comment)

While digging in, we aligned **`build_field_slug_map`** with the embedded-fields + parallel-fetch behavior (and added `test_build_field_slug_map_uses_embedded_phase_fields`).
- **`get_pipe_members`**: `(mresult.get("pipe") or {}).get("members")` for null pipe on membership checks.
- Regression: **`test_validate_ai_automation_prompt_null_pipe_does_not_raise`**.

Commits: `fc7d5dc` (sdk/mcp behavior), `ec9c87f` (tests/fixtures). CI-style unit run: `2572 passed` with `pytest -m "not integration"` (isolated from local `.env`).

Happy to split or adjust anything if you prefer a different warning shape or concurrency cap.

gbrlcustodio:
**thought (non-blocking, second pass):** The fixture-id pattern here is "same opaque token seeded into both the mock pipe and the behavior input, then assert validator output". A per-call factory would catch any coincidental-pass bug (e.g., production code leaking the constant as a default) without changing what any test asserts. Original review missed this; not blocking.

Move to factories (`make_pipe_id()`, `make_field_id()`):
- `packages/sdk/tests/_shared/ai_agent_test_payloads.py` `minimal_behavior_dict()` / `mock_api_behavior_response()`. Defaulting `pipe_id` / `field_id` args to factory calls inherits per-call uniqueness across most downstream tests.
- `packages/mcp/tests/tools/test_ai_agent_tools.py` `_behavior_update_card_on_pipe()` + `_pipe_graph_with_field()`. Textbook "both sides seeded from one constant" pattern; thread a factory value into both per test.
- `packages/mcp/tests/tools/test_ai_tool_helpers_slug_resolution.py` `_behavior_with_fields()`. Same pattern; file already mixes the fixtures with ad-hoc `\"111\"`/`\"555\"` literals, suggesting the constants weren't load-bearing.
- `EXAMPLE_AI_ACTION_REFERENCE_UUID*`: round-trip mocks, never asserted by literal value. Default to `str(uuid4())`.

Keep as constants:
- `EXAMPLE_FIELD_INTERNAL_IDS_BY_SLUG`: `test_build_field_slug_map_from_start_form_and_phases` asserts equality against this dict, so it needs to be a frozen shared reference.
- `EXAMPLE_FIELD_SLUG*`: symbolic slugs, not numeric IDs.
- `SKILL.md` examples in `skills/ai-agents/` and `skills/automations/`: docs need stable values.

gbrlcustodio (approved):

PR #252, opened 2026-05-28
refactor(infra): submodule split + SSRF tightening

gbrlcustodio:
## Summary

Three-commit refactor of `pipefy_infra` plus a tightening pass on the SSRF defenses, applied across the workspace.

- `92fd9c3` Splits the flat `pipefy_infra` package into bounded-context submodules (`config`, `security`) and drops the package-root re-exports. Every call site now does `from pipefy_infra import security` and calls through it, so the SSRF surface is greppable for audits.
- `d51d572` Closes an IPv4-mapped IPv6 SSRF gap (`::ffff:127.0.0.1` was bypassing the hand-rolled `_PRIVATE_NETWORKS` tuple) by switching to stdlib `ipaddress.is_*` properties; adds an empty-`addr_info` guard; migrates `get_event_loop()` to `get_running_loop()`; extracts the duplicated host-root check from auth/sdk settings into `security.assert_url_is_host_root`; reorders the scheme check before the hostname check in `validate_https_url`; case-normalizes `PIPEFY_KEYCHAIN_BACKEND`.
- `db41184` Applies the actionable findings from a max-effort `/simplify` review of the prior two commits. Headlines below.

## Behavior changes operators should know about

- **MCP attachment downloads are HTTPS-only by default.** `_validate_url_safe` now reads `PipefySettings.allow_insecure_urls` and threads it through, matching the policy switch on `base_url` and `auth_url`. Operators who depended on plain-http attachment fetches must set `PIPEFY_ALLOW_INSECURE_URLS=true` or supply the file via `file_content_base64`.
- **`PIPEFY_AUTH_URL` rejects query strings and fragments** at settings construction (path is still allowed for Keycloak-style realm URLs). A stray `?` or `#` from operator copy-paste used to pass and silently corrupt the downstream `.well-known/openid-configuration` concatenation.
- **`PIPEFY_KEYCHAIN_BACKEND` is now case-insensitive.** `AUTO`, `File`, `\tauto\n`, etc. normalize via `.strip().lower()` to match the `Literal["auto", "file"]` constraint instead of failing with a cryptic Literal-validation error.
- **`PIPEFY_BASE_URL` with repeated slashes (`https://app.pipefy.com//`) now fails at startup.** The new `assert_url_is_host_root` rejects any path beyond `''` or `'/'`; the prior inline `path.strip('/')` check accepted `//`, `///`, etc.
- **The literal-IP / DNS-result blocklist widens** to include multicast (`224.0.0.0/4`), reserved (`240.0.0.0/4`), and unspecified (`0.0.0.0`, `::`) ranges in addition to the prior private / loopback / link-local set. Error messages enumerate the blocked categories explicitly.
- **Empty-DNS-answer hosts now raise** instead of silently passing through `assert_hostname_resolves_to_public_ips`. `UnicodeError` (IDN encoding failures, labels >63 chars) is also wrapped as `ValueError` alongside `socket.gaierror`.

## Surface additions in `pipefy_infra.security`

- `validate_and_assert_public_url(url, *, field_label, allow_insecure=False) -> str` composes the sync gate, hostname parse, and DNS gate into one call and returns the validated hostname.
- `assert_url_has_no_query_or_fragment(url, *, field_label)` is the path-allowed sibling of `assert_url_is_host_root` (path-rejected), used by `auth_url`.

## Removed

- `pipefy_infra.strings` (single-function module, `strip_str` inlined back into the two settings classes).
- `derived_paths_hint` kwarg on `assert_url_is_host_root` (callers wrap their own context onto the `ValueError`).

## Test plan

- [x] `uv run --frozen pytest`: 2613 passed, 45 skipped, 47.79s (+17 new tests; the prior baseline was 2596).
- [x] `uv run --frozen ruff check` clean.
- [x] `uv run --frozen ruff format --check` clean.
- [x] Empirically verified the IPv4-mapped IPv6 gap in `d51d572` against `ipaddress.IPv6Address.is_loopback` on the deployed Python build.
- [x] Empirically verified `urlparse` whitespace handling (`urlparse(' https://x ').hostname == 'x'`) and `socket.getaddrinfo` `UnicodeError` raising on long IDN labels.
- [ ] Reviewer to spot-check the operator-visible behavior changes called out above against current deploy configs (especially `PIPEFY_BASE_URL` trailing slashes and any attachment downloaders pointed at plain-http hosts).


packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
```diff
@@ -168,6 +137,13 @@ class AttachmentTools:
 
     @staticmethod
     def register(mcp: FastMCP, client: PipefyClient) -> None:
+        # Read the policy switch once at registration: HTTPS-only by default,
+        # http + internal hosts permitted only when ``PIPEFY_ALLOW_INSECURE_URLS``
+        # is set. Matches the same toggle that gates ``base_url`` / ``auth_url``
+        # in ``PipefySettings`` / ``AuthSettings`` so the attachment surface
+        # does not silently accept plain http on a tightened deployment.
+        allow_insecure_urls = PipefySettings().allow_insecure_urls
```
adriannoes: `allow_insecure_urls` is captured from a fresh `PipefySettings()` at registration, while the rest of the server wires policy from the `Settings` / `PipefyClient` instance in the container. That’s fine when everything comes from env, but if someone ever passes `Settings(pipefy=PipefySettings(allow_insecure_urls=True))` without the env var, attachment downloads would stay HTTPS-only while the client might not. I’d thread the same `pipefy` object the container already has — e.g. pass it into `register` or expose settings on `PipefyClient` — so the attachment gate can’t drift from the client config.

CHANGELOG.md
```diff
@@ -48,7 +55,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 - **CLI**: `pyyaml` dependency of `pipefy-cli` (only consumer was the deleted `skills` command).
-- **SDK**: dropped the `pipefy_sdk.utils.url_ssrf` module and its `pipefy_sdk.utils` re-exports (`validate_https_service_endpoint_url`, `assert_hostname_is_not_internal`, `assert_hostname_resolves_to_public_ips`). The shared implementation now lives in `pipefy_infra`; import directly from there (`from pipefy_infra import validate_https_service_endpoint_url`).
+- **SDK**: dropped the `pipefy_sdk.utils.url_ssrf` module and its `pipefy_sdk.utils` re-exports. The shared implementation now lives in `pipefy_infra.security`; import the module (`from pipefy_infra import security`) and call through it (`security.validate_https_url(...)`).
+- **Auth / SDK**: the per-module `_URL_SHAPE_PATTERN` regex and `_strip_str` field-validator bodies are now sourced from `pipefy_infra` (`security.URL_SHAPE_PATTERN`, `pipefy_infra.strings.strip_str`); the previous private copies in `pipefy_auth.settings` and `pipefy_sdk.settings` are gone. Behaviour is unchanged; both models still register their own `_strip_str` classmethod against their own field tuples, but the actual strip semantics live in one place.
```
adriannoes: Small release-notes nit: this bullet still points at `pipefy_infra.strings.strip_str`, but this branch deletes `strings` and keeps strip inline in the settings validators. Easy fix so external readers aren’t sent to a module that no longer exists.

adriannoes (approved):
## Summary

Strong security-focused refactor: the `pipefy_infra` submodule split is clear, the IPv4-mapped IPv6 SSRF gap is closed with stdlib `ipaddress`, and attachment downloads now honor the same HTTPS / `PIPEFY_ALLOW_INSECURE_URLS` policy as the rest of the stack. I checked out `db41184` locally — `uv run ruff check .` and `uv run pytest -m "not integration"` (2613 passed) are green, matching CI.

No blockers from my side; happy to merge once the small follow-ups below are addressed (inline threads or a quick follow-up commit).

## Also noted

- **MCP tool docs:** The CHANGELOG correctly calls out HTTPS-only attachment downloads and the dev escape hatch, but `docs/mcp/tools/pipes-and-cards.md` and `database-tables.md` still only describe `file_url` vs `file_content_base64`. Worth a short note there so operators and agents routing via docs see the policy without digging through CHANGELOG.

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
gbrlcustodio: **deferred:** filed #253 as the proper fix and reverted the in-PR threading in 43e207d. Pushing URL-to-bytes resolution into the SDK makes `PipefyClient` the only reader of `allow_insecure_urls`, closing the drift gap structurally rather than by convention. Inline caveat in `attachment_tools.py` documents the gap until #253 lands. Env-driven deployments (every shipping consumer today) are unaffected.

CHANGELOG.md
gbrlcustodio: **resolved:** rewrote the bullet in eedc367. `_strip_str` body stays inlined as a 2-line classmethod on each settings model; only `URL_SHAPE_PATTERN` is shared via `pipefy_infra.security`. No more reference to the deleted `pipefy_infra.strings` module.

gbrlcustodio:
**resolved (F2):** added an HTTPS-only / `PIPEFY_ALLOW_INSECURE_URLS` clause to the `upload_attachment_to_card` row in `docs/mcp/tools/pipes-and-cards.md` and the `upload_attachment_to_table_record` row in `docs/mcp/tools/database-tables.md` in eedc367. Operators reading the tool docs now see the policy without digging into CHANGELOG.

gbrlcustodio:
**note (follow-up):** the eedc367 fix closes the drift gap but leaves the deeper layering issue intact: `PipefySettings.allow_insecure_urls` still has two readers because `PipefyClient` accepts only `file_bytes` and MCP does the URL fetch. Filed #253 to move URL-to-bytes resolution into the SDK so the client becomes the only consumer of the policy. Not blocking this PR.

PR #254, opened 2026-05-28
feat(portal): sub-portals CRUD and code review fixes (task 6)

adriannoes:
- Ships [Task 6.0](https://github.com/gbrlcustodio/pipefy-mcp-server/issues/211) sub-portal capabilities: SDK methods on Interfaces + internal_api, six MCP tools, six CLI `portal sub-portal` subcommands, parity rows (148 tools), and docs.
- Addresses code review findings (F1–F6) on the unhappy path: `PERMISSION_DENIED` from internal_api now surfaces as `PortalPermissionError` in the SDK and actionable MCP guidance (`create_portal` / `manage_portals`) instead of raw `ValueError` text with `correlation_id`.
- Adds regression tests for the realistic internal_api `ValueError` path, integration skip hygiene, and parametrized invalid-ID coverage (MCP + CLI).

## Code review fixes (this PR slice)

| Finding | Change |
|---------|--------|
| F1 | `unpublish_sub_portal` SDK docstrings distinguish unpublish vs detach vs delete |
| F2 | `_execute_internal_api_query_with_portal_errors` on four internal_api sub-portal methods |
| F3 | `map_portal_error_to_message` extracts `[code=…]` from any exception; strips diagnostic markers |
| F4 | MCP tool tests mock `ValueError` with `[code=PERMISSION_DENIED]` |
| F5 | Parametrized invalid-ID tests (MCP + CLI) |
| F6 | Integration test catches `(TransportQueryError, ValueError)` on publish/unpublish skip |


gbrlcustodio:
**issue (blocking):** `PortalService._internal_api_client` is never wired up, so 5 of the 6 new sub-portal tools fail at runtime.

`PipefyClient.__init__` constructs `PortalService(settings=settings, auth=auth)` at `packages/sdk/src/pipefy_sdk/client.py:123` (no `internal_api_client` kwarg → defaults to `None`). The setter at `client.py:143-149` only stores `self._internal_api_client = client`; nothing forwards it into `_portal_service`.

Runtime impact: every call to `update_sub_portal_element`, `publish_sub_portal`, `unpublish_sub_portal`, `delete_sub_portal_element`, `delete_sub_portal` hits `PortalService.execute_internal_api_query` → `raise ValueError("Internal API client is not configured.")` (`portal_service.py:152-155`). The new `_execute_internal_api_query_with_portal_errors` wrapper sees no `[code=PERMISSION_DENIED]` marker and re-raises; MCP wraps the bare string in `build_error_payload`. `create_sub_portal` is unaffected because it uses the Interfaces path.

For comparison, `AiAutomationService` solves the same chicken-and-egg via external construction + injection: the container builds `AiAutomationService(client=internal_client)` and calls `pipefy_client.set_ai_automation_service(...)` (`packages/mcp/src/pipefy_mcp/core/container.py:93-95`). `PortalService` doesn't follow that pattern (no `set_portal_service` analog) nor any other — it's left dangling.

Tests miss this:
- SDK unit tests construct `PortalService` directly with a mocked `internal_api_client` (`test_portal_service.py:99-108`).
- MCP/CLI tests mock `PipefyClient.<method>` wholesale.
- The live integration test catches `(TransportQueryError, ValueError)` as `pytest.skip(...)`, so the missing-client `ValueError` is indistinguishable from a real API error and the test skips green.

Fix options (in order of consistency with the codebase):
1. Mirror `AiAutomationService`: have the container construct `PortalService(internal_api_client=internal_client, ...)` and inject via a new `set_portal_service` setter on `PipefyClient`.
2. Thread `internal_api_client` into `PipefyClient.__init__` so `PortalService` is wired at construction.
3. Add `PortalService.set_internal_api_client(client)` and forward from `PipefyClient.set_internal_api_client`.

gbrlcustodio:
**suggestion (non-blocking):** `PortalVisibility = Literal["internal", "private", "public"]` at `packages/mcp/src/pipefy_mcp/tools/portal_tools.py:33` is redefined locally even though `pipefy_sdk.models.portal:11` already exports the same alias (re-exported in `__all__`). The file already imports `PortalElementType` from that module on line 13, so the path is established.

Drift risk: if the SDK adds a new value (e.g. `"unlisted"`), `UpdatePortalInput` and CLI flags accept it while the MCP tool signature continues to reject it at the JSON-schema layer with no test signal.

Pre-existing in the file (not introduced by this PR), so non-blocking — but a one-line cleanup while the portal surface is open:

```python
from pipefy_sdk.models.portal import (
    CreatePortalElementInput,
    PortalElementType,
    PortalVisibility,
    UpdatePortalElementInput,
)
```

and drop the local `PortalVisibility = Literal[...]` (plus the `Literal` import if unused elsewhere).

gbrlcustodio:
**suggestion (non-blocking):** Five sub-portal write tools in `packages/mcp/src/pipefy_mcp/tools/portal_tools.py` (`update_sub_portal_element` L797, `publish_sub_portal` L847, `unpublish_sub_portal` L886, `delete_sub_portal_element` L939, `delete_sub_portal` L998) share the same mechanical shape, with only five varying pieces (id set, debug string, client method, mutation key, failure verb):

```python
# PRELUDE — one block per id param
<id>, err = validate_tool_id(<id>, "<id>")
if err is not None:
    return err
# ... repeat for each id ...

await ctx.debug(f"<tool_name>: ...")

# (optionally) check_destructive_confirmation guard

try:
    result = await client.<method>(...)
    return build_error_payload(map_portal_error_to_message(exc))

if result.get("<MutationKey>", {}).get("success"):
    return build_success_payload(result, include_parsed=True)
return build_error_payload(
    f"Failed to <verb> ... '<id>'. Please try again or contact support."
)
```

About 14-16 lines of mechanical repetition per tool, ~80 LOC across the five.

The cost compounds with the other findings in this PR: the `result.get(<key>, {}).get("success")` AttributeError-on-explicit-null bug, a future correlation_id surfacing change, or any tweak to `map_portal_error_to_message`'s call shape each require five patch sites.

Sketch of the collapse with two small helpers in `portal_tool_helpers.py`:

```python
def validate_tool_ids(raw: dict[str, str]) -> tuple[dict[str, str] | None, dict[str, object] | None]:
    cleaned = {}
    for name, value in raw.items():
        ok, err = validate_tool_id(value, name)
        if err is not None:
            return None, err
        cleaned[name] = ok
    return cleaned, None


def finalize_mutation(
    result: dict[str, Any],
    mutation_key: str,
    failure_message: str,
) -> dict[str, object]:
    payload = result.get(mutation_key) or {}  # also fixes the explicit-null bug
    if payload.get("success"):
        return build_success_payload(result, include_parsed=True)
    return build_error_payload(failure_message)
```

Each tool then collapses to ~14 lines.

Honest counter: this is on the edge of "rule of three" territory. Defensible to ship as-is if portal sub-portals don't grow further; earns its keep the moment another internal_api `result.get(key).get("success")`-shaped tool appears.

adriannoes:
@gbrlcustodio, addressed the review feedback in b6848bd:

**Blocking**
- `PipefyClient.set_internal_api_client` now forwards to `PortalService.set_internal_api_client`, so sub-portal internal_api mutations work at runtime (MCP container + CLI auth paths unchanged).
- Regression: `test_set_internal_api_client_forwards_to_portal_service`.

**Non-blocking**
- `PortalVisibility` imported from `pipefy_sdk.models.portal` (removed local `Literal` alias).
- Sub-portal write tools deduplicated via `validate_tool_ids`, `finalize_internal_api_mutation`, and `run_sub_portal_internal_api_tool` (also fixes explicit-null mutation payloads).
- Integration: missing internal_api client fails instead of skipping green.

**Docs**
- Removed real org ids / PipeClaw examples from `docs/mcp/tools/portal.md`.

Portal/sub-portal unit tests green locally.

gbrlcustodio (approved):

Issue #253, opened 2026-05-28
refactor(mcp): drop URL ingestion from attachment uploads


gbrlcustodio:
## Context

Originally surfaced from PR #252 review as a layering issue: `PipefySettings.allow_insecure_urls` had two readers (`PipefyClient` and `pipefy_mcp.tools.attachment_tools`), and the proposal was to move URL-to-bytes resolution into the SDK so the client owned the SSRF policy end-to-end.

That framing assumed the MCP server would keep accepting `file_url` indefinitely. Re-examining the distribution model: this MCP is shipped as code that runs in the user's environment as a subprocess of the agent runtime. The trust boundary is the user; the server has the same filesystem and network access the user already has.

In that model the SSRF guard, 100 MiB cap, and redirect handling are not buying a security boundary. They are appropriate for a hosted MCP, not for a local subprocess where the user can already `cat` any file and `curl` any URL.

## New framing

Drop URL ingestion from MCP attachments entirely. Accept `file_path` instead. `file_content_base64` stays as the fallback for in-memory bytes the agent never wrote to disk. The drift gap closes because the MCP no longer has any reader of `allow_insecure_urls`.

## What ships

- SDK input models (`UploadAttachmentToCardInput`, `UploadAttachmentToTableRecordInput`): `file_url` removed, `file_path` added, `file_name` optional (inferred from basename when omitted).
- MCP attachment tools: drop `_validate_url_safe`, `_download_file_bytes`, `_resolve_file_bytes`, the size/redirect constants, the `httpx` import, the `PipefySettings` import. ~80 LoC of MCP-side download plumbing goes away.
- `packages/mcp/AGENTS.md` (with `CLAUDE.md` symlink) documents the `GATED:<PROFILE>` convention. `GATED:SELF_HOSTED` is tagged at the removal site so a future hosted profile can reintroduce URL ingestion behind a capability flag, with the SSRF policy injected from container settings rather than read from a fresh `PipefySettings()`.
- CLI `--file` gains `~` expansion for parity with MCP.
- New skill at `skills/attachments/pipefy-attachments/SKILL.md` walks the `file_path` vs `file_content_base64` decision and the local-subprocess trust model.

## What disappears vs the original proposal

- No SDK `FileSource` union, no `FileUrlSource`/`FileBytesSource` discriminator. The SDK's existing `file_bytes: bytes` contract is correct for a local-only world.
[...]

PR #255, opened 2026-05-28
refactor(mcp): drop attachment URL ingestion, accept file_path

gbrlcustodio:
## Summary

- MCP attachment tools accept only `file_path` (local filesystem path the MCP server reads). `file_url` and `file_content_base64` are gone. `file_name` is optional, inferred from the path basename.
- New `LocalFile` (in `pipefy_infra.filesystem`) owns the read pipeline: expand `~`, validate, optional size cap, read bytes. Every failure surfaces as `LocalFileError`, so MCP and CLI each catch one exception.
- New `Attachment` (in `pipefy_sdk.attachment`) binds a `LocalFile` to a target Pipefy field and runs the upload pipeline. After success it carries the result, `download_url`, and `storage_path`.
- SDK input models drop the source-XOR validator and `file_content_base64`; `file_path` is required (`NonBlankStr`); `file_name` stays optional and is resolved by `Attachment`. Drops the `cast(str, file_name)` in both tools.
- SSRF guard, redirect handling, and the 100 MiB download cap deleted from `attachment_tools.py` (the cap moved into `LocalFile` as a constructor param, applied via `MAX_ATTACHMENT_SIZE_BYTES`). Local-subprocess trust model justifies it.
- `packages/mcp/AGENTS.md` (with `CLAUDE.md` symlink) documents the `GATED:SELF_HOSTED` marker. The trust-model bullet about base64 is removed since the input mode is gone.
- CLI `--file` continues to support `~` expansion via the same `LocalFile`. CLI `~unknownuser` now reports `typer.BadParameter` instead of a `RuntimeError` traceback.
- `skills/attachments/pipefy-attachments/SKILL.md`: drops base64 sections; recommends writing generated bytes to a temp file and passing the path.

Addresses #253 review (PR #255):
- `~unknownuser` `RuntimeError` is now normalized to `LocalFileError` in `LocalFile.read()`. MCP returns the documented `step=file_read` envelope; CLI emits `typer.BadParameter`. Unit tests on both surfaces.
- Em dashes in `SKILL.md` "See also" and the directory-test docstring replaced with colons.
- Diff-anchored "mirrors the prior URL-download limit" comment dropped from `models/attachment.py`.
- `cast(str, data.file_name)` dissolves; `Attachment.name` derives at access time.

## Test plan

- [x] `uv run pytest packages` (2467 passed, 45 skipped)
- [x] `uv run ruff check packages` clean
- [x] `uv run ruff format --check packages` clean
- [ ] End-to-end MCP: `upload_attachment_to_card(file_path="~/foo.pdf", ...)`
- [ ] End-to-end MCP: oversize file (>100 MiB) rejected at `step=file_read`
- [ ] End-to-end MCP: `~unknownuser/foo.bin` returns `step=file_read` envelope
- [ ] End-to-end CLI: `pipefy attachment upload --file ~/foo.pdf ...`


packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
```diff
@@ -137,74 +51,65 @@ class AttachmentTools:
...
-                            file_url.strip(), allow_insecure=allow_insecure_urls
-                        ),
-                        None,
-                    )
+                # Match the SDK validator's whitespace-aware source check
+                # (`_source_nonempty`): a whitespace-only `file_path` is treated
+                # as no path, so the base64 source takes over.
+                if file_path and file_path.strip():
+                    await ctx.debug(f"{debug_prefix}: reading file_path")
+                    p = Path(file_path.strip()).expanduser()
```
adriannoes: `Path("~ghost/foo").expanduser()` raises `RuntimeError("Could not determine home directory.")` when `~<user>` names someone who is not a local account. `_read_local_or_decode` only catches `(OSError, binascii.Error, ValueError)`, so that `RuntimeError` escapes — on MCP it bypasses the `step=file_read` envelope the attachments skill documents and surfaces as a generic JSON-RPC error; on CLI (`attachment.py` around the same `expanduser()` call) it dumps a traceback instead of `typer.BadParameter`.

I'd either add `RuntimeError` to the except tuple or wrap `expanduser()` in a small helper that re-raises as `ValueError(f"Cannot expand ~ in {path}: unknown user")`. Fixed when `file_path="~ghostuser_xyz/foo.bin"` returns the documented envelope on MCP and a one-line `typer.BadParameter` on CLI, with unit tests on both surfaces.

adriannoes (commented):
Solid migration from `file_url` to local `file_path` / `file_content_base64`: layering is clean (SDK cap + MCP/CLI preflight), SSRF/httpx URL ingestion removal matches the local-subprocess trust model, tests and CI are green, and the new attachments skill plus `GATED:SELF_HOSTED` docs are helpful. I ran the attachment-focused unit suite locally on head `77994ce` (94 tests, ruff clean).

## Recommended before merge

One inline on `expanduser()` / unknown `~<user>` paths — those `RuntimeError`s skip the documented `step=file_read` envelope on MCP and become tracebacks on CLI. Small catch + tests on card, table, and CLI would close it.

## Also noted

For `file_content_base64`, we still `b64decode` before the 100 MiB cap. The input is already bounded by the MCP request, so this is not a blocker — an early length check on the cleaned base64 string (`len(cleaned) * 3 // 4`) would skip the largest allocation on oversize payloads.

skills/attachments/pipefy-attachments/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+- **`step=presigned_url`.** Organization id rejected, field id not an attachment, or Pipefy refused the request. Recovery: confirm `organization_id` with `get_organization` and that the field is actually an attachment field on the target card or record.
+- **`step=s3_upload`.** The presigned URL expired before PUT, or content/headers did not match what was signed. Recovery: retry the tool to obtain a fresh presigned URL. For unusually large files, expect Pipefy's storage policy to enforce the cap.
+- **`step=field_update`.** The field rejected the new attachment list (wrong type, missing permission). Recovery: confirm the field accepts attachments and that the caller has write access to the card or record.
+
+## See also
+
+- `skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md` — finding card ids and attachment field slugs.
+- `skills/database-tables/pipefy-database-tables/SKILL.md` — finding table record ids and field slugs.
```
Danielmoraisg: nit (writing standard): both "See also" bullets use an em dash (` — `). This same PR converted the `docs/mcp/tools/*.md` tables away from ` — ` to `:`, so this reads as inconsistent. Suggest a colon:

```
- `skills/.../SKILL.md`: finding card ids and attachment field slugs.
```

(It may match the em-dash style already used in other skill files' "See also" sections, so this is low priority. Flagging because the rest of the PR's prose avoids them.)

packages/mcp/tests/tools/test_attachment_tools.py
```diff
@@ -80,335 +73,359 @@ def attachment_session(attachment_mcp_server):
...
-    ):
-        result = await _download_file_bytes(
-            "https://example.com/a", allow_insecure=False
+async def test_upload_attachment_to_card_file_path_is_directory(
+    attachment_session,
+    mock_attachment_client,
+    extract_payload,
+    tmp_path: Path,
+):
+    """A directory is not a regular file — same file_read failure."""
```
Danielmoraisg: nit (writing standard): em dash in the docstring. `"A directory is not a regular file: same file_read failure."` reads the same without it.

packages/sdk/src/pipefy_sdk/models/attachment.py
```diff
@@ -12,6 +12,34 @@
 
 APPLICATION_OCTET_STREAM = "application/octet-stream"
 
+# Hard cap that mirrors the prior URL-download limit. Pipefy's direct-upload
```
Danielmoraisg: nit (comment): "mirrors the prior URL-download limit" is diff-anchored. It describes this value as a delta from code that was just removed, which goes stale once the URL path is forgotten. The rest of the comment is the keeper (Pipefy's direct-upload policy is undocumented; fail fast before issuing a presigned URL). Suggest dropping just that opening clause and starting at "Pipefy's direct-upload size policy is not documented...".

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
```diff
@@ -281,39 +190,41 @@ async def upload_attachment_to_card(
...
                     content_type=content_type,
                 )
             except ValidationError as exc:
                 return build_upload_error_payload(
-                    message=map_upload_error_to_message(exc, "validation"),
+                    message=map_upload_error_to_message(exc),
                     step="validation",
                 )
 
+            resolved_file_name = cast(str, data.file_name)
```
Danielmoraisg: observation (design, non-blocking): the `cast` is here because `data.file_name` is typed `str | None`, but `exactly_one_file_source` has already guaranteed a non-empty `str` by this point. A single field can't express "optional on input, required on output", so each reader pays with a `cast` (here and at line 275).

If you want to drop both casts, expose the resolved value as a typed member on the model instead of mutating `file_name`:

```python
from pydantic import PrivateAttr

file_name: str | None = None           # raw input, may be None
_resolved_file_name: str = PrivateAttr()

@model_validator(mode="after")
def exactly_one_file_source(self) -> Self:
    _raise_unless_exactly_one_file_source(self.file_path, self.file_content_base64)
    self._resolved_file_name = _resolve_file_name(self.file_name, self.file_path)
    return self

@property
def resolved_file_name(self) -> str:
    return self._resolved_file_name
```

Call sites become `file_name=data.resolved_file_name` (no cast). Side benefit: the current `self.file_name = _resolve_file_name(...)` mutation works only because `validate_assignment` is off in `model_config`; assigning a `PrivateAttr` bypasses validation, so that coupling goes away too.

Fine to leave as-is for this PR and treat as a follow-up.

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
```diff
@@ -349,34 +262,34 @@ async def upload_attachment_to_table_record(
...
```
Danielmoraisg: Same `cast` as line 203; the `resolved_file_name` property suggested there removes this one too.

Danielmoraisg (approved):

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
gbrlcustodio: **resolution** (fd0e1c0): path expansion moved into `LocalFile.read()` (`packages/infra/src/pipefy_infra/filesystem.py:62-65`), which catches `RuntimeError` from `expanduser()` and re-raises as `LocalFileError`. MCP returns the documented `step=file_read` envelope; CLI raises `typer.BadParameter`. Covered by `packages/infra/tests/test_filesystem.py::test_local_file_unknown_user_raises_normalized_error`.

skills/attachments/pipefy-attachments/SKILL.md
gbrlcustodio: **resolution** (fd0e1c0): both bullets now use colons.

packages/mcp/tests/tools/test_attachment_tools.py
gbrlcustodio: **resolution** (fd0e1c0): the test file was rewritten as part of the `LocalFile` + `Attachment` split. The offending docstring is gone.

packages/sdk/src/pipefy_sdk/models/attachment.py
gbrlcustodio: **resolution** (fd0e1c0): dropped the diff-anchored opener. The comment now starts at "Pipefy's direct-upload size policy is not documented..." (`packages/sdk/src/pipefy_sdk/models/attachment.py:14-16`).

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
gbrlcustodio: **resolution** (fd0e1c0): solved by a different route. Name resolution moved out of the Pydantic model into the new `Attachment` domain object (`packages/sdk/src/pipefy_sdk/attachment.py`). The input model is now pure validation (`file_name: str | None` stays optional, no mutation, no `PrivateAttr`), and the MCP/CLI surfaces pass `data.file_name` through unchanged. `Attachment.name` falls back to `file.name` when no explicit name is given, at access time, so no caller needs a `cast`.

packages/mcp/src/pipefy_mcp/tools/attachment_tools.py
gbrlcustodio: **resolution** (fd0e1c0): same fix as the line-203 thread. The cast at this line is gone too.

PR #261, opened 2026-06-01
feat(installer): one-command curl installer (install.sh)

### LINKED ISSUE #231: install: one-command curl installer (CLI + MCP + skills + client config)
## Outcome

One command installs the CLI, MCP server, skills (via `skills.sh`), and writes an MCP server entry for the user's host of choice.

```sh
curl -fsSL https://raw.githubusercontent.com/gbrlcustodio/pipefy-mcp-server/main/install.sh | sh
```

## Scope

**New files**

- `install.sh` (repo root) — POSIX shell script. Steps:
  1. Detect `uv`. If missing, prompt to install via the upstream `curl -LsSf https://astral.sh/uv/install.sh | sh`; `--yes` to skip the prompt.
  2. `uv tool install "pipefy-cli @ git+https://github.com/gbrlcustodio/pipefy-mcp-server"` and `uv tool install "pipefy-mcp-server @ git+https://github.com/gbrlcustodio/pipefy-mcp-server"`, pinned to a release tag (`@v0.2.0-beta.1` or newer). Resolve via `gh api repos/gbrlcustodio/pipefy-mcp-server/releases/latest` or a `VERSION=` constant near the top of the script. Switch to bare PyPI names (`uv tool install pipefy-cli pipefy-mcp-server`) once the project ships on PyPI at v1.0.
  3. Prompt to install skills via `npx skills add gbrlcustodio/pipefy-mcp-server --all -y`; skip if `npx` is unavailable or `--no-skills` is set.
  4. Prompt to register the MCP server in a client config. Support `--client claude-code|claude-desktop|cursor|none`. For `none`, print the JSON snippet for the user to paste manually.
  5. Print a next-step hint pointing at `/pipefy-login` (from #229) or the `uvx --from git+https://github.com/gbrlcustodio/pipefy-mcp-server pipefy auth login` fallback.

**Flags**

- `--yes` — skip all confirmation prompts.
- `--no-skills` — skip the `npx skills` step.
- `--client <id>` — `claude-code`, `claude-desktop`, `cursor`, or `none`.
- `--version <tag>` — pin to a specific release tag.
- `--prefix <path>` — passthrough to `uv tool install --tool-dir`.
- `--allow-root` — opt-in for root execution; refuse otherwise.

**Safety**

- `set -euo pipefail`.
- Print every command before executing.
- No `eval`.
- The only HTTP-pipe-to-shell is the upstream `uv` installer, behind an opt-in prompt.

**Docs**

- `README.md` — new "Quick install" subsection above the existing `uvx` / `uv tool install` blocks. Existing blocks remain as advanced paths.
- `docs/setup.md` — same.

**Replaces**

- `bootstrap.sh` — either delete (in favor of `install.sh --dev`) or rewrite as a thin shim that delegates. Decide during PR review.

## Acceptance criteria

- [ ] `bash install.sh --yes --no-skills --client none` in a fresh container exits 0; `pipefy --version` and `pipefy-mcp-server --help` both succeed; the JSON snippet for the chosen client is printed.
- [ ] Re-running the installer is idempotent (no double-install errors).
- [ ] `shellcheck install.sh` clean.

## Dependencies

- Should land after #229 so the post-install hint can reference `/pipefy-login`.
- Should land after #230 so the skills step uses `npx skills add` rather than the removed `pipefy skills` command.

gbrlcustodio:
## Summary

Adds a single POSIX `install.sh` at the repo root that installs the Pipefy CLI + MCP server end-to-end in one command, with no per-release maintenance on README install snippets.

```sh
curl -fsSL https://raw.githubusercontent.com/gbrlcustodio/pipefy-mcp-server/main/install.sh \
  | sh -s -- --client cursor
```

### How it works

- Resolves the latest GitHub Release tag via `GET /repos/<repo>/releases` (takes `.[0].tag_name`, which includes prereleases the `/releases/latest` endpoint hides). `--version vX.Y.Z` overrides.
- **Discovers the release's wheel assets** instead of composing wheel URLs in shell. The installer fetches `/repos/<repo>/releases/tags/<tag>`, parses every `.whl` `browser_download_url`, and splits them into the main wheel for each tool plus `--with` deps. No hardcoded package list, no PEP 440 normalization in `sed`. If the release ships an extra workspace package next month, the installer picks it up automatically.
- Runs `uv tool install --force` twice (CLI, MCP server), each with the other wheels passed as `--with`. After install, `pipefy-mcp-server` is on PATH, so client configs collapse to `{"command": "pipefy-mcp-server"}`.
- Optionally installs skills via `npx skills add <repo> -y` (skipped if `--no-skills` or `npx` missing).
- Writes the MCP server registration into the user's chosen client:
  - **Cursor / Claude Desktop**: JSON merge via inline `python3` heredoc, preserving any pre-existing `mcpServers` entries.
  - **Codex**: grep-guarded append of `[mcp_servers.pipefy]` to `~/.codex/config.toml`.
  - **Claude Code**: prints `/plugin marketplace add` + `/plugin install pipefy@pipefy` instructions (the installer can't write Claude Code's plugin store).
  - **none** (default if `--client` omitted): prints the JSON snippet to paste manually.

### Safety

- `set -eu`, no `pipefail` (not POSIX). Portable across `dash`, `zsh`-as-`sh`, BusyBox.
- Every external command printed (`+ uv tool install --force ...`) before execution; `--dry-run` skips execution entirely.
- No `eval`. No filesystem writes in dry-run.
- Refuses to run as `root` unless `--allow-root`.
- The only HTTP-pipe-to-shell call is the upstream `astral.sh/uv/install.sh`, behind a `confirm()` (or `--yes`).

### Flags

| Flag | Behavior |
|---|---|
| `--yes`, `-y` | Skip all confirmation prompts. |
| `--no-skills` | Skip the `npx skills add` step. |
| `--client <id>` | One of `claude-code\|claude-desktop\|cursor\|codex\|none`. |
| `--version <tag>` | Pin a specific GitHub Release tag (e.g. `v0.2.0-beta.2`). |
| `--prefix <dir>` | Exported as `UV_TOOL_DIR` for `uv tool install`. |
| `--allow-root` | Opt-in for root execution. |
| `--dry-run` | Print every command without executing. |
| `-h`, `--help` | Show help. |

### Docs

- `README.md` gains a **Quick install (recommended)** section above the existing client wiring; the previous JSON / TOML blocks stay below as **Manual MCP client wiring (advanced)**.
- `RELEASE.md` step 8 picks up a sanity-check bullet: each release run a `--dry-run` against `install.sh` on `main` and confirm `Resolved tag: vX.Y.Z` matches the just-cut tag. No per-release maintenance otherwise.
- `CHANGELOG.md` gains a single `### Added` entry under `[Unreleased]`.

## Test plan

- [x] `shellcheck --shell=sh install.sh` clean.
- [x] `sh -n install.sh` parses.
- [x] `./install.sh --help` prints the documented flag set.
- [x] `./install.sh --yes --no-skills --client none --dry-run --version v0.2.0-beta.1` enumerates the 3 wheels, prints the two `uv tool install` commands with `--with` populated correctly (each tool's main URL identified), and prints the manual snippet.
- [x] `./install.sh --yes --no-skills --client none --dry-run` (no `--version`) hits the live GitHub API and resolves the same tag.
- [x] `./install.sh --yes --no-skills --client cursor --dry-run` reports the JSON-merge intent for `~/.cursor/mcp.json`.
- [x] `./install.sh --yes --no-skills --client codex --dry-run` reports the TOML append intent.
- [x] `./install.sh --yes --no-skills --client claude-code --dry-run` prints the `/plugin` instructions.
- [x] `./install.sh --yes --client foo --dry-run` errors with exit 1 and the documented allowed-values list.
- [x] JSON merge round-trip on a `mcpServers` config containing an unrelated server preserves the unrelated entry and is idempotent across re-runs.
- [ ] End-to-end install on a fresh container, deferred to **after v0.2.0-beta.2 ships**: beta.1 only ships 3 of 5 wheels (`pipefy_cli`, `pipefy_mcp_server`, `pipefy_sdk`), so `uv` will fail to resolve `pipefy-auth` / `pipefy-infra` at install time. The first release with a self-consistent wheel set is `v0.2.0-beta.2` per `release.yml:69-84`. Live smoke runs as part of beta.2's release verification, not this PR's merge gate.

Closes #231.


install.sh
```diff
@@ -0,0 +1,421 @@
...
+        say "  Via Claude Code:     /pipefy:login"
+    fi
+}
+
+main() {
+    parse_args "$@"
+    refuse_root
+    detect_platform
+    case "$CLIENT" in
+        cursor|claude-desktop) require_python3 ;;
```
adriannoes: For `--client cursor` / `claude-desktop`, `require_python3` in `main()` runs even when `--dry-run` is set, so the script exits before it can print the JSON-merge preview — even though `json_merge_pipefy` never calls Python in dry-run mode. That makes dry-run unusable on images that only have `sh` and `curl` (common for CI sanity checks). I'd drop the early check in `main()` and keep the existing `require_python3` inside `json_merge_pipefy` after the dry-run guard, so preview and real install stay aligned.

install.sh
```diff
@@ -0,0 +1,421 @@
...
+EOF
+}
+
+write_client_config() {
+    case "$CLIENT" in
+        cursor)
+            json_merge_pipefy "$HOME/.cursor/mcp.json"
+            ;;
+        claude-desktop)
+            json_merge_pipefy "$(claude_desktop_config_path)"
```
adriannoes, line 347: On Linux with `--client claude-desktop`, the platform guard at L245 lives inside `$(claude_desktop_config_path)`. POSIX `set -e` doesn't propagate out of a `$( )` used as an argument, so the user sees the "Claude Desktop has no Linux build" message *and* a confusing `+ ensure mcpServers.pipefy in  (preserve…)` with a blank path, *and* "Install complete.", and the script exits 0. In real-run mode it instead reaches `python3 - ""` and falls over on a `pathlib.Path("")` `IsADirectoryError`. Easiest fix is to validate the `CLIENT:OS` combo up front in `main()` (e.g. `case "$CLIENT:$OS" in claude-desktop:Linux) err "…" ;; esac`) so the helper stays pure and the error happens before any side effect. Capturing into a variable first (`cdesk=$(claude_desktop_config_path)`) would also catch it via `set -e` on the assignment, but the up-front guard reads better.

adriannoes (changes requested):
## Summary

Solid installer work — wheel discovery, sibling-tool filtering, atomic JSON merge, and the iterative hardening in the commit history all show up well. Review included a local shell smoke pass on head `3e9f6ea` (dry-run paths, POSIX `set -e` repro for the claude-desktop case); CI is green.

Two shell-semantics issues are worth fixing before merge; both have inline threads.

## Required before merge

- **F1** — `require_python3` in `main()` blocks `--dry-run` for JSON clients on machines without Python, even though the merge path doesn't need it in dry-run. That breaks headless preview/CI sanity checks.
- **F4** — Linux + `--client claude-desktop`: `err` inside `$(claude_desktop_config_path)` doesn't abort the outer script, so dry-run can exit 0 with a blank config path (RELEASE.md step 8 on a Linux runner would look fine when it shouldn't).

## Also noted

- **F2** — When `--version` pins a tag, `npx skills add` still tracks the repo default; consider passing the same ref if the skills CLI supports it so pinned installs don't mix skill content with wheel versions.
- **F5** — If `mcpServers` exists but isn't a JSON object, the merge silently replaces it with `{}`; matching the root-object error path (`--client none`) would stay consistent with the "don't destroy user config" goal.

Coordinate the README quick-install path with **v0.2.0-beta.2** (first 5-wheel release) — your test plan already defers e2e until then; today's latest tag still resolves to beta.1 with three wheels.

install.sh
gbrlcustodio: Fixed in edfadf4: dropped the eager `require_python3` from `main()`. `json_merge_pipefy` already calls it after the dry-run guard, so dry-run on `sh`+`curl`-only images now reaches the preview.

install.sh
gbrlcustodio: Fixed in 3f6961e: moved the `CLIENT:OS` guard into `main()`, so Linux + `--client claude-desktop` errors before any side effect (no blank-path preview, no `python3 - ""` IsADirectoryError on real runs). Reduced `claude_desktop_config_path()` to just the Darwin printf, since `detect_platform` plus the new guard make the other branches unreachable.

adriannoes (approved):
Re-reviewed after your fixes on `0cf548f`. 

The round-1 items look good: dry-run no longer requires `python3` up front (`edfadf4`), and Linux + `--client claude-desktop` fails early in `main()` instead of falling through the `$( )` / `set -e` trap (`3f6961e`). CI is green; shell smoke on dry-run paths looks solid.

**Approved**, good to merge from my side.

## Optional follow-ups (not blocking)

- **Skills + `--version`:** when a release tag is pinned, `npx skills add` still tracks the repo default; passing the same ref (e.g. `@$TAG`) would keep skills aligned with the wheels if the CLI supports it.
- **JSON merge edge case:** if `mcpServers` exists but isn't an object, the merge still replaces it with `{}`; matching the root-object error path (`--client none`) would be a nice consistency polish later.

Live e2e install sanity still makes sense on **v0.2.0-beta.2** per your test plan — no change needed there.

PR #262, opened 2026-06-01
feat(automation): field_map guidance, event attributes catalog, and create preflight

adriannoes:
## Summary

- **Automation:** Document `field_map` for traditional `update_card_field` rules (numeric `internal_id`, not slug); add `get_automation_event_attributes` (SDK, MCP, CLI); preflight `field_map.fieldId` and `card_moved` + `move_single_card` transitions on **`create_automation` only** before GraphQL.
- **Agent contract:** Skill and `docs/mcp/tools/automations-and-ai.md` state create-only preflight and text-only move-transition errors (recovery via error message or `get_phase_allowed_move_targets`, not a `valid_destinations` envelope field).
- **Portal:** Expand `docs/mcp/tools/portal.md` and add `skills/portal-setup/pipefy-portal-setup/SKILL.md`.


docs/config.md
```diff
@@ -52,7 +52,7 @@ The same fields populate from environment variables in upper-case `PIPEFY_<NAME>
 
 | Variable | Default | Effect |
 |----------|---------|--------|
-| `PIPEFY_BASE_URL` | `https://app.pipefy.com` | Drives the four API endpoints (`graphql_url`, `internal_api_url`, `interfaces_graphql_url`, `service_account_url`) as computed properties. Set once for non-prod environments. |
+| `PIPEFY_BASE_URL` | `https://app.pipefy.com` | Drives the four API endpoints (`graphql_url`, `internal_api_url`, `interfaces_graphql_url`, `service_account_url`) as computed properties. Set once for non-prod environments. Portal tools use `interfaces_graphql_url` (`<base>/graphql/interfaces`) and `internal_api_url` (`<base>/internal_api`); see [`mcp/tools/portal.md`](mcp/tools/portal.md#endpoints). |
```
gbrlcustodio: nitpick(non-blocking): I don't think these add extra value here. If we need a tools vs schema table, which I don't think we do, it better be in a purposely created comparison table.

packages/mcp/tests/test_server.py
```diff
@@ -34,7 +34,17 @@ def client_session():
...
+    mock_container.initialize_services = AsyncMock()
+    mock_container.pipefy_client = MagicMock()
+
+    with (
+        patch("pipefy_mcp.server.settings", _MINIMAL_PIPEFY_SETTINGS),
+        patch(
+            "pipefy_mcp.server.ServicesContainer.get_instance",
+            return_value=mock_container,
+        ),
+    ):
```
gbrlcustodio: question (non-blocking): I’m confused about why this change was necessary. We didn’t modify the server initialization, so I don’t understand why it had to be altered.

packages/sdk/src/pipefy_sdk/automation_preflight.py
```diff
@@ -130,8 +133,169 @@ async def validate_traditional_automation_move_transition(
...
+            f'field_map fieldId "{field_id}" must be the numeric internal_id on pipe '
+            f"{action_pipe_id}, not the field slug used by update_card_field. {discovery}"
+        )
+    return (
+        f'field_map fieldId "{field_id}" was not found on pipe {action_pipe_id}. '
+        f"{discovery}"
+    )
+
+
+async def _load_action_pipe_internal_field_ids(
```
Danielmoraisg, line 225: This lazy phase-fetch is effectively unreachable. `GET_PIPE_QUERY` (`queries/pipe_queries.py:13-20`) always embeds `phases { fields { internal_id } }`, so `embedded_fields` is always a list. The branch only fires when a phase has an empty field list, where `get_phase_fields` returns the same empty list. So it adds a second network path and failure handler that production never needs. `test_validate_field_map_lazy_loads_phase_fields` passes only because it hand-builds a response with `fields: None`, which the real query never returns.

Either drop the branch and rely on the embedded fields, or, if this is deliberate future-proofing against a query change, note that and test against a realistic `get_pipe` response shape.

packages/sdk/src/pipefy_sdk/automation_preflight.py
```diff
@@ -130,8 +133,169 @@ async def validate_traditional_automation_move_transition(
...
+        for field in phase_data.get("fields") or []:
+            if not isinstance(field, dict):
+                continue
+            iid = field.get("internal_id")
+            if iid is not None:
+                internal_ids.add(str(iid))
+    return internal_ids
+
+
+async def validate_automation_field_map_field_ids(
```
Danielmoraisg, line 245: Two small design points, both non-blocking:

1. Decision and effect are mixed here: the function loads ids (I/O) and then classifies slug-vs-unknown and raises in the same body, so the branching can only be tested with a `get_pipe` mock. Extracting a pure `classify_field_map_field_ids(destination_ids, known_internal_ids) -> (field_id, reason) | None` would let the classification be unit-tested with no mocks, leaving this shell to load and raise.
2. Fail-fast: the slug-format check (`_DIGITS_ONLY_RE`) needs no pipe data but runs after `get_pipe`. A payload that is all slugs still pays a network round-trip before failing. Running the format check first avoids that.

packages/mcp/src/pipefy_mcp/tools/automation_tools.py
```diff
@@ -206,6 +216,32 @@ async def get_automation_events(
                 "Automation events catalog retrieved.",
             )
 
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=True, destructiveHint=False),
+        )
+        async def get_automation_event_attributes(ctx: Context) -> dict[str, Any]:
```
Danielmoraisg, line 222: This returns exactly one row today, because the backend `automationEventAttributes` type has a single field (`automationEventExecutionDatetime`). The genuinely useful token list for `field_map.value` (about 17 tokens: `created_at`, `finished_at`, `current_phase`, `due_date`, `id`, `title`, `assignees`, `labels`, `created_by`, copy-from-field by numeric internal_id, and more). The docstring is honest that those are not returned here, but an agent calling this to discover value tokens gets one.

Consider documenting the full  set, or making the single-row scope more prominent so agents do not treat this as the complete catalog.

skills/portal-setup/pipefy-portal-setup/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+  portals (main hub, pages, page elements, sub-portals, publish/unpublish).
+  Covers 20 MCP tools on Interfaces + internal_api. Not for pipes/cards.
+tags: [pipefy, portal, interfaces, sub-portal, pages, elements]
+---
+
+# Portal setup
+
+Configure an organization's Pipefy portal: bootstrap the main hub, add pages and widgets, wire and publish sub-portals. **20 MCP tools** (Interfaces GraphQL + internal_api for sub-portal wiring).
+
+Deep reference: [`docs/mcp/tools/portal.md`](../../docs/mcp/tools/portal.md). Parity matrix: [`docs/parity.md`](../../docs/parity.md). Env vars: [`docs/config.md`](../../docs/config.md).
```
Danielmoraisg: Broken relative links. This file is at `skills/portal-setup/pipefy-portal-setup/`, three levels deep, so repo-root docs need `../../../`. All three links here resolve to `skills/docs/...` (missing): `../../docs/mcp/tools/portal.md`, `../../docs/parity.md`, `../../docs/config.md`. Change them to `../../../docs/...`.

skills/portal-setup/pipefy-portal-setup/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+- After layout/element edits on a disposable page, main portal pages used in production still open in the builder (no HTTP 500).
+
+---
+
+## Failure modes
+
+| Symptom | Likely cause | Recovery |
+|---------|----------------|----------|
+| `PERMISSION_DENIED` on writes | Wrong org, missing `manage_portals`, or SA not joined on interface | Same org as `list_portals`; user runs portal admin join; try human admin token |
+| Reads OK, writes fail on “admin” org | Token is human on org A, numeric id is org B | Align `organization_uuid` with token membership |
```
Danielmoraisg: Curly double quotes around "admin". Use straight quotes.

skills/portal-setup/pipefy-portal-setup/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+| `create_portal_element` opaque / 500 | Interfaces instability on some orgs | `duplicate_portal_element` from existing widget on same page |
+| Portal viewer HTTP 500 | Orphan `layout` children or wrong layout shape | Copy/fix layout from `get_portal`; delete disposable smoke page |
+| Nested `success: false` | API rejected mutation | Read `error.message`; do not assume top-level success |
+| Validation on element | Wrong metadata keys or partial update | Full metadata blob; `linkName` / `name` per type |
+
+---
+
+## See also
+
+- [`docs/mcp/tools/portal.md`](../../docs/mcp/tools/portal.md) — endpoints, wire naming, maintainer introspection
```
Danielmoraisg: Both See-also links are broken from this depth. Line 326 `../../docs/mcp/tools/portal.md` resolves to `skills/docs/...` (use `../../../docs/...`). Line 327 `../introspection/pipefy-introspection/SKILL.md` resolves to `skills/portal-setup/introspection/...` (use `../../introspection/...`, matching the sibling-skill convention in `skills/automations/pipefy-automations/SKILL.md:225`).

docs/mcp/tools/portal.md
```diff
@@ -186,18 +213,39 @@ Additional element types may appear; treat unknown keys as opaque JSON.
...
+
+---
+
+## Maintainers (TDD and introspection)
+
+When changing portal SDK/MCP/CLI behavior:
+
+1. **Spec:** `.cursor/dev-planning/specs/portal-crud/` (PRD, `introspection-snapshot.md`, `tasks/tasks-portal-crud.md`).
+2. **TDD loop:** SDK unit tests → MCP tool tests → CLI tests; `uv run pytest -m "not integration" -k portal`; update [`docs/parity.md`](../../parity.md) in the same PR.
+3. **Live schema:** [`.cursor/skills/pipefy-graphql-introspection/SKILL.md`](../../../.cursor/skills/pipefy-graphql-introspection/SKILL.md) — point introspection at Interfaces or internal_api before adding mutations:
```
Danielmoraisg: This maintainers block points at uncommitted local tooling and a removed env var:

- This line links `.cursor/skills/pipefy-graphql-introspection/SKILL.md`; `.cursor/skills/` is not in the repo and is not gitignored, so the link is dead. Lines 237 and 243 also reference `.cursor/` paths.
- Line 242 uses `PIPEFY_GRAPHQL_URL`, which the codebase already removed (see CHANGELOG "Removed": set `PIPEFY_BASE_URL`). Settings silently ignore unknown vars, so the snippet would not work as written. Use `PIPEFY_BASE_URL`.

Decide whether the `.cursor/` tooling is meant to ship; if not, describe the introspection step without dead links.

docs/config.md
adriannoes: Already addressed in b6955ce — removed the portal endpoint note from the `PIPEFY_BASE_URL` row; portal endpoints stay in `portal.md` only.

packages/mcp/tests/test_server.py
adriannoes: Good catch — reverted in 9ace4f4. The mock was not needed; `test_register_tools` passes with only the settings patch (same as `dev`). It was added preemptively when wiring the new tool and should not have shipped.

packages/sdk/src/pipefy_sdk/automation_preflight.py
adriannoes: Agreed — dropped the lazy `get_phase_fields` branch in 9ace4f4. Preflight now relies solely on embedded phase fields from `get_pipe` (via `GET_PIPE_QUERY`). Removed `test_validate_field_map_lazy_loads_phase_fields` since it only passed with an unrealistic `fields: None` shape.

packages/sdk/src/pipefy_sdk/automation_preflight.py
adriannoes: Both addressed in 9ace4f4:

1. Extracted pure `find_non_numeric_field_map_field_id` and `find_invalid_field_map_field_id` with unit tests (no mocks).
2. Slug-format check runs before `get_pipe`; `test_validate_field_map_raises_for_slug_field_id` now asserts `get_pipe.assert_not_called()`.

packages/mcp/src/pipefy_mcp/tools/automation_tools.py
adriannoes: Addressed in 9ace4f4: tool docstring now leads with **"Not the full token list"** and names the single-row backend limit. Expanded the Common value tokens table in `docs/mcp/tools/automations-and-ai.md` (id, title, created_at, due_date, finished_at, current_phase, assignees, labels, created_by, copy-from-field, etc.) and cross-linked from the tools table.

skills/portal-setup/pipefy-portal-setup/SKILL.md
adriannoes: Fixed in 9ace4f4 — all three header links now use `../../../docs/...`.

skills/portal-setup/pipefy-portal-setup/SKILL.md
adriannoes: Fixed in 9ace4f4 — replaced curly quotes with backticks around `admin`.

skills/portal-setup/pipefy-portal-setup/SKILL.md
adriannoes: Fixed in 9ace4f4 — See also links now use `../../../docs/mcp/tools/portal.md` and `../../introspection/pipefy-introspection/SKILL.md` (sibling-skill convention).

docs/mcp/tools/portal.md
adriannoes: Fixed in 9ace4f4: removed `.cursor/skills/` dead links; maintainers section now points at `skills/introspection/pipefy-introspection/SKILL.md` and `gql-cli` examples using `$PIPEFY_BASE_URL/graphql/interfaces` and `$PIPEFY_BASE_URL/internal_api` (per `docs/config.md`).

Issue #230, opened 2026-05-25
skills: standardize install via skills.sh; drop CLI wheel bundling


gbrlcustodio:
## Outcome

Skills install via [`skills.sh`](https://github.com/vercel-labs/skills) (`npx skills add`), targeting 55+ agents (Claude Code, Cursor, Codex, OpenCode, …). The CLI wheel no longer carries Markdown; `sync_starter_pack.py` and its drift-check are removed.

```sh
npx skills add gbrlcustodio/pipefy-mcp-server                       # all skills
npx skills add gbrlcustodio/pipefy-mcp-server --skill pipes-and-cards
npx skills add gbrlcustodio/pipefy-mcp-server -g -a claude-code -y  # CI-friendly
```

## Scope

**Repo layout**

- `skills/<domain>/<skill>/SKILL.md` — already matches `skills.sh` expectations. Twelve `SKILL.md` files confirmed. No structural changes.

**Removals**

- `packages/cli/src/pipefy_cli/skills/` — delete the bundled `.md` copies and the `__init__.py`.
- `packages/cli/pyproject.toml` — drop the `[tool.hatch.build.targets.wheel].include = [\"src/pipefy_cli/skills/*.md\"]` line.
- `scripts/sync_starter_pack.py` — delete.
- CI step (if any) that runs `python scripts/sync_starter_pack.py --check` — remove.
- `packages/cli/src/pipefy_cli/commands/skills.py` (verify existence during the PR) — remove `pipefy skills list` / `pipefy skills show` commands and their tests; document under a "Removed" section in `CHANGELOG.md`.

**Audits**

- Spot-check `name` and `description` frontmatter on every `skills/**/SKILL.md` — `skills.sh` reads these for `--list` output.

**Docs**

- `skills/README.md` — replace the "Starter pack (shipped inside `pipefy-cli`)" block with `npx skills add` instructions.
- `README.md` "Agent skills" section — same swap.
- `docs/setup.md` — point at `npx skills add` instead of `pipefy skills show`.

**Future (out of scope here)**

- Submit the repo to `skills.sh`'s public catalog.

## Acceptance criteria

- [ ] `uv build` of the CLI package produces a wheel; `unzip -l dist/pipefy_cli-*.whl | grep skills/` returns empty.
- [ ] `npx skills add gbrlcustodio/pipefy-mcp-server --list` enumerates all twelve skills.
[...]

PR #264, opened 2026-06-01
ci: restore CLI wheel smoke on PR CI

### LINKED ISSUE #242: ci: restore CLI wheel smoke after removing pipefy skills list
## Why

The recent #230 cleanup deleted the only CI step that installed the `pipefy-cli` wheel and invoked the `pipefy` console script (`uv run pipefy skills list` in `skills-lint.yml`). The replacement coverage everyone has in mind — `test_cli_help_golden.py` plus the rest of the CLI test suite — imports `pipefy_cli.main` from the src layout (`pythonpath = [\"src\", ...]` in `packages/cli/pyproject.toml`). Src-layout imports bypass entry-point resolution and wheel metadata, so a class of packaging regressions is now invisible to PR-level CI:

- `[project.scripts] pipefy = pipefy_cli.main:app` typo'd or pointing at a renamed symbol.
- Runtime dep missing from `pyproject.toml` that the editable dev env happens to provide transitively.
- Module that the hatch wheel target silently excludes.
- Import-time side effect that breaks only when Python resolves the entry point against the installed dist-info.

The only surviving `pipefy --help` smoke is in `release.yml`, double-gated by `on.push.tags: v*` and `if: startsWith(github.ref_name, 'v1.') && !contains(github.ref_name, '-')`. The current `v0.2.0-beta.*` line never fires it. A packaging regression introduced today on `dev` would not be observed until the first `v1.*` stable cut — likely months out — or by an end user running `uvx --from git+... pipefy-cli` and hitting the failure cold.

This is a regression-detection gap, not an active bug. It matters because the project bills the CLI as a primary user surface (`pipefy-cli` workspace member, plugin slash command, install docs), so the cost of shipping a broken `pipefy` entry point to a user is meaningfully higher than the cost of a 5-second extra CI step.

## Acceptance

A regression that breaks `[project.scripts] pipefy` or removes a runtime dep the CLI needs at import time is caught by PR-level CI on `dev`, not at release-tag time.

Open to either implementation route (ungating `release.yml`'s smoke vs adding a `pipefy --version` step to `ci.yml`); the deciding factor is which path adds the smallest amount of CI surface for the coverage gained.

adriannoes:
## Summary
- Restore CLI packaging smoke in the PR `test` CI job.
- Build all workspace wheels (`uv build --all-packages --wheel`) and install CLI + deps in a clean venv.
- Invoke `pipefy --help` via the console script entry point (not src layout).

Closes #242. Removing `pipefy skills list` in #230 dropped the only PR-level entry-point check; `release.yml` smoke only runs on `v1.*` tags.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,13 +93,10 @@ jobs:
     - name: Lint infra (banned upward imports)
       run: cd packages/infra && uv run ruff check src
     - name: Build workspace wheels (packaging smoke)
-      run: |
-        mkdir -p dist-packaging-smoke
-        uv build --all-packages --wheel -o dist-packaging-smoke
+      run: uv build --all-packages --wheel -o dist-packaging-smoke
     - name: Smoke install CLI wheel (clean venv)
       run: |
         python -m venv .venv-cli-smoke
-        .venv-cli-smoke/bin/python -m pip install --upgrade pip
         .venv-cli-smoke/bin/pip install dist-packaging-smoke/pipefy_infra-*.whl \
           dist-packaging-smoke/pipefy_sdk-*.whl \
           dist-packaging-smoke/pipefy_auth-*.whl \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,18 @@ jobs:
       run: cd packages/auth && uv run ruff check src
     - name: Lint infra (banned upward imports)
       run: cd packages/infra && uv run ruff check src
-    - name: Build MCP wheel (packaging smoke)
-      run: uv build packages/mcp -o packages/mcp/dist
+    - name: Build workspace wheels (packaging smoke)
+      run: |
+        mkdir -p dist-packaging-smoke
+        uv build --all-packages --wheel -o dist-packaging-smoke
+    - name: Smoke install CLI wheel (clean venv)
+      run: |
+        python -m venv .venv-cli-smoke
+        .venv-cli-smoke/bin/python -m pip install --upgrade pip
+        .venv-cli-smoke/bin/pip install dist-packaging-smoke/pipefy_infra-*.whl \
+          dist-packaging-smoke/pipefy_sdk-*.whl \
+          dist-packaging-smoke/pipefy_auth-*.whl \
+          dist-packaging-smoke/pipefy_cli-*.whl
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -92,7 +92,18 @@ jobs:
       run: cd packages/auth && uv run ruff check src
     - name: Lint infra (banned upward imports)
       run: cd packages/infra && uv run ruff check src
-    - name: Build MCP wheel (packaging smoke)
-      run: uv build packages/mcp -o packages/mcp/dist
+    - name: Build workspace wheels (packaging smoke)
+      run: |
+        mkdir -p dist-packaging-smoke
```
gbrlcustodio: **suggestion (non-blocking):** drop this line; `uv build -o <dir>` creates the directory if absent.

Collapses the step to a one-liner: `run: uv build --all-packages --wheel -o dist-packaging-smoke`.

.github/workflows/ci.yml
```diff
@@ -92,7 +92,18 @@ jobs:
...
+        uv build --all-packages --wheel -o dist-packaging-smoke
+    - name: Smoke install CLI wheel (clean venv)
+      run: |
+        python -m venv .venv-cli-smoke
+        .venv-cli-smoke/bin/python -m pip install --upgrade pip
```
gbrlcustodio: **suggestion (non-blocking):** drop this line.

`python -m venv` on the repo's pinned Python ships a pip recent enough for these wheels, so the upgrade pays a PyPI round-trip per CI run for no observable behavior change. It also widens the failure surface: a transient PyPI outage flakes a step whose purpose is to validate local wheels.

`release.yml` has the same line; symmetry is the strongest argument for keeping it.

gbrlcustodio (approved):
LGTM. Two non-blocking nits inline.

.github/workflows/ci.yml
adriannoes: Done in 0b658e9 — step is now a one-liner: `uv build --all-packages --wheel -o dist-packaging-smoke`.

.github/workflows/ci.yml
adriannoes: Done in 0b658e9 — dropped the pip upgrade here. Left `release.yml` unchanged for symmetry as you noted.

PR #267, opened 2026-06-02
fix(install): prefer system python3 on macOS for uv tool venvs

gbrlcustodio:
## Summary

`uv tool install` defaults to python-build-standalone (PBS), whose macOS binaries lack the `Security.framework` entitlements `keyring.set_password` needs. The result a fresh installer hit today: `pipefy auth login` completes the OAuth dance against the IdP, then the keychain write fails with `PasswordSetError: Can't store password on keychain: (-25244, 'Unknown Error')` (`errSecMissingEntitlement`), and macOS prompts for keychain access several times before giving up.

On Darwin, probe `python3.{14,13,12,11,3}` on PATH (skipping uv-managed PBS paths), verify the candidate is >= 3.11, and pass `--python <path>` to `uv tool install`. Linux keeps the existing default since the PBS keychain bug is macOS-specific. A pre-set `UV_PYTHON` short-circuits the probe. If no system Python is found, the script warns and points at `PIPEFY_KEYCHAIN_BACKEND=file` as the runtime fallback.

## Test plan

- [x] Dry-run on Darwin with Homebrew python3 on PATH prints `Using system Python for tool venvs: /opt/homebrew/bin/python3.14` and the planned `uv tool install` command includes `--python /opt/homebrew/bin/python3.14`.
- [x] Dry-run on Darwin with `UV_PYTHON=/some/path` skips the probe and omits `--python` from the `uv tool install` command.
- [x] Real install on a fresh Mac (no prior pipefy keychain entries) completes login without prompts and writes to `pipefy` in Keychain Access.
- [ ] On Linux (CI runner), behavior unchanged: no probe, no `--python` override.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -185,21 +185,25 @@ pick_system_python() {
         ?*) uv_python_is_spec=1 ;;
     esac
 
-    # Under `curl | sh` the inherited PATH is often just /usr/bin and friends,
-    # so Homebrew's python3 at /opt/homebrew/bin (or /usr/local/bin on Intel)
-    # never gets probed and we fall through to PBS. Prepend the standard
-    # Homebrew prefixes on Darwin when they exist so the loop can find them.
-    for brew_dir in /opt/homebrew/bin /usr/local/bin; do
-        if [ -d "$brew_dir" ]; then
-            case ":$PATH:" in
-                *":$brew_dir:"*) ;;
-                *) PATH="$brew_dir:$PATH" ;;
-            esac
-        fi
+    # Build a probe-local PATH that includes Homebrew's standard prefixes. A
+    # `curl | sh` run from a non-interactive shell (CI, cron, freshly-spawned
+    # subshell whose rc hasn't loaded Homebrew's shellenv) can inherit a PATH
+    # of just /usr/bin and friends; without this, brew's python3 is invisible
+    # and the loop falls through to PBS. The augmented PATH is scoped to the
+    # probe so the rest of main() (uv, curl, python3, npx, pipefy lookups)
+    # sees the unmodified PATH.
+    # /usr/local/bin first, then /opt/homebrew/bin, so the Apple-Silicon-native
+    # prefix lands at the front; on Intel /opt/homebrew/bin typically doesn't
+    # exist and is skipped.
+    probe_path="$PATH"
+    for brew_dir in /usr/local/bin /opt/homebrew/bin; do
+        [ -d "$brew_dir" ] && probe_path="$brew_dir:$probe_path"
     done
 
+    keychain_hint="if 'pipefy auth login' later fails with keychain error -25244, set PIPEFY_KEYCHAIN_BACKEND=file or install Homebrew python3."
+
     for cmd in python3.14 python3.13 python3.12 python3.11 python3; do
-        path=$(command -v "$cmd" 2>/dev/null) || continue
+        path=$(PATH="$probe_path"; command -v "$cmd" 2>/dev/null) || continue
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -168,6 +169,38 @@ detect_uv() {
...
+    # `Security.framework` requires for keychain writes, so `pipefy auth login`
+    # later fails with `(-25244, 'Unknown Error')` (errSecMissingEntitlement).
+    # On Linux, uv's default Python is fine.
+    [ "$OS" = "Darwin" ] || return 0
+    # Honor UV_PYTHON only when it points at an absolute path (a user-pinned
+    # interpreter). A version spec like `UV_PYTHON=3.13` leaves uv free to
+    # resolve to PBS, which is the failure case this function exists to avoid.
+    case "${UV_PYTHON:-}" in /*) return 0 ;; esac
+
+    for cmd in python3.14 python3.13 python3.12 python3.11 python3; do
```
adriannoes, line 205: This probe only sees the current `PATH`. For `curl | sh`, that is often just `/usr/bin` and friends, so Homebrew under `/opt/homebrew/bin` never gets tried and we still fall through to PBS (same `-25244` path). Consider prepending `/opt/homebrew/bin` and `/usr/local/bin` on Darwin when those dirs exist, before this loop.

install.sh
```diff
@@ -168,6 +169,38 @@ detect_uv() {
...
+    # On macOS, prefer a system/Homebrew python3 over uv-managed
+    # python-build-standalone (PBS). PBS binaries lack the entitlements that
+    # `Security.framework` requires for keychain writes, so `pipefy auth login`
+    # later fails with `(-25244, 'Unknown Error')` (errSecMissingEntitlement).
+    # On Linux, uv's default Python is fine.
+    [ "$OS" = "Darwin" ] || return 0
+    # Honor UV_PYTHON only when it points at an absolute path (a user-pinned
+    # interpreter). A version spec like `UV_PYTHON=3.13` leaves uv free to
+    # resolve to PBS, which is the failure case this function exists to avoid.
+    case "${UV_PYTHON:-}" in /*) return 0 ;; esac
```
adriannoes: If `UV_PYTHON` is a version spec (e.g. `3.13`) and the probe finds no non-PBS interpreter on `PATH`, install still runs without `--python` and `uv` can resolve the spec to PBS. A targeted warn when `UV_PYTHON` is set but not `/*` and the loop exhausts would close that gap.

adriannoes (approved):
## Summary

Approved. The `realpath(sys.executable)` PBS filter and narrowing the `UV_PYTHON` early-return to absolute paths fix the shim bug that would have defeated `--python`.

## Also noted (inline)

- **F1:** Minimal `PATH` on `curl | sh` may skip Homebrew — prepending common macOS bin dirs would widen coverage.
- **F2:** `UV_PYTHON` version spec + empty probe can still land on PBS — worth an extra warn in that case.

Neither blocks merge.

## Verification

```bash
DRY_RUN=1 ./install.sh --dry-run   # Darwin: expect --python when Homebrew on PATH
```

install.sh
gbrlcustodio: Fixed in 7d0e979: prepend `/opt/homebrew/bin` and `/usr/local/bin` (when present) on Darwin before the probe loop.

install.sh
gbrlcustodio: Fixed in 7d0e979: track when `UV_PYTHON` is a non-absolute spec and emit a targeted warning if the probe exhausts without finding a non-PBS interpreter.

PR #268, opened 2026-06-02
chore(release): refresh uv.lock from bump_version.py and add pyproject lockstep

gbrlcustodio:
## Summary

Closes the release-time gap that broke CI today. `chore: release v0.2.0-beta.2.dev1` rewrote the five `__init__` files and root `pyproject.toml` but left `pipefy-workspace` in `uv.lock` pinned to the old version. The `test` job on dev caught it (14s failure on the release commit itself), but the release engineer didn't notice, and every subsequent PR off dev then hit `uv sync --locked` failures.

Two layered guards:

1. **`scripts/bump_version.py` now refreshes `uv.lock`.** After rewriting the version strings it shells out to `uv lock`. The release commit can't be authored without a fresh lockfile in the same change.

2. **CI lockstep check now includes root `pyproject.toml`.** `uv sync --locked` already catches lockfile drift downstream, but the lockstep step gives a clearer `version mismatch: sdk=X, pyproject=Y` error than uv's `lockfile needs to be updated` and short-circuits partial bumps before reaching the sync step.

## Test plan

- [x] `uv run python scripts/bump_version.py version=0.2.0-beta.2.dev1` (no-op bump) emits `Refreshing uv.lock...` and leaves the lockfile unchanged.
- [x] The extended lockstep snippet runs locally against the current tree and reports `OK: all in lockstep`.
- [x] CI on this PR passes both `lint` and `test`.
- [ ] Next real version bump (whenever beta.2 lands) goes through this script and the resulting commit includes the lockfile delta.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,38 +58,7 @@ jobs:
     - name: Install dependencies
       run: uv sync --locked --all-extras --dev
     - name: Verify package versions are in lockstep
-      run: |
-        uv run python <<'PY'
-        import re
-        from pathlib import Path
-
-        init_paths = {
-            "sdk": Path("packages/sdk/src/pipefy_sdk/__init__.py"),
-            "mcp": Path("packages/mcp/src/pipefy_mcp/__init__.py"),
-            "cli": Path("packages/cli/src/pipefy_cli/__init__.py"),
-            "auth": Path("packages/auth/src/pipefy_auth/__init__.py"),
-            "infra": Path("packages/infra/src/pipefy_infra/__init__.py"),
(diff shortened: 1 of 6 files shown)


adriannoes (approved):
This closes the release-time footgun cleanly: the bumper now runs `uv lock` after rewriting versions, and CI delegates lockstep checks to `scripts/bump_version.py verify` so the writer and reader cannot drift apart.

The second commit is a solid hardening pass — success output after `uv lock`, `[project]`-anchored regex for writes, `tomllib` for reads, and RELEASE.md aligned with actual behavior.

LGTM.

PR #260, opened 2026-06-01
feat(plugin): install via release wheel URLs

gbrlcustodio:
## Summary

- Switches `.mcp.json` and `commands/install.md` from `uvx --from git+...@latest#subdirectory=...` to direct GitHub Release wheel URLs. Eliminates the full git clone + setuptools build pass that every plugin install or `--refresh` paid per workspace member (sdk, auth, infra, mcp).
- Adds a step to `.github/workflows/release.yml` that, after creating the GitHub Release, patches `.mcp.json`, `commands/install.md`, `.claude-plugin/plugin.json`, and `.claude-plugin/marketplace.json` on `main` to point at the just-cut tag's wheels, and commits the result.
- Updates README MCP-client wiring blocks (Cursor, Claude Desktop, Codex) to show the same wheel-URL shape with `<TAG>` and `<VERSION>` placeholders. Curl installer (#231) is the path that resolves these automatically once it ships.

## Design notes

- **PEP 491 compliance.** Wheel filenames stay in canonical `pipefy_<pkg>-<version>-py3-none-any.whl` form. No renamed-stable copies. URLs are version-pinned, and the workflow rewrites them per release.
- **`pipefy-infra` is an explicit `--with`.** uvx will not auto-resolve a workspace-only transitive dep from PyPI pre-v1.
- **README intentionally not patched by the workflow.** Three reasons: dev/main divergence would create per-release merge conflicts, copy-pasting from non-`main` branches would surface stale URLs, and the README install blocks are documentation rather than a programmatic install path. Placeholders + the curl installer (#231) cover the user-facing path.
- **Branch protection on `main` is currently off** (`gh api repos/.../branches/main/protection` → 404), so the workflow can push directly. If branch protection lands later, swap the push for a `gh pr create` fallback.

## Test plan

- [x] `python3 -c "ver='0.2.0-beta.2'; print(ver.replace('-beta.', 'b').replace('-rc.', 'rc').replace('-alpha.', 'a'))"` returns `0.2.0b2`. Same derivation handles `-rc.`, `-alpha.`, plain `1.0.0`.
- [x] `.mcp.json` parses as valid JSON.
- [x] `release.yml` parses as valid YAML.
- [x] `uv run pytest` passes (2916 passed, 46 skipped).
- [ ] After this lands on `dev` and `v0.2.0-beta.2` ships, smoke `claude plugin install pipefy@pipefy` and `/pipefy:install` — both should resolve via `releases/download/...` URLs with no git clone.

## Out of scope

[...]

PR #270, opened 2026-06-02
feat(install): register Claude Code MCP server at user scope

gbrlcustodio:
## Summary

Stacked on #260. Removes the double materialization of the Pipefy MCP server on the Claude Code path and makes the installer the single owner of the Claude Code registration.

Previously `/pipefy:install` installed `pipefy_mcp_server` as a uv tool that the plugin's bundled `.mcp.json` (uvx) never referenced, so a plugin user who ran it ended up with two copies (the uvx-cached env plus an unused tool venv).

## What changed

- **`install.sh`**: the `--client claude-code` branch now registers the server via `claude mcp add pipefy --scope user -- pipefy-mcp-server` (idempotent: removes any prior user-scope entry first). By Claude Code's name precedence (`local > project > user > plugin`) the user-scope entry shadows the plugin's bundled entry, so only the installed binary spawns, on the same system Python the CLI is pinned to. `require_claude` errors cleanly if the `claude` CLI is absent.
- **`commands/install.md`**: `/pipefy:install` switches from `--client none` to `--client claude-code`, and the prose explains the override + required reload.
- **`.claude-plugin/plugin.json` + `hooks/check-server-version.sh`**: a `SessionStart` hook that nudges the user to re-run `/pipefy:install` when the installed server version drifts from the plugin's. No-ops for users on the pure plugin/uvx path (no installed binary).

## Design notes

- **The committed `.mcp.json` (uvx) is unchanged** and remains the zero-config default, so `/plugin install pipefy@pipefy` still works with no prior step. The override only kicks in once a user runs `/pipefy:install`.
- **Why user scope, not editing the plugin**: there's no supported way to disable just one plugin's MCP server; name-shadowing at a higher scope is the only lever, and it's deterministic per the documented precedence.
- **Why the drift hook**: shadowing means plugin auto-updates stop reaching the running server. The hook is the accepted mitigation for that staleness. It reads the plugin version (cheap sed on `plugin.json`) before spawning `pipefy-mcp-server --version`, so the Python cold start only happens for users who actually have the override installed.
- **The server only reads the keychain** (`load_session` -> `keyring.get_password`); the documented macOS `-25244` issue is a non-deterministic write during `pipefy auth login`. So this is about consistency with the CLI's system-Python install, not a known server-side failure.

## Test plan

- [x] `sh -n install.sh` and `sh -n hooks/check-server-version.sh`
- [x] `.claude-plugin/plugin.json` parses as valid JSON
- [x] `install.sh --dry-run --client claude-code` previews the `claude mcp add` registration
- [x] Hook verified in all three states: no binary (silent), version drift (nudge), version match (silent)
- [x] `uv run pytest` (2955 passed, 46 skipped)
- [ ] On a machine with Claude Code: after `/pipefy:install` + reload, confirm `claude mcp list` shows the user-scope `pipefy` and the plugin's uvx entry is suppressed

## Release sequencing

Targets #260's branch to keep the stack; retarget to `main` once #260 merges.


hooks/check-server-version.sh
```diff
@@ -0,0 +1,29 @@
...
+manifest="${CLAUDE_PLUGIN_ROOT:-}/.claude-plugin/plugin.json"
+[ -f "$manifest" ] || exit 0
+plugin=$(sed -n 's/.*"version"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' "$manifest" | head -n1)
+[ -n "$plugin" ] || exit 0
+
+installed=$(pipefy-mcp-server --version 2>/dev/null | tr -d '[:space:]') || exit 0
+[ -n "$installed" ] || exit 0
+
+if [ "$installed" != "$plugin" ]; then
+    echo "Pipefy: the installed MCP server ($installed) differs from this plugin ($plugin); re-run /pipefy:install to sync."
```
adriannoes, line 27: The SessionStart hook compares the installed server to `plugin.json`'s `"version"`, but release bumps don't update that file today (`bump_version.py verify` never checks it). On this branch the manifest is still `0.2.0-beta.1` while wheels are `0.2.0-beta.2`, so a user who just ran `/pipefy:install` from the latest release will get a drift nudge on every session even though they're on the release they installed.

I'd either add `.claude-plugin/plugin.json` to the same lockstep bump/verify path as the Python packages, or change what the hook compares so we only nudge when the running server is actually older than what the plugin expects—not when the manifest lagged a release cut.

adriannoes (commented):
## Summary

Reviewed the install/plugin diff locally (`sh -n`, `install.sh --dry-run --client claude-code`). The user-scope registration cleanly fixes double materialization (uv tool + plugin uvx) and the hook’s “cheap checks first” ordering is sensible. One version-sync gap is worth addressing before this ships broadly (inline on the hook).

## Also noted

- **F2 — install ordering:** `main()` installs both uv tools before `require_claude`, so a missing `claude` CLI leaves `pipefy` / `pipefy-mcp-server` on disk but exits before `claude mcp add`. Since `/pipefy:install` now passes `--client claude-code`, consider checking for `claude` before `install_tool`, or finishing with a clear warning instead of `err` after wheels are already installed.

hooks/check-server-version.sh
gbrlcustodio: Fixed in 8965e24: added `.claude-plugin/plugin.json` to the lockstep write and verify paths in `bump_version.py`, and synced the manifest to `0.2.0-beta.2`. The hook now compares against a manifest that tracks the release, so it only nudges when the installed binary actually lags. Went with the lockstep route over reworking the hook so the manifest stops drifting at the source.

gbrlcustodio:
Closing this; one piece is salvaged and the rest is deferred rather than dropped.

**Salvaged into #367:** the `bump_version.py` change that tracks `.claude-plugin/plugin.json`'s version in lockstep. That addressed a real, live drift (the manifest was stuck at `0.2.0-beta.1` while releases shipped `0.3.0-alpha.1`); it now moves with every bump and is asserted by `verify`. Rewritten for current `dev` and reconciled with the sibling-pinning work in the same file.

**Deferred (a design decision, not stale):** the installer's user-scope registration (`claude mcp add pipefy --scope user -- pipefy-mcp-server`) and the SessionStart drift hook. The double-materialization it targets on the Claude Code path still exists, so this is worth revisiting, but it assumed the git-subdirectory `.mcp.json` that #368 is changing to a PyPI install, so it's better re-evaluated on top of that. The drift hook also depends on the plugin.json lockstep, which is now in place via #367, so that groundwork exists if we pursue it.

This PR was stacked on #260, which is also being closed as superseded (PyPI publishing landed in #365).

Issue #273, opened 2026-06-03
Refactor InternalApiClient to accept a gql DocumentNode like other endpoints


gbrlcustodio:
## Summary

`InternalApiClient.execute_query` is the only GraphQL client in the SDK that accepts a raw query **string** and parses it with `gql()` internally. Every other endpoint accepts an already-parsed `gql()` `DocumentNode`, including the portal interfaces client, which is a plain `BasePipefyClient`. As a result, internal-API query modules store mutations as plain strings while every other query module stores `gql(...)` constants. We should make `InternalApiClient` accept a `DocumentNode` like the others and wrap the internal mutation constants in `gql()` at module load, removing the divergent signature and storage style.

## Current state

- `InternalApiClient.execute_query` overrides the base signature to take `str` and calls `parse_gql(query)` before delegating to `BasePipefyClient.execute_query` (`packages/sdk/src/pipefy_sdk/services/internal_api_client.py:68-71`, with a `# type: ignore[override]`).
- Internal mutation constants are plain strings:
  - `INTERNAL_DELETE_CARD_RELATION_MUTATION` (`packages/sdk/src/pipefy_sdk/queries/card_queries.py`)
  - portal sub-portal mutations (`packages/sdk/src/pipefy_sdk/queries/portal_internal_queries.py`)
- Call sites pass those strings: `PipefyClient.delete_card_relation` (`packages/sdk/src/pipefy_sdk/client.py:948`) and `PortalService` (`packages/sdk/src/pipefy_sdk/services/portal_service.py:276`).
- Everything else passes `gql(...)` `DocumentNode` constants, including the portal interfaces client, which is a plain `BasePipefyClient` pointed at `interfaces_graphql_url` (`portal_service.py:236-258`).

## Why

- Removes a special-cased signature override (`# type: ignore[override]`) that exists only to accept `str`.
- Gives one consistent query-definition style across the SDK (all `gql(...)` constants), so syntax errors surface at import time and tooling treats internal and public queries uniformly.
- The historical rationale in the docstrings ("sends raw GraphQL text via JSON POST", "the endpoint does not support schema validation that `gql()` provides") was inaccurate: the client already parses with `gql()` and uses the same `HTTPXAsyncTransport` as every other client. Those docstrings were recently corrected, but the underlying inconsistency remains.

## Proposed change

[...]

PR #274, opened 2026-06-03
fix(272): route AI automation create/update through public createAutomation

### LINKED ISSUE #272: AI automation create requires service-account credentials, but generate_with_ai is a public automation action
## Summary

Creating an AI automation (`generate_with_ai`) requires service-account credentials and is routed through the `/internal_api` endpoint. It should be creatable with a normal user-authenticated session, because `generate_with_ai` is a standard automation action exposed by the public automation API.

The CLI surfaces this as a hard stop:

```
AI Automation requires service-account credentials
(PIPEFY_SERVICE_ACCOUNT_CLIENT_ID, PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET).
Bearer --token mode does not attach the internal API client.
```

## Why this looks wrong

`generate_with_ai` is a first-class action in the public automation catalog, and the mutation used to create an AI automation is the same `createAutomation` mutation used for traditional automations.

1. `generate_with_ai` is returned by the public `get_automation_actions`:

   ```
   pipefy automation actions list --pipe <id>
   # -> [... {"id": "generate_with_ai", "enabled": true, "acceptedParameters": ["ai_params"], ...} ...]
   ```

2. The AI-automation create mutation is `createAutomation` with `action_id` and `action_params`, identical in shape to the public traditional-automation mutation:

   - `packages/sdk/src/pipefy_sdk/queries/ai_automation_queries.py:11` (`AI_CREATE_AUTOMATION_MUTATION`): `createAutomation(input: { name, action_id, event_id, action_repo_id, event_repo_id, action_params, event_params, condition })`
   - `packages/sdk/src/pipefy_sdk/queries/automation_queries.py` (public create): same `createAutomation`, same `action_id` / `action_params` fields.

So the only thing forcing service-account credentials is that the toolkit sends this mutation through `InternalApiClient` (the `/internal_api` endpoint) rather than the public `/graphql` path that the user session is already authenticated for.

## Where the requirement is enforced

- CLI gate: `packages/cli/src/pipefy_cli/commands/ai_automation.py:34-43` (`_require_ai_automation` raises when `client.ai_automation_available` is false, with the message above).
- Availability flag: `packages/sdk/src/pipefy_sdk/client.py:133-135` (`ai_automation_available` is true only when `_ai_automation_service` is set).
- Wiring: the AI-automation service is constructed on top of `InternalApiClient` (`/internal_api`):
  - `packages/mcp/src/pipefy_mcp/core/container.py:86-95`
  - `packages/sdk/src/pipefy_sdk/queries/ai_automation_queries.py:1-7` (docstring: mutations target the internal_api endpoint).

## Reproduce

```
# Authenticated as a normal user via `pipefy auth login` (no service account configured)
pipefy ai-automation validate-prompt --pipe <id> \
  --prompt "Classify %{<field_internal_id>} and write the result." \
  --field-ids '["<output_field_internal_id>"]'
# -> succeeds: {"success": true, "valid": true, ...}

pipefy ai-automation create ...
# -> blocked: "AI Automation requires service-account credentials ..."
```

Prompt validation works under a user session, but create does not, even though both ultimately go through the same `createAutomation` mutation.

## Expected

A user-authenticated session (stored session from `pipefy auth login`, or a bearer `--token`) can create and update `generate_with_ai` automations, the same way it can create traditional automations. Service-account credentials should not be required for this operation.

## Suggested fix

Route `create_ai_automation` / `update_ai_automation` through the public `create_automation` path with `action_id="generate_with_ai"` and `action_params.ai_params` (`{ value: <prompt>, fieldIds: [...] }`), using the session's normal auth, instead of the `/internal_api` client. If the internal endpoint is genuinely required for some sub-case, gate only that sub-case rather than the whole operation, and make the public path the default.

## Environment

- pipefy CLI / MCP version: `0.2.0-beta.2.dev1`
- Auth: stored session (`pipefy auth login`), no service account configured.

## Note

Surfaced while building a downstream plugin that provisions an AI triage automation as part of an install flow. Requiring service-account credentials makes an otherwise user-installable flow depend on separate internal credentials, which is the friction this issue is about.

gbrlcustodio:
Closes #272.

## Summary

`generate_with_ai` create/update were routed through the `/internal_api` endpoint (via `AiAutomationService`) and gated behind a "requires service-account credentials" check. A normal user session (stored session from `pipefy auth login`, or a bearer `--token`) could not create or update an AI automation, which broke downstream install flows that provision an AI triage automation.

The public `createAutomation` mutation already accepts `action_id="generate_with_ai"` with `action_params.aiParams` under the caller's normal auth (verified against the live API). This reroutes create/update onto the public path and removes the internal-API plumbing and the false gate. Read, list, and delete already used the public path.

## Changes

- Add `AutomationService.create_ai_automation` / `update_ai_automation` that build the `aiParams` envelope and delegate to the public `createAutomation` / `updateAutomation`, mirroring the existing `create_send_task_automation` pattern. `PipefyClient` delegates here.
- Delete `AiAutomationService` and `ai_automation_queries.py`; drop the `ai_automation_available` property, `set_ai_automation_service`, and the CLI/MCP wiring. `InternalApiClient` stays (still used by `delete_card_relation`).
- Remove the service-account gate from the CLI and MCP create/update paths; the MCP create path now uses the same `handle_automation_tool_graphql_error` as the other automation tools.
- Fix the stale CLI help text and the automations doc.
- Correct the `internal_api` query-module docstrings (`card_queries.py`, `portal_internal_queries.py`): the client parses with `gql()` itself, it does not bypass schema validation. (Follow-up refactor tracked in #273.)

No behavior change beyond removing the credential requirement.

## Verification

- `uv run pytest`: SDK 873 passed, CLI 305 passed, MCP 1320 passed. `ruff check` / `ruff format --check` clean.
- Live, under a normal `pipefy auth login` session (no service account) against pipe `307184972`: create, update (rename), get (confirmed `action_id=generate_with_ai` with `aiParams` intact), and delete all succeeded, including a run through a locally-built, system-installed `pipefy` wheel.


adriannoes (approved):
LGTM. Routing `generate_with_ai` through public `createAutomation` / `updateAutomation` fixes the false service-account gate and matches the existing `create_send_task_automation` pattern. I ran the full unit suite locally (2498 passed) on the PR head.

Optional follow-up: a CHANGELOG note that `AiAutomationService` is no longer exported from `pipefy_sdk` for anyone importing it directly.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
```diff
@@ -441,8 +397,6 @@ async def update_ai_automation(
                 debug: When True, append internal_api code and correlation_id to update failures (after sanitizing the main message).
```
adriannoes: Same stale wording as create: update failures now use `handle_automation_tool_graphql_error` on the public GraphQL path, not the deleted internal_api sanitizer. Worth aligning this `debug:` line with what the shared handler actually surfaces.

packages/sdk/src/pipefy_sdk/services/automation_service.py
```diff
@@ -334,6 +393,95 @@ async def create_send_task_automation(
             **extra_input,
         )
 
+    async def create_ai_automation(
```
adriannoes, line 396: This file is now 603 lines after folding in the AI create/update path — a bit over our usual 500-line split point. Not blocking for this fix, but since the AI helpers just moved here from the deleted internal service, consider extracting `_ai_params_payload`, `_extract_automation_id`, and the two `*_ai_automation` methods into a sibling module in a follow-up so the core adapter stays grep-friendly.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
```diff
@@ -402,8 +354,12 @@ async def create_ai_automation(
             try:
                 result = await client.create_ai_automation(validated)
             except Exception as exc:  # noqa: BLE001
-                return _ai_automation_api_failure_payload(
-                    exc, fallback=AI_AUTOMATION_CREATE_FAILED, debug=debug
+                return await handle_automation_tool_graphql_error(
```
adriannoes, line 357: Heads-up on the `debug:` docstring a few lines above (still references internal_api codes/correlation_id). Create now routes failures through `handle_automation_tool_graphql_error` on the public GraphQL path, so that Args line is stale — same fix as the update tool's docstring.

adriannoes (commented):
## Second review pass

We took another look at this PR after the initial approval — still looks good to merge. The routing change is clean, tests are solid, and the false service-account gate is correctly removed.

**Optional follow-ups** (non-blocking):

- **Stale MCP docstrings:** `create_ai_automation` and `update_ai_automation` still document `debug=True` as appending internal_api codes; both paths now use `handle_automation_tool_graphql_error` on the public GraphQL route. Inline notes on the update docstring and create error handler.
- **File size:** `automation_service.py` is 603 lines after folding in the AI helpers — consider extracting to a sibling module in a follow-up.
- **CHANGELOG:** worth an Unreleased bullet that `AiAutomationService` is no longer exported from `pipefy_sdk` for anyone importing it directly.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
gbrlcustodio: Fixed in b912df9: docstring now reads "append GraphQL error codes and correlation_id to update failures", matching what the shared handler surfaces.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
gbrlcustodio: Fixed in b912df9: same alignment applied to the create docstring.

packages/sdk/src/pipefy_sdk/services/automation_service.py
gbrlcustodio: Tracked in #278.

PR #276, opened 2026-06-03
refactor(sdk): make InternalApiClient accept a gql DocumentNode

### LINKED ISSUE #273: Refactor InternalApiClient to accept a gql DocumentNode like other endpoints

gbrlcustodio:
Closes #273.

## Summary

`InternalApiClient.execute_query` was the only GraphQL client in the SDK that accepted a raw query **string** and parsed it with `gql()` internally, carrying a `# type: ignore[override]` that existed solely to accept `str`. Every other endpoint, including the portal interfaces client (a plain `BasePipefyClient`), already accepts an already-parsed `gql()` `DocumentNode`. This made the internal-API query modules store mutations as plain strings while every other module stored `gql(...)` constants.

This change makes `InternalApiClient` consistent with the rest of the SDK:

- Deletes the `execute_query` override. Once the `parse_gql()` call is dropped it is a pure pass-through to `BasePipefyClient.execute_query(query: Any, …)`, so inheriting it is equivalent and removes the `# type: ignore[override]` and the now-unused `parse_gql` / `typing.Any` imports.
- Wraps the internal mutation constants in `gql(...)` at module load: `INTERNAL_DELETE_CARD_RELATION_MUTATION` (`card_queries.py`) and the three sub-portal mutations (`portal_internal_queries.py`).
- Refreshes the docstrings/comments that described the old raw-string behavior (and the AI-automation rationale, which is removed by #274).

No behavior change: syntax errors now surface at import time, and tooling treats internal and public queries uniformly.

## Base branch

Opened against `fix/272-ai-automation-public-path` (PR #274). That PR deletes `ai_automation_queries.py` and `ai_automation_service.py`, removing the third internal-client caller that still passed plain strings, so this branch's scope is exactly the card-relation and portal sub-portal callers. Retarget to `dev` once #274 merges.

## Tests

- `test_pipefy_facade.py`: the three substring assertions now run against `print_ast(INTERNAL_DELETE_CARD_RELATION_MUTATION)` (the canonical pattern used across the test suite), preserving the intent of pinning the snake_case input keys.
- `test_internal_api_client.py`: its tests passed raw query strings to `execute_query`; they now wrap in `gql(...)`.

## Verification

- Import-time check: all four constants parse as `DocumentNode`.
- `uv run python -m pytest` (SDK): 873 passed, 17 skipped.
- `ruff check` and `ruff format --check`: clean.

### DIFF
--- a/packages/sdk/src/pipefy_sdk/queries/card_queries.py
+++ b/packages/sdk/src/pipefy_sdk/queries/card_queries.py
@@ -265,16 +265,15 @@ UPDATE_FIELDS_VALUES_MUTATION = gql(
 )
 
 # ---------------------------------------------------------------------------
-# Internal API mutations (plain strings, not gql())
+# Internal API mutations
 #
 # ``deleteCardRelation`` is only available on the internal GraphQL schema
-# (core_api / internal_v1), not the public API. These are plain strings because
-# ``InternalApiClient.execute_query`` takes a raw string and parses it with
-# ``gql()`` itself before sending, unlike the public client which expects an
-# already-parsed ``DocumentNode``. Same pattern as ``portal_internal_queries.py``.
+# (core_api / internal_v1), not the public API. It runs through
+# ``InternalApiClient``, which accepts a ``gql()`` ``DocumentNode`` like every
+# other client.
 # ---------------------------------------------------------------------------
 
-INTERNAL_DELETE_CARD_RELATION_MUTATION = """
+INTERNAL_DELETE_CARD_RELATION_MUTATION = gql("""
 mutation deleteCardRelation(
   $childId: ID!,
   $parentId: ID!,
@@ -288,7 +287,7 @@ mutation deleteCardRelation(
     success
   }
 }
-"""
+""")
 
 __all__ = [
     "CREATE_CARD_MUTATION",
--- a/packages/sdk/src/pipefy_sdk/queries/portal_internal_queries.py
+++ b/packages/sdk/src/pipefy_sdk/queries/portal_internal_queries.py
@@ -1,36 +1,36 @@
-"""GraphQL mutation strings for portal sub-portals (internal_api endpoint).
+"""GraphQL mutations for portal sub-portals (internal_api endpoint).
 
-Mutations are plain strings (not ``gql()``) because
-``InternalApiClient.execute_query`` takes a raw string and parses it with
-``gql()`` itself before sending, unlike the public client which expects an
-already-parsed ``DocumentNode``.
+These run through ``InternalApiClient``, which accepts a ``gql()``
+``DocumentNode`` like every other client.
 """
 
 from __future__ import annotations
 
-UPDATE_SUB_PORTAL_ELEMENT_MUTATION = """
+from gql import gql
+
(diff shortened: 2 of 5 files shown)


adriannoes (approved):
Clean, focused refactor — dropping the `InternalApiClient` string override and wrapping internal mutations in `gql()` brings the SDK in line with every other client without changing runtime behavior. I checked out the branch locally and the full SDK unit suite (873 tests) plus ruff are green.

PR #277, opened 2026-06-03
feat: card-phase ergonomics (inventory, create_card phase_id, review fixes)

adriannoes:
This work closes gaps identified while **agent-seeding the Analytics pipe** (`306996636`) during the *chaos-pipe* exercise. Agents had to fall back to **`execute_graphql`**, and sometimes **`introspect_type` / `search_schema`**, because first-class MCP tools did not cover common card/phase workflows—even though the GraphQL API and parts of the SDK already supported them.

### What agents did with raw GraphQL (and why)

| Need | Typical escape hatch | Gap before this PR |
|------|----------------------|-------------------|
| Create a card in an orphan / non–start-form phase | `createCard` + `CreateCardInput.phase_id` via `execute_graphql`; `introspect_type("CreateCardInput")` | `create_card` had no `phase_id`; post-create `update_card` for title |
| Inventory cards per phase | `phase { cards_count }`, `phase { cards(first) }` | No MCP `get_phase_cards_count` / `get_phase_cards`; `get_cards` has no phase filter |
| Move cards safely | `move_card_to_phase` when destination ∈ allowed set | No proactive `get_phase_allowed_move_targets` tool (SDK existed; not exposed) |
| Pipe structure for seeding | Multiple `get_pipe` + ad-hoc phase queries | Start form not visible as `start_form_phase`; no `cards_count` on workflow phases |
| Labels | — | `create_label("red")` failed; API expects hex (`#RGB` / `#RRGGBB`) |
| Reports filtered by phase | Raw filter attempts | `ReportCardsFilter` shape easy to get wrong (top-level `current_phase` rejected by API) |

### What already worked without GraphQL

`create_phase`, field/condition setup, automations/AI agents (inactive), `move_card_to_phase` when the graph allowed it, `get_pipe` / `get_phase_fields`, and `create_card` on the start form with `skip_elicitation`.

### Goal

Let agents **seed, audit, and move cards on the happy path** using dedicated SDK + MCP + CLI surfaces (parity in `docs/parity.md`), without raw GraphQL for these flows. Phase transition **graph editing** stays UI-only (documented).

---

## Summary

- **SDK:** `get_pipe` inventory enrichment (`start_form_phase`, per-phase `cards_count`), phase card reads/counts, `create_card` with `phase_id`/`title`, `#RGB` label colors, and `ReportCardsFilter` preflight planner.
[...]

PR #280, opened 2026-06-03
feat(sdk): card-phase preflight planners (label hex, report filter)

adriannoes:
During **agent-seeding of the Analytics pipe** (`306996636`), agents used **`execute_graphql`** and **`introspect_type` / `search_schema`** when first-class tools lacked card/phase workflows. **Spec:** `.cursor/dev-planning/specs/archive/card-phase-ergonomics/`

**Stack** replacing closed [#277](https://github.com/pipefy/ai-toolkit/pull/277). Merge prior PRs in the stack before this one.

---

## Objective

Pure SDK planners (no GraphQL, no MCP/CLI): `normalize_label_color` (#RGB/#RRGGBB) and `validate_report_cards_filter` (ReportCardsFilter shape).

## Why now

Chaos-pipe hit `create_label("red")` and malformed report filters; validation belongs in a testable layer before tools call GraphQL.

## Improvements

- Fail-fast, grep-friendly errors for agents.
- Full unit coverage without network.


gbrlcustodio:
## Review

Two pure SDK planners (`normalize_label_color`, `validate_report_cards_filter`) with full unit coverage and no network. The logic is correct and the test suite is thorough. One finding, about naming rather than behavior. The errors-as-values contract was checked against both real consumers (the CLI adapter in #282 and the MCP adapter in #284) and is the right call; only the function name is out of step with it.

### Findings

1. **[minor] report_filter_preflight.py:36** name reads `validate_`, return type is an error string

   `validate_report_cards_filter(filter_obj) -> str | None` returns `None` when valid and an error message string when invalid. The errors-as-values shape is correct: the SDK layer must stay free of surface-specific exceptions and envelopes, and both consumers adapt the string cleanly (the CLI raises `typer.BadParameter`, the MCP tool wraps it in `build_report_error_payload`). The return type is not the problem.

   The name is. This package's other `validate_*` functions return a normalized value (`validate_page_size -> (int, err)`, `validate_tool_id -> (id, err)`) or raise (`validate_positional_id`). None return a bare error string, so `validate_` sets the wrong expectation here. The convention that fits is already in the consuming file: `report_tools.py` has `_blank_field_error(value, field) -> dict | None`, and both new consumers named their adapters `_report_filter_preflight_error` and `validate_report_filter_cli`. A `*_error` noun phrase matches the return.

   Suggested rename, keeping the `str | None` return:

   ```python
   def report_cards_filter_error(filter_obj: dict[str, Any] | None) -> str | None:
       """Return an error message if filter_obj is not a valid ReportCardsFilter, else None."""
   ```

   Call sites then read as what they do: `if (err := report_cards_filter_error(f)) is not None:`. The rename touches the function, its `__all__`, and the two consumer adapters in the stacked PRs #282 and #284, all unmerged, so the cost is contained to the stack.

adriannoes:
Thanks @gbrlcustodio — agreed on the naming, and rather than renaming in place here, the rename lands in #291 together with the SDK-boundary move you proposed on #284: `validate_report_cards_filter` -> `report_cards_filter_error`, consumed by `ReportService._prepared_report_cards_filter`, which raises `ValueError` with the planner message. The errors-as-values shape stays, the name now matches the `str | None` return, and the two consumer adapters this rename would have touched in #282/#284 are deleted by the same move.

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
```diff
@@ -0,0 +1,142 @@
...
+            f"{path}.value must be a string when set, received {type(value).__name__}."
+        )
+
+    field_type = node.get("type")
+    if field_type is not None and (
+        not isinstance(field_type, str) or not field_type.strip()
+    ):
+        return (
+            f"{path}.type must be a non-empty string when set, received {field_type!r}."
+        )
```
Danielmoraisg: minor: query "operator" and "type" are only checked as non-empty strings, but both are closed enums in the schema. ReportCardsFilterQueryOperators is eq, not_eq, contains, not_contains, gt, gte, lt, lte, exists, not_exists, unknown, period; ReportCardsFilterFieldTypes is email, number, string, group_select, select, date, boolean (verified by live introspection). The group operator is enum-checked at lines 61-70, so the validator is asymmetric: {"field": "x", "operator": "equals"} passes preflight and then fails at GraphQL with exactly the opaque error this preflight exists to prevent. The error text at line 115 already lists enum values without enforcing them. Suggest enforcing both enums, or documenting why they are left open.

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
```diff
@@ -0,0 +1,142 @@
...
+    if filter_obj is None:
+        return None
+    if not isinstance(filter_obj, dict):
+        return f"filter must be a JSON object (ReportCardsFilter), received {type(filter_obj).__name__}"
+    return _validate_filter_group(filter_obj, path="filter")
+
+
+def _validate_filter_group(node: dict[str, Any], *, path: str) -> str | None:
+    if path == "filter":
+        for rejected in REPORT_CARDS_FILTER_REJECTED_ROOT_KEYS:
```
Danielmoraisg: nit: iterating the frozenset directly makes the reported key nondeterministic when more than one rejected key is present, since set iteration order depends on string hash randomization. Iterate sorted(REPORT_CARDS_FILTER_REJECTED_ROOT_KEYS), matching the sorted(unknown) at line 58.

packages/sdk/src/pipefy_sdk/label_color.py
```diff
@@ -0,0 +1,19 @@
...
+
+def normalize_label_color(value: str) -> str:
+    """Normalize label color to ``#RRGGBB`` (accepts ``#RGB`` or ``#RRGGBB``)."""
+    stripped = value.strip()
+    if _LABEL_COLOR_HEX_3_RE.fullmatch(stripped):
+        r, g, b = stripped[1], stripped[2], stripped[3]
+        return f"#{r}{r}{g}{g}{b}{b}".upper()
+    if _LABEL_COLOR_HEX_6_RE.fullmatch(stripped):
+        return f"#{stripped[1:].upper()}"
+    raise ValueError(f"expected #RGB or #RRGGBB hex color, received {stripped!r}")
```
Danielmoraisg, line 19: nit: the two sibling planners use different error contracts: this one raises ValueError while report_filter_preflight.py:38 returns str | None. #291 later unifies the surface by raising at the SDK boundary, so this is just a consistency wrinkle within this PR; fine to leave if the stack lands as is.

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
```diff
@@ -0,0 +1,142 @@
...
+    "operator": "and",
+    "queries": [
+        {
+            "field": "current_phase",
+            "operator": "eq",
+            "type": "select",
+            "value": "<phase_id>",
+        }
+    ],
+}
```
Danielmoraisg: nit: EXAMPLE_PHASE_FILTER is a mutable module-level dict exported in __all__ and later embedded in error payloads (#284). Any consumer mutation poisons every subsequent error message. MappingProxyType or a factory function would close that off.

Danielmoraisg (approved):

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
adriannoes: Addressed in e13fea0: added `REPORT_CARDS_FILTER_QUERY_OPERATORS` and `REPORT_CARDS_FILTER_FIELD_TYPES` (values from live introspection) and reject invalid query `operator` / `type` with the same enum-style messages as group `operator`.

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
adriannoes: Addressed in e13fea0: rejected-root-key loop now iterates `sorted(REPORT_CARDS_FILTER_REJECTED_ROOT_KEYS)` for deterministic reporting.

packages/sdk/src/pipefy_sdk/label_color.py
adriannoes: Acknowledged — leaving `normalize_label_color` as `ValueError` here; #291 unifies raising at the SDK boundary so MCP/CLI share one contract.

packages/sdk/src/pipefy_sdk/report_filter_preflight.py
adriannoes: Addressed in e13fea0: `EXAMPLE_PHASE_FILTER` is now a `MappingProxyType` over a factory body so top-level mutation cannot poison shared error payloads; tests assert read-only behavior.

PR #281, opened 2026-06-03
feat(sdk): phase card reads, create_card phase_id/title, get_pipe inventory

adriannoes:
During **agent-seeding of the Analytics pipe** (`306996636`), agents used **`execute_graphql`** and **`introspect_type` / `search_schema`** when first-class tools lacked card/phase workflows. **Spec:** `.cursor/dev-planning/specs/archive/card-phase-ergonomics/`

**Stack** replacing closed #277. Merge prior PRs in the stack before this one.

**Depends on:** #280

---

## Objective

SDK + facade for the GraphQL capabilities agents used via raw queries:

- `get_phase_cards`, `get_phase_cards_count`, `get_phase_allowed_move_targets`
- `create_card(..., phase_id=, title=)` on `CreateCardInput`
- `get_pipe` selects `cards_count` on workflow phases (single GraphQL round trip)

## Improvements

- No more `phase { cards_count }` / `createCard` escape hatches for these reads/writes.
- `get_phase(phase_id)` returns `{phase_id, phase_name, cards_count}`; `get_phase_cards_count` delegates to it.
- Start-form intake stays on public fields: `start_form_fields` + `startFormPhaseId` (no `start_form_phase` node — private API).

## Review follow-up

Addresses @gbrlcustodio review (2026-06-09): dropped `pipe_inventory` enrichment and renamed `get_phase_cards_count_payload` → `get_phase`.


packages/sdk/src/pipefy_sdk/services/pipe_service.py
```diff
@@ -40,7 +44,25 @@ def __init__(
...
+            return result
+        start_form_row: dict | None = None
+        start_id = pipe.get("startFormPhaseId")
+        if start_id is not None:
+            start_id_str = str(start_id)
+            phases = pipe.get("phases") or []
+            if not any(
+                isinstance(p, dict) and str(p.get("id")) == start_id_str for p in phases
+            ):
+                start_form_row = await self._fetch_phase_cards_count_row(start_id)
```
gbrlcustodio: The start form phase is private API, so `get_pipe` should not surface it. This enrichment adds a `start_form_phase` node (`{id, name, cards_count}`) to every `get_pipe` result, which leaks that private entity to agents. The public way to describe the intake is `start_form_fields`, which `GET_PIPE_QUERY` already returns in the same round trip, so the intake view should be built from those, not from a start form phase node.

The leak is the reason to drop it. Two further points reinforce it:

- The `cards_count` on that node is private and unused, and is unreliable on the start form anyway: this PR's own docstrings (`get_pipe` and the `get_phase_cards_count` tool) warn that "Start-form `cards_count` may be 0 while cards still exist." So it surfaces a number callers are told not to trust.
- Producing the node costs an extra GraphQL round trip here (`_fetch_phase_cards_count_row`), because the start form phase is not in `GET_PIPE_QUERY`'s `phases` selection (confirmed against a live pipe). That second query runs on every `get_pipe`, including the lightweight paths: `get_labels`, AI validation context, and delete previews, none of which want it.

Since `phases` already excludes the start form, the in-list removal branch and the `cards_count` coercion in `enrich_pipe_get_pipe_inventory` are no-ops as well. So the only load-bearing change in this area is adding `cards_count` to the `phases` selection in `GET_PIPE_QUERY`. The rest can be removed: `pipe_inventory.py`, this conditional second query, both `ValueError` guards, and the `start_form_phase` node.

packages/sdk/src/pipefy_sdk/services/pipe_service.py
```diff
@@ -237,13 +269,47 @@ async def get_phase_cards_count(self, phase_id: str | int) -> int:
...
         """
-        variables = {"phase_id": str(phase_id)}
-        result = await self.execute_query(GET_PHASE_CARDS_COUNT_QUERY, variables)
-        phase = (result or {}).get("phase")
-        if not isinstance(phase, dict) or phase.get("cards_count") is None:
-            raise ValueError("phase.cards_count missing from response")
+        phase = await self._fetch_phase_cards_count_row(phase_id)
         return int(phase["cards_count"])
 
+    async def get_phase_cards_count_payload(self, phase_id: str | int) -> dict:
```
gbrlcustodio: The `_payload` suffix is the issue here. It is a one-off (the only `*_payload` method in the SDK), and elsewhere in the codebase "payload" means a GraphQL request/response body, not "a labeled dict for tools." The suffix also does not convey what distinguishes this from the `get_phase_cards_count` int sibling: that it adds the phase id and name. The two cannot be told apart by name.

Suggested fix: rename to a canonical phase-node getter that returns the same shape, with no new parameters.

```python
async def get_phase(self, phase_id) -> dict:
    # {phase_id, phase_name, cards_count}
```

Keep `get_phase_cards_count -> int` as a thin scalar that delegates to it (single fetch).

I considered folding `get_phase_fields` into this via an `include_fields`/`required_only` pair, but that is the wrong shape: `required_only` is meaningful only when fields are requested, so a separate boolean is inert unless `include_fields=True` and admits an invalid combination (`required_only=True, include_fields=False`). Patching that with a single mode parameter (a `Literal` or `Enum`) would work, but it pushes the method toward a return shape that varies by argument, which is the same smell that keeps `get_phase_cards` (a paginated connection) out. `required_only` is a fields-specific filter, so it belongs on the fields-specific method.

So leave `get_phase_fields(phase_id, required_only=False)` as its own method. Two cohesive getters (`get_phase` for the count node, `get_phase_fields` for the form) read better than one getter with a mode argument, and the codebase already favors a plain `include_fields: bool` for optional related data (`get_card`, `get_phase_cards`) rather than filter-mode flags.

adriannoes:
Thanks @gbrlcustodio — both points addressed in `9ef6586`, with stack PRs #282–#285 updated to match.

**`start_form_phase` / `get_pipe`:** dropped the enrichment entirely. `get_pipe` is a single `GET_PIPE_QUERY` again; `pipe_inventory.py` removed. Only load-bearing change is `cards_count` on workflow `phases`. Start-form intake stays on `start_form_fields` / `startFormPhaseId`.

**`get_phase` rename:** `get_phase_cards_count_payload` → `get_phase`; `get_phase_cards_count` delegates. `get_phase_fields` left separate. MCP/CLI now call `client.get_phase`.

Holding merge until we get another review pass on the stack.

packages/sdk/src/pipefy_sdk/queries/pipe_queries.py
```diff
@@ -133,12 +134,45 @@
     query GetPhaseCardsCount($phase_id: ID!) {
```
gbrlcustodio: Follow-up to the `get_phase` rename: the backing query and private helper still carry count-only names, but neither is count-only anymore.

`GET_PHASE_CARDS_COUNT_QUERY` selects `phase { id, name, cards_count }` and now backs `get_phase`, which returns the whole row. `_fetch_phase_cards_count_row` returns that same row and is consumed by both `get_phase` and `get_phase_cards_count`. The count is one field on a phase node, not the purpose of the query.

Suggested rename to match the `get_phase` standard:
- `GET_PHASE_CARDS_COUNT_QUERY` -> `GET_PHASE_QUERY` (GraphQL op `GetPhaseCardsCount` -> `GetPhase`)
- `_fetch_phase_cards_count_row` -> `_fetch_phase_row`

`get_phase_cards_count -> int` stays as is: it is a genuine count accessor that delegates to the helper. Touch points are contained: this query module (and its `__all__`), the import and two call sites in `pipe_service.py`, and two references in `test_pipe_service.py`.

adriannoes:
Thanks @gbrlcustodio — addressed in `4fdab2c`: renamed `GET_PHASE_CARDS_COUNT_QUERY` → `GET_PHASE_QUERY`, GraphQL op `GetPhase`, and `_fetch_phase_cards_count_row` → `_fetch_phase_row` to match the `get_phase` surface.

packages/sdk/tests/services/test_pipe_service.py
```diff
@@ -549,14 +581,107 @@ async def test_get_phase_cards_count_raises_when_missing(mock_settings):
...
+        return_value={
+            "phase_id": str(phase_id),
+            "phase_name": "Doing",
+            "cards_count": 42,
+        }
+    )
+
+    result = await service.get_phase_cards_count(phase_id)
+
+    service.get_phase.assert_awaited_once_with(phase_id)
```
Danielmoraisg: nit: this test replaces `service.get_phase` with an `AsyncMock` and asserts the await, which tests internal wiring rather than a contract. The observable behavior (payload in, 42 out) is already covered by `test_get_phase_cards_count_returns_native_scalar`; this one only adds coupling to the private delegation choice.

Danielmoraisg (approved):

packages/sdk/tests/services/test_pipe_service.py
adriannoes: Thanks @Danielmoraisg — agreed. Removed `test_get_phase_cards_count_delegates_to_get_phase` in 6fba81c; `test_get_phase_cards_count_returns_native_scalar` keeps the public contract (GET_PHASE_QUERY in, scalar 42 out).

adriannoes:
All review feedback on this PR is addressed:

- **@gbrlcustodio** — `start_form_phase` enrichment dropped (`9ef6586`); `get_phase_cards_count_payload` → `get_phase` with `GET_PHASE_QUERY` / `_fetch_phase_row` rename (`4fdab2c`). Threads resolved.
- **@Danielmoraisg** — removed the delegation wiring test; contract coverage stays on `test_get_phase_cards_count_returns_native_scalar` (`6fba81c`).

SDK unit tests green locally (`927 passed`, `-m "not integration"`). Ready for merge into `feat/card-phase-erg/1-preflight-planners`.

PR #282, opened 2026-06-03
feat(mcp,cli): phase inventory tools and CLI parity

adriannoes:

**Depends on:** https://github.com/pipefy/ai-toolkit/pull/281

---

## Objective

Expose SDK phase reads on MCP + CLI so agents can audit/move without GraphQL:

- MCP: `get_phase_cards`, `get_phase_cards_count`, `get_phase_allowed_move_targets`
- CLI: `pipefy phase cards`, `cards-count`, `allowed-moves`
- Clamp `phase cards --first` to 1–500 (F5)
- MCP `create_label` / `update_label` hex via planner (needs PR1)

## Improvements

- Discover allowed move targets **before** `move_card_to_phase` fails.
- Count/list cards in one phase without pipe-wide `get_cards` title search.

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/_common.py
+++ b/packages/cli/src/pipefy_cli/commands/_common.py
@@ -69,7 +69,9 @@ _CARDS_PAGE_SIZE_MIN = 1
 _CARDS_PAGE_SIZE_MAX = 500
 
 
-def validate_cards_page_size(first: int) -> int:
+def validate_cards_page_size(first: int | None) -> int | None:
+    if first is None:
+        return None
     if first < _CARDS_PAGE_SIZE_MIN or first > _CARDS_PAGE_SIZE_MAX:
         raise typer.BadParameter(
             f"--first must be between {_CARDS_PAGE_SIZE_MIN} and "
--- a/packages/cli/src/pipefy_cli/commands/_common.py
+++ b/packages/cli/src/pipefy_cli/commands/_common.py
@@ -13,6 +13,7 @@ import typer
 from gql.transport.exceptions import TransportError, TransportQueryError
 from pipefy_sdk import PipefyClient, PipefySettings, stream_bytes
 from pipefy_sdk.exceptions import PipefyError
+from pipefy_sdk.report_filter_preflight import validate_report_cards_filter
 
 from pipefy_cli.auth import (
     AuthContext,
@@ -47,6 +48,40 @@ def validate_positional_id(value: str) -> str:
 def resource_id_argument(*, help: str) -> Any:
     """Typer ``Argument`` for resource ids when ``ignore_unknown_options`` is enabled."""
     return typer.Argument(..., help=help, callback=validate_positional_id)
+
+
+def validate_optional_resource_id(value: str | None, label: str) -> str | None:
+    if value is None:
+        return None
+    cleaned = value.strip()
+    if not cleaned:
+        raise typer.BadParameter(
+            f"Invalid '{label}': provide a non-empty string or positive integer."
+        )
+    if cleaned.startswith("-") and cleaned[1:].isdigit():
+        raise typer.BadParameter(f"Invalid '{label}': provide a positive integer.")
+    if cleaned.isdigit() and int(cleaned) <= 0:
+        raise typer.BadParameter(f"Invalid '{label}': provide a positive integer.")
+    return cleaned
+
+
+_CARDS_PAGE_SIZE_MIN = 1
+_CARDS_PAGE_SIZE_MAX = 500
+
+
+def validate_cards_page_size(first: int) -> int:
+    if first < _CARDS_PAGE_SIZE_MIN or first > _CARDS_PAGE_SIZE_MAX:
+        raise typer.BadParameter(
+            f"--first must be between {_CARDS_PAGE_SIZE_MIN} and "
+            f"{_CARDS_PAGE_SIZE_MAX} (inclusive)."
+        )
+    return first
+
+
+def validate_report_filter_cli(filter_obj: dict[str, Any] | None) -> None:
+    message = validate_report_cards_filter(filter_obj)
+    if message is not None:
+        raise typer.BadParameter(message)
 
 
 def export_poll_max_rounds(
--- a/packages/cli/src/pipefy_cli/commands/phase.py
+++ b/packages/cli/src/pipefy_cli/commands/phase.py
@@ -44,10 +44,10 @@ def phase_get(
 
 
 @phase_app.command(
-    "allowed-moves",
+    "targets",
     context_settings=ID_POSITIONAL_CONTEXT_SETTINGS,
 )
-def phase_allowed_moves(
+def phase_targets(
     ctx: typer.Context,
     phase_id: str = resource_id_argument(
         help="Source phase id (card's current phase)."
@@ -68,10 +68,10 @@ def phase_allowed_moves(
 
 
 @phase_app.command(
-    "cards-count",
+    "count",
     context_settings=ID_POSITIONAL_CONTEXT_SETTINGS,
 )
-def phase_cards_count(
+def phase_count(
     ctx: typer.Context,
     phase_id: str = resource_id_argument(help="Phase id."),
     json_out: bool = typer.Option(
--- a/packages/cli/src/pipefy_cli/commands/phase.py
+++ b/packages/cli/src/pipefy_cli/commands/phase.py
@@ -88,7 +88,7 @@ def phase_cards_count(
     """
 
     async def factory(client: PipefyClient):
-        return await client.get_phase_cards_count_payload(phase_id)
+        return await client.get_phase(phase_id)
 
     run_cli_command(ctx, json_out, factory)
 
--- a/packages/cli/src/pipefy_cli/commands/phase.py
+++ b/packages/cli/src/pipefy_cli/commands/phase.py
@@ -6,6 +6,10 @@ from typing import Any
 
 import typer
 from pipefy_sdk import PipefyClient
+from pipefy_sdk.phase_inventory import (
+    get_phase_not_found_message,
+    is_get_phase_not_found_error,
+)
 
 from pipefy_cli.commands._common import (
     ID_POSITIONAL_CONTEXT_SETTINGS,
@@ -88,7 +92,12 @@ def phase_count(
     """
 
     async def factory(client: PipefyClient):
-        return await client.get_phase(phase_id)
+        try:
+            return await client.get_phase(phase_id)
+        except ValueError as exc:
+            if is_get_phase_not_found_error(exc):
+                raise ValueError(get_phase_not_found_message(phase_id)) from exc
+            raise
 
     run_cli_command(ctx, json_out, factory)
 
--- a/packages/cli/src/pipefy_cli/commands/phase.py
+++ b/packages/cli/src/pipefy_cli/commands/phase.py
@@ -13,6 +13,7 @@ from pipefy_cli.commands._common import (
     parse_json_object,
     resource_id_argument,
     run_cli_command,
+    validate_cards_page_size,
 )
 
 phase_app = typer.Typer(help="Pipe phase operations.", no_args_is_help=True)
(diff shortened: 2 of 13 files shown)


gbrlcustodio (commented):
Two CLI naming notes on the new phase commands.

packages/cli/src/pipefy_cli/commands/phase.py
```diff
@@ -42,6 +43,97 @@ async def factory(client: PipefyClient):
...
+    """List phases a card may move to from this phase (UI transition rules)."""
+
+    async def factory(client: PipefyClient):
+        return await client.get_phase_allowed_move_targets(phase_id)
+
+    run_cli_command(ctx, json_out, factory)
+
+
+@phase_app.command(
+    "cards-count",
```
gbrlcustodio: `cards-count` reads as a noun-noun and stands as a second verb beside `cards` over the same resource. The count is a projection of those cards, not a separate object, so the name labels a derived value rather than an action.

Recommend `phase count <phase_id>`: it reads as an action ("count cards in this phase"), stays in the phase group, and drops the noun-noun.

Two alternatives considered and not recommended:
- `phase cards --count`: a flag that switches the return from a card list to a scalar. That is the return-shape mode flag rejected on the `get_phase` thread in #281; the same objection applies.
- `card count --phase-id <id>`: move it to the card app and scope by option. Only worth it if counting will ever be scoped by something other than a phase. Today it cannot: there is no cheap pipe-total scalar, so `--phase-id` would be effectively required and the command would be this one with more ceremony and no new capability.

MCP stays `get_phase_cards_count` either way. CLI verbs already diverge from tool names (`card list` vs `get_cards`), so parity is intact.

packages/cli/src/pipefy_cli/commands/phase.py
```diff
@@ -42,6 +43,97 @@ async def factory(client: PipefyClient):
     run_cli_command(ctx, json_out, factory)
 
 
+@phase_app.command(
+    "allowed-moves",
```
gbrlcustodio: This command returns a list of phases: the destination phases a card can move to (`allowed_phases` of `{id, name}`, from `phase.cards_can_be_moved_to_phases`). `allowed-moves` names the action, but what comes back is phases.

Recommend renaming to `targets` (`phase targets <phase_id>`): it names the returned entity (target phases) and aligns with the MCP tool suffix `get_phase_allowed_move_targets`. `destinations` is a fine alternative if you want it to read without knowing the tool name. Both beat `allowed-moves`, which labels the edges rather than the phases returned.

packages/mcp/tests/tools/test_pipe_tool_helpers.py
```diff
@@ -23,6 +23,7 @@
     UserCancelledError,
     _filter_editable_field_definitions,
     _filter_fields_by_definitions,
+    _merge_phase_and_start_form_field_values,
```
gbrlcustodio: This import (and the `test_merge_phase_and_start_form_field_values_keeps_both_subsets` test it feeds) references `_merge_phase_and_start_form_field_values`, which is not defined in `pipe_tool_helpers.py` on this branch. The function lands in #283. On #282 alone the import raises `ImportError`, so the whole module fails to collect:

```
ImportError: cannot import name '_merge_phase_and_start_form_field_values'
from 'pipefy_mcp.tools.pipe_tool_helpers'
```

Verified by running `uv run python -m pytest packages/mcp/tests/tools/test_pipe_tool_helpers.py` against this PR head: 1 error during collection, 0 tests run. CI is red on #282 in isolation.

Move this import and the test to #283, next to the function they exercise.

packages/cli/src/pipefy_cli/commands/_common.py
```diff
@@ -49,6 +50,40 @@ def resource_id_argument(*, help: str) -> Any:
...
+    if cleaned.isdigit() and int(cleaned) <= 0:
+        raise typer.BadParameter(f"Invalid '{label}': provide a positive integer.")
+    return cleaned
+
+
+_CARDS_PAGE_SIZE_MIN = 1
+_CARDS_PAGE_SIZE_MAX = 500
+
+
+def validate_cards_page_size(first: int) -> int:
```
gbrlcustodio: `validate_cards_page_size` duplicates the existing private `_validate_cards_page_size` in `card.py` (same 1-500 bounds, same `typer.BadParameter` message). The `phase cards` test even notes "parity with `card list`", but `card list` still calls its own private copy (`card.py:164`), so the two now coexist.

Either migrate `card.py` to this shared validator and delete `_validate_cards_page_size`, or, if they must stay separate, say why. The signatures diverge: the `card.py` version takes `int | None` and passes None through; this one takes `int`. To share cleanly, this should accept `int | None` too, or `card list` keeps its None guard at the call site.

gbrlcustodio (commented):
One blocker (test fails to collect on this PR in isolation) and one duplication nit.

packages/cli/src/pipefy_cli/commands/_common.py
gbrlcustodio: Withdrawing this: #283 deletes `_validate_cards_page_size` and migrates `card list` to this shared validator, handling the None case at the call site. The duplication is transient within this PR, not a standing defect.

adriannoes:
Thanks @gbrlcustodio — addressed in `d096020`: moved the `_merge_phase_and_start_form_field_values` test to #283 (fixes collection on this PR in isolation). Renamed CLI to `pipefy phase count` and `pipefy phase targets`.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
```diff
@@ -343,6 +354,172 @@ async def clone_pipe(
...
+            try:
+                payload = await client.get_phase(phase_id_str)
+            except Exception as exc:  # noqa: BLE001
+                return handle_pipe_config_tool_graphql_error(
+                    exc,
+                    "Get phase cards count failed.",
+                    debug=debug,
+                    resource_kind="phase",
+                    resource_id=phase_id_str,
+                )
```
Danielmoraisg, line 456: minor: `get_phase_cards_count` handles phase-not-found inconsistently with its two siblings. The SDK raises `ValueError("phase.cards_count missing from response")` when `phase` is null; the broad `except Exception` routes it to `handle_pipe_config_tool_graphql_error`, but NOT_FOUND enrichment never fires (no GraphQL code, and `_looks_like_not_found` matches only the markers "not found"/"does not exist"), so the agent sees the raw internal message with no `NOT_FOUND` code. `get_phase_allowed_move_targets` and `get_phase_cards` both return "Phase {id} not found or access denied." with `code=NOT_FOUND`. The same root cause makes CLI `phase count` exit 2 with the opaque message while `phase targets`/`cards` print `{"phase": null}` with exit 0. Worth fixing in the same pass

packages/cli/src/pipefy_cli/commands/_common.py
```diff
@@ -49,6 +50,40 @@ def resource_id_argument(*, help: str) -> Any:
...
+    if first < _CARDS_PAGE_SIZE_MIN or first > _CARDS_PAGE_SIZE_MAX:
+        raise typer.BadParameter(
+            f"--first must be between {_CARDS_PAGE_SIZE_MIN} and "
+            f"{_CARDS_PAGE_SIZE_MAX} (inclusive)."
+        )
```
Danielmoraisg: minor: this shared `validate_cards_page_size` duplicates `card.py`'s private `_validate_cards_page_size` (card.py:45-52), including a second copy of the min/max constants (`_CARDS_PAGE_SIZE_MIN/MAX` vs `_CARDS_FIRST_MIN/MAX`). `card list` was not refactored to use the shared helper, so the 1-500 rule now lives in two CLI places (three counting MCP's `MAX_PAGE_SIZE`). Point `card list` at this helper while it is being introduced.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
```diff
@@ -343,6 +354,172 @@ async def clone_pipe(
...
+                phase_id: Phase ID. Discover via: ``get_pipe(pipe_id).phases[].id``
+                    for workflow phases, or ``get_pipe(pipe_id).startFormPhaseId``
+                    for the start-form phase (not listed in ``phases[]``).
+                first: Max cards per page (1-500). Default 50.
+                after: Cursor from ``pageInfo.endCursor`` of a previous call.
+                include_fields: When True, include each card's custom fields.
+                debug: When True, append GraphQL codes and correlation_id to errors.
+
+            Returns:
+                GraphQL ``phase`` object with ``cards`` connection (``edges``, ``pageInfo``,
```
Danielmoraisg: nit: the Returns section says the tool returns the GraphQL `phase` object with a `cards` connection, but it actually returns the query-root dict `{"phase": {...}}` (legacy) or that dict under `data` (envelope). Also at lines 1067 and 1132 the label docstrings say hex `#RRGGBB` while `normalize_label_color` also accepts `#RGB`; the error message reveals the wider contract.

packages/mcp/src/pipefy_mcp/tools/registry.py
```diff
@@ -20,6 +20,7 @@
 from pipefy_mcp.tools.table_tools import TableTools
 from pipefy_mcp.tools.webhook_tools import WebhookTools
 
+# 152 tools — keep in sync with docs/parity.md row count and README MCP totals.
```
Danielmoraisg: nit: this comment uses an em dash ("# 152 tools — keep in sync..."); replace with plain punctuation. The hardcoded `152` is also rot-prone; consider dropping the count and keeping just the "keep in sync with parity.md and README" note.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
```diff
@@ -343,6 +354,172 @@ async def clone_pipe(
...
+                readOnlyHint=True,
+            ),
+        )
+        async def get_phase_cards_count(
+            phase_id: PipefyId,
+            debug: bool = False,
+        ) -> dict[str, Any]:
+            """Return the native ``Phase.cards_count`` for a phase (fast inventory).
+
+            Uses the schema scalar — no card enumeration. On the **start-form** phase,
```
Danielmoraisg: nit: em dash in the docstring ("Uses the schema scalar — no card enumeration."). Replace with plain punctuation, e.g. "Uses the schema scalar; no card enumeration."

Danielmoraisg (approved):

packages/cli/src/pipefy_cli/commands/_common.py
adriannoes: Fixed in 011ec76: `card list` now uses shared `validate_cards_page_size` from `_common.py` (accepts `int | None`); removed duplicate `_validate_cards_page_size` and constants from `card.py`.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
adriannoes: Fixed in 011ec76: `get_phase_cards` Returns now documents the query-root `{"phase": ...}` shape; `create_label` / `update_label` color docs now say `#RGB` or `#RRGGBB`.

packages/mcp/src/pipefy_mcp/tools/registry.py
adriannoes: Fixed in 011ec76: dropped the hardcoded count and em dash; comment is now a plain sync note.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
adriannoes: Fixed in 011ec76: replaced em dash with semicolon in the `get_phase_cards_count` docstring.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tools.py
adriannoes: Fixed in 011ec76: added `pipefy_sdk.phase_inventory` to map `get_phase` ValueError to NOT_FOUND in MCP `get_phase_cards_count` and CLI `phase count` (exit 2 with the same message as siblings). Regression tests added.

PR #283, opened 2026-06-03
feat(mcp,cli): create_card phase_id and title on CreateCardInput

adriannoes:

**Depends on:** https://github.com/pipefy/ai-toolkit/pull/282

---

## Objective

**#1 chaos-pipe pain:** cards in orphan/non–start-form phases via `createCard` + `introspect_type("CreateCardInput")`.

- MCP `create_card(phase_id=…)` with `validate_tool_id`, merged start-form + phase field filtering
- Pass `title` on `CreateCardInput`; `title_warning` only on real API mismatch (F1)
- CLI `card create --phase-id` (F3)
- Docstrings: `start_form_phase.id` discovery (F4)

## Improvements

- Seed 12 cards in a Limbo phase with tools only (success metric from PRD).

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/card.py
+++ b/packages/cli/src/pipefy_cli/commands/card.py
@@ -56,7 +56,7 @@ def _apply_create_card_title_warning(
 ) -> dict[str, Any]:
     if not requested_title:
         return result
-    card_node = result.get("createCard", {}).get("card")
+    card_node = (result.get("createCard") or {}).get("card")
     if not isinstance(card_node, dict):
         return result
     if card_node.get("title") == requested_title:
--- a/packages/cli/src/pipefy_cli/commands/card.py
+++ b/packages/cli/src/pipefy_cli/commands/card.py
@@ -51,6 +51,24 @@ def _parse_fields_json(raw: str | None) -> dict[str, Any] | list[dict[str, Any]]
     raise typer.BadParameter("--fields must be a JSON object or array")
 
 
+def _apply_create_card_title_warning(
+    result: dict[str, Any], *, requested_title: str | None
+) -> dict[str, Any]:
+    if not requested_title:
+        return result
+    card_node = result.get("createCard", {}).get("card")
+    if not isinstance(card_node, dict):
+        return result
+    if card_node.get("title") == requested_title:
+        return result
+    warned = dict(result)
+    warned["title_warning"] = (
+        "Card created but title was not applied as expected "
+        f"(response title={card_node.get('title')!r}, requested={requested_title!r})."
+    )
+    return warned
+
+
 def _parse_field_updates_json(raw: str | None) -> list[dict[str, Any]] | None:
     if raw is None or raw.strip() == "":
         return None
@@ -258,7 +276,8 @@ def card_create(
             create_kwargs["phase_id"] = phase_id
         if title:
             create_kwargs["title"] = title
-        return await client.create_card(pipe_id, payload, **create_kwargs)
+        result = await client.create_card(pipe_id, payload, **create_kwargs)
+        return _apply_create_card_title_warning(result, requested_title=title)
 
     run_cli_command(ctx, json_out, factory)
 
--- a/packages/cli/src/pipefy_cli/commands/card.py
+++ b/packages/cli/src/pipefy_cli/commands/card.py
@@ -25,6 +25,7 @@ from pipefy_cli.commands._common import (
     resource_id_argument,
     run_cli_command,
     validate_cards_page_size,
+    validate_optional_resource_id,
 )
 
 card_app = typer.Typer(help="Card operations.", no_args_is_help=True)
@@ -223,12 +224,20 @@ def card_create(
         None,
         "--title",
         "-t",
-        help="Optional title (applied via update after create).",
+        help="Optional title (sent on CreateCardInput when set).",
+    ),
+    phase_id: str | None = typer.Option(
+        None,
+        "--phase-id",
+        help=(
+            "Target phase id (CreateCardInput.phase_id). "
+            "Creates the card in that phase instead of the start form."
+        ),
     ),
     fields_json: str | None = typer.Option(
         None,
         "--fields",
-        help="JSON object or array: start-form field values for createCard.",
+        help="JSON object or array: field values for createCard.",
     ),
     json_out: bool = typer.Option(
         False,
@@ -237,19 +246,19 @@ def card_create(
         help="Print machine-readable JSON to stdout.",
     ),
 ) -> None:
-    """Create a card in a pipe (start-form fields via --fields JSON when required)."""
+    """Create a card in a pipe (start-form or phase fields via --fields JSON when required)."""
 
     fields = _parse_fields_json(fields_json)
+    phase_id = validate_optional_resource_id(phase_id, "phase_id")
 
     async def factory(client: PipefyClient):
         payload = fields if fields is not None else {}
-        result = await client.create_card(pipe_id, payload)
-        card_node = (result.get("createCard") or {}).get("card") or {}
-        cid = card_node.get("id")
-        if title and cid:
-            await client.update_card(str(cid), title=title)
-            card_node["title"] = title
-        return result
+        create_kwargs: dict[str, Any] = {}
+        if phase_id is not None:
+            create_kwargs["phase_id"] = phase_id
+        if title:
+            create_kwargs["title"] = title
+        return await client.create_card(pipe_id, payload, **create_kwargs)
 
     run_cli_command(ctx, json_out, factory)
 
--- a/packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
@@ -111,9 +111,8 @@ class PipeTools:
             ``get_start_form_fields`` and pass all required values.
 
             With ``phase_id``, the card is created in that phase (including orphan
-            phases that are not the start form). Discover via:
-            ``get_pipe(pipe_id).phases[].id`` or ``get_pipe(pipe_id).startFormPhaseId``.
-            For agent seeding in a specific phase, set
+            phases that are not the start form). Discover workflow phase IDs via
+            ``get_pipe(pipe_id).phases[].id``. For agent seeding in a specific phase, set
             ``phase_id`` and ``skip_elicitation=true``. When ``fields`` is non-empty,
             keys are filtered against ``get_phase_fields(phase_id)`` and
             ``get_start_form_fields(pipe_id)`` so pipes that still require start-form
@@ -140,10 +139,10 @@ class PipeTools:
                 required_fields_only: If True, only elicit required fields. Default: False.
                 skip_elicitation: When True, bypass interactive elicitation and send
                     ``fields`` directly to the API. Recommended for AI agent workflows.
-                phase_id: Optional target phase ID. Skips start-form elicitation; when
-                    ``fields`` is non-empty, validates against phase and start-form
-                    field definitions. Discover via: ``get_pipe(pipe_id).phases[].id``
-                    or ``get_pipe(pipe_id).startFormPhaseId``.
+                phase_id: Optional target phase ID. Interactive elicitation prompts
+                    start-form fields (matching the Pipefy UI); phase field values are
+                    merged from ``fields`` when provided. For agent seeding, set
+                    ``skip_elicitation=true``. Discover via: ``get_pipe(pipe_id).phases[].id``.
             """
             card_data = fields or {}
             can_elicit = supports_elicitation(ctx)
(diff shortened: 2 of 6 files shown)


packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -121,48 +132,119 @@ async def create_card(
...
+                await ctx.debug(
+                    f"Expected fields for phase {phase_id}: {expected_fields}"
+                )
+                await ctx.debug(
+                    f"Expected start-form fields for pipe {pipe_id}: "
+                    f"{start_form_expected_fields}"
+                )
+                await ctx.debug(f"Provided fields: {fields}")
+
+                if can_elicit and expected_fields and not skip_elicitation:
```
gbrlcustodio: In the phase branch, the interactive path elicits `expected_fields`, which are the phase fields from `get_phase_fields`. Pipefy prompts only start-form fields at card creation, even when the card is created in a specific phase; phase fields are filled later from within the phase. So this path prompts fields the product never shows at creation, and it drops any start-form values the caller pre-passed, since the elicited result is keyed to phase fields only.

The skip-elicitation sibling merges both sets via `_merge_phase_and_start_form_field_values`, and the API accepts that (see `test_create_card_with_phase_id_keeps_start_form_and_phase_fields`), so whether start-form values survive depends on whether elicitation fires. On pipes that require start-form values on `CreateCardInput`, the interactive path can build a payload missing them and fail at the API.

Fix: elicit `start_form_expected_fields` to mirror the UI, then merge pre-passed phase values back. Phase fields stay a programmatic concern, sent but not prompted. Keying the elicit condition on the start-form set also lets the form fire when no `fields` were pre-passed, matching the non-phase path.

Scope is interactive use only; agent seeding via `skip_elicitation=True` is unaffected and correct.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -121,48 +132,119 @@ async def create_card(
...
-            )
-
-            await ctx.debug(f"Expected fields for pipe {pipe_id}: {expected_fields}")
-            await ctx.debug(f"Provided fields: {fields}")
-
             card_data = fields or {}
             can_elicit = supports_elicitation(ctx)
 
-            if can_elicit and not skip_elicitation:
+            if phase_id is not None:
```
gbrlcustodio, line 150: Non-blocking. The two top-level branches repeat the same shape: fetch start-form definitions, filter to editable, debug-log, then elicit-or-filter with identical `try/except MalformedFieldDefinitionError / UserCancelledError`. The start-form fetch and the elicit block are duplicated, and the field decisions are interleaved with the awaits, so the branching cannot be unit-tested without a live client and ctx.

Once the start-form-vs-phase elicit fix lands, the phase branch only adds "also fetch phase fields and merge," so the two branches collapse into one path with an optional phase augmentation. Consider extracting a pure `plan_create_card_fields(provided, start_form_defs, phase_defs, can_elicit, skip_elicitation)` that returns what to elicit and what to send, with the awaits kept at the edges. That turns the phase-vs-start-form difference into data rather than a duplicated control-flow branch, and gives the decision logic direct unit tests.

packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -248,19 +244,19 @@ def card_create(
...
-        if title and cid:
-            await client.update_card(str(cid), title=title)
-            card_node["title"] = title
-        return result
+        create_kwargs: dict[str, Any] = {}
+        if phase_id is not None:
+            create_kwargs["phase_id"] = phase_id
+        if title:
+            create_kwargs["title"] = title
+        return await client.create_card(pipe_id, payload, **create_kwargs)
```
gbrlcustodio: Nit, MCP/CLI parity. The MCP `create_card` emits `title_warning` when the API stores a title different from the one requested (Pipefy derives the title from a start-form field, so a passed `title` can be overridden). The CLI `card create` forwards `title` and returns the raw result with no equivalent check, so a CLI user whose title was overridden gets no signal. The gap persists through #284 and #285. The raw result does carry the response `title`, so it is recoverable from `--json`, but there is no explicit warning. Either mirror the MCP verify-and-warn here or record the divergence in `docs/parity.md`.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -659,7 +739,12 @@ async def get_pipe(
 
             Returns:
                 dict: GraphQL response containing a ``pipe`` object with ``id``, ``name``,
-                ``phases``, ``labels``, ``start_form_fields``, and related metadata from the API.
+                ``phases`` (workflow phases only; each includes ``cards_count``),
+                ``startFormPhaseId`` (start form is not listed in ``phases``),
+                ``labels``, ``start_form_fields``, and related metadata from the API.
+
+                For start-form inventory, use ``get_phase`` / ``get_phase_cards`` on
```
gbrlcustodio: On the #281 thread the conclusion was that the start-form phase is private Pipefy API: its card count is not visible to users and surfacing it to an agent has no use. The structural leak from that thread (the enrichment round-trip) was removed and `cards_count` is now selected natively on the workflow phases. This docstring reopens the same path: it advertises `startFormPhaseId` and tells the agent to call `get_phase` / `get_phase_cards` on it for start-form inventory. That is the private start-form phase reached through its id, and the trailing caveat about its unreliable `cards_count` is exactly the value #281 said to keep away from agents.

This is the cross-PR contract point: merging this as-is settles the #281 question in the opposite direction, by docstring rather than by decision. Recommend dropping the start-form-inventory guidance (point at `start_form_fields` for the intake form instead) and reconsidering whether `startFormPhaseId` should be surfaced at all. `create_card` already targets the start form when `phase_id` is omitted, so it is not needed as a create target. The field-exposure call itself belongs on the SDK surface in #281; this comment is to avoid hardening the leak into the tool contract first.

packages/mcp/tests/tools/test_pipe_tools.py
```diff
@@ -593,6 +797,35 @@ async def test_get_pipe_forwards_pipe_id_to_client(
         payload = extract_payload(result)
         assert payload["pipe"]["name"] == "My Pipe"
 
+    @pytest.mark.parametrize("client_session", [None], indirect=True)
+    async def test_get_pipe_returns_phase_cards_count(
```
gbrlcustodio: Nit. This mocks `client.get_pipe` to return a payload that already contains `startFormPhaseId` and per-phase `cards_count`, then asserts they come back. The tool only validates the id and forwards the SDK result, so this checks passthrough, not that `GET_PIPE_QUERY` selects those fields. If someone drops `cards_count` from the query selection, this stays green because the mock still supplies it. The `all(p["id"] != "100" ...)` assertion checks the hand-written fixture too; the tool does no phase filtering, so it cannot fail.

The regression risk is at the query layer, where the repo already guards selections via `print_ast` (see `GET_PHASE_CARDS_COUNT_QUERY` in `test_pipe_service.py`). Add the equivalent for `GET_PIPE_QUERY`:

```python
def test_get_pipe_query_selects_inventory_fields():
    query_text = print_ast(GET_PIPE_QUERY)
    assert "cards_count" in query_text
```

Keep this forwarding test if you want passthrough coverage, but rename it to say so and drop the `p["id"] != "100"` assertion. If the `startFormPhaseId` exposure is removed per the docstring finding above, the query assertion covers only `cards_count`.

gbrlcustodio (commented):
Review of create_card phase_id/title. One corrected behavior note, one refactor note, a parity nit, and two findings that intersect the #281 start-form-phase decision.

adriannoes:
Thanks @gbrlcustodio — addressed in `2e0d0b8`: interactive `create_card` with `phase_id` now elicits start-form fields (UI parity) and merges phase values programmatically; dropped start-form phase inventory hints from `get_pipe`; added `title_warning` on `pipefy card create`; simplified the `get_pipe` passthrough test.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
gbrlcustodio: Verified on 2e0d0b8: the phase branch now elicits the start-form fields and merges any pre-passed phase values back through `_merge_phase_and_start_form_field_values`, with the elicit condition keyed on the start-form set. Matches the intended fix. Resolving.

packages/cli/src/pipefy_cli/commands/card.py
gbrlcustodio: Verified on 2e0d0b8: `_apply_create_card_title_warning` mirrors the MCP verify-and-warn and `card_create` returns it, so a CLI user whose title was overridden now gets the same signal. Resolving.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
gbrlcustodio: Verified on 2e0d0b8: the `get_pipe` docstring now points at `start_form_fields` for intake and no longer routes agents to `get_phase`/`get_phase_cards` on `startFormPhaseId`. The #281 leak is not reopened by this contract. Resolving.

packages/mcp/tests/tools/test_pipe_tools.py
gbrlcustodio: Verified on 2e0d0b8: `test_get_pipe_query_selects_cards_count_on_phases` guards the selection via `print_ast(GET_PIPE_QUERY)`, and the passthrough test was renamed `test_get_pipe_forwards_sdk_payload` with the `p["id"] != "100"` assertion dropped. Resolving.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
gbrlcustodio: Following up: this one is still open. It is a non-blocking suggestion, so no change is required to merge, but could you either extract the shared plan step or leave a note here that the duplicated branches are intentional? Then we can close it out.

adriannoes: Deferring this one deliberately: the duplicated branches stay as-is in this PR. The elicit-or-filter shape is the most delicate interactive surface in the server, and collapsing the two branches into a pure plan_create_card_fields planner deserves its own focused PR with manual Cursor MCP smoke validation rather than riding along here. Tracked in #290 with your proposed shape as the starting point. Feel free to resolve.

packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -61,6 +49,24 @@ def _parse_fields_json(raw: str | None) -> dict[str, Any] | list[dict[str, Any]]
     raise typer.BadParameter("--fields must be a JSON object or array")
 
 
+def _apply_create_card_title_warning(
+    result: dict[str, Any], *, requested_title: str | None
+) -> dict[str, Any]:
+    if not requested_title:
+        return result
+    card_node = result.get("createCard", {}).get("card")
```
Danielmoraisg: minor: `result.get("createCard", {}).get("card")` raises `AttributeError` when the response is `{"createCard": null}` (GraphQL null on partial failure). The code this replaces was defensive: `(result.get("createCard") or {}).get("card") or {}`. The `isinstance` guard on the next line runs too late to help. Restore the `or {}` guard.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -121,48 +131,122 @@ async def create_card(
...
+                try:
+                    form_fields = await client.get_start_form_fields(
+                        pipe_id, required_fields_only
+                    )
+                except MalformedFieldDefinitionError as exc:
+                    return tool_error(str(exc))
+
+                expected_fields = _filter_editable_field_definitions(
+                    form_fields.get("start_form_fields", [])
+                )
```
Danielmoraisg, line 218: minor: the phase branch (above) and this non-phase branch each duplicate the `get_start_form_fields` fetch plus an identical `try/except MalformedFieldDefinitionError / UserCancelledError` elicitation block, and the handler closure is now about 120 lines of branching. The pure merge step was extracted, but the decision of which defs to fetch, whether to elicit, and what to filter could be one shared path (fetch start-form once, fetch phase defs conditionally, single elicitation block), keeping the tool a thin shell. Refactor candidate, not blocking.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -93,26 +94,35 @@ async def create_card(
...
+            With ``phase_id``, the card is created in that phase (including orphan
+            phases that are not the start form). Discover workflow phase IDs via
+            ``get_pipe(pipe_id).phases[].id``. For agent seeding in a specific phase, set
+            ``phase_id`` and ``skip_elicitation=true``. When ``fields`` is non-empty,
+            keys are filtered against ``get_phase_fields(phase_id)`` and
+            ``get_start_form_fields(pipe_id)`` so pipes that still require start-form
+            values on ``CreateCardInput`` receive them alongside phase fields.
+
+            Before ``move_card_to_phase``, call ``get_phase_allowed_move_targets`` on
+            the card's current phase; transition rules are UI-only (Pipe settings ->
```
Danielmoraisg: nit: this paragraph exists to satisfy the discovery-hints audit, but as written it is about a different tool (`move_card_to_phase`), and "transition rules are UI-only (Pipe settings -> Phase -> Connections)" is cryptic in the `create_card` docstring. Anchor it to the create-then-move workflow, e.g. "If you plan to move the card afterwards, call `get_phase_allowed_move_targets` first."

Danielmoraisg (approved):

packages/cli/src/pipefy_cli/commands/card.py
adriannoes: Fixed in 598f553 — restored `(result.get("createCard") or {}).get("card")` and mirrored the same guard in the MCP `create_card` path. Added regression tests for `{"createCard": null}` in CLI and MCP.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
adriannoes: Agreed this is the right direction — deferring to #290 (same extraction as gbrlcustodio's thread). Keeping the duplicated branches in this PR to avoid mixing refactor with the phase_id feature surface. Non-blocking; resolving.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
adriannoes: Fixed in 598f553 — rewrote the paragraph to anchor on the create-then-move workflow instead of the cryptic UI-only transition rules note.

PR #284, opened 2026-06-03
feat(mcp,cli): report filter preflight and CLI label hex

adriannoes:

**Depends on:** https://github.com/pipefy/ai-toolkit/pull/283

---

## Objective

Close remaining **boundary** gaps from chaos-pipe:

- Run `validate_report_cards_filter` on **all** MCP/CLI paths that accept `filter` (create/update/export pipe + org reports) — F6
- CLI `label create/update` uses `normalize_label_color` (F7)

## Improvements

- Top-level `current_phase` mistake caught before GraphQL with a clear message.
- `#F00` works like the live API; `red` rejected locally.

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/_common.py
+++ b/packages/cli/src/pipefy_cli/commands/_common.py
@@ -13,7 +13,11 @@ import typer
 from gql.transport.exceptions import TransportError, TransportQueryError
 from pipefy_sdk import PipefyClient, PipefySettings, stream_bytes
 from pipefy_sdk.exceptions import PipefyError
-from pipefy_sdk.report_filter_preflight import validate_report_cards_filter
+from pipefy_sdk.label_color import normalize_label_color
+from pipefy_sdk.report_filter_preflight import (
+    normalize_report_cards_filter,
+    validate_report_cards_filter,
+)
 
 from pipefy_cli.auth import (
     AuthContext,
@@ -80,10 +84,30 @@ def validate_cards_page_size(first: int | None) -> int | None:
     return first
 
 
-def validate_report_filter_cli(filter_obj: dict[str, Any] | None) -> None:
-    message = validate_report_cards_filter(filter_obj)
+def validate_report_filter_cli(filter_obj: dict[str, Any] | None) -> dict[str, Any] | None:
+    if filter_obj is None:
+        return None
+    normalized = normalize_report_cards_filter(filter_obj)
+    message = validate_report_cards_filter(normalized)
     if message is not None:
         raise typer.BadParameter(message)
+    return normalized
+
+
+def validate_label_name_cli(name: str) -> str:
+    nm = name.strip()
+    if not nm:
+        typer.echo("--name must be non-empty.", err=True)
+        raise typer.Exit(2)
+    return nm
+
+
+def validate_label_color_cli(color: str) -> str:
+    try:
+        return normalize_label_color(color)
+    except ValueError as exc:
+        typer.echo(str(exc), err=True)
+        raise typer.Exit(2) from exc
 
 
 def export_poll_max_rounds(
--- a/packages/cli/src/pipefy_cli/commands/_common.py
+++ b/packages/cli/src/pipefy_cli/commands/_common.py
@@ -99,8 +99,7 @@ def validate_report_filter_cli(
 def validate_label_name_cli(name: str) -> str:
     nm = name.strip()
     if not nm:
-        typer.echo("--name must be non-empty.", err=True)
-        raise typer.Exit(2)
+        raise typer.BadParameter("--name must be non-empty.")
     return nm
 
 
@@ -108,8 +107,7 @@ def validate_label_color_cli(color: str) -> str:
     try:
         return normalize_label_color(color)
     except ValueError as exc:
-        typer.echo(str(exc), err=True)
-        raise typer.Exit(2) from exc
+        raise typer.BadParameter(str(exc)) from exc
 
 
 def export_poll_max_rounds(
--- a/packages/cli/src/pipefy_cli/commands/_common.py
+++ b/packages/cli/src/pipefy_cli/commands/_common.py
@@ -84,7 +84,9 @@ def validate_cards_page_size(first: int | None) -> int | None:
     return first
 
 
-def validate_report_filter_cli(filter_obj: dict[str, Any] | None) -> dict[str, Any] | None:
+def validate_report_filter_cli(
+    filter_obj: dict[str, Any] | None,
+) -> dict[str, Any] | None:
     if filter_obj is None:
         return None
     normalized = normalize_report_cards_filter(filter_obj)
--- a/packages/cli/src/pipefy_cli/commands/label.py
+++ b/packages/cli/src/pipefy_cli/commands/label.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 
 import typer
 from pipefy_sdk import PipefyClient
-from pipefy_sdk.label_color import normalize_label_color
 
 from pipefy_cli.commands._common import (
     ID_POSITIONAL_CONTEXT_SETTINGS,
     confirm_destructive,
     resource_id_argument,
     run_cli_command,
+    validate_label_color_cli,
+    validate_label_name_cli,
 )
 
 label_app = typer.Typer(help="Pipe label operations.", no_args_is_help=True)
@@ -56,18 +57,8 @@ def label_create(
 ) -> None:
     """Create a label on a pipe."""
 
-    nm = name.strip()
-    if not nm:
-        typer.echo("--name must be non-empty.", err=True)
-        raise typer.Exit(2)
-    if not color.strip():
-        typer.echo("--color must be non-empty.", err=True)
-        raise typer.Exit(2)
-    try:
-        col = normalize_label_color(color)
-    except ValueError as exc:
-        typer.echo(str(exc), err=True)
-        raise typer.Exit(2) from exc
+    nm = validate_label_name_cli(name)
+    col = validate_label_color_cli(color)
 
     async def factory(client: PipefyClient):
         return await client.create_label(pipe_id, nm, col)
@@ -92,18 +83,8 @@ def label_update(
 ) -> None:
     """Update a label (name and color are both required by Pipefy)."""
 
-    nm = name.strip()
-    if not nm:
-        typer.echo("--name must be non-empty.", err=True)
-        raise typer.Exit(2)
-    if not color.strip():
-        typer.echo("--color must be non-empty.", err=True)
-        raise typer.Exit(2)
-    try:
-        col = normalize_label_color(color)
-    except ValueError as exc:
-        typer.echo(str(exc), err=True)
-        raise typer.Exit(2) from exc
+    nm = validate_label_name_cli(name)
+    col = validate_label_color_cli(color)
 
     async def factory(client: PipefyClient):
         return await client.update_label(label_id, name=nm, color=col)
--- a/packages/cli/src/pipefy_cli/commands/label.py
+++ b/packages/cli/src/pipefy_cli/commands/label.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import typer
 from pipefy_sdk import PipefyClient
+from pipefy_sdk.label_color import normalize_label_color
 
 from pipefy_cli.commands._common import (
     ID_POSITIONAL_CONTEXT_SETTINGS,
@@ -46,17 +47,27 @@ def label_create(
     pipe_id: str = typer.Option(..., "--pipe", help="Pipe id."),
     name: str = typer.Option(..., "--name", "-n", help="Label name."),
     color: str = typer.Option(
-        ..., "--color", "-c", help="Label color (per Pipefy API)."
+        ...,
+        "--color",
+        "-c",
+        help="Label color as hex #RRGGBB (e.g. #E50000, #FF0000).",
     ),
     json_out: bool = typer.Option(False, "--json", "-j"),
 ) -> None:
     """Create a label on a pipe."""
 
     nm = name.strip()
-    col = color.strip()
-    if not nm or not col:
-        typer.echo("--name and --color must be non-empty.", err=True)
+    if not nm:
+        typer.echo("--name must be non-empty.", err=True)
         raise typer.Exit(2)
+    if not color.strip():
+        typer.echo("--color must be non-empty.", err=True)
+        raise typer.Exit(2)
+    try:
+        col = normalize_label_color(color)
+    except ValueError as exc:
+        typer.echo(str(exc), err=True)
+        raise typer.Exit(2) from exc
 
     async def factory(client: PipefyClient):
         return await client.create_label(pipe_id, nm, col)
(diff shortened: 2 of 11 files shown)


packages/mcp/src/pipefy_mcp/tools/report_tools.py
```diff
@@ -390,6 +404,9 @@ async def create_pipe_report(
             if err is not None:
                 return err
             err = _blank_field_error(name, "name")
+            if err is not None:
+                return err
+            err = _report_filter_preflight_error(filter)
```
gbrlcustodio: After this PR the preflight is the sole gate on all six report create/update/export paths; before it, the SDK forwarded `filter` to GraphQL whenever it was not None (`report_service.create_pipe_report` / `update_pipe_report` use `if filter is not None`). `validate_report_cards_filter` enforces a strict positive `ReportCardsFilter` schema, so any shape it does not model is now rejected client-side. Worth verifying against the API that the validator is not stricter than what GraphQL accepts:

- Numeric `value`: a query `{"field": "current_phase", "operator": "eq", "value": 987654321}` (phase id as a JSON number) is rejected with "value must be a string when set, received int." This contradicts the repo's own convention that numeric JSON values are coerced to strings to match the API (`docs/mcp/tools/pipes-and-cards.md:18`). Coercing the id here would match the rest of the tool surface; hard-rejecting surprises agents that pass a bare number.
- Empty `{}`: previously forwarded to GraphQL; now rejected with "filter.operator is required." If `{}` is how a caller clears a report filter on update, that path no longer reaches the API.
- List values and extra keys: the validator allows only a string `value` and a fixed key set ({field, operator, value, type, label, id}). Confirm no `ReportCardsFilter` operator takes a list value and no supported query key falls outside that set; otherwise valid filters are blocked.

The validator itself lives in `report_filter_preflight.py` (added in #280, unchanged here), so the fix may belong there, but the behavior activates with this PR's wiring. Two directions: align the validator to the API (coerce numeric ids, accept the value/key shapes GraphQL takes), or keep the preflight scoped to the specific naive mistake it targets (the rejected top-level keys like `current_phase`) rather than imposing a full positive schema on a previously-passthrough input.

packages/mcp/src/pipefy_mcp/tools/report_tools.py
```diff
@@ -34,6 +35,15 @@ def _blank_field_error(value: str, field: str) -> dict[str, Any] | None:
     return None
 
 
+def _report_filter_preflight_error(
```
gbrlcustodio: The PR's goal is to run filter validation on every path that accepts `filter`, but the check is wired at 12 hand-edited call sites (6 MCP tools here, 6 CLI commands in the same PR). All twelve converge on the SDK report service methods (`create_pipe_report`, `update_pipe_report`, `create_organization_report`, `update_organization_report`, `export_pipe_report`, `export_organization_report`), which forward the filter to GraphQL with no validation. A 7th report path, a new tool/command, or any direct SDK consumer (a script or notebook) must remember to import and call the preflight or it silently regresses to the exact naive-`current_phase` bug this PR exists to prevent. An invariant that must hold on all paths cannot be enforced by reviewing a leaf.

`normalize_label_color` has the same shape: it is now applied in MCP `pipe_config_tools` and CLI `label.py`, but the SDK `create_label`/`update_label` still forward `color` untouched. Pushing both the filter preflight and the color normalization down to the SDK service boundary would make the invariant hold by construction for every caller, and let the per-site guards and the duplicated normalization be deleted rather than maintained and re-added per path.

packages/cli/src/pipefy_cli/commands/label.py
```diff
@@ -46,17 +47,27 @@ def label_create(
...
     """Create a label on a pipe."""
 
     nm = name.strip()
-    col = color.strip()
-    if not nm or not col:
-        typer.echo("--name and --color must be non-empty.", err=True)
+    if not nm:
+        typer.echo("--name must be non-empty.", err=True)
         raise typer.Exit(2)
+    if not color.strip():
```
gbrlcustodio: The `if not color.strip(): ... raise typer.Exit(2)` guard is redundant. `normalize_label_color` already strips and raises `ValueError` on empty or whitespace input, and the following `try/except ValueError` converts that to `Exit(2)`. So the dedicated empty-color check only changes the message wording for the empty case and is otherwise unreachable as a distinct outcome.

The full name-check plus color-check plus normalize ladder is also copy-pasted verbatim in `label_update` (lines 95-103). A shared helper in `_common.py` (where the `BadParameter`-based validators already live) would collapse both copies and remove the redundant guard, so color rules and exit codes cannot drift between the two commands.

packages/mcp/src/pipefy_mcp/tools/report_tools.py
```diff
@@ -378,6 +388,10 @@ async def create_pipe_report(
         ) -> dict[str, Any]:
             """Create a pipe report. Use column `name` in `fields`; filter field names from `get_pipe_report_filterable_fields`.
 
+            ``filter`` must use ReportCardsFilter shape (``operator`` + ``queries``), not a
```
gbrlcustodio, line 397: The "filter must use ReportCardsFilter shape, not a top-level `current_phase` key" guidance was added to only `create_pipe_report` and `update_pipe_report`. The other four filter-accepting tools (`create_organization_report`, `update_organization_report`, `export_pipe_report`, `export_organization_report`) got only the generic "ReportCardsFilter shape" line. An agent calling one of those four sees no warning against the naive `{current_phase: [...]}` shape, builds the wrong filter, and hits the runtime preflight rejection the docstring on the sibling tools would have steered it away from. Same invariant, inconsistent docs across the paths that enforce it.

gbrlcustodio (commented):
One correctness risk to verify against the API, one altitude note, one cleanup, one docs gap. The filter-narrowing items are plausible (the validator is stricter than the previously-passthrough path), not confirmed regressions: the repo has no ReportCardsFilter schema to check against.

adriannoes:
Thanks @gbrlcustodio — addressed in `293a18f`: integer `filter.queries[].value` scalars coerce to strings before preflight; shared label name/color CLI validators; consistent ReportCardsFilter shape docstrings on all six filter paths. SDK-boundary enforcement deferred to a follow-up (consistent with boundary-only validation in #289).

gbrlcustodio:
## Consolidating the four filter and label threads

The four open threads on this PR are one structural issue, not four. Filter and color validation is a decision (is this input valid, and what is its normalized form?) wired at twelve effect sites: the six MCP report tools and the six CLI commands. Every one of those sites forwards through a single SDK layer: `report_service` (`create_pipe_report`, `update_pipe_report`, `create_organization_report`, `update_organization_report`, `export_pipe_report`, `export_organization_report`) and `pipe_config_service` (`create_label`, `update_label`). The validation sits one layer too high.

### The move

Push normalize and validate down to those SDK methods, and let each surface only shape the error.

```python
# report_service.py: one decision point that all six methods call
def _prepare_filter(filter_obj):
    if filter_obj is None:
        return None
    normalized = normalize_report_cards_filter(filter_obj)
    msg = validate_report_cards_filter(normalized)
    if msg is not None:
        raise ValueError(msg)  # or a small typed subclass for precise catches
    return normalized
# each method: input_obj["filter"] = _prepare_filter(filter)

# pipe_config_service.py: normalize at the funnel
async def create_label(self, pipe_id, name, color):
    color = normalize_label_color(color)  # already raises ValueError on bad input
    ...
async def update_label(self, label_id, **attrs):
    if "color" in attrs:
        attrs["color"] = normalize_label_color(attrs["color"])
```

Then delete the twelve leaf guards. MCP catches the error and returns `build_report_error_payload`; the CLI catches it and exits non-zero. The color path already raises `ValueError`, which both surfaces already map.

### What this resolves

- All-paths invariant (the wiring thread): holds by construction. A seventh report path, a new tool, or a direct SDK consumer is covered without remembering to call a helper.
- Label duplication and the redundant empty-color guard: gone. No per-surface color helper, no copy-paste between create and update.
- The `current_phase` docstring guidance: stops being load-bearing. The central validator returns one clear, shape-pointing message, so a per-tool docstring line becomes optional rather than something repeated on six tools (and re-broken, see below).
- The numeric-versus-API strictness question: not answered by this move, but localized. It collapses from a property of twelve sites to one seam next to the GraphQL call, which is where to verify it and relax it.

### Trade-offs

- The SDK methods would raise on invalid input instead of forwarding. That is the more correct boundary and keeps the decision inside the owned abstraction, but it is a contract change and the SDK tests move with it.
- CLI error timing shifts from parse-time (`typer.BadParameter` in a parameter validator) to call-time inside the command body. Same non-zero exit and message; a thin catch can preserve the `BadParameter` feel if that matters.
- Larger diff than the current fix commit, but it nets less code: one decision point replacing twelve, with the leaf guards and duplicated normalization deleted.

### Needs fixing regardless of the above

The docstring change in 293a18f introduced a regression. The four newly-documented tools use f-string docstrings (`f"""..."""`). Python does not treat an f-string as a docstring, so `__doc__` is `None` for `create_organization_report`, `update_organization_report`, `export_pipe_report`, and `export_organization_report`, and their MCP descriptions are now empty (confirmed with `ast.get_docstring(...) == None`). A variable cannot go into a docstring at all; even `"text" + VAR` yields `None`. Inline the guidance as a plain string literal the way `create_pipe_report` and `update_pipe_report` already do. Tracked on that thread.

packages/mcp/src/pipefy_mcp/tools/report_tools.py
gbrlcustodio: Numeric `value` coercion is addressed in 293a18f (`normalize_report_cards_filter`, normalize-then-validate). The remaining question, whether the validator is stricter than the API on empty `{}` and list values, is folded into the consolidated recommendation: it becomes a single SDK-level seam next to the GraphQL call rather than a property of twelve sites. Resolving in favor of https://github.com/pipefy/ai-toolkit/pull/284#issuecomment-4672322470.

packages/mcp/src/pipefy_mcp/tools/report_tools.py
gbrlcustodio: This is the core of the consolidated write-up just posted, restated and expanded there with the concrete SDK-boundary sketch and trade-offs. Not addressed in code yet; tracking it in https://github.com/pipefy/ai-toolkit/pull/284#issuecomment-4672322470. Resolving this thread in favor of that.

packages/cli/src/pipefy_cli/commands/label.py
gbrlcustodio: Addressed in 293a18f: `validate_label_name_cli` / `validate_label_color_cli` extracted to `_common.py` and used by both commands, and the redundant `if not color.strip()` guard is gone. The consolidated note (https://github.com/pipefy/ai-toolkit/pull/284#issuecomment-4672322470) would push color normalization one layer further down into the SDK, which removes even these CLI helpers, but the duplication this thread raised is resolved. Resolving.

packages/mcp/src/pipefy_mcp/tools/report_tools.py
gbrlcustodio: Partly actioned, but it introduced a regression. The four added docstrings are f-strings (`f"""..."""`), and Python does not treat an f-string as a docstring, so `__doc__` is None for `create_organization_report`, `update_organization_report`, `export_pipe_report`, and `export_organization_report` (confirmed via `ast.get_docstring(...) == None`). Their MCP descriptions are now empty, and the interpolated guidance is silently discarded. A variable cannot go into a docstring at all (even `"text" + VAR` yields None); inline the guidance as a plain literal the way `create_pipe_report` / `update_pipe_report` do. Leaving this open. The consolidated note (https://github.com/pipefy/ai-toolkit/pull/284#issuecomment-4672322470) explains why centralizing at the SDK removes the need to hand-edit these six docstrings in the first place.

adriannoes: Fixed in b7dbc4e: the four f-string docstrings are now plain literals with the ReportCardsFilter guidance inlined (matching create_pipe_report / update_pipe_report), and the now-unused _REPORT_FILTER_SHAPE_DOC constant is removed. Added regression tests asserting every report tool exposes a non-empty MCP description and that all six filter paths carry the current_phase warning. The SDK-boundary consolidation is being implemented as a stacked follow-up PR on this series - reply on the consolidated note coming with the PR link.

adriannoes:
Thanks @gbrlcustodio — you convinced us. The consolidated move is implemented in #291 (PR 7 of this stack, stacked on #285): `ReportService` now normalizes the filter and raises `ValueError` via a single `_prepared_report_cards_filter` decision point wired into all six methods, `PipeConfigService.create_label`/`update_label` normalize the hex color, and the twelve leaf guards are deleted. MCP maps `ValueError` to the same `build_report_error_payload` envelope; the CLI maps it to exit code 2 (`typer.BadParameter` on report commands), so messages and exit codes are unchanged. The #280 naming finding rides along there (`validate_report_cards_filter` -> `report_cards_filter_error`).

The f-string docstring regression was fixed separately on this PR in b7dbc4e (plain literals + regression tests asserting non-empty descriptions and the `current_phase` guidance on all six filter paths) — per your note, with the central validator in place that guidance is now informative rather than load-bearing.

packages/cli/src/pipefy_cli/commands/_common.py
```diff
@@ -78,10 +82,32 @@ def validate_cards_page_size(first: int) -> int:
...
+
+def validate_label_name_cli(name: str) -> str:
+    nm = name.strip()
+    if not nm:
+        typer.echo("--name must be non-empty.", err=True)
+        raise typer.Exit(2)
+    return nm
+
+
+def validate_label_color_cli(color: str) -> str:
```
Danielmoraisg, line 106: nit: this helper (and `validate_label_name_cli`) echoes raw text and `raise typer.Exit(2)`, while `validate_report_filter_cli` at line 85 raises `typer.BadParameter` (rendered with usage text). Both exit 2 and each follows the precedent of the code it replaced, but a user who hits both gets two different error formats. Worth unifying when convenient.

Danielmoraisg (approved):

packages/cli/src/pipefy_cli/commands/_common.py
adriannoes: Addressed in 26c86b3: `validate_label_name_cli` and `validate_label_color_cli` now raise `typer.BadParameter` like `validate_report_filter_cli`, so all three share the same error format (exit 2 with usage context).

adriannoes:
Rebased onto latest `feat/card-phase-erg/4-mcp-cli-create-card-phase` (skipped duplicate create_card commit) and pushed **26c86b3** addressing @Danielmoraisg's nit: label name/color validators now use `typer.BadParameter` like `validate_report_filter_cli`.

**Tests run locally:**
- `uv run pytest packages/cli/tests/test_label_webhook_commands.py packages/cli/tests/test_report_pipe_filter_commands.py packages/mcp/tests/tools/test_report_tools.py packages/sdk/tests/test_report_filter_preflight.py packages/sdk/tests/test_label_color.py -m "not integration"` — **94 passed**
- Full suite: 3065 passed; 6 failures are pre-existing stack drift (parity/docs rows for phase tools + CLI help golden from stacked PRs 6+), unrelated to this PR's scope.

Ready for manual merge when CI is green.

PR #279, opened 2026-06-03
chore(140): point plugin manifest ownership at pipefy org

gbrlcustodio:
## What

Repoint the two ownership fields in the plugin manifests at the Pipefy org and round out the marketplace plugin entry.

- `marketplace.json` `owner`: was `{ name: "Gabriel Custodio", url: https://github.com/gbrlcustodio }`, now `{ name: "Pipefy", url: https://github.com/pipefy }`.
- `plugin.json` `author`: set name to `Pipefy` and add `url: https://github.com/pipefy`, mirroring the marketplace owner.
- `marketplace.json` plugin entry: add `homepage` and `repository` (`https://github.com/pipefy/ai-toolkit`), matching the values already in `plugin.json`.

## Why

The repo URL repoint in #140 rewrote owner-path URLs but its scope was "URL paths only," so it left these ownership fields on the personal account. After the org transfer the marketplace owner is Pipefy, and a half-migrated `owner` (personal name + org URL) reads as inconsistent. `owner.name` and `owner.url` now describe the same entity.

Note the deliberate distinction: `owner`/`author` now attribute ownership to the org, which is correct post-transfer. Distribution package names are untouched, consistent with #140.

## Scope

Manifest fields only. The untracked `skills-lock.json` and the historical `CHANGELOG.md` references are left as-is.

Issue #140, opened 2026-05-19
Rename to `ai-toolkit` and transfer to the Pipefy GitHub org


gbrlcustodio:
Tracking issue for moving this repo from `gbrlcustodio/pipefy-mcp-server` to `pipefy/ai-toolkit` (org handle TBD). Rename + transfer should land in the same window to avoid two URL changes for consumers.

### Pre-transfer
- [ ] Confirm target org handle and the owner who will accept the transfer
- [ ] Sign-off from Pipefy engineering / platform leadership
- [ ] Legal / brand review — README disclaimer ("not Pipefy's official integration") needs to come out; license headers updated to reflect org ownership

### Transfer mechanics
- [ ] Rename `pipefy-mcp-server` → `ai-toolkit`
- [ ] Transfer ownership to the Pipefy org
- [ ] Reconfigure branch protection rules (not all settings carry over on transfer)
- [ ] Re-add Actions secrets / environment variables (don't transfer)
- [ ] Verify webhooks still resolve

### Docs / code sweep
- [ ] README CI badge URL (currently `gbrlcustodio/pipefy-mcp-server`)
- [ ] README `<owner>/pipefy-labs` placeholders in install commands (4 spots) and `feedback & issues` link
- [ ] README release link (`/releases/tag/v0.1.0-beta.1`)
- [ ] Repo description on GitHub
- [ ] `docs/` internal references
- [ ] `pipefy-labs` umbrella naming in README — keep, replace with "Pipefy AI Toolkit", or drop?
- [ ] Grep for any remaining hardcoded `gbrlcustodio/pipefy-mcp-server` references

### Communication
- [ ] Notify open contributors before transfer (existing PRs follow automatically, but local remotes need refetching)
- [ ] CHANGELOG / release notes entry
- [ ] Update external surfaces (MCP registry listings, future PyPI project URLs, blog posts)

### Open questions
- **Repo rename ≠ package rename.** The MCP package is `pipefy-mcp-server` and the CLI is `pipefy-cli` — both intentional consumer-facing names. Should they stay (recommended — renaming the package breaks every existing install script) or move to something like `pipefy-ai-toolkit-mcp`?
- Coincide with v1.0 + PyPI cutover, or do them as separate events?

gbrlcustodio:
### Pre-staged rename PR + transfer sequence

[...]

PR #287, opened 2026-06-03
release: v0.2.0-beta.3

### LINKED ISSUE #272: AI automation create requires service-account credentials, but generate_with_ai is a public automation action

gbrlcustodio:
Promote `dev` to `main` to cut **v0.2.0-beta.3**.

After merge, the `v0.2.0-beta.3` and `latest` tags are placed on the resulting merge commit on `main`, and the Release workflow builds and attaches the five wheels (`pipefy_cli`, `pipefy_mcp_server`, `pipefy_sdk`, `pipefy_auth`, `pipefy_infra`).

## Contents since v0.2.0-beta.2
- **fix #272 (#274)**: AI automation create/update no longer require service-account credentials; routed through the public `createAutomation`/`updateAutomation`. Includes the SDK surface move `AiAutomationService` → `AutomationService`.
- **chore #140 (#279)**: plugin manifest ownership repointed to the Pipefy org (repo metadata).
- **chore**: release prep (version bump + CHANGELOG) in #286.

See `CHANGELOG.md` `[0.2.0-beta.3]` for the user-facing summary.


Danielmoraisg (approved):

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
```diff
@@ -402,8 +354,12 @@ async def create_ai_automation(
             try:
                 result = await client.create_ai_automation(validated)
             except Exception as exc:  # noqa: BLE001
-                return _ai_automation_api_failure_payload(
-                    exc, fallback=AI_AUTOMATION_CREATE_FAILED, debug=debug
+                return await handle_automation_tool_graphql_error(
```
adriannoes, line 357: When create/update started using `handle_automation_tool_graphql_error`, we lost the old stable fallback for failures whose text is only `[code=…]` / `[correlation_id=…]` markers. The handler already falls back to `"Automation request failed."` when there are no extracted strings, but after `strip_internal_api_diagnostic_markers` the joined message can still end up empty. A small guard (`if not base.strip(): base = "Automation request failed."`) in `automation_tool_helpers.py` would match what `enrich_behavior_error` does for the same shape — not a merge blocker for this release. Easy to confirm with a unit test on the helper using `ValueError(" [code=X] [correlation_id=Y]")`.

adriannoes (approved):
## Summary

LGTM for promoting `dev` to `main` as **v0.2.0-beta.3**. The #272 fix correctly routes AI automation create/update through public `createAutomation` / `updateAutomation`, removes the false service-account gate, and keeps MCP/CLI thin over `AutomationService`. I checked out head `5986cea` locally: `ruff check .` passed and **2498** unit tests passed (`pytest -m "not integration"`). CI is green.

## Also noted

- **F1 (follow-up):** See inline on `create_ai_automation` — optional guard in `handle_automation_tool_graphql_error` when marker-stripping leaves an empty message (unlikely on the public GraphQL path).

## Verification

```bash
uv run ruff check .
uv run pytest -m "not integration" -q
```

Internal review notes: `.cursor/dev-planning/code-review/pr-287-release-v0-2-0-beta-3.md` (pipefy-mcp-server workspace).

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
gbrlcustodio: Tracked in #288.

PR #285, opened 2026-06-03
docs: card-phase ergonomics parity, skills, and README

adriannoes:
Analytics pipe seeding (`306996636`) required **`execute_graphql`** for phase inventory and phase-local creates. **Spec:** `.cursor/dev-planning/specs/archive/card-phase-ergonomics/` — **Stack** after [#277](https://github.com/pipefy/ai-toolkit/pull/277) (closed).

**Depends on:** https://github.com/pipefy/ai-toolkit/pull/284

---

## Objective

Document the **tool-first** agent path so future runs skip GraphQL:

- `docs/parity.md`, MCP pipes-and-cards + reports reference
- `pipefy-pipes-and-cards` skill: *Seed pipe across phases* workflow
- README tool count (152), CHANGELOG, marketplace `ai-toolkit` metadata

## Improvements

- Skills/CI lint align with shipped tool names.
- Agents routed away from `execute_graphql` for these flows.

PR #291, opened 2026-06-10
refactor(sdk,mcp,cli): enforce report filter and label color validation at SDK boundary

adriannoes:
## Summary

- Implements the consolidated review proposal from #284 (https://github.com/pipefy/ai-toolkit/pull/284#issuecomment-4672322470): ReportCardsFilter preflight and label color normalization move from twelve effect sites (six MCP report tools + six CLI commands) into the SDK funnel they all forward through.
- `ReportService` create/update/export methods now normalize the filter and raise `ValueError` with the planner message; `PipeConfigService.create_label`/`update_label` normalize the hex color. The twelve leaf guards are deleted; MCP maps `ValueError` to the same `build_report_error_payload` envelope and the CLI maps it to exit code 2 (`typer.BadParameter` on report commands), so messages and exit codes are unchanged.
- Also lands the #280 naming finding: `validate_report_cards_filter` -> `report_cards_filter_error`, matching its `str | None` errors-as-values return.

PR 7 of the card-phase stack, stacked on #285. 

The diff exceeds the usual 10-file guideline because it is one structural move (one decision point replacing twelve guards, with relocated test coverage); it nets less code at the surfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - **MCP / CLI (label color)**: `create_label` and `update_label` validate `color` as hex `#RGB` or `#RRGGBB` before GraphQL (e.g. `red` is rejected with `expected #RGB or #RRGGBB hex color, received 'red'`).
+- **SDK**: report `filter` preflight and label `color` normalization now run at the SDK boundary — `ReportService` create/update/export methods and `create_label`/`update_label` normalize input and raise `ValueError` when invalid (one decision point replacing per-surface guards). MCP tools and CLI commands only map the error; messages and exit codes are unchanged.
 - **Docs / install**: canonical GitHub repository is [`pipefy/ai-toolkit`](https://github.com/pipefy/ai-toolkit). Install snippets, `install.sh`, MCP setup links, Claude plugin metadata, and `.mcp.json` now point at `github.com/pipefy/ai-toolkit` (GitHub may still redirect older URLs).
 
 ## [0.2.0-beta.2] - 2026-06-02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **MCP / CLI**: `get_phase_allowed_move_targets` lists valid destination phases for `move_card_to_phase` (GraphQL `phase.cards_can_be_moved_to_phases`). MCP returns normalized `{phase_id, phase_name, allowed_phases}`; CLI: `pipefy phase targets <phase_id>`.
 - **SDK / MCP / CLI**: `get_phase_cards_count` reads native `Phase.cards_count` (MCP/CLI return `{phase_id, phase_name, cards_count}` via `get_phase`). `get_phase_cards` lists cards in a phase via `Phase.cards` pagination (`first` default 50, `after`, optional fields). CLI: `pipefy phase count <phase_id>`, `pipefy phase cards <phase_id> --first --after`.
 - **SDK / MCP / CLI**: `get_pipe` selects `cards_count` on each workflow phase in `phases[]` (start-form intake stays on `start_form_fields`).
-- **SDK / MCP / CLI**: report create/update/export paths preflight `filter` as nested `ReportCardsFilter` (`operator` + `queries`); reject naive top-level keys such as `current_phase` before GraphQL; coerce integer `value` scalars to strings.
+- **SDK / MCP / CLI**: report create/update/export paths preflight `filter` as nested `ReportCardsFilter` (`operator` + `queries`) in the SDK before GraphQL; reject naive top-level keys such as `current_phase`; coerce integer `value` scalars to strings. MCP and CLI surface the same error; CLI rejects invalid filters before auth.
 
 ### Changed
 
-- **MCP / CLI (label color)**: `create_label` and `update_label` validate `color` as hex `#RGB` or `#RRGGBB` before GraphQL (e.g. `red` is rejected with `expected #RGB or #RRGGBB hex color, received 'red'`).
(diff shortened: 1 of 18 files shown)


packages/cli/src/pipefy_cli/commands/report_pipe.py
```diff
@@ -139,18 +139,20 @@ def report_pipe_create(
...
         raise typer.BadParameter("--formulas must be a JSON array")
 
     async def factory(client: PipefyClient):
-        return await client.create_pipe_report(
-            pipe,
-            name.strip(),
-            fields=fields_list,
-            filter=filt,
-            formulas=formulas_val,
+        return await bad_parameter_on_value_error(
```
Danielmoraisg: minor: validation timing moved from option-parse time to after auth resolution. `bad_parameter_on_value_error` wraps the client call inside `factory`, which runs after `get_authenticated_client`; that function exits 2 on missing auth and eagerly refreshes stored sessions over the network. Before this PR, an invalid `--filter` was rejected before any auth work. Now an unauthenticated user with a bad filter sees the auth error instead, and authenticated users pay a token refresh before a purely local rejection. Same pattern in `report_org.py` and `label.py`. The PR body says messages and exit codes are unchanged, but this is a real behavior change for the unauthenticated/bad-input case.

packages/cli/tests/test_label_webhook_commands.py
```diff
@@ -50,7 +53,7 @@ def test_label_create_rejects_color_name_before_api(
         )
     assert result.exit_code == 2
     assert "expected #RGB or #RRGGBB hex color, received 'red'" in result.stderr
-    mock_client.create_label.assert_not_called()
+    mock_client.create_label.assert_awaited_once()
 
 
 def test_label_create_passes_normalized_hex(
```
Danielmoraisg: nit: test names are now stale after the move. `test_label_create_passes_normalized_hex` now asserts the CLI forwards the raw lowercase `"#ff0000"`, the opposite of its name (normalization happens in the SDK, which is stubbed here). The `..._rejects_color_name_before_api` cases (lines 87 here, and the MCP `..._before_graphql` test) now stub the client to raise and assert it was awaited, so they test error mapping, not rejection before the API. Rename them to match what they verify.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - **MCP / CLI (label color)**: `create_label` and `update_label` validate `color` as hex `#RGB` or `#RRGGBB` before GraphQL (e.g. `red` is rejected with `expected #RGB or #RRGGBB hex color, received 'red'`).
+- **SDK**: report `filter` preflight and label `color` normalization now run at the SDK boundary — `ReportService` create/update/export methods and `create_label`/`update_label` normalize input and raise `ValueError` when invalid (one decision point replacing per-surface guards). MCP tools and CLI commands only map the error; messages and exit codes are unchanged.
```
Danielmoraisg: nit: this Changed entry describes a refactor of behavior introduced by the entry directly above it, both under Unreleased; no released user ever saw the per-surface placement, and the entry itself says "messages and exit codes are unchanged" (i.e. no observable change). Amend the existing unreleased entries instead of recording an internal move. The line also has an em dash ("at the SDK boundary — `ReportService`"); replace with plain punctuation.

docs/mcp/tools/reports.md
```diff
@@ -5,7 +5,7 @@ Pipe reports and organization reports: discovery, CRUD, single pipe report fetch
 ## Cross-cutting patterns
 
 - Build `ReportCardsFilter` using `get_pipe_report_columns` and `get_pipe_report_filterable_fields`; use `introspect_type` for uncommon inputs.
-- **Filter shape:** nested `operator` (`and` | `or`) plus `queries` (and optional `groups`). Do **not** pass a top-level `current_phase` array — that is not valid GraphQL input. `create_pipe_report` / `update_pipe_report` (MCP and CLI) validate structure before calling the API.
+- **Filter shape:** nested `operator` (`and` | `or`) plus `queries` (and optional `groups`). Do **not** pass a top-level `current_phase` array — that is not valid GraphQL input. Filter structure is validated in the SDK before calling the API; MCP and CLI surface the same error.
```
Danielmoraisg: nit: "Filter structure is validated in the SDK before calling the API" is passive; "The SDK validates filter structure before calling the API" is tighter. (The em dash earlier in this line predates this PR.)

Danielmoraisg (approved):

Danielmoraisg (approved):

packages/cli/src/pipefy_cli/commands/report_pipe.py
adriannoes: Fixed in c9fbca0: CLI calls `prepare_report_cards_filter_cli` / `normalize_label_color_cli` at option-parse time (same SDK planners as the service layer). Invalid filter/color now fails before `get_authenticated_client`. Added regression tests without auth (`test_report_pipe_create_rejects_invalid_filter_without_auth`, `test_label_create_rejects_non_hex_color_without_auth`).

packages/cli/tests/test_label_webhook_commands.py
adriannoes: Fixed in c9fbca0: renamed CLI tests (`test_label_create_rejects_non_hex_color`, `test_label_create_forwards_normalized_color_to_sdk`, etc.) and MCP test (`test_label_color_maps_sdk_value_error__no_integration`). Invalid-color CLI cases now use `assert_not_called()` since validation is pre-auth.

CHANGELOG.md
adriannoes: Fixed in c9fbca0: removed the internal-move Changed entry; amended the existing Unreleased Added/Changed bullets to mention SDK validation and CLI pre-auth rejection.

docs/mcp/tools/reports.md
adriannoes: Fixed in c9fbca0: switched to active voice ("The SDK validates filter structure before calling the API").

PR #293, opened 2026-06-11
chore(sdk): enforce explicit TLS verification on GraphQL transport

adriannoes:
## Summary

- Add `verify=True` explicitly when constructing `HTTPXAsyncTransport` in `BasePipefyClient.execute_query`.
- Add a unit test asserting TLS verification is enabled on every transport build.
- Document the change under `[Unreleased]` in `CHANGELOG.md`.

`validate_https_url` on settings is intentionally deferred to a follow-up PR.

## Test plan

- [x] `uv run pytest packages/sdk/tests -m "not integration"` — 874 passed
- [x] `uv run ruff check .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Releases are versioned in lockstep across workspace members (`pipefy-sdk`, `pipe
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Changed
+
+- **SDK**: GraphQL `HTTPXAsyncTransport` now sets `verify=True` explicitly so TLS certificate verification cannot be disabled by implicit defaults.
+
 ## [0.2.0-beta.3] - 2026-06-03
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ Releases are versioned in lockstep across workspace members (`pipefy-sdk`, `pipe
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Changed
+
+- **SDK**: GraphQL `HTTPXAsyncTransport` now sets `verify=True` explicitly so TLS certificate verification cannot be disabled by implicit defaults.
```
Danielmoraisg: `verify=True` is already httpx's default, so this pins intent and adds a regression guard rather than closing a gap. "cannot be disabled by implicit defaults" implies the default was unsafe, which is not the case. Suggest: "sets `verify=True` explicitly to pin TLS verification on, independent of httpx's default."

Danielmoraisg (approved):

CHANGELOG.md
adriannoes: Good catch — updated CHANGELOG to say we pin intent rather than fix an unsafe default.

PR #294, opened 2026-06-11
chore(deps): upgrade pydantic to 2.13.4

adriannoes:
## Summary

- Bump `pydantic>=2.13.4` in `packages/sdk`, `packages/mcp`, `packages/auth`, and `packages/infra`.
- Refresh `uv.lock` (`pydantic` 2.13.4, `pydantic-core` 2.46.4, `pydantic-settings` 2.14.1).
- CHANGELOG entry under `[Unreleased]`.

Stacked on #PR1 — merge after `chore/enforce-tls-verification`.

## Test plan

- [x] `uv run pytest -m "not integration"` — 2947 passed
- [x] `uv run pytest -m "not integration" -W error::DeprecationWarning` — 2947 passed
- [x] `uv run ruff check .`


packages/sdk/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "httpx>=0.27.0",
     "openpyxl>=3.1.5",
     "pipefy-infra",
-    "pydantic>=2.11.3",
+    "pydantic>=2.13.4",
```
Danielmoraisg: Unbounded upper bound: `pydantic>=2.13.4` can pull a future 3.x on the next unconstrained resolve. It matches the existing repo convention so it is not a regression, but `pydantic>=2.13.4,<3` is the standard guard if you want it. Same applies to the auth/infra/mcp pyproject files in this PR.

Danielmoraisg (approved):

packages/sdk/pyproject.toml
adriannoes: Added `pydantic>=2.13.4,<3` in sdk, mcp, auth, and infra pyprojects.

PR #295, opened 2026-06-11
chore(deps): upgrade gql to 4.0.0

adriannoes:
## Summary

- Bump `gql[httpx]>=4.0.0,<5` in `packages/sdk` and refresh `uv.lock`.
- Adapt SDK tests for gql 4: `gql()` constants are `GraphQLRequest` objects (use `.document` for AST introspection); HTTP timeouts surface as `TransportConnectionFailed` (subclass of `TransportError`, so existing CLI catches remain valid).
- `session.execute(query, variable_values=...)` preserved; `verify=True` unchanged from PR1.
- CHANGELOG entry under `[Unreleased]`.

Stacked on pydantic upgrade branch — merge after `chore/upgrade-pydantic-2.13`.

## Test plan

- [x] `uv run pytest -m "not integration"` — 2947 passed
- [x] `uv run ruff check .`
- [x] `uv run pytest -m integration` — 46 skipped (no `PIPEFY_*` creds in CI agent env)

## Manual smoke (post-merge)

- Cursor MCP: invoke a read-only tool (e.g. `get_me`) against a configured server.
- Optional: `npx @modelcontextprotocol/inspector uv --directory . run pipefy-mcp-server` for transport debugging.

### DIFF
--- a/packages/sdk/src/pipefy_sdk/base_client.py
+++ b/packages/sdk/src/pipefy_sdk/base_client.py
@@ -34,8 +34,11 @@ def unwrap_relay_connection_nodes(connection: Any) -> list[dict[str, Any]]:
 def _graphql_request_with_variables(
     query: Any, variables: dict[str, Any]
 ) -> GraphQLRequest:
-    """Bind variables on a fresh request so shared ``gql()`` constants stay immutable."""
-    return GraphQLRequest(query, variable_values=variables or None)
+    """Bind variables on a fresh request so shared ``gql()`` constants stay immutable.
+
+    An empty ``variables`` dict omits ``variable_values`` (``None``) on the request.
+    """
+    return GraphQLRequest(query, variable_values=variables if variables else None)
 
 
 class BasePipefyClient:
--- a/packages/sdk/src/pipefy_sdk/base_client.py
+++ b/packages/sdk/src/pipefy_sdk/base_client.py
@@ -5,6 +5,7 @@ from collections.abc import Callable
 from typing import Any, ClassVar
 
 from gql import Client
+from gql.graphql_request import GraphQLRequest
 from gql.transport.exceptions import TransportQueryError
 from gql.transport.httpx import HTTPXAsyncTransport
 from graphql import GraphQLSchema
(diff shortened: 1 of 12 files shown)


packages/sdk/src/pipefy_sdk/base_client.py
```diff
@@ -30,6 +31,13 @@ def unwrap_relay_connection_nodes(connection: Any) -> list[dict[str, Any]]:
     return nodes
 
 
+def _graphql_request_with_variables(
+    query: Any, variables: dict[str, Any]
+) -> GraphQLRequest:
+    """Bind variables on a fresh request so shared ``gql()`` constants stay immutable."""
+    return GraphQLRequest(query, variable_values=variables or None)
```
Danielmoraisg: `variables or None` converts an empty dict to `None`, a small undocumented change from the previous `variable_values={}`. Behaviorally equivalent for gql (falsy variables are omitted from the payload), so low risk. An empty-variables test would lock it down; the current unit test only covers the non-empty case.

Danielmoraisg (approved):

packages/sdk/src/pipefy_sdk/base_client.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any, ClassVar
 
 from gql import Client
+from gql.graphql_request import GraphQLRequest
```
Danielmoraisg, line 8: Now that `gql()` returns a `GraphQLRequest`, a few docstrings that still call its output a "DocumentNode" are inaccurate (these files aren't in this diff, so flagging from the nearest anchor):
- `internal_api_client.py:35-36`
- `card_queries.py:273`
- `portal_internal_queries.py:3-4`

Cosmetic only, but worth a one-line sweep while the gql-4 context is fresh.

packages/sdk/src/pipefy_sdk/base_client.py
adriannoes: Added a unit test asserting empty `{}` omits `variable_values` (None), documented the behavior in `_graphql_request_with_variables`, and switched to `variables if variables else None` for clarity.

packages/sdk/src/pipefy_sdk/base_client.py
adriannoes: Updated `portal_internal_queries.py` and `card_queries.py` to reference `GraphQLRequest` instead of `DocumentNode`.

PR #319, opened 2026-06-19
ci(release): toggle-gated PyPI publishing for all release tags (+ docs)

gbrlcustodio:
## What

Make PyPI publishing a separate, toggle-gated CI job that covers every release tag (betas included), and reconcile `RELEASE.md` with the workflow.

## Why

`RELEASE.md` claimed only `pipefy-cli` and `pipefy-mcp-server` publish to PyPI, gated to `v1.` tags. Two problems:

1. **Which packages.** `release.yml` stages and uploads all five wheels, and the dependency closure forces that: `pipefy-cli`/`pipefy-mcp-server` depend on `pipefy-sdk`/`pipefy-auth`, which depend on `pipefy-infra`. Published wheels declare these as bare, unpinned deps (the `[tool.uv.sources]` workspace mapping applies only in dev), so a public `pip install` only resolves if every package in the closure is on PyPI.
2. **When.** The `v1.`-only gate was a stand-in for the PyPI account not being provisioned yet, not a deliberate policy. The intended end state publishes all release tags, including the `v0.2.0-beta.*` line.

## Changes

**`.github/workflows/release.yml`**

- Split publishing into a separate `publish-pypi` job (`needs: release`) under a dedicated `pypi` GitHub Actions environment, with `id-token: write`. The `release` job keeps `contents: write` and now uploads `dist/` as an artifact the publish job consumes. Least-privilege per job; the top-level `permissions` block is removed.
- Replace the `startsWith('v1.') && !contains('-')` step gate with a job-level toggle: `if: vars.PYPI_PUBLISH_ENABLED == 'true'`. Off by default (job skips), so betas and finals don't attempt PyPI until the account and the five pending publishers exist. The sec team flips the variable to turn it on, no code change.

**`RELEASE.md`**

- Reframed around the toggle instead of the v0.x/v1 split: every tag builds + creates a GitHub Release; PyPI publishing layers on top when `PYPI_PUBLISH_ENABLED` is `true`, for all release tags.
- Repository-setup section lists all five distributions, the single-batch failure mode, pending publishers for the first publish (owner `pipefy`, repo `ai-toolkit`, workflow `release.yml`, environment `pypi`), the toggle, and the optional environment protection rules.

## Sec-team / maintainer prerequisites (before flipping the toggle)

1. Provision the PyPI account/org and create five **pending publishers** (`pipefy-sdk`, `pipefy-mcp-server`, `pipefy-cli`, `pipefy-auth`, `pipefy-infra`), all pointing at owner `pipefy`, repo `ai-toolkit`, workflow `release.yml`, environment `pypi`. All five names are currently available on PyPI.
2. Create the `pypi` GitHub Actions environment (add protection rules if you want an approval gate).
3. Set repo variable `PYPI_PUBLISH_ENABLED=true`. One missing publisher fails the whole release, so do this only once all five exist.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - "v*"
 
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   release:
     runs-on: ubuntu-latest
@@ -71,7 +75,8 @@ jobs:
           from pathlib import Path
 
           dist = Path("dist")
-          for prefix in ("pipefy_sdk", "pipefy_mcp_server", "pipefy_cli", "pipefy_auth", "pipefy_infra"):
+          expected = ("pipefy_sdk", "pipefy_mcp_server", "pipefy_cli", "pipefy_auth", "pipefy_infra")
+          for prefix in expected:
               wheels = list(dist.glob(f"{prefix}-*.whl"))
               if len(wheels) != 1:
                   print(
@@ -79,6 +84,17 @@ jobs:
                       file=sys.stderr,
                   )
                   sys.exit(1)
+          # Bound the published set: any wheel beyond the five known packages
+          # (e.g. a new workspace member) must fail here rather than be silently
+          # published, since the publish step copies every wheel in dist/.
+          all_wheels = list(dist.glob("*.whl"))
+          if len(all_wheels) != len(expected):
+              print(
+                  f"expected exactly {len(expected)} wheels in dist/, got "
+                  f"{sorted(w.name for w in all_wheels)!r}",
+                  file=sys.stderr,
+              )
+              sys.exit(1)
           PY
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v2
@@ -92,6 +108,7 @@ jobs:
           name: dist
           path: dist/
           if-no-files-found: error
+          retention-days: 1
 
   publish-pypi:
     needs: release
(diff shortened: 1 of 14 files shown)


gbrlcustodio:
Closing this in favor of the changes that already landed plus two focused follow-ups.

The core goal here (publish all five wheels to PyPI on every release tag) shipped in #365 via a simpler path: drop the `v1.`-only gate from the three PyPI steps, keep a single job, publish over OIDC/Trusted Publishing. It is live: `0.3.0-alpha.1` published all five distributions to PyPI, and trusted publishers are configured for each.

Given that, a rebase of this branch is not worth it: the diff is stale against the #361 SDK rename (`pipefy_sdk` -> `pipefy`, pins `==0.2.0-beta.3`) and would need conflict resolution across ~14 files, and its off-by-default `PYPI_PUBLISH_ENABLED` toggle would re-gate publishing that now works (its rationale, "while the PyPI account is being provisioned", no longer holds).

The genuinely valuable pieces are carried forward, rewritten for the current name and version:

- #367 -- pin workspace siblings to the lockstep version in published metadata, plus the wheel-set bound (fail when `dist/` holds an unexpected wheel).
- #366 -- `skip-existing: true` on the PyPI upload (idempotent re-runs) and reopening the changelog `[Unreleased]` section.

Not carried forward (can be a separate hardening PR if wanted): splitting `publish-pypi` into its own job under a dedicated `pypi` environment with per-job least-privilege permissions and an optional approval gate.

Issue #305, opened 2026-06-18
Hosted-safe attachment path


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Provide an attachment-upload path that does not depend on the server's local filesystem: a client-supplied file URL with SSRF and size guards, an in-body upload, or a client-side presigned upload. Gate the local-file-path variant to the stdio profile.

## Acceptance criteria

An attachment can be uploaded in the hosted profile without server filesystem access; the local-file-path variant is unreachable in the hosted profile; the tool is then marked remote-safe.

## Dependencies

Depends on the default-deny allowlist (#304).

adriannoes:
## Field report: the same gap bites the stdio profile through claude.ai connectors

A user hit this with the local stdio server, so the problem is not exclusive to the hosted profile. Setup: claude.ai in the browser with code execution enabled, plus the Pipefy MCP connector running on the user's Windows machine. The agent generated a file inside its Anthropic-hosted sandbox (`/mnt/user-data/outputs/...`) and had no way to hand the bytes to the server. Reproducing that call shape against a path the server cannot see returns:

```json
{"success":false,"error":{"message":"File not found or not a regular file: /mnt/user-data/outputs/felipe-repro.txt"},"step":"file_read"}
```

The agent then tried base64 input (dropped in 6007155c along with the URL mode; #253's text said base64 would stay as the in-memory fallback, but the final refactor removed it too) and a Google Drive URL (dropped in #253), and ended up asking the user to download the file from the chat and reply with a local path. That works, but it costs several turns and a manual step per file.

The acceptance criterion "an attachment can be uploaded without server filesystem access" therefore needs to hold for two topologies, not one:

1. Hosted profile (the original scope here).
2. Local stdio server whose MCP client is not co-located with the environment that generates files (claude.ai + connector).

## What we verified (2026-07-02)

**The client-side presigned flow already works with existing tools.** Ran end to end against the PipeClaw test org (302398434), card 1406588268, field `egress_test_attachment`:

1. `execute_graphql` with `createPresignedUrl` (organizationId, fileName, contentType) returns the S3 PUT url and a download url.
2. `curl -X PUT --upload-file ...` to the presigned url: HTTP 200.
3. `update_card_field` accepts the raw storage path (the url's path component without the leading slash, `orgs/<uuid>/uploads/<uuid>/<file>`).
[...]

Issue #302, opened 2026-06-18
Request-scoped identity


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Remove any global identity resolved at startup. Resolve the caller's validated token per request via the MCP auth context and a callable resolver on the shared API client, so concurrent callers act as their own users.

## Acceptance criteria

A concurrency-isolation test with many concurrent calls across distinct identities shows zero cross-identity responses, using a discriminator the identities actually differ on.

## Dependencies

Builds on the resource-server role (#301).

Issue #301, opened 2026-06-18
OAuth resource-server role


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Serve OAuth protected-resource metadata (RFC 9728); return `401` with a `WWW-Authenticate` challenge on missing or invalid tokens; validate the inbound bearer on every request (signature via JWKS, issuer, expiry, and audience). The server never resolves a global service identity.

## Acceptance criteria

The protected-resource metadata endpoint returns correct data; an unauthenticated request returns `401` with the challenge; a valid token passes; tampered, expired, and wrong-audience tokens are rejected (both directions tested).

## Dependencies

Feeds request-scoped identity, which consumes the validated token.

Issue #300, opened 2026-06-18
HTTP (Streamable) transport profile and combined --remote launch flag


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Add a Streamable HTTP transport mode alongside the existing stdio transport (bind config, long-lived process). Register tools once on the shared app, not in a per-session lifespan. Add a single `--remote` launch flag (or subcommand) that starts the server in the hosted profile: HTTP transport plus the default-deny remote tool profile, with a safety interlock that refuses to serve the full tool surface over public HTTP unless an explicit escape hatch is set.

## Acceptance criteria

Server boots in HTTP mode and a client completes a tool call; the stdio path is unchanged; `--remote` sets both transport and the remote profile; starting public HTTP without the remote profile is refused unless the escape hatch is set.

## Dependencies

Pairs with request-scoped identity and the resource-server role; the remote profile flag itself is implemented in the default-deny allowlist issue.

PR #327, opened 2026-06-22
feat(mcp): HTTP (streamable) transport profile and --remote flag

gbrlcustodio:
## What

Adds a Streamable HTTP transport behind a `--remote` launch flag (issue #300), an `McpSettings` model for the MCP-only knobs (moved out of the SDK's `PipefySettings`), and a loopback-only bind guard for the HTTP transport.

This is now rebased onto `dev`: the register-once foundation (#329) and the per-request client injection via `Context` (#330) have both landed, so the diff here is only the HTTP transport work.

## How

- `main.py`: `--remote` / `--host` / `--port` parsing; dispatches to `run_server`.
- `server.py`: a single `run_server(*, http=False, host=None, port=None, remote_mode=None)` is the serve entry point for both transports. `http=False` (default) speaks stdio; `http=True` builds a dedicated app (no constructor `lifespan`), initializes services once, and serves over Streamable HTTP. Both register via the shared `_register_pipefy_tools`. `_assert_loopback_http_bind` refuses any non-loopback HTTP bind (see Scope). `build_pipefy_mcp_server` remains the app factory the in-process tool tests use.
- `settings.py`: new `McpSettings` (`unified_envelope`, `remote_mode`, `host`, `port`); `env_prefix="PIPEFY_MCP_"` keeps the `PIPEFY_MCP_*` env vars stable.
- The two MCP flags move out of `pipefy_sdk.PipefySettings` into `McpSettings`.

Tools take no client at registration: each resolves the live client per request from the lifespan context via `get_pipefy_client(ctx)` (the model from #330), so the shared registration path works unchanged across both transports and across a per-session lifespan re-entry under Streamable HTTP.

## Scope

HTTP mode is unauthenticated foundation work: no inbound bearer validation (#301) and no per-request on-behalf-of identity (#302), so every call runs as the single identity resolved at startup. Because of that, the HTTP transport refuses any non-loopback bind (`_assert_loopback_http_bind`) regardless of the tool profile, the remote-safe subset is no safer over an unauthenticated network bind. Loopback also keeps the filesystem tools coherent (the attachment uploads read a local `file_path`, which only resolves when the server shares the client's disk; remote-safe file inputs are #305). `--remote` still selects the default-deny remote profile and is for local/validation use; a network bind is revisited when inbound auth lands (#301).

## Verification

- `ruff check` + `ruff format --check` clean on `packages/mcp` and `packages/sdk`.
- `packages/mcp` suite green (1384 passed, 16 integration deselected), including the HTTP acceptance test that drives a real Streamable HTTP round-trip.


Danielmoraisg (approved):
LGTM!

PR #329, opened 2026-06-22
refactor(mcp): register tools at construction, reserve lifespan for resources

gbrlcustodio:
## What

Move Pipefy tool registration out of the FastMCP `lifespan` and into construction time, and reduce the lifespan to its resource role (initialize services, yield the container as `lifespan_context`). This is the foundation increment that aligns the server with the official FastMCP lifespan contract; the HTTP transport work (#300) will build on it.

## Why

Registration lived inside the `lifespan`, which under Streamable HTTP runs **per session** (per request in stateless mode), not once per process. Registering there re-registered on every session and required repeat-visit bookkeeping (`prepare_/mark_/cleanup`) to remove-and-re-add Pipefy tools so they would rebind to a freshly initialized client. That machinery is a workaround for putting registration in the wrong place.

The official FastMCP pattern keeps registration at construction and uses the lifespan only for resources accessed per request via `lifespan_context`. This change adopts the structure without rewriting tool bodies.

## How

- `server.py`: `build_pipefy_mcp_server` -> `_register_pipefy_tools` registers once, up front. The `lifespan` now only initializes services and yields the container.
- `core/container.py`: new `PipefyClientProxy` resolves `container.pipefy_client` on each access, so tools can bind once at registration and still follow a service re-initialization (a fresh client) without re-registering.
- `tools/registry.py`: `register_tools` binds the proxy; the "client is None" guard is gone (registration no longer needs a live client).
- Deletes `core/pipefy_tool_lifecycle.py` (repeat-visit machinery), no longer needed.
- Tests: replaced the repeat-visit/lifespan-registration tests with construction-time registration tests and proxy unit tests.

## Scope

Per-request on-behalf-of identity (tools reading the client from `lifespan_context` per request) is intentionally **not** in this PR; it is the follow-up that makes the hosted profile multi-user (#302). This PR keeps the process-global single-identity model, which is correct for stdio and for the current unauthenticated HTTP foundation.

## Verification

- `ruff check` + `ruff format --check` clean on `packages/mcp`.
- Full `packages/mcp` suite green: 1371 passed, 16 skipped.

## Follow-up cleanups

Small, no-behavior-change simplifications to the registration mechanism and serve init, layered on the same branch:

- `main.py`: drop the duplicate `"Starting Pipefy MCP server"` log (`run_server` already owns it) and the now-unused logger.
- `tools/registry.py`: data-drive `register_tools` via a module-level `_TOOLSETS` tuple instead of 16 repeated `.register` calls; `register_tools` now returns `None` (no caller used the old `self.mcp` return); collapse the over-defensive `_snapshot_tool_names` to a direct comprehension matching the bare `_tool_manager` access already used in `retain_only`.
- `core/container.py`: declare `pipefy_client` in `__init__` rather than as a shared class-level default.

The process-global `ServicesContainer` singleton is left in place. Replacing it with build-site ownership was explored but reverted: it needed a closure-based lifespan factory (FastMCP fixes the lifespan signature, so the owned container can't be passed as an argument), which added more indirection than the singleton it removed. The cleaner endpoint is #330, which drops `PipefyClientProxy` and reads the client per request from `lifespan_context`, letting the lifespan own a local container with no closure and no global.


packages/mcp/src/pipefy_mcp/core/container.py
```diff
@@ -90,3 +90,28 @@ async def initialize_services(self, settings: Settings) -> None:
             allow_insecure_urls=pipefy.allow_insecure_urls,
         )
         self.pipefy_client.set_internal_api_client(internal_client)
+
+
+class PipefyClientProxy:
```
gbrlcustodio, line 97: Capturing why this proxy exists, so a future reader understands it before deleting it.

Two facts in this PR make some lazy, per-request client resolution mandatory:

1. Tools register at construction (`_register_pipefy_tools`).
2. At construction time `container.pipefy_client` is still `None`. The lifespan calls `initialize_services`, which assigns a fresh `PipefyClient` instance, and under Streamable HTTP the lifespan re-enters per session, so that instance is swapped.

A tool that captured a concrete client at registration would capture `None`, and even after init would hold a stale instance across re-entry. In this PR the proxy is one stable handle that resolves `container.pipefy_client` on each access, with a single centralized None-check.

The proxy is an interim bridge, not the end state. It is removed in #330 (refactor/mcp-context-injection, stacked on this branch): tools stop closing over a client and instead take `ctx: Context` and call `get_pipefy_client(ctx)` per request, reading `ctx.request_context.lifespan_context.pipefy_client`. That is the official FastMCP lifespan/Context pattern, it is the plumbing for per-request on-behalf-of identity (#302), and it also removes the `cast(PipefyClient, ...)` wart at the `register()` boundary, since `register(mcp)` no longer passes a client at all.

So: keep the proxy as-is for #329; expect it to disappear when #330 lands.

Danielmoraisg (approved):
LGTM!

Issue #333, opened 2026-06-23
refactor(sdk): constructor-level DI for services via an injected GraphQLExecutor


gbrlcustodio:
## Summary

Invert the GraphQL execution seam in the SDK so services receive their executor through the constructor instead of inheriting it from `BasePipefyClient`. This lets every service rely on constructor-level dependency injection and removes the need for tests to monkeypatch `execute_query`.

## Motivation

Today 15 of the 16 SDK services inherit `BasePipefyClient` and call inherited `self.execute_query`, which builds the httpx transport and runs the query in the same method. Because the execution is inherited rather than injected, there is no collaborator to substitute, so tests reach into the constructed object and replace the method (`service._interfaces_client.execute_query = AsyncMock(...)`). Services already inject their other collaborators (`auth`, `pipe_service`, `card_service`, `table_service`); GraphQL execution is the one dependency that is not.

This violates the project's coding principles: decision and effect are mixed in one method, services depend on the concrete httpx/gql stack rather than an owned abstraction, and tests bind to implementation details instead of a contract.

## Proposed change

1. Define a narrow owned interface, `GraphQLExecutor` (Protocol), with `async def execute_query(query, variables) -> dict`. It exposes only what services need; nothing about httpx or gql leaks through it.
2. Make the current `BasePipefyClient` the adapter that implements it (rename to e.g. `HttpxGraphQLExecutor`). It keeps the deliberate per-call fresh-transport creation, the fetched-schema reuse cache, and the optional `on_graphql_error` formatter. This is the only place that touches httpx/gql.
3. Convert services from inheritance to composition: each takes `executor: GraphQLExecutor` (plus its existing collaborators) and calls `self._executor.execute_query(...)`. `PortalService` takes the three executors it needs (public, Interfaces, internal) instead of building two and receiving one.
4. Keep the composition root in `PipefyClient.__init__` (optionally a small `build_executors(settings, auth)` helper): build one executor per endpoint from the shared `(settings, auth)` and inject. One `auth` instance preserves the single-token-cache invariant.
5. Tests inject a fake executor implementing the protocol. No `@patch`, no method replacement on constructed objects.

## Principle alignment

- Separate decisions from effects: the adapter owns the HTTP effect; services reduce to pure query and variable building, one executor call, and pure result mapping.
- Depend on abstractions you own: services depend on `GraphQLExecutor`; a transport or vendor change touches one file.
[...]

PR #334, opened 2026-06-23
refactor(sdk,mcp,cli): consolidate internal API client construction and ownership

gbrlcustodio:
## What

Consolidate how the SDK's internal API client is built and owned. The facade now constructs it like every other endpoint client, the services that use it receive it at construction as a required dependency, and the one facade method that did internal-API I/O directly is now pure delegation.

Pipefy has three GraphQL endpoints, each a `BasePipefyClient` differing only by URL (public, Interfaces, internal). The facade built the public and Interfaces clients in place from the shared `(settings, auth)` but threaded the internal one in via `set_internal_api_client`, so every caller (the MCP container, the CLI) repeated the same wiring and tests reached for a setter that existed only for them. `delete_card_relation` was the lone facade method that wasn't pure delegation, because the card/relation services had nowhere to host an internal-API call.

## Changes

- **Construct in `__init__`**: `PipefyClient.__init__` builds the single shared `InternalApiClient` from `(settings, auth)` and injects it into the services that need it. Removed `set_internal_api_client` from `PipefyClient` and `PortalService`. MCP `core/container.py` and CLI `cli/auth.py` collapse to a single `PipefyClient(settings, auth=...)` call.
- **Move `delete_card_relation` onto `RelationService`**: it now lives next to its sibling `create_card_relation`, and `INTERNAL_DELETE_CARD_RELATION_MUTATION` moves from `queries/card_queries.py` to `queries/relation_queries.py`. The facade method becomes one-line delegation, so the facade holds no I/O of its own.
- **Require the internal client**: with the facade always injecting it, the `InternalApiClient | None` parameters and their None guards on `RelationService` and `PortalService` were unreachable in production. The parameter is now non-optional on both, and the dead guards plus the orphaned `INTERNAL_API_CLIENT_NOT_CONFIGURED` constant are gone.
- **Remove `internal_api_available`**: it was already always `True` in production (the internal client was wired unconditionally regardless of auth tier), so its guards in the CLI `relation card delete` command and the MCP `delete_card_relation` tool were reachable only from tests. Removed the property and both guards.
- **Tests**: migrated to the seam the other endpoint clients already use (construct the client, then replace `execute_query` on it), instead of `@patch` and the setter. Added a `RelationService`-level test for `delete_card_relation`; dropped the tests that exercised the now-removed guards.

## Verification

- SDK 958 passed / 17 skipped; MCP 1383 passed / 16 skipped; CLI 320 passed / 13 skipped.
- `rg "internal_api_available|set_internal_api_client|INTERNAL_API_CLIENT_NOT_CONFIGURED"` returns nothing.
- Behavior-preserving: the internal client was always wired before, so neither the removed `internal_api_available` guard nor the removed None guards ever fired in production.

## Notes

- Rebased onto `dev` now that #327 (`feat/mcp-http-transport`) has merged; this PR targets `dev` directly.
- Clears the groundwork #333 (constructor-level DI via an injected `GraphQLExecutor`) was waiting on: every service now owns its operations and is handed the endpoint clients it needs, and the facade is pure delegation, which is the executor refactor's end state for the facade.

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/relation.py
+++ b/packages/cli/src/pipefy_cli/commands/relation.py
@@ -156,7 +156,7 @@ def relation_card_delete(
     yes: bool = typer.Option(False, "--yes", "-y", help="Skip confirmation prompt."),
     json_out: bool = typer.Option(False, "--json", "-j"),
 ) -> None:
-    """Remove a card relation (requires service-account credentials; internal API)."""
+    """Remove a card relation (internal API, requires OAuth)."""
 
     confirm_destructive(
         yes=yes,
--- a/packages/cli/src/pipefy_cli/commands/relation.py
+++ b/packages/cli/src/pipefy_cli/commands/relation.py
@@ -7,7 +7,6 @@ from pipefy_sdk import PipefyClient
 
 from pipefy_cli.commands._common import (
     ID_POSITIONAL_CONTEXT_SETTINGS,
-    authenticated_client_from_ctx,
     confirm_destructive,
     parse_json_object,
     resource_id_argument,
@@ -158,17 +157,6 @@ def relation_card_delete(
     json_out: bool = typer.Option(False, "--json", "-j"),
 ) -> None:
     """Remove a card relation (requires service-account credentials; internal API)."""
-
-    client = authenticated_client_from_ctx(ctx)
-    if not client.internal_api_available:
-        typer.echo(
-            "delete_card_relation requires service-account credentials "
-            "(PIPEFY_SERVICE_ACCOUNT_CLIENT_ID, "
-            "PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET). "
-            "The deleteCardRelation mutation is only available on the internal API.",
-            err=True,
-        )
-        raise typer.Exit(2)
 
     confirm_destructive(
         yes=yes,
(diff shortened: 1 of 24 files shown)


packages/cli/src/pipefy_cli/commands/relation.py
```diff
@@ -159,17 +158,6 @@ def relation_card_delete(
 ) -> None:
     """Remove a card relation (requires service-account credentials; internal API)."""
```
Danielmoraisg: Now that the `internal_api_available` guard is gone, nothing enforces what this docstring still states ("requires service-account credentials").

This is not a regression. The guard keyed off `internal_api_available`, which was always `True` because the internal client was wired regardless of auth tier, so a bearer-token user never saw the friendly message anyway and instead hit a raw internal-API auth error. Removing it is behavior-preserving.

The open question is whether we want a tier-aware preflight check for `delete_card_relation` (this CLI command and the MCP tool in `pipe_tools.py`) as a follow-up, or whether the raw API error is acceptable. Flagging so the decision is explicit rather than implied by the removal.

Danielmoraisg (dismissed):

packages/cli/src/pipefy_cli/commands/relation.py
gbrlcustodio: Fixed in b01566e: the service-account note was inaccurate to begin with. Reworded to "(internal API, requires OAuth)" to match the SDK `delete_card_relation` docstring, since no service account is required.

Danielmoraisg (approved):

adriannoes (approved):

PR #335, opened 2026-06-23
refactor(sdk): constructor-level DI for services via an injected GraphQLExecutor

### LINKED ISSUE #333: refactor(sdk): constructor-level DI for services via an injected GraphQLExecutor

gbrlcustodio:
Closes #333.

## What this does

Inverts the GraphQL execution seam. SDK services no longer inherit `execute_query` from `BasePipefyClient`; they receive a narrow owned `GraphQLExecutor` through their constructor and call `self._executor.execute_query(...)`. The httpx/gql stack lives behind a single adapter, `HttpxGraphQLExecutor`, and tests inject a fake executor instead of monkeypatching `execute_query` on constructed objects.

This addresses three coding-principle violations in the old design: decision and effect were mixed in one inherited method, services depended on the concrete httpx/gql stack rather than an owned abstraction, and tests bound to implementation details rather than a contract.

## How it works

- New `pipefy_sdk.graphql_executor` module: the `GraphQLExecutor` Protocol (one `async execute_query(query: DocumentNode, variables) -> dict`) and `HttpxGraphQLExecutor`, the renamed `BasePipefyClient` and sole adapter. It is a pure transport: it takes a resolved endpoint `url`, an `auth`, a `cache_schema` flag, and an optional `on_graphql_error` mapper, and holds no `PipefySettings`. Its `execute_query` body (transport-per-call, lock-guarded fetched-schema cache, `on_graphql_error` mapping) is unchanged from the old base client.
- The relay helper moves to `pipefy_sdk.utils.relay`; the internal-endpoint error formatter moves to `pipefy_sdk.internal_api_errors`. Both are transport-free homes.
- All 16 services convert to `__init__(self, *, executor, ...)`. `auth` is dropped from every service; `settings` survives only on `WebhookService`; collaborators (`pipe_service`, `card_service`, ...) become required, dropping the self-construct fallbacks.
- The composition root builds one executor per endpoint via `build_executors(settings, auth) -> Executors` and injects them: every public service gets the public executor, `RelationService` gets `(public, internal)`, `PortalService` gets `(public, interfaces, internal)`. `build_executors` is the one place that resolves each endpoint URL from settings, so the adapter stays agnostic to endpoint topology.

## Breaking change

The `InternalApiClient` class and its `pipefy_sdk` export are removed. The internal endpoint is now an `HttpxGraphQLExecutor` built with `url=internal_api_url` plus the `[code=…][correlation_id=…]` error formatter, produced by `build_executors`. Documented under `[Unreleased]` in the changelog. Accepted on the beta SDK.

## Behavioral deltas worth a reviewer's eye

- **One public executor shared across services.** Each service previously held its own `BasePipefyClient`; public services now share a single executor. The transport is built fresh per call, so the only shared state is the per-instance fetched-schema cache (off by default), and one cache for the one public schema is the correct model. All three executors still share one `auth` instance, preserving the single-token-cache invariant.
- **Required collaborators.** `MemberService` and `WebhookService` no longer self-construct `PipeService` / `CardService` fallbacks; callers must supply them, which the composition root already does.

## Concurrency

MCP clients issue parallel tool calls, so one shared public executor serves concurrent `execute_query` coroutines. This is safe: a fresh `HTTPXAsyncTransport` per call (no shared connection state), query constants that are never mutated (variables bind onto a fresh request), and a schema cache double-checked under a lock. A single service instance already served concurrent calls before this change; sharing one executor across services widens the same pattern.

## Tests

Every service test injects a `mock_executor` (a `MagicMock(spec=GraphQLExecutor)` with an `AsyncMock` `execute_query`) and asserts on it. No `service.execute_query = AsyncMock` monkeypatches remain. The base-client adapter test is renamed to `test_httpx_graphql_executor.py`.

## Verification

- Full unit suite green (`uv run python -m pytest -m "not integration"`): 3109 passed.
- `uv run ruff check .` and `uv run ruff format .` clean.
- Integration tests skip cleanly without credentials.
- Smoke: `pipefy --help` and the MCP composition root wire correctly.
- Final greps confirm zero `InternalApiClient` references (outside the changelog note), zero `.execute_query = AsyncMock` monkeypatches, zero `BasePipefyClient` references.


packages/sdk/tests/services/test_internal_api_executor.py
Danielmoraisg, line 1: Should-fix: rename this file. `InternalApiClient` is gone, and this file now exercises `build_executors(...).internal` (the docstring was updated to say so). The sibling `test_pipefy_base_client.py` was renamed to `test_httpx_graphql_executor.py` to track its new subject; doing the same here, for example `test_internal_api_executor.py`, keeps the test layout consistent and avoids sending a reader looking for a class that no longer exists.

packages/sdk/src/pipefy_sdk/internal_api_errors.py
```diff
@@ -0,0 +1,21 @@
+"""Error envelope for the internal_api GraphQL endpoint.
```
Danielmoraisg, line 1: Nit: now that this is imported across module boundaries (`client.py` pulls it into `build_executors`), the leading underscore misrepresents its scope. It reads as private-to-this-module but is part of the composition seam. Consider `format_internal_api_error`. Behavior is preserved verbatim from the deleted `InternalApiClient`, which is the right call.

packages/sdk/src/pipefy_sdk/services/portal_service.py
Danielmoraisg, line 1: Nit: the `__init__` docstring still mentions a shared `auth` ("pass a shared `auth`... so OAuth token caching is not duplicated"), but the constructor no longer takes `auth`; it takes the three executors. The sentence describes composition-root behavior, which is defensible, but a reader of the signature sees no `auth` parameter. Consider rewording to describe the injected executors instead.

packages/sdk/tests/services/pipefy/test_portal_service.py
Danielmoraisg, line 1: Nit (coverage): dropping `test_portal_service_interfaces_client_uses_interfaces_graphql_url` removed the only assertion that `PortalService` is wired to the interfaces URL rather than the main one. That was a private-internals assertion, so removing it aligns with testing contracts, and URL correctness is still covered at the executor level in `test_httpx_graphql_executor.py`. But nothing in the unit suite now asserts that each service receives the role-correct executor. If you want that signal back, a small behavioral construction test (assert which injected mock each role-specific call routes to) would close the gap without touching internals.

CHANGELOG.md
Danielmoraisg, line 1: Nit: this bullet packs the DI mechanism, the protocol seam, the rename, the breaking `InternalApiClient` removal, the collaborator changes, and the test strategy into one entry. It is accurate and reads cleanly, but splitting the Breaking sub-clause into its own line would scan better for someone skimming the changelog. Keep a Changelog does not require it, so optional.

packages/sdk/tests/services/test_internal_api_executor.py
gbrlcustodio: Fixed in 549cd61: renamed to test_internal_api_executor.py, matching the executor it exercises and the test_httpx_graphql_executor.py rename.

packages/sdk/src/pipefy_sdk/internal_api_errors.py
gbrlcustodio: Fixed in de361f0: dropped the underscore (format_internal_api_error). I also moved the module out of services/ to the package root next to graphql_executor.py, since it is executor-layer error formatting wired at the composition root rather than a service concern.

packages/sdk/src/pipefy_sdk/services/portal_service.py
gbrlcustodio: Fixed in b4105ff: reworded the class docstring to say the executors are injected and that the shared-auth token caching is composition-root behavior, so the signature (which takes no auth) is no longer contradicted.

packages/sdk/tests/services/pipefy/test_portal_service.py
gbrlcustodio: Fixed in efa9a9c: added test_build_executors_routes_each_endpoint_to_its_url asserting public/interfaces/internal each target their own URL, and removed the orphaned URL constants. The existing interfaces/internal routing tests already assert which injected executor each call uses.

CHANGELOG.md
gbrlcustodio: Fixed in 64f8f93: split the Breaking sub-clause out into its own bullet (now `**SDK (breaking)**:`) so it scans on its own line.

Danielmoraisg (approved):

PR #337, opened 2026-06-24
feat(sdk,mcp,cli): send client telemetry headers on Pipefy API requests

### LINKED ISSUE #336: feat: send client telemetry headers on Pipefy API requests
## Summary

Add client-identifying telemetry headers to every outbound request to the Pipefy GraphQL API so usage patterns (which client, which version, which surface) can be tracked server-side.

## Motivation

Today the SDK sends only `Authorization: Bearer <token>` plus the implicit `Content-Type: application/json` that `gql` sets. There is no `User-Agent`, no client/version identifier, and no per-request correlation id. As a result Pipefy has no way to distinguish traffic coming from the MCP server, the CLI, or direct SDK use, nor to attribute load to a toolkit version.

## Current state

- Single header chokepoint for auth: the `httpx.Auth` classes in `packages/auth/src/pipefy_auth/bearer.py` (`StaticBearerAuth`, `CallableBearerAuth`, `RefreshableBearerAuth`). They inject only `Authorization`.
- Transport is built per query in `packages/sdk/src/pipefy_sdk/base_client.py` (`HTTPXAsyncTransport`), with no `headers=` argument.
- A version constant already exists but is not sent on the wire: `packages/sdk/src/pipefy_sdk/__init__.py` (`0.2.0-beta.3`).
- The attachment / S3 upload path in `packages/sdk/src/pipefy_sdk/services/attachment_service.py` uses a raw `httpx.AsyncClient` and bypasses the auth classes entirely, so any header added only to the auth classes would not cover it.

## Proposed headers

- `User-Agent`: `pipefy-sdk/<version> (<surface>)`, where `<surface>` is one of `mcp` / `cli` / `sdk`. Identifies client and entry point in one field.
- `X-Client-Name` / `X-Client-Version`: optional structured split if Pipefy prefers parsed fields over a UA string.
- `X-Request-ID`: per-request UUID for correlation and tracing.

The GraphQL operation name already travels in the request body, so a separate operation header is not needed for coarse usage buckets.

## Design considerations

- Prefer adding these at the transport level rather than inside the bearer auth classes, so both the GraphQL path and the attachment upload path are covered and so client identity stays separate from auth.
- The surface tag (`mcp` / `cli` / `sdk`) needs to be threaded from the composition root, since the SDK itself does not know which front end instantiated it.
- Version should be sourced from the existing package version constant rather than hardcoded.

## Acceptance criteria

- Every outbound Pipefy request (GraphQL and attachment upload) carries `User-Agent` and `X-Request-ID`.
- The `User-Agent` reflects the real package version and the originating surface.
- Coverage is verified by a contract-level test asserting the headers are present on the request.

Danielmoraisg:
## Summary

Outbound Pipefy API requests now carry client-identifying telemetry headers so the API can attribute traffic by surface and version.

- `User-Agent`: `pipefy-sdk/{version} ({surface})`
- `X-Client-Name`: the surface (`mcp`, `cli`, or `sdk`)
- `X-Client-Version`: the package version

`{surface}` is the calling client's identity. It is a `PipefyClient` constructor argument (threaded to `build_executors`), set at each composition root: the MCP `ServicesContainer` passes `mcp`, the CLI `get_authenticated_client` passes `cli`, and direct SDK use keeps the `sdk` default. The version reads from `pipefy_sdk.__version__`.

## Coverage

| Path | Headers |
|------|---------|
| GraphQL (Public, Interfaces, Internal API) | full set |
| OAuth handshake (`signin.pipefy.com`) | `User-Agent: pipefy-auth/{version}` + `X-Client-Name: auth` + `X-Client-Version` |

Telemetry rides only the GraphQL path (the Pipefy API). The S3 attachment upload and the automation-jobs export download go to non-Pipefy hosts, so they carry no telemetry: the API never receives those requests, and the headers would only disclose the client surface and version to third parties.

The Internal API executor is one of the three built by `build_executors`, so the resolved header set reaches it with no extra wiring. The OAuth handshake identifies the auth component rather than an API surface: `X-Client-Name` is `auth` and the `User-Agent` product is `pipefy-auth` (the package that owns those requests, with its own version). It carries no surface, because the surface does not reach the auth layer without threading through the security-sensitive refresh path.

## How it works

The header builder is a pure function in `pipefy_infra.telemetry`, shared so the SDK and auth packages cannot drift on the `User-Agent` format; `pipefy_sdk.telemetry` re-exports it.

- GraphQL: `build_executors(settings, auth, surface=...)` resolves the header set once and passes it to all three `HttpxGraphQLExecutor`s, which attach it on the per-call `HTTPXAsyncTransport`.
- OAuth: `pipefy_auth._http.http_client` sets the auth header set on the fresh client.

The surface is a constructor argument rather than a setting, so it is never read from env or TOML and `PIPEFY_CLIENT_SURFACE` cannot forge it. A genuinely new surface adds a `ClientSurface` literal member.

## Scope note

`X-Request-ID` from the issue's acceptance criteria is intentionally omitted. With it gone the header set is fully static.

## Tests

- Pure builder unit tests, including the surface-less OAuth `User-Agent`.
- Contract tests for the GraphQL and OAuth transports run the real httpx client and swap only the network transport for `httpx.MockTransport`, asserting the telemetry headers land on the actual request with the bearer/OAuth auth still applied.
- Facade tests assert the surface threads from `PipefyClient`/`build_executors` to the executors, defaults to `sdk`, and that `PIPEFY_CLIENT_SURFACE` cannot forge it; container and CLI tests assert each composition root stamps its surface.

Closes #336


packages/sdk/src/pipefy_sdk/base_client.py
```diff
@@ -92,6 +94,9 @@ async def execute_query(self, query: Any, variables: dict[str, Any]) -> dict:
             auth=self._auth,
             timeout=Timeout(timeout=self.GRAPHQL_REQUEST_TIMEOUT_SECONDS),
             verify=True,
+            headers=telemetry_headers(
+                surface=self.settings.client_surface, version=__version__
```
gbrlcustodio: **issue (blocking): built on a seam that has been removed from `dev`**

This reads `self.settings.client_surface` on every call inside `BasePipefyClient`. The executor-DI PR (#335) has landed on `dev`: it renamed `base_client.py` to `graphql_executor.py`, made `HttpxGraphQLExecutor` a pure transport with no `settings`, and dropped `settings` from the services. `ObservabilityService` and `AttachmentService` no longer carry a `settings` attribute, so these reads no longer compile against current `dev`. This branch needs rebasing onto the executor seam.

Rebased, the feature gets simpler and rides only the GraphQL path. `build_executors(settings, auth)` is already the single place that reads settings and builds all three executors:

- give `HttpxGraphQLExecutor.__init__` a resolved `headers: dict[str, str] | None = None` and attach it to the transport, the same way it already takes `url` and `cache_schema`;
- have `build_executors` call `telemetry_headers(surface=settings.client_surface, version=__version__)` once and pass it to all three executors.

That builds the header dict once per executor instead of per request, keeps `self.settings.client_surface` out of every transport, and reads `__version__` in one place instead of scattered across modules. The non-GraphQL paths (S3 upload, export download) should not carry these headers at all (see the attachment-service comment).

packages/sdk/src/pipefy_sdk/settings.py
```diff
@@ -139,6 +141,16 @@ def settings_customise_sources(
         ),
     )
 
+    client_surface: ClientSurface = Field(
```
gbrlcustodio: **issue (blocking): client_surface should not be operator config**

`client_surface` is exposed as configurable: `PIPEFY_CLIENT_SURFACE`, a TOML key, a docs row, default `sdk`. It is an identity stamp fixed by which entry point runs, not something an operator varies per deployment. The MCP and CLI roots already force their value (init kwarg and the `model_validate` dict outrank env), so the env var is inert for those surfaces. Its only reachable effect is on direct SDK use, where a caller can set `PIPEFY_CLIENT_SURFACE=mcp` and inject false `mcp` traffic into the attribution this feature exists to produce. The `Literal["mcp", "cli", "sdk"]` also blocks an embedder from tagging a new surface, so there is no escape-hatch value.

Keep it a programmatic field that composition roots set, defaulting to `sdk`, but remove the env and TOML binding (a `validation_alias` the env source cannot reach, or otherwise exclude it from the env source), drop the docs row, and drop the env-source tests. A real new surface needs a `Literal` member anyway, so nothing is lost.

gbrlcustodio (commented):
Client telemetry headers for the SDK, MCP, and CLI surfaces. The header builders are pure and well tested, and routing the GraphQL path through one transport seam is the right altitude. Inline comments mark two items to resolve before merge (the dependency on the `base_client` seam that #335 removes, and exposing `client_surface` as operator config) plus two smaller notes.

packages/mcp/src/pipefy_mcp/core/container.py
```diff
@@ -41,7 +41,11 @@ async def initialize_services(self, settings: Settings) -> None:
         Args:
             settings: Application settings with Pipefy credentials.
         """
-        pipefy = settings.pipefy
+        # model_validate (not model_copy) re-checks the stamped surface and reads
+        # only this dict, so it never re-loads env/TOML.
```
gbrlcustodio: **suggestion (non-blocking): stamp with model_copy instead of dump and re-validate**

`PipefySettings.model_validate({**settings.pipefy.model_dump(), "client_surface": "mcp"})` round-trips every field through a dump and a full re-validation to change one value. `settings.pipefy.model_copy(update={"client_surface": "mcp"})` does the same with less work. The comment justifies `model_validate` as re-checking the stamped surface, but the surface here is the hardcoded literal `mcp`, which cannot be invalid, and the round-trip also re-runs the SSRF host check on `base_url` for no benefit.

packages/sdk/src/pipefy_sdk/services/attachment_service.py
```diff
@@ -117,7 +122,10 @@ def __init__(
         self._card_service = card_service
         self._table_service = table_service
         self._s3_uploader: S3Uploader = s3_uploader or HttpxS3Uploader(
-            allowed_host_pattern=_ALLOWED_UPLOAD_HOST_RE
+            allowed_host_pattern=_ALLOWED_UPLOAD_HOST_RE,
+            client_headers=telemetry_headers(
```
gbrlcustodio: **suggestion (blocking): drop telemetry headers from the non-GraphQL paths**

Telemetry should ride only the GraphQL path (the Pipefy API). The S3 upload here and the export download in `observability_export_csv` target non-Pipefy hosts (S3, and whatever the export URL redirects to); the API never receives those requests, so the headers buy no attribution and only disclose the client surface and version to third parties.

Remove the `client_headers` param on `HttpxS3Uploader`, the `headers` params on `stream_bytes` / `download_bytes`, and the `telemetry_headers` calls in `attachment_service` / `observability_service`. That also drops the matching contract tests for those two paths.

(The OAuth handshake is left as is: it goes to the Pipefy IdP host and carries its own auth-component header set, not this one.)

gbrlcustodio (changes requested):

packages/sdk/src/pipefy_sdk/base_client.py
Danielmoraisg: Done. Rebased onto the executor seam from #335. The feature no longer touches `base_client.py` (gone) and rides only the GraphQL path:

- `HttpxGraphQLExecutor.__init__` takes a resolved `headers: dict[str, str] | None = None` and attaches it to the transport, alongside `url` and `cache_schema`.
- `build_executors` resolves the header set once and passes it to all three executors.

No more per-request `settings` read, and `__version__` is read in one place.

packages/sdk/src/pipefy_sdk/settings.py
Danielmoraisg: Went further than "keep the field, drop the env binding": `surface` is now a `PipefyClient`/`build_executors` constructor argument, not a `PipefySettings` field. It is no longer a config input at all, so the `PIPEFY_CLIENT_SURFACE` forge vector is closed structurally (nothing to exclude from env/TOML), the `model_copy` stamp from your container comment dissolves, and the feature stops touching `settings.py` and the CLI config. A regression test (`test_env_var_cannot_forge_surface`) asserts a default client still stamps `sdk` with `PIPEFY_CLIENT_SURFACE=mcp` set.

This deviates from the "keep it a programmatic field" suggestion. Happy to revert to a field + env-source block if you'd prefer that shape.

packages/mcp/src/pipefy_mcp/core/container.py
Danielmoraisg: Dissolved by the constructor-argument change. The container no longer clones settings to stamp the surface: it passes `settings.pipefy` through untouched and hands `surface="mcp"` to `PipefyClient`, so there is no `model_dump`/`model_validate` (or `model_copy`) round-trip, and no SSRF re-check on `base_url`.

packages/sdk/src/pipefy_sdk/services/attachment_service.py
Danielmoraisg: Done. The reworked feature rides only the GraphQL path. `attachment_service` and `observability_service` are untouched, there is no `client_headers`/`headers` plumbing on the S3 uploader or `stream_bytes`/`download_bytes`, and the contract tests for those two paths are gone. OAuth is left as-is (Pipefy IdP host, its own auth-component header set).

gbrlcustodio (approved):

Issue #312, opened 2026-06-18
Verify the production API accepts the user-facing issuer


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Present a real user token to the production API endpoint and confirm it accepts tokens from the user-facing issuer directly (the same-audience assumption). If it rejects on issuer, the design shifts toward a token alternative.

## Acceptance criteria

A documented result confirming or refuting direct acceptance; if refuted, a follow-up to adopt a token alternative.

## Dependencies

None. Should run early: a negative result reshapes the token strategy (#310/#311).

gbrlcustodio:
## Result: confirmed

A real user token from the user-facing issuer is accepted directly by the API. The same-audience assumption holds, so no token exchange or token alternative is needed (#310/#311 not triggered) and the OBO design in #297 stands.

Validated end to end against staging. One requirement surfaced: tokens must carry a resolvable user identity claim, otherwise the API returns 401. Tracked as a follow-up.

Issue #307, opened 2026-06-18
Rate limiting and abuse protection


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Treat the server as a load amplifier and enforce limits server-side: per-identity rate limiting (keyed on the authenticated subject and client, with a burst allowance for legitimate fan-out), per-user concurrency caps, downstream connection-pool limits and a circuit breaker that honors upstream `429`s, cost-weighting for expensive tools, and `429` + `Retry-After`. Place volumetric and unauthenticated controls at the edge or WAF.

## Acceptance criteria

A single identity flooding the server is throttled with `429` + `Retry-After`; the concurrency cap sheds beyond a max queue depth; an upstream `429` trips the breaker; limits are configurable.

## Dependencies

Builds on the resource-server role (#301) for the authenticated subject.

mocha06:
Dependency from the remote write-exposure policy (#442):

Write tools are now exposed on the remote profile relying only on API-side rate limits plus per-user identity — no per-user write governance in the MCP server itself. The tool-call middleware seam (`packages/mcp/src/pipefy_mcp/core/tool_middleware.py`) is where per-identity rate limiting, per-user concurrency caps, and cost-weighting from this issue will attach; today it carries only structured logging.

So this issue is the tracked prerequisite for scaling remote write exposure beyond the current conservative, API-permission-bounded set (the organization service-account tools are the first writes on the seed). The policy and this deferral are documented in `packages/mcp/AGENTS.md` under "Write tools on the remote profile".

Issue #303, opened 2026-06-18
Configurable transport host allowlist


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Make the transport allowed hosts and origins configurable so the server survives being fronted by a proxy or load balancer. DNS-rebinding protection otherwise rejects the public Host header with `421 Misdirected Request`.

## Acceptance criteria

Behind a proxy with a public Host header the server does not return `421`; the default localhost-only behavior is preserved when the allowlist is unset.

## Dependencies

Builds on the HTTP transport profile (#300).

PR #338, opened 2026-06-24
feat(mcp): OAuth resource-server profile for the HTTP transport

### LINKED ISSUE #301: OAuth resource-server role

gbrlcustodio:
Closes #301.

Makes the MCP HTTP transport (#300) optionally act as an OAuth 2.0 resource server. With the profile enabled it validates the inbound bearer on every request (RS256 via JWKS, issuer, expiry, and optional audience), serves the RFC 9728 protected-resource metadata, and returns `401` + `WWW-Authenticate` on a missing or invalid token. A malformed token is rejected (401), never surfaced as a 500. The server never resolves a global service identity for inbound validation.

This unblocks #302 (request-scoped identity, which consumes the validated token from the MCP auth context) and #307 (rate limiting keyed on the subject).

## In scope

- Per-request inbound bearer validation: signature against the issuer JWKS, issuer match, and a required `exp`; audience is checked when enabled.
- RFC 9728 protected-resource metadata and the `401` / `WWW-Authenticate` challenge, both served by FastMCP.
- Relaxing the loopback bind guard once the profile is enabled, since every request then carries a validated bearer.

## Out of scope

- Request-scoped on-behalf-of identity (#302). Even with a validated bearer, every call still runs as the single identity resolved at startup. The design does not preclude it: FastMCP auto-installs `AuthContextMiddleware`, so the validated token lands in `get_access_token()`.
- The configurable host / Origin allowlist for a proxied deployment (#303). Until it lands, an off-loopback bind has no DNS-rebinding protection, so treat off-loopback as a trusted-network deployment.

## Design

The validation core and the transport adapter are split across packages so the `mcp -> auth` dependency arrow stays one-way and no mcp-SDK type enters auth:

- `packages/auth` owns `JwtValidator`, a transport-agnostic primitive: it resolves the JWKS URL (explicit override or via OIDC discovery), builds a caching `PyJWKClient` once, decodes RS256 tokens, and returns raw claims or raises `TokenValidationError`. OIDC discovery now also exposes `jwks_uri`. Adds the `pyjwt[crypto]` dependency.
- `packages/mcp` owns `JwtTokenVerifier`, a thin adapter over `JwtValidator` that implements FastMCP's `TokenVerifier` protocol and maps validated claims onto the SDK's `AccessToken`. `ResourceServerSettings` (`settings.rs`, env prefix `PIPEFY_MCP_RS_`) carries the inbound issuer, audience, scopes, and public `resource_server_url`. `_build_resource_server_auth` wires the verifier and FastMCP's `AuthSettings` into the app.

`verify_audience` defaults `False` for the same-audience interim, which runs before the IdP issues an `aud` claim (open question #312). Production flips it on once the IdP maps `aud`; tests cover both directions.

## Configuration

Enable with `PIPEFY_MCP_RS_ENABLED=1`. Required when enabled: `PIPEFY_MCP_RS_ISSUER_URL` (inbound issuer that signs caller tokens) and `PIPEFY_MCP_RS_RESOURCE_SERVER_URL` (this server's public canonical URL, e.g. `https://mcp.pipefy.com/mcp`). Optional: `AUDIENCE`, `VERIFY_AUDIENCE`, `REQUIRED_SCOPES`, `JWKS_URI`, `ALLOW_INSECURE_URLS`.

## Tests

- auth: `jwks_uri` discovery parsing (present, absent, insecure rejected and allowed); `JwtValidator` over an in-test RSA keypair (valid, tampered, unknown key, expired, wrong issuer, wrong audience both directions, discovery-resolved `jwks_uri`).
- mcp: `JwtTokenVerifier` claim mapping (client_id fallback, list and string scopes, fractional `exp`, unmappable claims and validation failure return `None`); resource-server integration over `httpx.ASGITransport` (401 + `WWW-Authenticate`, metadata route 200, stdio has no auth, build wires auth); loopback guard relaxes when enabled.

All suites green; lint and format clean.

### DIFF
--- a/packages/auth/src/pipefy_auth/settings.py
+++ b/packages/auth/src/pipefy_auth/settings.py
@@ -375,4 +375,124 @@ class AuthSettings(BaseSettings):
         return self
 
 
-__all__ = ["AuthSettings"]
+class JwtValidationSettings(BaseSettings):
+    """How this process validates an inbound bearer when it acts as a resource server.
+
+    :class:`AuthSettings` configures the *outbound* side: how this process
+    authenticates *to* Pipefy. This is the inbound counterpart that feeds
+    :class:`~pipefy_auth.JwtValidator`. It is a separate model, not fields on
+    :class:`AuthSettings`, because only a transport running the OAuth
+    resource-server profile (today, the MCP ``--remote`` HTTP transport) consumes
+    it: a CLI that only authenticates outbound never carries resource-server knobs.
+
+    ``issuer_url`` is an override. Left unset, the consumer falls back to the
+    issuer this process already logs into (the :class:`OidcClient` issuer): in a
+    single-realm deployment the IdP that signs caller tokens is the same one we
+    authenticate to, so it need not be configured twice. Set it only when inbound
+    and outbound issuers diverge (token exchange, multi-tenant federation).
+
+    ``env_prefix="PIPEFY_JWT_"`` keeps these distinct from ``AuthSettings``'
+    ``PIPEFY_*`` vars; ``allow_insecure_urls`` is the one exception, aliased to the
+    shared ``PIPEFY_ALLOW_INSECURE_URLS`` so the whole auth domain has a single
+    insecure-URL posture rather than a per-model toggle.
+    """
+
+    model_config = SettingsConfigDict(
+        env_prefix="PIPEFY_JWT_",
+        env_file=".env",
+        env_file_encoding="utf-8",
+        extra="ignore",
+        populate_by_name=True,
+    )
+
+    @classmethod
+    def settings_customise_sources(
+        cls,
+        settings_cls: type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> tuple[PydanticBaseSettingsSource, ...]:
+        # Precedence: init_kwargs > env > dotenv > config.toml > file_secret.
+        return (
+            init_settings,
+            env_settings,
+            dotenv_settings,
+            PipefyTomlConfigSource(settings_cls),
+            file_secret_settings,
+        )
+
+    issuer_url: str | None = Field(
+        default=None,
+        description=(
+            "Override for the inbound OIDC issuer that signs caller tokens "
+            "(env: PIPEFY_JWT_ISSUER_URL). Left unset, the consumer falls back "
+            "to the issuer this process logs into; set it only when inbound and "
+            "outbound issuers diverge. The JWKS endpoint is resolved from the "
+            "issuer's discovery document unless jwks_uri is given."
+        ),
+    )
+
+    audience: str | None = Field(
+        default=None,
+        description=(
+            "Expected token audience (env: PIPEFY_JWT_AUDIENCE). Required when "
+            "verify_audience is true."
+        ),
+    )
+
+    verify_audience: bool = Field(
+        default=False,
+        description=(
+            "When true (env: PIPEFY_JWT_VERIFY_AUDIENCE), reject tokens whose "
+            "aud does not include audience. Defaults false for the same-audience "
+            "interim, which runs before the IdP issues an aud claim."
+        ),
+    )
+
+    jwks_uri: str | None = Field(
+        default=None,
+        description=(
+            "Explicit JWKS endpoint override (env: PIPEFY_JWT_JWKS_URI). When "
+            "unset, resolved from the issuer's discovery document."
+        ),
+    )
+
+    allow_insecure_urls: bool = Field(
+        default=False,
+        validation_alias=AliasChoices("PIPEFY_ALLOW_INSECURE_URLS"),
+        description=(
+            "When true (env: PIPEFY_ALLOW_INSECURE_URLS, shared with the outbound "
+            "side), inbound-validation URLs may use http:// and internal hosts; "
+            "local development only."
+        ),
+    )
+
+    def resolve_issuer_url(self, default: str | None) -> str | None:
+        """The inbound issuer: the explicit override, else ``default`` (the login issuer)."""
+        return self.issuer_url or default
+
+    @model_validator(mode="after")
+    def _validate_configuration(self) -> Self:
+        # issuer_url may carry a realm path (Keycloak-style), so only a query or
+        # fragment that would corrupt the .well-known concatenation is forbidden.
+        # The stripped value is persisted: surrounding whitespace in an env var
+        # would otherwise survive into jwt.decode(issuer=...), which compares the
+        # iss claim exactly and would reject every token.
+        if self.verify_audience and not self.audience:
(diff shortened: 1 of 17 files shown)


packages/auth/src/pipefy_auth/settings.py
```diff
@@ -375,4 +375,124 @@ def _validate_endpoint_urls(self) -> Self:
...
+    model_config = SettingsConfigDict(
+        env_prefix="PIPEFY_JWT_",
+        env_file=".env",
+        env_file_encoding="utf-8",
+        extra="ignore",
+        populate_by_name=True,
+    )
+
+    @classmethod
+    def settings_customise_sources(
```
gbrlcustodio, line 409: Developer note: this `settings_customise_sources` is identical across all five settings models. The stacked follow-up #339 extracts it (with the shared `model_config` keys) into a `PipefyBaseSettings` base in `pipefy_infra.config`, so this override is removed there. Intentional duplication in this PR.

packages/auth/src/pipefy_auth/settings.py
```diff
@@ -375,4 +375,124 @@ def _validate_endpoint_urls(self) -> Self:
...
+        description=(
+            "When true (env: PIPEFY_ALLOW_INSECURE_URLS, shared with the outbound "
+            "side), inbound-validation URLs may use http:// and internal hosts; "
+            "local development only."
+        ),
+    )
+
+    def resolve_issuer_url(self, default: str | None) -> str | None:
+        """The inbound issuer: the explicit override, else ``default`` (the login issuer)."""
+        return self.issuer_url or default
```
gbrlcustodio, line 474: Developer note: #339 canonicalizes the resolved issuer here, dropping a trailing slash so it matches the token `iss` claim under `jwt.decode`'s exact comparison.

packages/auth/src/pipefy_auth/settings.py
```diff
@@ -375,4 +375,124 @@ def _validate_endpoint_urls(self) -> Self:
...
+        # compares iss exactly, and jwks_uri into the JWKS fetch. A realm path is
+        # allowed, so only a query or fragment (which would corrupt URL building)
+        # is rejected.
+        for label in ("issuer_url", "jwks_uri"):
+            value = getattr(self, label)
+            if value is None:
+                continue
+            stripped = value.strip()
+            setattr(self, label, stripped)
+            security.assert_url_has_no_query_or_fragment(stripped, field_label=label)
```
gbrlcustodio, line 491: Developer note: #339 replaces this inline strip + shape-check + HTTPS/SSRF sequence with a reusable `security.sanitize_url(...)` helper shared across the settings models.

packages/mcp/src/pipefy_mcp/settings.py
```diff
@@ -88,6 +91,99 @@ def settings_customise_sources(
...
+    model_config = SettingsConfigDict(
+        env_prefix="PIPEFY_MCP_RS_",
+        env_file=".env",
+        env_file_encoding="utf-8",
+        extra="ignore",
+        populate_by_name=True,
+    )
+
+    @classmethod
+    def settings_customise_sources(
```
gbrlcustodio, line 122: Developer note: same duplicated `settings_customise_sources` as the auth models. #339 extracts it into the shared `PipefyBaseSettings` base, removing this override.

packages/mcp/src/pipefy_mcp/settings.py
```diff
@@ -88,6 +91,99 @@ def settings_customise_sources(
...
+    @model_validator(mode="after")
+    def _validate_configuration(self) -> Self:
+        if self.resource_server_url is None:
+            return self
+        # Persist the stripped value: surrounding whitespace in an env var would
+        # otherwise survive into the RFC 9728 resource identifier. The /mcp
+        # endpoint path is expected, so only a query or fragment is forbidden.
+        stripped = self.resource_server_url.strip()
+        self.resource_server_url = stripped
+        security.assert_url_has_no_query_or_fragment(
```
gbrlcustodio, line 178: Developer note: #339 collapses this inline strip + shape-check + HTTPS/SSRF sequence into the shared `security.sanitize_url(...)` helper.

gbrlcustodio (commented):
Developer notes for reviewers: the settings code this PR introduces is refactored by the stacked follow-up #339; the inline notes below point to what moves there. The duplicated source-chain and inline URL validation here are intentional in this PR and cleaned up in #339.

packages/mcp/src/pipefy_mcp/server.py
```diff
@@ -107,37 +158,50 @@ def build_pipefy_mcp_server(
...
-    anyone who can reach the port, so HTTP is restricted to loopback. Relax this
-    guard when inbound auth lands (#301). (The filesystem tools, e.g. the
-    attachment uploads, also only make sense on loopback, where the server shares
-    the client's disk.)
+    Once the resource-server profile is configured, every request carries a
+    validated bearer, so a non-loopback bind is allowed. The configurable host /
+    Origin allowlist for a proxied deployment is #303.
     """
-    if host in _LOOPBACK_HOSTS:
+    if host in _LOOPBACK_HOSTS or resource_server_configured:
```
gbrlcustodio: Developer note: off-loopback exemption gates only on `resource_server_configured`, so with `verify_audience` off (the default) the bind accepts any token the issuer signed, and every call still runs as the single startup identity. This is intentional and not addressed in this PR, and per-request on-behalf-of identity is tracked separately as #302. 

Tightening the exemption to require `verify_audience`/`required_scopes` is deferred until then. Flagging so it is not re-raised as a finding here.

packages/auth/src/pipefy_auth/verification.py
```diff
@@ -0,0 +1,127 @@
...
+        The discovery path reaches the IdP through
+        :func:`fetch_provider_metadata`, which enforces this gate on the issuer
+        before trusting its advertised ``jwks_uri``. An explicit override skips
+        discovery, so the primitive enforces the same invariant here rather than
+        handing an unchecked key-fetch URL (``http://`` or an internal host) to
+        :class:`~jwt.PyJWKClient`. A realm path is allowed; only a query or
+        fragment is rejected.
+        """
+        stripped = jwks_uri.strip()
+        security.assert_url_has_no_query_or_fragment(stripped, field_label="jwks_uri")
```
gbrlcustodio, line 100: Developer note: this inline strip + shape-check + HTTPS/SSRF sequence is the same one #339 collapses into `security.sanitize_url(...)`. The stacked follow-up routes this jwks_uri override through that single chokepoint alongside the settings models, so this hand-rolled form is replaced there.

gbrlcustodio (commented):
Developer note for reviewers (not review feedback). Same forward-reference to #339 as the other inline notes on this PR.

PR #339, opened 2026-06-25
refactor(settings): shared base classes and auth env namespacing

gbrlcustodio:
## Summary

Two related settings refactors, plus the two review-driven fixes they grew out of, split off from the OAuth resource-server PR (#338) into their own PR.

- **`sanitize_url`**: extract the strip + shape + HTTPS/SSRF gate into one reusable `pipefy_infra.security.sanitize_url` that every settings model's validator calls, replacing the per-model inline checks.
- **Shared settings base classes**: `PipefyBaseSettings` (env/`.env` config plus the `init > env > dotenv > config.toml > file_secret` source chain) and `InsecureUrlSettings` (the shared `allow_insecure_urls` field). The five settings models across infra/sdk/auth/mcp now inherit these instead of each repeating the source chain and config keys. Closes the review finding on settings boilerplate duplication.
- **`PIPEFY_AUTH_` namespacing (breaking)**: `AuthSettings` moves off the bare `PIPEFY_` prefix onto its own `PIPEFY_AUTH_` domain, so its own knobs no longer sit in the global env namespace alongside the SDK foundation. The deployment-wide `PIPEFY_BASE_URL` / `PIPEFY_ALLOW_INSECURE_URLS` and the curated credential names are kept through explicit aliases.
- **Trailing-slash fix**: the resolved inbound issuer drops a trailing slash so it matches the token `iss` claim exactly.

## Breaking change

- `PIPEFY_AUTH_URL` -> `PIPEFY_AUTH_ISSUER_URL`
- `PIPEFY_DISABLE_STORED_SESSION` -> `PIPEFY_AUTH_DISABLE_STORED_SESSION`
- `PIPEFY_KEYCHAIN_BACKEND` -> `PIPEFY_AUTH_KEYCHAIN_BACKEND`
- TOML keys `auth_url` / `auth_client_id` -> `issuer_url` / `client_id`
- The legacy `PIPEFY_OAUTH_CLIENT` / `PIPEFY_OAUTH_SECRET` aliases and their deprecation shim are removed outright (no grace period).

`AuthSettings` uses `extra="ignore"`, so a stale old name is dropped silently. Full table and migration steps in `docs/MIGRATION.md`.

## Testing

Per-package suites all green: infra 143, sdk 954, auth 164, cli 317, mcp 1398. ruff and ruff-format clean.

## Note on stacking

This PR is stacked on #338 (OAuth resource-server): the base-class extraction touches `JwtValidationSettings`, which #338 introduces. It targets `feat/mcp-oauth-resource-server` and shows only the four settings commits. Merge #338 first; GitHub then retargets this PR to `dev`.


gbrlcustodio:
Superseded by #342, which reworks this into a library/application split with one injected DeploymentConfig (the libraries become env-free pure value objects; all env reading moves to the application edge). Same operator-facing env contract / breaking renames.

PR #343, opened 2026-06-30
refactor(auth): parse credentials into a ResolvedAuth sum type

gbrlcustodio:
## Summary

Reworks the auth credential resolution to follow parse-don't-validate. `resolve_pipefy_auth` now parses the precedence chain into a closed sum type instead of building an `httpx.Auth` directly, so the winning tier is kept in the type rather than recovered after the fact.

- New `ResolvedAuth = StaticTokenAuth | ServiceAccountAuth | StoredSessionAuth` in `pipefy_auth.resolver`. Each variant carries its tier name as a `ClassVar`.
- `resolve_pipefy_auth(...) -> ResolvedAuth | None` is the parse (same precedence and short-circuit behavior, including the keychain probe for the stored-session tier).
- `build_httpx_auth(resolved) -> httpx.Auth` is the builder, total over the sum: no `None` branch, no fallthrough.
- Call sites (`mcp.core.container`, `cli.auth`) read `resolved.tier` and call `build_httpx_auth(resolved)` instead of going through the `tier_for` isinstance reverse lookup.

Behavior is preserved: `PipefyClient` still receives the same `httpx.Auth` it did before; only the internal seam changed.

`tier_for` and its `_TIER_BY_AUTH_TYPE` table are now redundant in `src` (nothing builds an `httpx.Auth` it then needs to classify) but are kept in this PR since they are public API with their own tests. Removing them is a follow-up.

Also documents the principle in `AGENTS.md` as a companion to the existing "Type validation at boundaries, not inside" section.

## Out of scope

The half-configured service-account case (client_id set without client_secret silently demoting to the next tier) lives on `AuthSettings`, not the resolver, and is left untouched here.

## Testing

- `uv run python -m pytest packages/auth/tests` -> 175 passed
- `uv run python -m pytest packages/cli/tests` -> 319 passed, 13 skipped
- `uv run python -m pytest packages/mcp/tests` -> 1384 passed, 16 skipped
- `uv run python -m pytest packages/sdk/tests` -> 964 passed, 17 skipped
- `uv run ruff check` on touched files -> clean

## Docs

`AGENTS.md` updated (new "Parse, don't validate" subsection). No config or tool behavior changed.


packages/auth/src/pipefy_auth/resolver.py
```diff
@@ -41,34 +41,35 @@ class ServiceAccount:
...
+    """Resolved static-token tier: a pre-issued bearer, stripped and non-blank."""
+
+    token: str
+
+
+@dataclass(frozen=True)
+class ServiceAccountAuth:
+    """Resolved service-account tier: the OAuth2 client-credentials inputs."""
 
+    credentials: ServiceAccount
```
adriannoes, line 55: StaticTokenAuth and ServiceAccountAuth pick up the default dataclass repr, so anything that prints `resolved` — a pytest assertion failure, an accidental `logger.debug(resolved)`, a REPL session — shows the bearer token or `client_secret` in plain text. 

Worth being upfront that this mirrors an existing gap rather than introducing a brand new one: `pipefy_auth.ServiceAccount` has carried `client_secret` in its default repr since before this PR, and `pipefy_cli.BearerToken` does the same for the raw token value, both already exported as public API. 

Since this refactor is already adding two more public types to the same family, it feels like the right moment to redact all of them together (`repr=False` on the secret fields, or a custom `__repr__`) rather than letting a third instance of the same gap ship. Not something I'd hold this PR for, but worth a quick follow-up so the pattern doesn't keep spreading — you'd know it's fixed when `repr(StaticTokenAuth("x"))`, `repr(ServiceAccountAuth(...))`, `repr(ServiceAccount(...))`, and `repr(BearerToken(...))` all redact their secret field.

packages/auth/tests/test_resolver.py
```diff
@@ -89,8 +89,8 @@ def test_service_account_wins_when_no_static_token_and_short_circuits_stored(
     resolved = resolve_pipefy_auth(
         service_account=_service_account(), oidc_client=_oidc()
     )
-    assert isinstance(resolved, OAuth2ClientCredentials)
-    assert tier_for(resolved) == SERVICE_ACCOUNT_TIER
+    assert isinstance(resolved, ServiceAccountAuth)
```
adriannoes, line 102: Tests assert `isinstance(resolved, StaticTokenAuth)` / `isinstance(resolved, ServiceAccountAuth)` but stop there — `build_httpx_auth`'s mapping for these two tiers only gets indirect coverage downstream (e.g. the CLI test that checks `PipefyClient` got a `StaticBearerAuth`). 

The stored-session tier already has direct coverage further down (`build_httpx_auth(resolved)` asserted as `RefreshableBearerAuth`), so it'd be a small, consistent addition to do the same here: assert `build_httpx_auth(resolve_pipefy_auth(static_token=...))` is a `StaticBearerAuth`, and the service-account equivalent is an `OAuth2ClientCredentials`. That closes the loop on the `match` in `build_httpx_auth` being the actual transport contract, not just `resolve_pipefy_auth`'s parsing.

adriannoes (dismissed):
## Summary

This is a clean execution of parse-don't-validate for the credential precedence chain: `resolve_pipefy_auth` now returns a closed `ResolvedAuth` sum type instead of an `httpx.Auth`, and `build_httpx_auth` is the single, exhaustive (`assert_never`-backed) place that builds the transport object. 

I checked out `refactor/auth-resolved-auth-sum-type` @ `efb81411` locally, ran `uv run pytest packages/auth/tests/test_resolver.py packages/cli/tests/test_auth.py -m "not integration" -q` (29 passed) and `uv run ruff check` on the touched packages (clean); CI is green on both lint and test. No blockers — approving.

## Also noted

- **Secrets in repr** (inline note on `resolver.py`) — `StaticTokenAuth` / `ServiceAccountAuth` use the default dataclass repr, so the bearer token / `client_secret` print in plain text on a pytest failure or stray `repr(resolved)`. Worth flagging that this mirrors an existing gap in `ServiceAccount` and `BearerToken` (both already public, both pre-dating this PR) rather than a regression this PR introduces — but since two more public types just joined the same family, it's a good moment to redact all four together. Not blocking.
- **`build_httpx_auth` coverage** (inline note on `test_resolver.py`) — the static-token and service-account arms of the `match` only get indirect coverage today; the stored-session arm already asserts `build_httpx_auth(resolved)` directly. A couple of parametrized assertions would close that gap and lock the transport contract, not just the parsing.

packages/auth/src/pipefy_auth/resolver.py
gbrlcustodio: Fixed in 2f937e3: added `field(repr=False)` to all four secret fields. `StaticTokenAuth.token`, `ServiceAccount.client_secret` (which also covers `ServiceAccountAuth` via its nested `credentials`), and `pipefy_cli.BearerToken.value`. Verified `repr()` of each redacts the secret.

packages/auth/tests/test_resolver.py
gbrlcustodio: Fixed in cc60e8d: added two direct tests asserting `build_httpx_auth(resolve_pipefy_auth(static_token=...))` is a `StaticBearerAuth` and the service-account equivalent is an `OAuth2ClientCredentials`, matching the stored-session coverage.

PR #344, opened 2026-06-30
refactor(auth,cli): parse auth methods into ResolvedAuth

gbrlcustodio:
Follows #343, which gave `resolve_pipefy_auth` a `ResolvedAuth` sum type but left its sibling `detect_pipefy_tiers` returning `list[str]` tier names. This finishes the migration.

## detect returns ResolvedAuth

The `list[str]` gap forced four awkward spots in the `pipefy auth status` path, all now resolved:

- `tier == STATIC_TOKEN_TIER` string compare, now a match on the variant in `to_display_source`
- a `# type: ignore[misc]` cast of the detected list to `DisplaySource`, now typed end to end
- an `auth.oidc_client is None` re-check raising an "invariant broken" `RuntimeError`, now gone: the OIDC client is read off a `StoredSessionAuth` that holds it by construction
- two ways to ask "is stored-session involved" (string membership vs isinstance), now isinstance throughout

`DisplaySource` moved next to the functions that produce it (`pipefy_cli.auth`), which also broke the import cycle behind the cast. The locked `auth status` JSON wire schema is unchanged.

A follow-up `/simplify` pass found the command still detected twice: it flattened to display strings, then re-resolved to recover the typed winner, running the precedence chain and its keychain probe twice. It now derives the display list, the active method (the precedence-first head), and the stored-session populate from one `list[ResolvedAuth]`. `resolve_pipefy_auth` keeps its own short-circuiting path: it must not probe the keychain when a higher-precedence token already wins, so it is not folded into the detect function. The now-dead `STATIC_TOKEN_TIER` / `SERVICE_ACCOUNT_TIER` / `STORED_SESSION_TIER` constants are removed (no consumers after the sum-type migration).

## Naming: "auth method" vs "tier"

`detect_pipefy_tiers` returned `list[ResolvedAuth]` but the CLI assigned it to a `tiers` variable from a `detect_cli_sources` function: one value wearing two labels, neither saying what it is. Each value is a way to authenticate, so it is now named that:

- `detect_pipefy_tiers` -> `detect_pipefy_auth_methods`
- `detect_cli_sources` -> `detect_cli_auth_methods` (it returns the resolved methods, not display sources)
- locals/loop vars: `tiers`/`tier` -> `auth_methods`/`method`

"tier" is kept where it means precedence rank ("short-circuits lower tiers"); "source" is kept for the CLI display labels (`DisplaySource`, `auth_source`). Kind and rank are now distinct words.

## Tests

`packages/auth` 175 passed, `packages/cli` 319 passed / 13 skipped, ruff clean. The wire-output tests and the display-label tests (on the now-public `to_display_source`) pass unchanged.

### DIFF
--- a/packages/auth/tests/test_resolver.py
+++ b/packages/auth/tests/test_resolver.py
@@ -201,6 +201,31 @@ def test_detect_skips_tiers_without_credentials(monkeypatch):
 
 
 @pytest.mark.unit
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        {
+            "static_token": "T",
+            "service_account": _service_account(),
+            "oidc_client": _oidc(),
+        },
+        {"service_account": _service_account(), "oidc_client": _oidc()},
+        {"oidc_client": _oidc()},
+        {},
+    ],
+)
+def test_detect_head_is_the_resolved_method(monkeypatch, kwargs):
+    """``auth status`` treats ``detect(...)[0]`` as active; both draw from the
+    same generator, so its head equals ``resolve(...)`` (and empty ⟺ ``None``)."""
+    monkeypatch.setattr(
+        "pipefy_auth.resolver.load_session", lambda **_: _stored_session()
+    )
+    methods = detect_pipefy_auth_methods(**kwargs)
+    resolved = resolve_pipefy_auth(**kwargs)
+    assert (methods[0] if methods else None) == resolved
+
+
+@pytest.mark.unit
 def test_missing_auth_message_mentions_all_tiers():
     msg = missing_auth_message()
     assert "PIPEFY_TOKEN" in msg
--- a/packages/auth/tests/test_resolver.py
+++ b/packages/auth/tests/test_resolver.py
@@ -14,9 +14,6 @@ from pipefy_auth.bearer import (
 from pipefy_auth.identity import OidcClient
 from pipefy_auth.refresh import RefreshError
 from pipefy_auth.resolver import (
-    SERVICE_ACCOUNT_TIER,
-    STATIC_TOKEN_TIER,
-    STORED_SESSION_TIER,
     ServiceAccount,
     ServiceAccountAuth,
     StaticTokenAuth,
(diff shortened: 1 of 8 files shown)


packages/auth/tests/test_resolver.py
```diff
@@ -182,23 +179,25 @@ def test_detect_lists_every_configured_tier_in_precedence_order(monkeypatch):
...
     )
-    assert tiers == [
-        STATIC_TOKEN_TIER,
-        SERVICE_ACCOUNT_TIER,
-        STORED_SESSION_TIER,
+    assert [type(method) for method in methods] == [
+        StaticTokenAuth,
+        ServiceAccountAuth,
+        StoredSessionAuth,
     ]
```
adriannoes, line 191: The `auth status` path now treats `auth_methods[0]` as the active method without a second resolve call, which is sound as long as detect and resolve stay in sync. I'd add one parametrized test that compares `detect_pipefy_auth_methods(...)[0]` with `resolve_pipefy_auth(...)` for the common credential combos — cheap insurance if someone edits one gate and not the other. You'll know it's covered when that test fails if the two functions diverge.

adriannoes (dismissed):
Clean finish to the `ResolvedAuth` migration from #343: detect returns typed variants end-to-end, `auth status` runs a single detection pass, and the string compares / `# type: ignore` / duplicate keychain probe are gone.

No blockers; wire JSON contract for `auth status` is unchanged and existing `test_auth_status.py` scenarios still cover precedence and masking.

## Also noted

- `auth status` trusts `auth_methods[0]` without re-calling `resolve_pipefy_auth`. 
- a parametrized parity test (`detect(...)[0]` ≡ `resolve(...)`) would catch future gate drift between the two functions. Left as an optional follow-up, not merge-blocking.

packages/auth/tests/test_resolver.py
gbrlcustodio: Fixed in 54db156. I went a step past the suggested test and made the invariant structural: `resolve_pipefy_auth` and `detect_pipefy_auth_methods` now both pull from one lazy `_detected_methods` generator, with resolve taking its first item and detect taking all of them. So `detect(...)[0] == resolve(...)` holds by construction rather than by convention, and the two cannot drift when someone edits a gate. resolve's short-circuit still falls out of generator laziness, so the stored-session keychain read stays unevaluated when a higher tier wins (the `mock_load.calls == 0` tests confirm it). Kept a parametrized test as executable documentation of the contract.

PR #345, opened 2026-06-30
refactor(auth): fold the inbound audience pair into an AudiencePolicy sum type

gbrlcustodio:
Follows #344's observation that "verify_audience requires audience" is a sum type wearing two fields, and finishes that thought for the inbound-validation config.

## The loose shape

`JwtValidationSettings` carries the inbound audience config as two co-dependent fields, `audience: str | None` and `verify_audience: bool`, whose only legal states are "skip the audience check" or "verify against a present audience". Nothing in the types rules out the third combination, and `JwtValidator.validate()` re-derived the decision on every request (the `audience if verify else None` ternary plus a parallel `verify_aud` flag).

## The sum type

`AudiencePolicy = SkipAudience | RequireAudience`. `RequireAudience` cannot exist without its value, so "verify against nothing" is unrepresentable. `JwtValidator` takes the policy and resolves it into the `jwt.decode` kwargs once in `__init__`, since the policy is fixed and `validate()` is the per-request hot path. The match is total with `assert_never`, mirroring `build_httpx_auth` on the outbound side.

## Where each rule lives

- `JwtValidationSettings` stays a pure reader for the two env vars (`PIPEFY_JWT_AUDIENCE`, `PIPEFY_JWT_VERIFY_AUDIENCE`); env/dotenv/toml layering is unchanged. A `model_validator` rejects verify-without-audience at construction, alongside the existing URL checks, so the illegal pair never escapes the reader.
- The MCP resource-server composition root (`build_resource_server_auth`) folds the pair into an `AudiencePolicy` and builds the `JwtValidator` directly.

## Scope

This is deliberately the audience slice only. It does not add a `Resolved*` config object: the audience posture is the one part that needed a domain type, and a frozen bag of bare-`str` URL fields would carry no guarantee its constructor did not enforce (a hand-written instance with a bad URL would still construct). Turning the issuer and JWKS URLs into a self-validating type is a separate change, kept orthogonal to this one.

## Tests

`packages/auth` 177 passed, `packages/mcp` 1386 passed, ruff clean. The verify-without-audience misconfiguration is asserted at settings construction; the fold outcome (skip by default, required audience when verifying) is pinned in the resource-server builder tests. Issuer resolution and the "no issuer" `RuntimeError` are unchanged.

### DIFF
--- a/packages/auth/src/pipefy_auth/__init__.py
+++ b/packages/auth/src/pipefy_auth/__init__.py
@@ -60,10 +60,8 @@ from pipefy_auth.verification import (
     AudiencePolicy,
     JwtValidator,
     RequireAudience,
-    ResolvedJwtValidation,
     SkipAudience,
     TokenValidationError,
-    resolve_jwt_validation,
 )
 
 __all__ = [
@@ -71,9 +69,7 @@ __all__ = [
     "AuthSettings",
     "JwtValidationSettings",
     "RequireAudience",
-    "ResolvedJwtValidation",
     "SkipAudience",
-    "resolve_jwt_validation",
     "CallableBearerAuth",
     "DEFAULT_AUTH_CLIENT_ID",
     "DiscoveryPolicy",
--- a/packages/auth/src/pipefy_auth/__init__.py
+++ b/packages/auth/src/pipefy_auth/__init__.py
@@ -56,11 +56,24 @@ from pipefy_auth.storage import (
     load_session,
     store_session,
 )
-from pipefy_auth.verification import JwtValidator, TokenValidationError
+from pipefy_auth.verification import (
+    AudiencePolicy,
+    JwtValidator,
+    RequireAudience,
+    ResolvedJwtValidation,
+    SkipAudience,
+    TokenValidationError,
+    resolve_jwt_validation,
+)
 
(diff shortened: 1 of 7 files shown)


adriannoes (approved):
Clean follow-through on the #344 sum-type direction for inbound JWT validation. `AudiencePolicy = SkipAudience | RequireAudience` makes "verify against nothing" unrepresentable, `JwtValidator` resolves `jwt.decode` kwargs once in `__init__`, and the fold at `build_resource_server_auth` fits the per-app composition root pattern in `AGENTS.md`. 

Dropping `ResolvedJwtValidation` is a sensible scope cut — bare URL strings would not be self-guaranteeing yet.

Review included local checkout+tests on `c44f0ef`: ruff clean, 43 targeted tests + 181 auth package tests passed. CI lint + test green. **Approve-ready** — no blockers.

## Also noted

- The strip-then-validate path for all-whitespace `audience` (`"   "` → `""` → rejected when `verify_audience=True`) is documented in `_validate_audience` but not covered by a regression test. 
- One cheap test with `audience="   "` and `verify_audience=True` would pin the env-var footgun without blocking merge.

Danielmoraisg (approved):

PR #349, opened 2026-07-02
feat: add automation execution metrics capability

rebecaventura:
## Summary

Adds an automation execution metrics capability across the SDK, MCP, and CLI. It reports per-automation metrics (`totalRuns`, `successRate`, `failureRate`, `averageDuration`, `lastRun`) over a rolling window (`FIFTEEN_MINUTES`, `SIXTY_MINUTES`, `TWELVE_HOURS`, `TWENTY_FOUR_HOURS`), backed by the `automations.executionMetrics` GraphQL field.

The `automations` query returns partial results: it can carry metrics for the automations a token may read alongside per-automation errors (for example `PERMISSION_DENIED`) for those it may not, in one response. The capability surfaces this as a partial success, returning a normalized `{automations, partial_errors, page_info}` payload rather than failing outright, so one denied automation in a batch does not drop the rest. A response with no readable data at all is treated as a failure, not an empty success.

## What's included

Capability parity across all three surfaces (per the "Adding a New Capability" checklist in AGENTS.md):

- **SDK**: `GET_AUTOMATION_EXECUTION_METRICS_QUERY`, `ObservabilityService.get_automation_execution_metrics`, and the `PipefyClient` passthrough.
- **MCP**: a read-only `get_automation_execution_metrics` tool, validating the organization id, the optional automation id list, the optional repo id, the period, and the page size.
- **CLI**: `pipefy usage execution-metrics`.
- **Docs and skills**: the `docs/parity.md` row and the observability `SKILL.md` entry.

## Partial-execution handling

Supporting the partial path required a fix in the GraphQL executor. gql 4 raises `TransportQueryError` whenever a response carries GraphQL errors, even for a partial success, so an approach relying on `get_execution_result=True` never returned partial data. `execute_query_allow_partial` now reconstructs the partial data and the raw error dicts from the exception into an owned `PartialQueryResult`, which keeps gql and graphql-core types out of the service layer.

Two smaller cleanups came with it: the period values are single-sourced from one SDK constant that the MCP validation set and the CLI help both derive from, and a shared `validate_optional_tool_id_list` helper replaces a hand-rolled list-of-ids validation loop.

## Pagination

The server caps the `automations` connection at 50 per page, so an unbounded query silently dropped everything past the first 50 automations. The query, service, client, MCP tool, and CLI now take `first` and `after` (the tool bounds `first` at the server's 50-per-page cap), and the normalized payload carries `page_info` (`hasNextPage`, `endCursor`) so callers page through the rest with `after`.

## Testing performed

- `uv run ruff check` and `uv run ruff format --check`: clean.
- `uv run pytest`: full suite green, including the SDK executor and observability service tests, the MCP tool tests, the CLI help golden, and the tool-parity drift guards.
- MCP surface through an in-memory session with a fake client: happy path, partial success (the payload carries `partial_errors` with `automation_id`), total failure (an error payload, not an empty success), pagination (`first`/`after` forwarded, `page_info` returned), and input validation for period, empty id list, empty organization id, and out-of-range page size.
- Real API through `pipefy usage execution-metrics`:
  - A mixed list (one readable automation plus one denied id) returned the readable automation's metrics alongside `partial_errors: [{automation_id, code: PERMISSION_DENIED, correlation_id}]`.
  - An unknown organization returned an error rather than an empty success.
  - Pagination: `--first 1` returned one automation with `page_info.hasNextPage: true`; paging with `--after <endCursor>` advanced to the next automation. `--first 51` came back with 50 and `hasNextPage: true`, confirming the server's page cap.

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/observability_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/observability_tools.py
@@ -32,6 +32,9 @@ _VALID_EXECUTION_METRICS_PERIODS = frozenset(AUTOMATION_EXECUTION_METRICS_PERIOD
 # Pagination
 _MIN_PAGE_SIZE = 1
 _MAX_PAGE_SIZE = 100
+# The automations connection caps a page at 50; asking for more is rejected up front
+# rather than silently returning 50.
+_MAX_EXECUTION_METRICS_PAGE_SIZE = 50
 
 # CSV / export download limits
 _MIN_CSV_CHARS = 256
@@ -415,15 +418,19 @@ class ObservabilityTools:
             automation_ids: list[PipefyId] | None = None,
             repo_id: PipefyId | None = None,
             period: str = "SIXTY_MINUTES",
+            first: int = 30,
+            after: str | None = None,
             debug: bool = False,
         ) -> dict[str, Any]:
-            """Get execution metrics (totalRuns, successRate, failureRate, averageDuration, lastRun) for one or more automations over a rolling window. Returns metrics for the automations you can access plus a `partial_errors` list naming any that were denied (PERMISSION_DENIED); this is a partial result, not a hard failure. `period`: FIFTEEN_MINUTES, SIXTY_MINUTES (default), TWELVE_HOURS, or TWENTY_FOUR_HOURS.
+            """Get execution metrics (totalRuns, successRate, failureRate, averageDuration, lastRun) for one or more automations over a rolling window. Returns metrics for the automations you can access plus a `partial_errors` list naming any that were denied (PERMISSION_DENIED); this is a partial result, not a hard failure. Results are paginated: `page_info` carries `hasNextPage` and `endCursor`, and the server caps a page at 50 automations, so pass `endCursor` back as `after` to fetch the rest. `period`: FIFTEEN_MINUTES, SIXTY_MINUTES (default), TWELVE_HOURS, or TWENTY_FOUR_HOURS.
 
             Args:
                 organization_id: Organization ID (numeric org id, same as in the Pipefy URL).
                 automation_ids: Optional automation IDs to fetch metrics for. Omit to fetch metrics for every automation in the organization (optionally scoped by repo_id). When provided, it must contain at least one ID.
                 repo_id: Optional pipe/repo ID to scope the query.
                 period: AutomationExecutionMetricsPeriod (FIFTEEN_MINUTES, SIXTY_MINUTES, TWELVE_HOURS, TWENTY_FOUR_HOURS). Defaults to SIXTY_MINUTES, matching the API.
+                first: Page size, 1 to 50 (default 30).
+                after: Cursor from a previous page's page_info.endCursor.
                 debug: When True, append GraphQL codes and correlation_id to errors.
             """
             client = get_pipefy_client(ctx)
@@ -445,12 +452,21 @@ class ObservabilityTools:
                         f"{sorted(_VALID_EXECUTION_METRICS_PERIODS)}."
                     ),
                 )
+            if not _MIN_PAGE_SIZE <= first <= _MAX_EXECUTION_METRICS_PAGE_SIZE:
+                return build_observability_error_payload(
+                    message=(
+                        f"Invalid 'first': must be between {_MIN_PAGE_SIZE} and "
+                        f"{_MAX_EXECUTION_METRICS_PAGE_SIZE}."
+                    ),
+                )
             try:
                 raw = await client.get_automation_execution_metrics(
                     organization_id,
                     normalized_ids,
                     repo_id=normalized_repo_id,
                     period=period,
+                    first=first,
+                    after=after,
                 )
             except ValueError as exc:
                 return build_observability_error_payload(message=str(exc))
--- a/packages/mcp/src/pipefy_mcp/tools/observability_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/observability_tools.py
@@ -22,6 +22,14 @@ from pipefy_mcp.tools.validation_helpers import validate_tool_id
 
 _VALID_PERIODS = {"current_month", "last_month", "last_3_months"}
 
+# AutomationExecutionMetricsPeriod enum: rolling windows for execution metrics.
+_VALID_EXECUTION_METRICS_PERIODS = {
+    "FIFTEEN_MINUTES",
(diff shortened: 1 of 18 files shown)


packages/sdk/src/pipefy_sdk/graphql_executor.py
```diff
@@ -27,6 +40,18 @@ async def execute_query(
...
 
 
+class PartialGraphQLExecutor(GraphQLExecutor, Protocol):
+    """:class:`GraphQLExecutor` plus a partial-tolerant execute path.
+
+    For services whose queries mix ``data`` with per-node ``errors`` in one
+    response; everything else depends on the narrower protocol.
+    """
+
+    async def execute_query_allow_partial(
```
Danielmoraisg, line 50: **note:** We have created this issue to later improve the executor so it has only one entrypoint independent of the response type https://github.com/pipefy/ai-toolkit/issues/354

packages/mcp/src/pipefy_mcp/tools/observability_tools.py
```diff
@@ -399,6 +410,112 @@ async def get_ai_credit_usage(
...
+                    event_id=event_id,
+                    active=active,
+                    search=normalized_search,
+                    sort_by=sort_by,
+                    sort_order=sort_order,
+                    period=period,
+                    first=first,
+                    after=after,
+                )
+            except ValueError as exc:
```
adriannoes, line 505: Total failures (unknown org, no org access) come back as `ValueError` from the SDK and land here, which returns a plain message and never looks at `debug`. 

The docstring above promises that `debug=True` appends GraphQL codes and correlation_id to errors, but that only happens on the transport branch below. Per-automation denials are fine: they already carry `correlation_id` in `partial_errors`. For this top-level failure path, either narrow the docstring or route through the debug-aware handler, and a small test (like `test_get_ai_credit_usage_value_error_from_client`) would lock in the envelope.

skills/observability/pipefy-observability/SKILL.md
```diff
@@ -35,6 +35,7 @@ Monitor AI agent and automation execution, usage stats, credit consumption, and
 | `get_automation_logs_by_repo` | — (CLI v0.3+) | Yes | Automation logs filtered by pipe UUID. |
 | `get_agents_usage` | — (CLI v0.3+) | Yes | Org-level AI agent execution count and trends. |
 | `get_automations_usage` | — (CLI v0.3+) | Yes | Org-level automation execution stats. |
+| `get_automation_execution_metrics` | `pipefy usage execution-metrics` | Yes | Per-automation execution metrics (totalRuns, success/failure rate, avg duration, lastRun) over a rolling window; partial success returns `partial_errors` for denied ids. |
```
adriannoes, line 38: Nice addition. While you're in this file: the frontmatter and the intro on line 12 still say "10 MCP tools", but the table now has 11 entries including this row. Agents use that count to scope the domain, so a quick bump to 11 in both places would keep routing accurate.

adriannoes (dismissed):
Reviewed with a local checkout: ruff clean, 113 targeted tests green, CI green. This is a solid, mergeable capability (SDK + MCP + CLI + parity). The partial GraphQL path (`PartialQueryResult`, single failure decision in `_normalize_execution_metrics_result`) is well tested and matches gql 4 behavior. No blockers from my side.

Daniel already left a note on [#354](https://github.com/pipefy/ai-toolkit/issues/354) for collapsing the executor entrypoints; I won't repeat that.

## Worth a quick follow-up in this PR

**1. Skill tool count** (inline on `SKILL.md`): the table now has 11 tools but the frontmatter and intro still say 10. Easy fix while you're in the file.

**2. CLI wiring test:** `test_usage_credits_invokes_client` already guards the credits command. A matching smoke test for `usage execution-metrics` would catch a broken import or wrong facade method in CI.

**3. Total-failure path on the MCP tool:** when the org lookup fails, the SDK raises `ValueError` and the tool handles it with `build_observability_error_payload(message=str(exc))`, which ignores `debug` even though the docstring says `debug=True` appends GraphQL codes and correlation_id. The transport-error branch does honor `debug`; per-automation denials already surface `correlation_id` in `partial_errors`. Either narrow the docstring for total failures or route that message through the debug-aware handler, and add a test mirroring `test_get_ai_credit_usage_value_error_from_client`.

## Not asking to change in this PR

CLI enum/range validation (`period`, `first`, etc.): I checked and no CLI command validates closed enums today; only structural checks via `BadParameter`. Asking for it here would introduce a pattern the CLI doesn't have elsewhere. Fine as a platform follow-up if you want MCP and CLI to fail fast the same way.

## For consumers

Partial success is intentional: an all-denied batch returns `success: true` with `automations: []` and the denials in `partial_errors`. Callers should read both fields, not only `automations`.

packages/mcp/src/pipefy_mcp/tools/observability_tools.py
rebecaventura: Hi Adrianno! Thanks for the code review!

Narrowed the [docstring](https://github.com/pipefy/ai-toolkit/pull/349/changes/0c3a8c0c4c55ddde94ad685be5ed73f793d487fc#diff-d359c08a4d9f3e33cfcb50bc140d933be00c2015a9ea07a29fd4d82323252db4R447) and added a [test](https://github.com/pipefy/ai-toolkit/pull/349/changes/0c3a8c0c4c55ddde94ad685be5ed73f793d487fc#diff-ac9c8eaedaf41b103fdb320325bfe187f7105c19d9e407a2218379180a57f88aR590) :)

skills/observability/pipefy-observability/SKILL.md
rebecaventura: Oh, right! Fixed the count on lines [6](https://github.com/pipefy/ai-toolkit/pull/349/changes/0c3a8c0c4c55ddde94ad685be5ed73f793d487fc#diff-3d70803fe8ab35458427727f6813a1a23c3a4100714bf86e2b7014f40d8c9c89R6) and [12](https://github.com/pipefy/ai-toolkit/pull/349/changes/0c3a8c0c4c55ddde94ad685be5ed73f793d487fc#diff-3d70803fe8ab35458427727f6813a1a23c3a4100714bf86e2b7014f40d8c9c89R12)

adriannoes (approved):
Follow-ups look good, thanks for addressing the review, @rebecaventura and @Danielmoraisg :)

Issue #346, opened 2026-07-01
Harden "parse, don't validate": self-guaranteeing types + per-app runtime


gbrlcustodio:
## Status (updated 2026-07-11)

About a third of this tracker has landed since it was filed. The parse-don't-validate doctrine it used to enumerate now lives in the repo docs (see Principles below), so this issue is trimmed to the concrete, still-open auth work and no longer duplicates the invariant catalogue.

Landed:
- Phase 1: the JWT verify-without-audience fail-fast validator (`JwtValidationSettings._validate_audience`).
- Phase 2: `(audience, verify_audience)` collapsed into the `AudiencePolicy` sum type; `ResolvedJwtValidation` / `resolve_jwt_validation` deleted (the inbound resolve dissolved).
- Phase 3: `McpRuntime.for_profile` wires eagerly at boot and owns the process engine and the inbound verifier.

## Problem (residual)

Some parsed types still encode intent without enforcing it, and one projection path still silently degrades:
- `resolver.py` sum-type members (`StaticTokenAuth`, `ServiceAccountAuth`, `StoredSessionAuth`) document invariants ("stripped and non-blank") but no `__post_init__` enforces them, so a hand-written instance can be invalid. The `Resolved`/domain name is still a convention, not a guarantee.
- `AuthSettings.to_service_account()` / `to_oidc_client()` remain on-demand projections that return `None` on a half-configured triple; consumers (`mcp/core/runtime.py`, `cli/commands/_common.py`) call them directly instead of receiving a projected domain type from a runtime.
- The CLI has no runtime, so it re-derives auth from settings ad hoc.

## Principles (canonical home is the docs, not this issue)

The self-guaranteeing-types plus per-app-runtime doctrine this issue used to spell out now lives in:
- `AGENTS.md`: "type validation at boundaries", "parse, don't validate", "parsed types are self-guaranteeing", "composition: the per-app runtime".
- `docs/architecture.md`: type ownership at boundaries, ports, the per-app composition root.
- `docs/target-architecture.md`: HOW-1 (parse at the boundary, interior total) and HOW-2 (effects built once in the runtime, decisions as sum types, fail fast).

Do not re-attempt (decisions specific to this work, not captured in the docs):
- A custom `PydanticBaseSettingsSource` + inner reader to make a settings field the sum type: it only hides the loose pair; the domain sum type already makes the illegal state unrepresentable.
[...]

PR #351, opened 2026-07-02
feat(mcp): request-scoped identity via app-scoped runtime

### LINKED ISSUE #302: Request-scoped identity

gbrlcustodio:
## What

Wires per-request on-behalf-of identity for the MCP `remote` profile's resource-server transport (#302). An app-scoped `McpRuntime` owns one shared `PipefyClient`; each request resolves its identity from its own validated inbound bearer, so one server serves concurrent callers as themselves rather than as a single startup identity. The stdio/local profile is unchanged: it still runs as the one credential resolved at startup.

Closes #302.

## Changes

- **Request-scoped identity via an app-scoped runtime.** Replaces the per-request DI container (`core/container.py`) with a single `McpRuntime` (`core/runtime.py`) built once at server construction. Tools resolve the client per call from the lifespan context, so identity is request-scoped without re-registering the tool table. Under the hosted profile a stateless `RequestContextBearerAuth` reads each caller's validated bearer from the request context; under stdio the runtime holds one startup credential.
- **Composition-root auth resolution.** Credential resolution and its fail-fast move out of the runtime into `StartupIdentity.from_configured_credential`, called by `_select_auth_source` at the composition root. Both identity variants (`StartupIdentity`, `RequestScopedIdentity`) now carry an `httpx.Auth`, so the runtime wires the shared client uniformly with no per-variant branching.
- **Server built from one injected resolved `Settings`.** `resolve_mcp_settings` returns the full `Settings`, `build_pipefy_mcp_server(settings)` reads the profile/host/port from it, and `server.py` reads no process globals on the serve path. This mirrors the CLI's composition-root pattern (`resolve_cli_settings`).
- **Fail fast when the `remote` profile has no resource server.** Launching `remote` without `PIPEFY_MCP_RS_RESOURCE_SERVER_URL` now raises at startup with an actionable message instead of silently degrading to the startup credential (there would be no per-request identity to act as).
- **Dropped the unused `CallableBearerAuth` adapter.** It had no production construction site; `RefreshableBearerAuth` already covers the per-request-provider behaviour plus the 401 refresh-and-retry the stored-session path needs.

## Testing

- `packages/mcp`: full suite green (unit; the live-integration tests skip cleanly without credentials).
- `packages/auth`: full suite green.
- Full non-integration suite green from the repo root.

## Notes

This is the stepping stone toward the single per-app runtime tracked in #346. A follow-up (`refactor/pipefy-engine-session`) takes this further into an engine/session split in the SDK. The module-level `settings = Settings()` singleton survives only for two per-request readers (unified-envelope check, graphql timeout) that have no settings handle yet; threading those through the runtime is a separable follow-up.


packages/mcp/src/pipefy_mcp/auth/request_identity.py
```diff
@@ -0,0 +1,65 @@
...
+
+def require_request_bearer() -> str:
+    """Return the validated bearer token for the in-flight MCP request.
+
+    FastMCP's ``AuthContextMiddleware`` stores the ``AccessToken`` it validated
+    in a per-request contextvar; this reads it back. Raises when no token is
+    present (a call ran outside the resource-server request scope), so a missing
+    identity fails loudly instead of issuing an unauthenticated Pipefy call.
+    """
+    access = get_access_token()
```
adriannoes: This reads `get_access_token()`, which pulls from `auth_context_var` set by `AuthContextMiddleware` in the ASGI request task. Under stateful streamable HTTP (the default: we never set `stateless_http=True`), the tool handler runs in the session's long-lived server task spawned at `initialize`. That task keeps the contextvar snapshot from init, not the bearer on the current `tools/call`.

I reproduced on the real app: init with token A, `tools/call` with token B on the same session, outbound header is still A. The per-message `request_ctx` already carries the current Starlette `Request` with the right `.user.access_token`, but this path does not use it.

Read the bearer from the current MCP request context (or opt into `stateless_http=True` with that documented and tested). A stateful integration test with two different bearers on one session is the acceptance check for #302.

adriannoes (changes requested):
## Summary

I checked out `3e849bf3` locally: ruff clean, 1585 unit tests green, CI green. The refactor toward `McpRuntime` as composition root looks sound and the remote fail-fast without a resource server is a good guard.

The blocker is request-scoped identity under the transport we actually ship. Streamable HTTP runs in stateful mode (`stateless_http` defaults to false and we never override it). The MCP server task is spawned during `initialize` and later `tools/call` messages run inside that same task. `RequestContextBearerAuth` reads `get_access_token()`, which comes from `auth_context_var` set by `AuthContextMiddleware` in the ASGI request task. That context does not update inside the long-lived server task, so outbound Pipefy calls keep replaying the session initializer's bearer.

I reproduced this on the real app: initialize with token A, `tools/call` on the same session with token B, outbound `Authorization` is still A. That breaks mid-session token refresh (stale outbound token leads to persistent 401s) and creates a confused-deputy path if another principal reuses a session id. Loopback binding limits exposure today, but the profile is meant to serve concurrent remote callers and does not deliver per-request identity yet.

## Required before merge

Read the token from the current MCP message context (e.g. the Starlette `Request` on `request_ctx`, where `.user.access_token.token` is set per message) instead of `get_access_token()`. Alternatively serve with `stateless_http=True` and document that choice.

Add a transport-level test: init with bearer A, `tools/call` with bearer B on the same stateful session, assert outbound header is B. This test fails today and is the acceptance check for #302. The existing 24-task unit test only exercises contextvars in the same asyncio task, so it stays green while this bug is live.

## Verification

```bash
uv run ruff check packages/mcp packages/auth
uv run pytest packages/mcp/tests packages/auth/tests -m "not integration" -q
```

Stateful OBO probe (fails today): init session with bearer A, `tools/call` with bearer B on same session, capture outbound `Authorization` header.

adriannoes (dismissed):
The blocker is addressed in 08bb66da. 

Reading the bearer from per-message `request_ctx` instead of the session-frozen `auth_context_var` is the right fix, and `test_stateful_session_task_reads_current_message_bearer` reproduces the two-task topology we called out.

Good to mark ready once you're happy with the draft state, @gbrlcustodio 👍🏻

adriannoes (approved):
Blocker from the earlier review is addressed in 08bb66da: bearer is read from per-message `request_ctx`, and `test_stateful_session_task_reads_current_message_bearer` covers the stateful session topology.

Follow-up commits clean up composition root (`run_server` takes resolved Settings) and docs (CHANGELOG, AGENTS.md). CI green, 1587 unit tests pass locally on caad107c.

LGTM for merge from my side.

PR #352, opened 2026-07-02
fix(sdk): accept both key spellings in automation extra_input and fix --extra help

### LINKED ISSUE #275: automation create --extra help says camelCase, but the API requires snake_case action_params
## Summary

`pipefy automation create --extra` documents its JSON object as "extra `CreateAutomationInput` fields (camelCase)", but the API expects snake_case field names. Passing the documented camelCase `actionParams` is rejected; the snake_case `action_params` works. The help text points users at the failing form.

## Reproduce

Against any pipe, create a `move_single_card` automation and supply the action params.

Documented form (camelCase), fails:

```
pipefy automation create --pipe <id> --name "move" \
  --trigger-id card_created --action-id move_single_card --no-active \
  --extra '{"actionParams":{"to_phase_id":"<phase_id>"}}'
# -> Variable $input of type CreateAutomationInput! was provided invalid value
#    for actionParams (Field is not defined on CreateAutomationInput)
```

snake_case form, succeeds:

```
pipefy automation create --pipe <id> --name "move" \
  --trigger-id card_created --action-id move_single_card --no-active \
  --extra '{"action_params":{"to_phase_id":"<phase_id>"}}'
# -> {"createAutomation": {"automation": {"id": "...", "active": false}, "error_details": null}}
```

## Where the wording is

The `--extra` option help on `automation create` describes the object as `CreateAutomationInput` fields in camelCase. The actual input uses snake_case (`action_params`, `event_params`, `action_id`, `event_id`), consistent with the mutation in `packages/sdk/src/pipefy_sdk/queries/automation_queries.py` and the SDK preflight that reads `action_params` (`packages/sdk/src/pipefy_sdk/automation_preflight.py:101`).

## Expected

Either:

- Fix the help text to say snake_case and use the API field names (`action_params`, `event_params`), or
- Accept camelCase in `--extra` and normalize it to snake_case before sending (the SDK already reads both spellings in preflight, so only the outgoing mutation variables need normalizing).

A first-class affordance for the common case would also help, for example `--to-phase <phase_id>` for `move_single_card`, so callers do not need to hand-build `action_params` at all.

## Environment

- pipefy CLI version: `0.2.0-beta.2`
- Auth: stored session (`pipefy auth login`).

Danielmoraisg:
Closes #275.

## Problem

`pipefy automation create --extra` documented its JSON object as "extra `CreateAutomationInput` fields (camelCase)", but the top-level keys are passed verbatim into the mutation variables. Introspecting `CreateAutomationInput` and `UpdateAutomationInput` shows the types mix naming styles: most fields are snake_case (`action_params`, `event_params`, `action_id`, `event_id`, `event_repo_id`, `action_repo_id`, `scheduler_frequency`) while a few are camelCase (`clientMutationId`, `responseSchema`, `schedulerCron`, `searchFor`). So the documented `actionParams` was rejected by the API, and a blanket case converter would break the camelCase-canonical fields.

## Change

- New pure helper `pipefy_sdk.automation_input.normalize_automation_input_keys`: an explicit alias map that rewrites top-level keys of `extra_input` to the API field names, in both directions (`actionParams` -> `action_params`, `scheduler_cron` -> `schedulerCron`). Nested payloads pass through verbatim because their casing is part of each param's own contract (`action_params.taskParams` is legitimately camelCase inside a snake_case field). When both spellings are present, the API-name key wins.
- `PipefyClient.create_automation` and `update_automation` apply the helper before preflight and the mutation, so the CLI and MCP surfaces both accept either spelling.
- CLI `--extra` help on `automation create` and `automation update` now names the API fields and states that camelCase aliases are accepted; the help golden fixture is regenerated.
- MCP tool and SDK docstrings that claimed "camelCase keys" for these inputs are corrected.

The issue also suggests a first-class `--to-phase` option for `move_single_card`; that is left as a follow-up since it is an affordance rather than part of this defect.

## Verification

Ran the issue's repro with the patched CLI against a scratch pipe in a test organization (created and deleted for this purpose):

- The previously failing documented form now succeeds: `--extra '{"actionParams":{"to_phase_id":"..."}}'` creates the rule, and `automation get` shows `action_params.to_phase_id` persisted.
- The snake_case form still works unchanged.
- `automation update --extra '{"actionParams":{"to_phase_id":"..."}}'` also works and the new phase id is visible on the rule.

Full test suite: 3102 passed. New unit tests cover the alias map (both directions, precedence, nested payloads untouched, input not mutated) and the facade wiring.

### DIFF
--- a/docs/mcp/tools/automations-and-ai.md
+++ b/docs/mcp/tools/automations-and-ai.md
@@ -8,7 +8,7 @@ Traditional automations (if/then rules) and AI-powered automations and agents. *
 
 Ten tools manage Pipefy traditional automations: if/then rules bound to a pipe via the standard GraphQL API.
 
-**Tip:** For **send-a-task** rules (`send_a_task` action), use `create_send_task_automation` (pipe, trigger, task title, recipients) instead of hand-building `action_params.taskParams` on `create_automation`. For other actions, call `get_automation_events` (global event catalog) and `get_automation_actions` with the target pipe (`repoId`) before `create_automation` to pick valid `trigger_id` / `action_id` values. Writes accept optional `extra_input` (camelCase API keys) and `debug=true` on errors.
+**Tip:** For **send-a-task** rules (`send_a_task` action), use `create_send_task_automation` (pipe, trigger, task title, recipients) instead of hand-building `action_params.taskParams` on `create_automation`. For other actions, call `get_automation_events` (global event catalog) and `get_automation_actions` with the target pipe (`repoId`) before `create_automation` to pick valid `trigger_id` / `action_id` values. Writes accept optional `extra_input` (top-level keys are snake_case, e.g. `action_params`, `event_params`; mirror `get_automation` output) and `debug=true` on errors.
 
 | Tool | Read-only | Role |
 |------|-----------|------|
@@ -107,7 +107,7 @@ For `card_moved` + `move_single_card`, when `extra_input` includes source and de
 1. **Discover** — `get_automation_events(pipe_id)`, `get_automation_actions(pipe_id)`, `get_automation_event_attributes()` for official `field_map` tokens; field ids via `get_start_form_fields` / `get_phase_fields`.
 2. **Create disabled** — `create_automation` with `active: false` and the `extra_input` payload above.
 3. **Verify** — `get_automation(automation_id)` and confirm `action_params.field_map` round-trip.
-4. **Enable** — `update_automation` with `extra_input: { "active": true }` (or camelCase `active` per API).
+4. **Enable** — `update_automation` with `extra_input: { "active": true }`.
 
 ---
 
--- a/packages/cli/src/pipefy_cli/commands/automation.py
+++ b/packages/cli/src/pipefy_cli/commands/automation.py
@@ -105,9 +105,8 @@ def automation_create(
         None,
         "--extra",
         help=(
-            "Optional JSON object: extra CreateAutomationInput fields, using the API "
-            "field names (e.g. action_params, event_params). camelCase aliases such "
-            "as actionParams are accepted and rewritten."
+            "Optional JSON object: extra CreateAutomationInput fields. Top-level keys "
+            "are snake_case (e.g. action_params, event_params); camelCase is also accepted."
         ),
     ),
     json_out: bool = typer.Option(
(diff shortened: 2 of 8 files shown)


adriannoes (dismissed):
## Summary

Clean, well-scoped fix for #275. The explicit bidirectional alias map in `automation_input.py` is the right tradeoff for GraphQL's mixed naming — better than a blanket case converter that would break camelCase-canonical fields like `schedulerCron`. Normalization at the `PipefyClient` facade automatically covers CLI and MCP without duplicating logic, and the second commit sensibly teaches agents one canonical spelling (snake_case top-level, mirror `get_automation`) while keeping the normalizer as a silent safety net.

Review included local checkout (`a5cb6f89`) and tests: ruff pass, 967 SDK unit tests pass. CI lint + test green.

## Also noted

- **F1 (`docs/mcp/tools/automations-and-ai.md:11`):** The MCP tool docstrings now teach snake_case top-level keys (mirror `get_automation`), but this reference doc still opens with "extra_input (camelCase API keys)". Agents that read the doc instead of the tool schema will keep sending the wrong shape — same class of confusion as #275. Worth aligning: snake_case canonical at the top level, optional footnote that aliases still work.

## Verification

```bash
uv run pytest packages/sdk/tests/test_automation_input.py -q
uv run pytest packages/sdk/tests -m "not integration" -q
```

adriannoes:
**F1 — `docs/mcp/tools/automations-and-ai.md:11`**

The MCP tool docstrings now teach snake_case top-level keys (mirror `get_automation`), but this doc still opens with "extra_input (camelCase API keys)". Agents that read the reference doc instead of the tool schema will keep sending the wrong shape — the same class of confusion as #275.

I'd align this file with the tool docstrings: snake_case canonical at the top level, optional footnote that aliases still work. You'll know it's right when line 11 no longer leads with camelCase.

adriannoes (approved):
## Re-check after `5d55bfd`

Follow-up commit addresses F1: `docs/mcp/tools/automations-and-ai.md` now teaches snake_case top-level `extra_input` (mirror `get_automation`) and drops the misleading camelCase `active` aside. Docs-only delta; runtime unchanged.

Re-verified locally on `5d55bfd`: 967 SDK unit tests pass. CI lint + test green.

Original assessment unchanged — clean fix for #275, merge-ready once marked out of draft.

PR #358, opened 2026-07-03
feat(cli): table field create, update, delete

### LINKED ISSUE #355: CLI parity: table field create, update, delete
## Scope

CLI-only. Add `pipefy table field` (create / update / delete) wrapping existing SDK methods. Flip three `docs/parity.md` rows from deferred to shipped: `create_table_field`, `update_table_field`, `delete_table_field`.

**Branch:** `feat/cli-table-field-commands` → `dev`

## Commands

- `pipefy table field create <table_id> --label L --type T` → `create_table_field`
- `pipefy table field update <field_id> [--table T] [--label L] [--extra JSON]` → `update_table_field` (only non-null keys)
- `pipefy table field delete <field_id> --table T` → `delete_table_field` (destructive: confirm or `--yes`)

Help must distinguish table columns from phase fields (`pipefy field`).

## Acceptance

- [ ] `pipefy table field --help` lists create, update, delete
- [ ] Tests in `packages/cli/tests/test_table_field_commands.py` (happy path per command + delete abort on denied confirm)
- [ ] `docs/parity.md` rows shipped; `skills/database-tables` CLI table updated
- [ ] `CHANGELOG.md` Unreleased entry; CLI help golden refreshed
- [ ] `uv run pytest packages/cli/tests` and ruff clean

## Out of scope

SDK and MCP changes.

adriannoes:
## Summary

- Add `pipefy table field` subcommand group (create / update / delete) over existing SDK facade methods
- Flip three `docs/parity.md` rows to shipped: `create_table_field`, `update_table_field`, `delete_table_field`
- Update database-tables skill CLI references and CLI help golden fixture
- Add `CHANGELOG.md` Unreleased entry

Closes #355

Part of milestone **CLI parity closeout** (PR 1 of 3).

## Test plan

- [x] `uv run pytest packages/cli/tests/test_table_field_commands.py -v` (4 passed)
- [x] `uv run pytest packages/cli/tests/test_cli_help_golden.py -v` (2 passed)
- [x] `uv run pipefy table field --help` lists create, update, delete


packages/cli/src/pipefy_cli/commands/table.py
```diff
@@ -149,3 +155,92 @@ async def factory(client: PipefyClient):
...
+    if label is not None:
+        attrs["label"] = label
+    attrs.update(extra)
+    if not attrs:
+        raise typer.BadParameter(
+            "Provide at least one of: --label, --extra (non-empty)."
+        )
+
+    async def factory(client: PipefyClient):
+        return await client.update_table_field(field_id, table_id=table, **attrs)
```
Danielmoraisg: **Silent wrong-write when `--extra` carries a reserved key.**

`--extra` is merged straight into `attrs` (line 216) and splatted into `update_table_field(field_id, table_id=table, **attrs)`. Two failure modes:

- `--extra '{"id":"999"}'` lands `id` in `attrs`. The SDK sets `payload["id"] = field_id` then overwrites it in its merge loop, so the update **silently retargets a different field** than the positional `field_id`. No error, wrong write. This is the realistic trap: copy a field object into `--extra`.
- `--extra '{"table_id":"42"}'` collides with the explicit `table_id=table` kwarg and raises `TypeError: got multiple values for keyword argument 'table_id'`, which `run_cli_command` does not catch, so it escapes as a raw traceback (exit 1).

The MCP path guards exactly these via `_UPDATE_TABLE_FIELD_EXTRA_RESERVED = {"id"}` and by popping `table_id` out of `extra_input` (`packages/mcp/src/pipefy_mcp/tools/table_tools.py:59`). This CLI re-implements the merge without the guard.

Suggest a shared helper in `_common.py` (e.g. `build_update_attrs(explicit, extra, *, reserved)`) that strips reserved keys before splatting, used by both this and `create` below, rather than two hand-rolled merges.

packages/cli/src/pipefy_cli/commands/table.py
```diff
@@ -149,3 +155,92 @@ async def factory(client: PipefyClient):
...
+
+    extra = parse_json_object(extra_json, "--extra") or {}
+    lab = label.strip()
+    ft = field_type.strip()
+    if not lab or not ft:
+        typer.echo("--label and --type must be non-empty.", err=True)
+        raise typer.Exit(2)
+
+    async def factory(client: PipefyClient):
+        return await client.create_table_field(table_id, lab, ft, **extra)
```
Danielmoraisg, line 189: **`--extra` reserved keys crash `create`.**

`create_table_field(table_id, lab, ft, **extra)` splats `--extra` alongside the positional args, so:

- `--extra '{"label":"X"}'` → `TypeError: got multiple values for argument 'label'`
- `--extra '{"table_id":"99"}'` → `TypeError: got multiple values for argument 'table_id'`

`run_cli_command` catches `PipefyError`/`TransportError`/`ValueError` but not `TypeError`, so the user sees an unhandled traceback. (`{"type": ...}` in extra does not crash but silently overrides `--type`.)

The MCP tool filters these with `_CREATE_TABLE_FIELD_EXTRA_RESERVED = {"table_id", "label", "type"}` (`table_tools.py:58`). Routing this call through the same reserved-key helper suggested on the `update` command fixes both spots at once.

packages/cli/src/pipefy_cli/commands/table.py
```diff
@@ -149,3 +155,92 @@ async def factory(client: PipefyClient):
...
+
+@table_field_app.command("update", context_settings=ID_POSITIONAL_CONTEXT_SETTINGS)
+def table_field_update(
+    ctx: typer.Context,
+    field_id: str = resource_id_argument(help="Table field id."),
+    table: str | None = typer.Option(
+        None,
+        "--table",
+        help="Table id containing this field (recommended; required by API).",
+    ),
```
Danielmoraisg, line 202: **`--table` is optional here but required by the API.**

`delete` (below) makes `--table` required, and the API requires `table_id` for updates too. As written, `update f1 --label X` (no `--table`) sends a payload without `table_id` and defers to a raw server-side rejection. The MCP tool returns a clean "table_id is required by the API" message before calling. Consider making `--table` required for parity with `delete`, or validating it here.

Note this is independent of the reserved-key collision above: requiring `--table` does not prevent a user from also passing `table_id` inside `--extra`, so both need addressing.

packages/cli/src/pipefy_cli/commands/table.py
```diff
@@ -149,3 +155,92 @@ async def factory(client: PipefyClient):
...
+    json_out: bool = typer.Option(False, "--json", "-j"),
+) -> None:
+    """Create a field (column) on a database table."""
+
+    extra = parse_json_object(extra_json, "--extra") or {}
+    lab = label.strip()
+    ft = field_type.strip()
+    if not lab or not ft:
+        typer.echo("--label and --type must be non-empty.", err=True)
+        raise typer.Exit(2)
```
Danielmoraisg, line 185: **Reuse + consistency (non-blocking).**

Two things here:

1. This block is near-identical to `field_create` in `field.py` (same `--label`/`--type` options, same `.strip()` + empty-check). A shared helper (e.g. `require_nonempty_label_type`) would collapse the duplication across table fields and phase fields.
2. Error idiom is inconsistent within this PR: `create` uses `typer.echo(...) + raise typer.Exit(2)` while `update` uses `typer.BadParameter`. `BadParameter` is the idiomatic Typer path (exit 2 + usage context). Standardizing on it reads better. Relatedly, `update` never validates `--label ""` non-empty, so an empty label passes through there while `create` rejects it. Pick one behavior.

packages/cli/src/pipefy_cli/commands/table.py
```diff
@@ -149,3 +155,92 @@ async def factory(client: PipefyClient):
...
+        None,
+        "--table",
+        help="Table id containing this field (recommended; required by API).",
+    ),
+    label: str | None = typer.Option(None, "--label", "-l", help="New field label."),
+    extra_json: str | None = typer.Option(
+        None,
+        "--extra",
+        help="JSON object merged into UpdateTableFieldInput.",
+    ),
```
Danielmoraisg, line 221: **Design question (non-blocking): typed flags vs `--extra` on update.**

The MCP `update_table_field` exposes `label`, `description`, `required`, `options` as first-class args; here only `--label` is typed and everything else goes through `--extra` JSON. The minimalism matches `table update`, so it is defensible, but it is worth a conscious decision: `--description`/`--required` are common enough that pushing them into hand-written JSON is a rough edge, and it is exactly that `--extra` path that triggers the collision issues above. If you keep `--extra`-only, the reserved-key helper carries the safety; if you add typed flags, you shrink the JSON surface people actually touch.

packages/cli/tests/test_table_field_commands.py
Danielmoraisg, line 1: **Coverage gap that maps to the `--extra` findings.**

These tests exercise only happy paths and delete-deny; no test passes `--extra` at all. That is why the reserved-key collision and silent `id`-override on `update` would ship unnoticed. Whatever you decide on those, add cases with a reserved key in `--extra` (e.g. `update ... --extra '{"id":"999"}'` and `create ... --extra '{"label":"X"}'`) plus the `update` "no attrs provided" `BadParameter` path.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+- **CLI**: `pipefy table field create|update|delete` manage database table columns, closing the three deferred table-field rows in the MCP↔CLI parity matrix (`create_table_field`, `update_table_field`, `delete_table_field`). `create` takes positional table id plus `--label`/`--type` (pass-through to `CreateTableFieldInput`); `update` sends only provided keys (`--table` recommended, required by the API); `delete` is destructive (`--yes` or interactive confirm; `--table` required).
```
Danielmoraisg: **Nit (optional): dense prose.** This entry stacks several parentheticals into one long sentence. Splitting the three subcommands into short clauses (or a sub-list) would scan more easily. Purely cosmetic.

packages/cli/src/pipefy_cli/commands/table.py
adriannoes: Addressed in 8885813. merge_extra_attrs in _common.py strips id from --extra before the SDK call. 

When --table is set, table_id is reserved in extra too. Typed flags override extra. 

Tests: 
- test_table_field_update_strips_id_from_extra, 
- test_table_field_update_prefers_table_flag_over_extra_table_id. 

Smoke tested on PipeClaw org.

packages/cli/src/pipefy_cli/commands/table.py
adriannoes: Addressed in 8885813. create passes --extra through merge_extra_attrs with reserved {table_id, label, type} before splatting. Test: test_table_field_create_strips_reserved_keys_from_extra. Verified live on PipeClaw.

packages/cli/src/pipefy_cli/commands/table.py
adriannoes: Kept --table optional to match MCP: table_id can come from --extra when the flag is omitted (test_table_field_update_accepts_table_id_from_extra). merge_extra_attrs prevents table_id in extra from colliding when --table is set. No local preflight for missing table_id; API error path unchanged. 

Happy to require --table in a follow-up if you want parity with delete.

packages/cli/src/pipefy_cli/commands/table.py
adriannoes: Added --description, --required/--not-required, and --options on update (3ecea62). --extra remains for uncommon UpdateTableFieldInput keys.

packages/cli/src/pipefy_cli/commands/table.py
adriannoes: Deferred a shared require_nonempty_label_type helper; duplication with field_create predates this PR and is out of scope. Kept typer.Exit(2) on create to match field_create. Empty --label on update still passes through, same as MCP update_table_field.

packages/cli/tests/test_table_field_commands.py
adriannoes: Added six tests: reserved keys on create/update, table_id from extra, --table priority over extra table_id, typed flags, and BadParameter when no attrs are provided.

CHANGELOG.md
adriannoes: Done in ed4de16. Split the Unreleased entry into short clauses per subcommand.

Danielmoraisg (approved):

Issue #362, opened 2026-07-03
SDK: export editable field filter helpers for MCP/CLI parity


adriannoes:
## Context

`fill_card_phase_fields` (MCP) and `pipefy card fill` (CLI) both filter user-provided field values to editable phase field IDs before `update_card`. The logic is duplicated today:

- MCP: `packages/mcp/src/pipefy_mcp/tools/pipe_tool_helpers.py` (`_filter_editable_field_definitions`, `_filter_fields_by_definitions`)
- CLI: `packages/cli/src/pipefy_cli/commands/card.py` (same semantics, reimplemented locally)

This was an explicit tradeoff in the cli-parity-closeout task 3.0 PR because the CLI cannot import `pipefy_mcp.*` and should not import private `pipefy_sdk.services` / `pipefy_sdk.queries` paths.

## Proposal

Promote the two pure filter helpers to the **public** `pipefy_sdk` surface (e.g. `pipefy_sdk.field_filters` or similar), export via `__init__.py` / `__all__`, and consume from both MCP and CLI.

## Semantics to preserve

- Non-dict field definitions are skipped.
- Definitions **without** an `editable` key count as editable (`field_def.get("editable", True)`).
- Value filter keeps only keys whose `field_id` is in `{field_def["id"]}` for editable definitions.

## Acceptance

- Single implementation used by MCP `fill_card_phase_fields` / `create_card` skip-elicitation paths and CLI `card fill`.
- Unit tests on the SDK helpers (no mocks needed).
- No behavior change vs current duplicated copies (regression tests in MCP + CLI packages stay green).

## Related

- Parent parity work: cli-parity-closeout task 3.0 (`fill_card_phase_fields` → `pipefy card fill`)
- Filed from code review of `feat/cli-card-fill`

adriannoes:
## Update (PR #363 review)

PR #363 now promotes the **filter predicates** to the public SDK (`pipefy_sdk.field_filters`: `filter_editable_field_definitions`, `filter_fields_by_definitions`, `skipped_field_ids`). MCP and CLI import from there; unit tests live in `packages/sdk/tests/test_field_filters.py`.

[...]

PR #363, opened 2026-07-03
feat(cli): add card fill for non-interactive phase field updates

### LINKED ISSUE #357: CLI parity: card fill (non-interactive)
## Scope

CLI-only. Add `pipefy card fill` for deferred row `fill_card_phase_fields`. No elicitation — direct path equivalent to MCP `skip_elicitation=True`.

**Branch:** `feat/cli-card-fill` → `dev`

## Design (locked)

**Option A:** `get_phase_fields(phase_id)` → filter input `fields` to editable field IDs (local ~15-line filter in CLI; do not import `pipefy_mcp`). Missing `editable` on a definition counts as editable. Non-editable keys are dropped silently (MCP parity). Then `update_card(card_id, field_updates=...)`.

## Command

`pipefy card fill <card_id> --phase <phase_id> --fields '{"field_id":"value"}' [--required-only]`

## Acceptance

- [ ] After filter, `update_card` receives expected `field_updates`
- [ ] Empty or all-non-editable input returns `{"success": true, "message": "No fields to update."}` without calling `update_card`
- [ ] Tests in `test_card_commands.py` (happy path + filtered non-editable field + no-fields path)
- [ ] `docs/parity.md` row shipped; review `docs/mcp/tools/pipes-and-cards.md` and `skills/pipes-and-cards` for CLI mention
- [ ] Ruff clean

## Out of scope

SDK/MCP changes; promoting filter helpers to public SDK (optional follow-up).

adriannoes:
## Summary

- Adds `pipefy card fill <card_id> --phase <phase_id> --fields '{"…"}'` as CLI parity for MCP `fill_card_phase_fields` (non-interactive / `skip_elicitation` equivalent).
- Filters `--fields` to editable phase field IDs locally (Option A); returns `{"success": true, "message": "No fields to update."}` without calling `update_card` when nothing remains after filtering.
- Flips `fill_card_phase_fields` to `shipped` in `docs/parity.md`; updates CHANGELOG, SKILL, and MCP docs.

Closes #357

## Follow-up

- Deduplicate editable-field filter helpers via public SDK export (Option C): #362

## Test plan

- [x] `uv run pytest packages/cli/tests/test_card_commands.py -k fill -v` (5 passed)
- [x] `uv run pytest packages/cli/tests -q` (322 passed)
- [x] Manual E2E: `pipefy card fill` against Demo PM3 AI pipe (empty fields no-op, editable field update, unknown field filtered)


packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -366,6 +393,48 @@ async def factory(client: PipefyClient):
...
+            phase_fields_result.get("fields", [])
+        )
+        field_data = _filter_fields_by_definitions(fields, expected_fields)
+        if not field_data:
+            return {"success": True, "message": "No fields to update."}
+        field_updates = [
+            {"field_id": field_id, "value": value}
+            for field_id, value in field_data.items()
+        ]
+        return await client.update_card(card_id, field_updates=field_updates)
```
Danielmoraisg: **Silent no-op on filtered-out keys, plus a divergence from the MCP tool it mirrors.**

Two related points on this filter-then-update block:

1. **Unmatched keys are dropped silently.** Any `--fields` key that is not an editable phase field is filtered out with no warning; the command then either partially updates or returns `{"success": true, "message": "No fields to update."}` with exit 0. A CLI user has no elicitation step to catch this, so a typo'd field id reads as success. Concrete trigger: `card fill 99 --phase 100 --fields '{"stauts":"done"}'` → "No fields to update", exit 0. Consider echoing the dropped keys to stderr (or a `skipped` list in the JSON) so the caller knows.

2. **Divergence from `fill_card_phase_fields` when the phase reports no editable fields.** The MCP tool only runs `_filter_fields_by_definitions` inside `elif expected_fields:` — when `expected_fields` is empty it sends the raw `fields` unfiltered. This command filters unconditionally (line 426), so with an empty `expected_fields` it always no-ops. Reachable via `--required-only` on a phase with no required fields: MCP applies the values, CLI returns "No fields to update." The CLI behavior is arguably safer (never sends unfiltered), but it is a behavior difference from the tool this claims parity with — decide if it is intentional, and if so the "MCP skip_elicitation equivalent" docstring slightly oversells it.

Minor, same block: when `--fields '{}'` (or all keys filter out), `get_phase_fields` is still called before the no-op. Since this path never elicits, you could short-circuit on empty `fields` before the network round trip.

packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -93,6 +94,32 @@ def _split_csv_ids(raw: str | None) -> list[str | int] | None:
...
+    field_definitions: list[dict[str, Any]],
+) -> dict[str, Any]:
+    if not fields:
+        return {}
+    editable_ids = {field_def["id"] for field_def in field_definitions}
+    return {
+        field_id: value
+        for field_id, value in fields.items()
+        if field_id in editable_ids
+    }
```
Danielmoraisg: **Reuse / altitude (partly tracked by #362).** `_filter_editable_field_definitions` and `_filter_fields_by_definitions` are verbatim copies of the helpers in `packages/mcp/src/pipefy_mcp/tools/pipe_tool_helpers.py:418`, and the `card_fill` factory below duplicates the whole MCP orchestration (`get_phase_fields` → filter editable → filter by keys → no-op-or-`update_card`). #362 proposes exporting the two helpers, but the copy here is the orchestration too, not just the predicates. The deeper fix is a single SDK method (e.g. `client.fill_card_phase_fields(card_id, phase_id, fields, required_only)`) that both the MCP tool and this command call, so the editable-default and empty-`expected_fields` semantics live in exactly one place. That also closes the divergence flagged on the fill command.

packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -93,6 +94,32 @@ def _split_csv_ids(raw: str | None) -> list[str | int] | None:
...
+    return editable_fields
+
+
+def _filter_fields_by_definitions(
+    fields: dict[str, Any],
+    field_definitions: list[dict[str, Any]],
+) -> dict[str, Any]:
+    if not fields:
+        return {}
+    editable_ids = {field_def["id"] for field_def in field_definitions}
```
Danielmoraisg: Nit (defensive): `field_def["id"]` raises `KeyError` if a definition dict lacks `id`, and `run_cli_command` does not catch `KeyError`, so it would surface as a traceback rather than a clean error. `_filter_editable_field_definitions` guarantees dicts but not the `id` key. The real API always returns `id`, so low priority — `field_def.get("id")` (skipping falsy) would harden it. Same shape exists in the MCP helper, so worth fixing in the shared version if #362 lands.

packages/cli/src/pipefy_cli/commands/card.py
```diff
@@ -366,6 +393,48 @@ async def factory(client: PipefyClient):
...
+    phase_id: str = typer.Option(
+        ...,
+        "--phase",
+        help="Phase id whose fields to fill.",
+    ),
+    fields_json: str = typer.Option(
+        ...,
+        "--fields",
+        help='JSON object of field id to value, e.g. \'{"campo":"v"}\'.',
+    ),
```
Danielmoraisg, line 390: Nit (discoverability): the `card` group now has two field-value input shapes — `card update --field-updates '[{"field_id":...,"value":...}]'` (JSON array) and `card fill --fields '{"id":"value"}'` (JSON object). Each mirrors its MCP tool, so it is defensible, but a user moving between the two commands has to remember which shape applies. Worth a line in the help or SKILL cross-referencing them or unifying them as they seem to serve the same purpose.

packages/cli/src/pipefy_cli/commands/card.py
adriannoes: Thanks for calling this out, really helpful for CLI ergonomics.

Addressed in the latest push:

1. **Filtered keys**: when `--fields` keys are dropped by the filter, the JSON response now includes `skipped_field_ids` (both on full no-op and on partial updates). A typo like `{"stauts":"done"}` still exits 0 but makes it clear nothing was applied.

2. **Empty `expected_fields`**: we are keeping the stricter CLI behavior on purpose (task 3.0 acceptance). Documented in the command docstring and the `parity.md` note rather than aligning to MCP pass-through in this PR.

3. **Empty `--fields`**: we short-circuit before `get_phase_fields` when the parsed object is `{}`.

Covered by `test_card_fill_typo_reports_skipped_field_ids` and the updated empty-fields test.

packages/cli/src/pipefy_cli/commands/card.py
adriannoes: Agreed, the duplication was worth fixing now for the predicates.

In `df7b73c` we promoted `filter_editable_field_definitions`, `filter_fields_by_definitions`, and `skipped_field_ids` to the public SDK (`pipefy_sdk.field_filters`); MCP and CLI both import from there now, with unit tests in `packages/sdk/tests/test_field_filters.py`.

The full orchestration (`get_phase_fields` → filter → no-op / `update_card`) is still duplicated by design in this PR.. larger surface area. 

I expanded #362 with what remains so we can unify flow and semantics (including empty-`expected_fields`) in a follow-up rather than widening #357 further.

packages/cli/src/pipefy_cli/commands/card.py
adriannoes: Good catch, agreed it is worth hardening even if the API always sends `id` today.

`filter_fields_by_definitions` in `pipefy_sdk.field_filters` now uses `field_def.get("id")` and skips definitions without an id, so we avoid a `KeyError` bubbling up as a traceback. Covered by `test_filter_fields_by_definitions_skips_definitions_without_id` in the SDK test suite.

packages/cli/src/pipefy_cli/commands/card.py
adriannoes: Makes sense, the two shapes mirror their MCP tools, but the CLI UX benefit of a cross-reference is real.

Added a line to `card fill` `--help` pointing to `card update --field-updates` (JSON array) for ad-hoc updates without phase discovery, and updated the SKILL row for `fill_card_phase_fields` with the same distinction. Golden help regenerated accordingly.

adriannoes:
@Danielmoraisg — could you take another look when you have a moment?

We pushed follow-up commits addressing your review (skipped_field_ids, SDK `field_filters` dedup, doc/help updates) and merged latest `dev` (CHANGELOG conflict resolved). CI is green on the latest push.

Thanks again for the thorough review.

Danielmoraisg (approved):

PR #361, opened 2026-07-03
refactor(sdk): rename pipefy-sdk distribution to pipefy

gbrlcustodio:
## Why

PyPI refuses the name `pipefy-sdk`: its ultranormalized form (`pipefysdk`) collides with an unrelated third-party package, so the SDK cannot be published under that name. This blocks publishing the SDK (and therefore `pipefy-mcp-server`) to PyPI (see #248).

Rename the distribution to `pipefy`: it is free, clears the similarity guard, and reads as the first-party flagship. A code assessment ruled out merging `sdk` + `auth` + `infra` (the SDK does not import `auth`, and `auth` carries the heavy OAuth/keychain deps), so `pipefy-infra` and `pipefy-auth` keep their names.

## What

Distribution rename only. The import module stays `pipefy_sdk`, the directory stays `packages/sdk`, and the version machinery (`bump_version.py`, the release tag-match) and ruff import bans are untouched.

- `packages/sdk/pyproject.toml` `name` -> `pipefy`.
- Root `pyproject.toml` `[tool.uv.sources]` + dev group; `cli`/`mcp` dependencies and `[tool.uv.sources]`.
- `release.yml` and `ci.yml` wheel-name globs (`pipefy_sdk-*` -> `pipefy-*`); the SDK wheel now builds as `pipefy-<version>-*.whl`.
- Consumption docs and install snippets: `.mcp.json`, root/`mcp`/`cli`/`sdk`/`infra` READMEs, `commands/install.md`, `docs/*`, `RELEASE.md`.
- Three source docstrings/help strings (CLI help + entry docstring; the MCP re-export docstring corrected to the import module `pipefy_sdk`).
- `CHANGELOG.md` `[Unreleased]` entry; `uv.lock` regenerated.
- `AGENTS.md`: doc-map / project-structure updates plus a new guideline for the gradual `pipefy_sdk` -> `pipefy` import-module migration.

## Testing

- `uv sync`, `uv run ruff check .` / `ruff format --check .`, `uv run python scripts/bump_version.py verify` all pass.
- `uv run pytest -m "not integration"`: 3171 passed, 0 failures (branch rebased onto `dev`). The CLI help golden reflects the one-line help-text change.
- `uv build --all-packages`: five wheels, SDK now `pipefy-0.2.0b4-*.whl`. Verified each `release.yml` prefix glob still matches exactly one wheel (`pipefy-*.whl` matches only the SDK, since the others are `pipefy_*`).
[...]

PR #365, opened 2026-07-03
ci(release): publish to PyPI on every tag and bump to 0.3.0-alpha.1

gbrlcustodio:
## What

Enable an alpha release off `dev` that publishes to PyPI:

- **`ci(release)`**: drop the `startsWith(github.ref_name, 'v1.') && !contains(github.ref_name, '-')` gate from the three PyPI steps (Stage / Smoke install / Upload). The workflow now uploads all five workspace wheels via Trusted Publishing on every `v*` tag, including 0.x and pre-release tags. Pre-release versions upload as pre-releases (installable only with `--pre`).
- **`chore(release)`**: bump the lockstep version `0.2.0-beta.4` -> `0.3.0-alpha.1` (minor bump for the breaking changes under `[Unreleased]`, first alpha of the 0.3.0 line).
- **`RELEASE.md`**: document the every-tag publishing policy and the trusted-publisher setup for all five distributions.

## Why

The previous gate published to PyPI only for `v1.` tags with no pre-release suffix, so a 0.x alpha cut off `dev` produced a GitHub Release but never reached PyPI. Relaxing the gate lets us exercise the full PyPI pipeline from the alpha line.

## Stacking

This branch is stacked on `refactor/rename-sdk-dist-to-pipefy` (#361), so it carries none of that PR's diff here and cannot conflict with it. #361 must land first (its wheel-name globs are what make the `pipefy-*.whl` SDK upload match). When #361 merges, GitHub auto-retargets this PR's base to `dev`.

## Before tagging a release

- Configure PyPI Trusted Publishers for all five distributions (`pipefy`, `pipefy-cli`, `pipefy-mcp-server`, `pipefy-auth`, `pipefy-infra`); use the pending-publisher flow for first upload of a new name.
- Rename the `CHANGELOG.md` `## [Unreleased]` heading to `## [0.3.0-alpha.1] - <date>` at tag time (the release-notes step needs the matching heading).
- Do not roll the `latest` moving tag for a pre-release; `latest` drives default installs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,7 +88,6 @@ jobs:
           files: dist/*.whl
           body_path: CHANGELOG_RELEASE_BODY.md
       - name: Stage PyPI wheels (SDK + CLI + MCP + Auth + Infra)
-        if: startsWith(github.ref_name, 'v1.') && !contains(github.ref_name, '-')
         run: |
           mkdir -p dist-pypi
           shopt -s nullglob
@@ -105,7 +104,6 @@ jobs:
           fi
           cp "${sdk[0]}" "${cli[0]}" "${mcp[0]}" "${auth[0]}" "${infra[0]}" dist-pypi/
       - name: Smoke install staged wheels (clean venv)
-        if: startsWith(github.ref_name, 'v1.') && !contains(github.ref_name, '-')
         run: |
           python -m venv .venv-pypi-smoke
           .venv-pypi-smoke/bin/python -m pip install --upgrade pip
@@ -113,7 +111,6 @@ jobs:
           .venv-pypi-smoke/bin/pipefy --help
           .venv-pypi-smoke/bin/pipefy-mcp-server --help
       - name: Upload to PyPI
-        if: startsWith(github.ref_name, 'v1.') && !contains(github.ref_name, '-')
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages-dir: dist-pypi/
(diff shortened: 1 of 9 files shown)


adriannoes (dismissed):
## Summary

Solid, focused release-infra PR. Removing the three identical `v1.`-only gates unblocks the full PyPI pipeline for 0.x pre-releases without touching unrelated workflow logic — exactly the right way to exercise Trusted Publishing before v1. I checked out the head locally: `bump_version.py verify`, ruff, and 3176 unit tests all pass.

Merge **#361** first (wheel glob `pipefy-*.whl` and distribution rename). After this lands on `dev`, mark ready and let CI run on the retargeted branch before the first `v0.3.0-alpha.*` tag.

## What worked well

- Minimal diff, maximum effect: three `if:` removals unlock PyPI for every `v*` tag.
- `RELEASE.md` clearly documents every-tag publishing and `--pre` semantics for pre-releases.
- Lockstep version bump across all five packages plus `pyproject.toml` and `uv.lock`.
- PR body is exemplary: merge order, trusted publishers, CHANGELOG at tag time, and not rolling `latest` on pre-releases.

## Also noted

- **Step 6** in `RELEASE.md` still rolls `latest` unconditionally; you already warned about skipping that for pre-releases in the PR description — worth adding that one-line caveat to the checklist when you have a moment (not blocking).

gbrlcustodio:
Thanks for the review and the local verification, @adriannoes.

Addressed the non-blocking note: step 6 now carries a caveat to skip rolling the `latest` tag for pre-release tags, since `latest` drives default installs and must track the newest stable release, not an alpha/beta/rc (commit 1f85b58).

Also rebased onto `dev` now that #361 has landed, so the branch is current. The rebase plus the caveat commit dismissed your approval, so a re-stamp would be appreciated when you have a moment. The substance is unchanged: the three `if:` removals and the lockstep bump to `0.3.0-alpha.1`, plus the one-line step-6 caveat you flagged.

Trusted publishers are confirmed in place for all five distributions. Once CI is green and this is re-approved, the plan is to merge to `dev`, cut the `## [0.3.0-alpha.1]` CHANGELOG section, then tag `v0.3.0-alpha.1` to exercise the PyPI pipeline.

PR #367, opened 2026-07-04
build(release): pin workspace siblings to the lockstep version

gbrlcustodio:
## Why

Published wheels declared their workspace siblings unpinned (`pipefy`, `pipefy-auth`, `pipefy-infra`). The `[tool.uv.sources]` workspace mapping only applies in-repo, so a public `pip install pipefy-cli==X` could resolve a sibling to any newer version on PyPI instead of the one built and tested alongside it. All five distributions share one lockstep version and are only ever tested together at that version, so the published dependency closure has to resolve to it too.

Two related gaps let a release ship wrong with no signal:

- The Claude plugin manifest (`.claude-plugin/plugin.json`) carries the version the marketplace shows, but nothing kept it in lockstep. It had already drifted to `0.2.0-beta.1` while the workspace shipped `0.3.0-alpha.1`.
- The release wheel check confirmed each known package built one wheel, but not that no other wheel existed. A new workspace member would build a wheel that never gets staged for PyPI, shipping a partial release unnoticed.

## What

- Pin each package's workspace siblings to the exact lockstep version in its published metadata. The set to pin is derived from each package's real `[project.dependencies]`, so a new inter-package dependency is pinned automatically.
- Track the plugin manifest version in lockstep: written on every bump, asserted by `verify`.
- Fail the release build when `dist/` holds a wheel outside the known workspace packages, forcing a deliberate decision before a new package publishes to PyPI.
- Everything `bump_version.py` reads or writes is derived from a source of truth another tool already enforces: workspace membership from `[tool.uv.workspace]`, each `__version__` file from `[tool.hatch.version].path`, sibling pins from `[project.dependencies]`. Adding a member needs no hand-maintained list edits.

CI runs `bump_version.py verify`, which folds the sibling pins, the plugin manifest, the root `pyproject.toml`, and `uv.lock` into one all-must-match comparison (canonicalized via `packaging.version.Version`, so `==0.3.0-alpha.1` matches uv's normalized `0.3.0a1`). A duplicate or decoy version/pin occurrence fails loudly rather than being silently rewritten. `uv lock --locked` stays clean, since the pins do not move the lockfile and workspace sources still resolve in dev.


adriannoes (dismissed):
Solid release-metadata fix. 

Pinning workspace siblings to the lockstep version in published `pyproject.toml` metadata closes a real gap: `[tool.uv.sources]` only applies in-repo, so a public `pip install pipefy-cli==X` could previously pull newer siblings from PyPI. I checked out the branch locally — 35 unit tests pass, `bump_version.py verify` is clean, and the built `pipefy_cli` wheel carries `Requires-Dist: pipefy==0.3.0-alpha.1` (and `pipefy-auth`) as intended. 

Sharing one regex between `write_dep_pins` and `verify_lockstep` is a nice maintainability win, and the release workflow wheel-count bound is a cheap guard against a sixth workspace member shipping unnoticed.

**What worked well**

- `workspace_dep_pin_re` serves both writer and verifier — no duplicate hand-maintained patterns.
- Regex tests cover the scary cases (`pipefy` vs `pipefy-infra`, `[tool.uv.sources]` keys, description prose) plus a guard that `WORKSPACE_DEP_PINS` never lists a package's own name.
- Plugin manifest folded into lockstep closes the drift you called out (`0.2.0-beta.1` vs `0.3.0-alpha.1`).

**Also noted**

- The automation table in `RELEASE.md` still describes the pre-PR `verify` scope (five `__version__` strings + root `pyproject.toml`). Worth a quick update to mention sibling `==` pins, `.claude-plugin/plugin.json`, and the release workflow's exact-five-wheels guard — saves the next release cutter a head-scratch when verify fails on a pin.

gbrlcustodio:
Good catch on the automation table drift. Fixed in 009d7069: the reference now covers the sibling `==` pins, `.claude-plugin/plugin.json`, and `uv.lock` that verify asserts, plus the release workflow's exact-five-wheels guard.

adriannoes (dismissed):
Thanks for the follow-up — the `RELEASE.md` automation table now matches the expanded `verify` scope (sibling pins, plugin manifest, `uv.lock`, and the five-wheel guard). No further notes from my side; still good to merge.

adriannoes (approved):
Looked at the six commits since the docs follow-up. The move to derive package lists, `INIT_PATHS`, and sibling pins from `uv.workspace` / hatch config / declared `[project.dependencies]` removes the hand-maintained maps we were worried about, and `_sole_match` / `_sub_exactly_one` close the silent-first-match hole. CI green; `bump_version.py verify` passes locally. Good to merge.

PR #370, opened 2026-07-06
fix(sdk): accept both key spellings in automation extra_input and fix --extra help

Danielmoraisg:
## What

Brings the `fix/275-automation-extra-input-case` changes onto `dev`.

## Why

PR #352 was merged into `main` instead of `dev`, so `dev` was missing the fix. This puts the same three commits on `dev`:

- `3100c68` fix(sdk): accept both key spellings in automation extra_input and fix --extra help
- `a5cb6f8` docs(automation): teach one canonical extra_input spelling, stop advertising aliases
- `5d55bfd` docs(automation): align automations reference with snake_case extra_input guidance

## Notes

A `git merge-tree` dry run against `dev` reported no conflicts. Merge with **Rebase and merge** to keep history linear.

### DIFF
--- a/docs/mcp/tools/automations-and-ai.md
+++ b/docs/mcp/tools/automations-and-ai.md
@@ -8,7 +8,7 @@ Traditional automations (if/then rules) and AI-powered automations and agents. *
 
 Ten tools manage Pipefy traditional automations: if/then rules bound to a pipe via the standard GraphQL API.
 
-**Tip:** For **send-a-task** rules (`send_a_task` action), use `create_send_task_automation` (pipe, trigger, task title, recipients) instead of hand-building `action_params.taskParams` on `create_automation`. For other actions, call `get_automation_events` (global event catalog) and `get_automation_actions` with the target pipe (`repoId`) before `create_automation` to pick valid `trigger_id` / `action_id` values. Writes accept optional `extra_input` (camelCase API keys) and `debug=true` on errors.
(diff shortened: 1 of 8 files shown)


gbrlcustodio (approved):

adriannoes (approved):
## Summary

This is a clean backport of the three commits already merged to `main` via #352, bringing the #275 fix onto `dev`. I checked out the PR head locally (`5d55bfd`): ruff and targeted pytest passed (normalizer unit tests, facade wiring, CLI help golden).

The diff matches what we approved on #352: a pure `normalize_automation_input_keys()` planner in the SDK, wired through `PipefyClient.create_automation` / `update_automation` before preflight and the mutation, plus corrected CLI `--extra` help and aligned MCP/docs guidance.

## What worked well

- Normalization lives in the SDK facade, so MCP and CLI stay thin and share one path.
- The alias map is bidirectional and well tested (precedence, nested pass-through, input immutability).
- Normalization runs before preflight, which fixes the #275 repro (`actionParams` only) without duplicating logic.
- MCP docstrings and the automations reference now teach snake_case top-level keys and mirroring `get_automation` output.

## Also noted

- **F1 (optional follow-up):** If a caller sends both `action_params: {}` and a populated `actionParams`, the normalizer drops the alias whenever the canonical key is present—even when that value is an empty dict. Preflight used to fall back to `actionParams` because `{}` is falsy, but after normalization only the empty `action_params` remains. This is a rare edge case and does not block this backport; a regression test (or treating an empty canonical dict as absent) would close the loop.

Issue #309, opened 2026-06-18
Per-user observability and quotas


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Add audit logging and quota accounting keyed to the authenticated subject rather than to the process.

## Acceptance criteria

Each tool call is attributable to a user identity in logs; per-user quota counters exist.

## Dependencies

Builds on request-scoped identity (#302).

PR #378, opened 2026-07-08
feat(mcp): add tool-call middleware seam

### LINKED ISSUE #374: Tool-call middleware seam

## Scope

Add a documented middleware (interceptor) chain around tool invocation, so cross-cutting concerns can be layered without swapping the server's internal request handlers.

Today the only way to observe or govern a tool call is to replace the low-level `CallToolRequest` handler on the server. That is a single private slot: it is undocumented, and the next feature that needs the same seam overwrites the previous one. Several accepted and proposed items all need this exact interception point:

- per-tool logging and metrics (#373, #309)
- per-user quotas and rate limiting (#309, #307)
- cost weighting for expensive tools (#307)
- downstream protection: circuit breaking and honoring the API's 429s (#307)

Each of these is "run something around a tool invocation with the caller's validated identity in hand." A single ordered middleware chain turns all of them into registered middleware rather than internal extension.

Proposed shape (illustrative, not prescriptive):

```python
async def middleware(ctx: ToolCallContext, call_next):
    # ctx exposes: tool name, arguments, the validated auth context
    # (subject, client id, scopes) from request-scoped identity, and a request id
    result = await call_next(ctx)   # runs inner middleware, then the tool
    # after: result error flag, duration
    return result
```

Middleware runs in registration order (outer to inner) around the handler, may short-circuit (for example, a quota middleware returning an error result without calling `call_next`), and has read access to the validated auth context so it never re-decodes the bearer.

## Acceptance criteria

- A public, documented way to register tool-call middleware; the internal request-handler dict is no longer the extension surface.
- Middleware receives the tool name, arguments, validated caller identity (subject, client id, scopes), and a request id correlating to the HTTP request.
- Middleware runs in a defined order and can short-circuit a call (return an error result without invoking the tool).
- The structured tool logging in #373 is reimplemented as a middleware on this chain, demonstrating the seam end to end.
- Registering middleware does not require importing or mutating server internals.

## Dependencies

Builds on request-scoped identity (#302, closed) for the auth context passed to middleware. Unblocks the tool-call half of per-user observability and quotas (#309), rate limiting and abuse protection (#307), and is the seam #373 should be built on.

gbrlcustodio:
Closes #374.

## What

Adds a public, ordered middleware chain around MCP tool invocation so cross-cutting concerns (logging, quotas, rate limiting, cost weighting, downstream 429/circuit-breaking) layer as composed middleware instead of each overwriting FastMCP's single private `request_handlers[CallToolRequest]` slot.

## How

- `core/tool_middleware.py` defines the contract (`ToolCallContext`, `ToolCallMiddleware`/`CallNext`, `short_circuit_error`), `compose` (outer-to-inner), and `install_tool_call_middleware`, which wraps the handler once at build time (once per app: a second install raises; a no-op when the list is empty).
- The composition root owns the middleware list: `server.py`'s `default_tool_middlewares(settings)` builds the per-profile list and hands it to `install_tool_call_middleware(app, ...)`. The runtime stays out of middleware entirely and owns only the engine, identity, and inbound auth, so `core` no longer depends on `observability`.
- `auth/request_identity.py` gains `caller_identity(request)` returning the validated caller's client id and scopes (the type `ctx.identity` carries), sharing one `_authenticated_user` reader with `require_request_bearer`.
- `observability/tool_log_middleware.py` ships the first consumer: one structured JSON line per call (tool, outcome, duration, argument keys, client id, request id), emitted to stderr so it never corrupts the stdio JSON-RPC stream. It is seeded by default only under the remote profile; that is a default, not a capability boundary, since the chain installs on every profile.

## Design notes

- Middleware runs in list order, may short-circuit with a canonical `tool_error` envelope (`isError=True`, signalling the tool never ran), and reads identity from the per-message request context, never `auth_context_var`.
- Middleware sees raw un-coerced arguments (FastMCP registers the terminal with `validate_input=False`); `argument_keys` is bounded and values-free, and the bearer and argument values are never logged.
- The log outcome distinguishes normal control flow (`cancelled` on client disconnect, `elicitation` on a URL-elicitation signal) from `error`, so neither inflates error metrics.
- The private `CallToolRequest` slot is no longer the extension surface; the wrap is pinned to `mcp==1.25.0` and fails loud if that contract changes.

## Deferred / follow-up

- Subject (the end-user, distinct from client id) is deferred to per-user quotas (#309), the consumer that keys on it. This is a scope decision, not a feasibility one: both profiles can source subject when it lands (remote from the validated `sub` claim, local by decoding the configured JWT credential once at startup), so the AC's subject field arrives with that consumer, across both profiles.
- This supersedes the tool-line handler wrap in #373's structured logging; when #373 lands, its tool-call logger folds onto this chain, and its `sub`-preserving token subtype is what the remote half of subject would reuse. #373's "logs on stdout" acceptance criterion should reconcile to stderr, since this chain also runs under the stdio transport where stdout carries the JSON-RPC stream.
- Governance short-circuits currently log as `outcome=error`; a distinct `denied` outcome (so quota and rate-limit rejections do not read as failures) lands with the quota consumer.

## Testing

- New unit tests for compose ordering, short-circuit, exception propagation and the outcome taxonomy (cancelled/elicitation/error), the short-circuit envelope shape, the once-per-app install (second install raises) and the empty-chain no-op, `caller_identity` (including the local-over-HTTP no-user-scope case), `default_tool_middlewares` per profile, and the tool-log middleware (fields, privacy, stdout silence).
- Full `packages/mcp` suite: 1434 passed, 9 skipped (the no-creds live tests). Verified the real build path wraps the handler under remote and is a no-op under local.

### DIFF
--- a/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
+++ b/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
@@ -3,8 +3,8 @@
 The first consumer of the tool-call middleware chain (see
 :mod:`pipefy_mcp.core.tool_middleware`): it emits one JSON line per call with the
 tool name, outcome, duration, the caller's client id, argument key names, and a
-request id for correlation. It proves a real cross-cutting concern rides the seam
-without touching FastMCP internals.
+request id for correlation. Being the first consumer, it also exercises the
+chain end to end.
 
 Privacy: never logs the bearer, and never logs argument values, only their
 (bounded) key names.
--- a/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
+++ b/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
@@ -12,6 +12,14 @@ Privacy: never logs the bearer, and never logs argument values, only their
 Output goes through the standard ``logging`` module, which writes to stderr. It
 must never write to stdout: the stdio transport frames JSON-RPC over stdout, so a
 stray log line there corrupts the protocol.
+
+Line integrity depends on the process's root log handler, which this module does
+not configure. FastMCP's default ``RichHandler`` wraps at width 80 in a non-TTY,
+which would split an event across physical lines; and a deployment that
+reconfigures root logging before construction can drop these INFO lines. The
+hosted logging wiring (the structured-log emitter) owns installing a plain,
+non-wrapping formatter at INFO so the one-line contract holds; until then a host
+that cares must install one.
 """
 
 from __future__ import annotations
--- a/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
+++ b/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
@@ -16,40 +16,38 @@ stray log line there corrupts the protocol.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import time
 from typing import Literal
 
-from mcp import types
+from mcp import UrlElicitationRequiredError, types
 
 from pipefy_mcp.core.tool_middleware import CallNext, ToolCallContext
 
 logger = logging.getLogger("pipefy_mcp.observability.tool_call")
 
-ToolCallOutcome = Literal["ok", "error"]
+# "cancelled" (client disconnect) and "elicitation" (the tool is asking the client
+# to visit a URL) are normal control flow, kept out of "error" so an error-rate
+# alert is not tripped by them. A governance short-circuit still reads as "error":
+# it returns isError=True and is indistinguishable from a tool-reported failure at
+# the result boundary.
+ToolCallOutcome = Literal["ok", "error", "cancelled", "elicitation"]
 
 
 def _outcome(result: types.ServerResult) -> ToolCallOutcome:
     """Map a terminal result to an outcome.
 
-    Reads ``isError`` off the result's ``CallToolResult`` root. The attribute is
-    read defensively: an experimental ``CreateTaskResult`` root has no ``isError``,
-    which should read as ``ok`` rather than raise.
+    Reads ``isError`` off the result's root defensively: an experimental
+    ``CreateTaskResult`` root has no ``isError`` and reads as ``ok`` rather than
+    raising.
     """
-    root = getattr(result, "root", None)
-    return "error" if getattr(root, "isError", False) else "ok"
+    return "error" if getattr(result.root, "isError", False) else "ok"
 
 
 def _emit(event: dict[str, object]) -> None:
-    """Write one structured line via logging (stderr), never stdout.
-
-    Serialization runs before ``logger.info``, so it would otherwise be paid even
-    when the line is discarded; the ``isEnabledFor`` guard skips it when the level
-    is off.
-    """
-    if not logger.isEnabledFor(logging.INFO):
-        return
+    """Write one structured line via logging (stderr), never stdout."""
     logger.info(json.dumps(event, separators=(",", ":")))
 
 
@@ -60,12 +58,13 @@ async def tool_log_middleware(
 
     A tool body error surfaces as a result with ``isError=True`` (FastMCP's
     terminal turns tool exceptions into an error result), so the common path logs
-    from the returned result. The exceptions that DO propagate through the chain,
-    ``CancelledError`` (client disconnect / ``notifications/cancelled``) and
-    ``UrlElicitationRequiredError``, are logged as errors and re-raised: swallowing
-    them would break cancellation and elicitation. ``BaseException`` (not
-    ``Exception``) is caught so a ``CancelledError`` cannot skip the handler and let
-    ``finally`` emit a stale ``ok``.
+    from the returned result. The two exceptions that DO propagate through the
+    chain each get their own non-error outcome and are re-raised: ``CancelledError``
+    (client disconnect / ``notifications/cancelled``) and
+    ``UrlElicitationRequiredError`` (the client must visit a URL to continue).
+    Swallowing either would break cancellation and elicitation. ``BaseException``
+    (not ``Exception``) is caught so a ``CancelledError`` cannot skip the handler
+    and let ``finally`` emit a stale ``ok``.
     """
     started_at = time.perf_counter()
     outcome: ToolCallOutcome = "ok"
@@ -73,19 +72,28 @@ async def tool_log_middleware(
         result = await call_next(ctx)
         outcome = _outcome(result)
         return result
+    except asyncio.CancelledError:
+        outcome = "cancelled"
+        raise
+    except UrlElicitationRequiredError:
+        outcome = "elicitation"
+        raise
     except BaseException:
         outcome = "error"
         raise
     finally:
-        duration_ms = round((time.perf_counter() - started_at) * 1000, 3)
-        _emit(
-            {
-                "event": "tool_call",
-                "tool": ctx.tool_name,
-                "outcome": outcome,
-                "duration_ms": duration_ms,
-                "arg_keys": list(ctx.argument_keys),
-                "client_id": ctx.identity.client_id,
-                "request_id": ctx.request_id,
-            }
-        )
+        # Guard here, not inside _emit, so the event dict is not built when the
+        # line would be discarded.
+        if logger.isEnabledFor(logging.INFO):
+            duration_ms = round((time.perf_counter() - started_at) * 1000, 3)
+            _emit(
+                {
+                    "event": "tool_call",
+                    "tool": ctx.tool_name,
+                    "outcome": outcome,
+                    "duration_ms": duration_ms,
+                    "arg_keys": list(ctx.argument_keys),
+                    "client_id": ctx.identity.client_id,
+                    "request_id": ctx.request_id,
+                }
+            )
--- a/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
+++ b/packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
@@ -1,0 +1,91 @@
+"""Structured per-tool-call logging as a tool-call middleware.
+
+The first consumer of the tool-call middleware chain (see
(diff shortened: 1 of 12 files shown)


gbrlcustodio:
## Follow-up: make the seam publicly registerable, and use it to make FastMCP an implementation detail

Context: I consumed this branch end to end in a real remote deployment (a thin Kubernetes serving layer over `pipefy-mcp-server`) to dogfood the seam. It works: `build_pipefy_mcp_server` installs the chain and the tool logger fires under the remote profile, on `mcp==1.28.1` as well as the pinned 1.25.0. Two concrete gaps surfaced, and they point at a larger opportunity that I think is the real value of this work.

### Gap 1: the seam is installed but not externally registerable

The PR body says "`McpRuntime.register_tool_middleware` is the public registration seam" and "`server.py` reads `runtime.tool_middlewares`." Neither is in the code. `runtime.py` has no middleware references, `server.py` seeds a hardcoded `default_tool_middlewares(settings)`, and `install_tool_call_middleware` raises on a second install. So a consumer of `build_pipefy_mcp_server` cannot add its own middleware through the public path: you get the built-in logger or you reach back into `app._mcp_server.request_handlers`, which is exactly the private slot this PR set out to retire.

Concrete motivating case: the deployment needed per-tool Prometheus metrics (a second consumer of the same seam). With no registration point, the only options were to re-wrap the private handler (defeats the PR) or drop the metrics (what I did). The seam solved the collision problem internally but did not expose the solution.

### Gap 2: `ToolCallContext` leaks the SDK type it is meant to hide

`ToolCallContext` carries `req: types.CallToolRequest` and exposes `arguments` as the SDK-typed map. So a middleware author writes against `mcp.types`, and the wrap targets `app._mcp_server.request_handlers[CallToolRequest]` pinned to `mcp==1.25.0` ("fails loud if that contract changes"). A downstream that consumes this package resolved `mcp==1.28.1` in practice. It happened to still work, but that means the SDK version and the low-level handler shape are currently consumer-facing concerns, not internal ones.

### The opportunity: FastMCP as an implementation detail

Today FastMCP is effectively the public API. `build_pipefy_mcp_server` returns a `FastMCP`, and every extension reaches into it: `custom_route`, `_mcp_server.request_handlers`, `settings.transport_security`, `streamable_http_app()`. This seam is the first chance to invert that. The test for whether the boundary holds: could you upgrade the `mcp` SDK across a breaking change, or swap FastMCP for another MCP server implementation, without any consumer editing code? Right now the answer is no. It should be yes.

A shape that gets there, and that also simplifies the current surface rather than widening it:

```python
@dataclass(frozen=True)
class Plugin:
    """A composable unit of extension. Every field optional; compose many."""
    tool_middlewares: Sequence[ToolCallMiddleware] = ()
    http_middlewares: Sequence[ASGIMiddleware] = ()   # applied outer-to-inner
    routes: Sequence[Route] = ()                       # e.g. /healthz, /metrics
    # session_store: SessionStore | None = None        # the stateless/scaling seam, later

def build_app(settings: Settings, *plugins: Plugin) -> ASGIApp:
    """Assemble a ready-to-serve ASGI app from settings and plugins.

    Folds a built-in core plugin (resource-server wiring, transport security from
    settings, the tool logger under remote) with the caller's plugins, installs the
    tool-call chain once, mounts routes, wraps http middleware, returns the ASGI app.
    No bind and no loopback guard: those are a serving-time policy, not a build concern.
    """
```

Why this is both simpler and more extensible:

- One public entry point that returns a finished ASGI app. `custom_route`, `streamable_http_app()`, and post-construction mutation of `transport_security` all leave the public surface. FastMCP stops being the API.
- The built-in logger stops being a privileged hardcoded path and becomes the first plugin, composed the same way a consumer's plugin is. `default_tool_middlewares` becomes `core_plugin(settings)`.
- Plugins compose (observability, quotas, a consumer's metrics) with no plugin aware of the others, and adding a capability is a new field or a new plugin, never a change to `build_app`'s signature. That is the Open/Closed property a `tool_middlewares=` parameter alone would not give.
- The `mcp` pin and the `request_handlers` reach become purely internal: adapted once, in one place, when the SDK bumps, invisible downstream.

For this to actually hold, `ToolCallContext` must be pipefy-owned: `tool_name: str`, `arguments: Mapping[str, Any]`, `identity: CallerIdentity`, `request_id`. The raw `CallToolRequest` should be removed from the public contract or demoted to an explicitly-unstable escape hatch (see below). Otherwise the seam still binds middleware authors to `mcp.types` and the boundary leaks.

How three consumers collapse onto it:

```python
# CLI
serve_http(build_app(settings), bind=settings.bind)   # loopback guard lives in serve_http

# the remote deployment wrapper, replacing its monkeypatch + mutation entirely
build_app(settings, Plugin(
    tool_middlewares=[prometheus_tool_middleware],
    routes=[Route("/healthz", health), Route("/metrics", metrics)],
    http_middlewares=[AccessLogMiddleware],
))

# tests
build_app(settings, Plugin(tool_middlewares=[capture]))
```

### The discipline this requires

Hiding FastMCP means pipefy_mcp owns every bit of surface it exposes, and anything it does not surface (elicitation, progress, resources, prompts, future SDK features) is unreachable without breaking the abstraction. So the design should include one deliberate, documented, "unstable, may change with the SDK" escape hatch back to the underlying app for the long tail. Without it, "implementation detail" becomes "cage" and people reach around the boundary again.

### Suggested scope for this PR vs later

- In this PR: make registration public (fold consumer plugins/middlewares into the install alongside the built-ins, with a documented order), and make `ToolCallContext` pipefy-owned so the seam does not leak `mcp.types`. These are the two things that make the seam actually usable by a second consumer.
- Follow-up: the full `Settings`/`Plugin`/`build_app` shape and demoting FastMCP behind it, plus `serve_http` owning the bind policy. Keep `build_pipefy_mcp_server(settings)` as a thin shim for one release.

### Related, separable

Consuming this branch also required pinning `pipefy`, `pipefy-auth`, and `pipefy-infra` to the branch, because the `workspace = true` siblings are ahead of the identically-versioned PyPI wheels (the branch SDK adds `PipefyEngine`, which `runtime.py` imports; the PyPI `0.3.0a1` wheel does not, so it crashes at import). "Open to extension" also means "installable by an extender," which the shared non-immutable `0.3.0a1` version currently blocks. That is a packaging concern separate from this PR, worth its own issue.

packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
```diff
@@ -0,0 +1,99 @@
...
+    Reads ``isError`` off the result's root defensively: an experimental
+    ``CreateTaskResult`` root has no ``isError`` and reads as ``ok`` rather than
+    raising.
+    """
+    return "error" if getattr(result.root, "isError", False) else "ok"
+
+
+def _emit(event: dict[str, object]) -> None:
+    """Write one structured line via logging (stderr), never stdout."""
+    logger.info(json.dumps(event, separators=(",", ":")))
```
adriannoes, line 59: One thing I hit checking the logger end to end: the "one JSON line per call" contract doesn't survive the default handler config. Nothing in `run_server` / `main.py` configures logging, so the only config in the process is FastMCP's `configure_logging` (from `FastMCP.__init__`), which prefers `RichHandler` whenever `rich` is importable — and it always is here via `mcp[cli]` → `typer` → `rich`. In a non-TTY (container logs) Rich wraps at width 80, so a realistic `tool_call` event lands in the aggregator as several physical lines with tokens split mid-word. 

Related: the `isEnabledFor(INFO)` guard only passes because FastMCP sets the root logger to INFO at construction, so a deployment that configures root logging first can silently drop every tool-call line. stdout is safe either way (`Console(stderr=True)`), so this is parseability, not protocol corruption. 

Is the plan for the #373 structured-logging emitter to own the handler config when it lands on this seam? If so, a one-line note in the module docstring ("line integrity depends on the root handler; hosted wiring must install a plain formatter") would keep the contract honest until then; otherwise this middleware probably wants its own non-propagating stderr handler.

adriannoes (dismissed):
## Summary

This lands the tool-call middleware seam the hosted stack needs: an ordered chain around FastMCP’s single `CallToolRequest` slot, with short-circuit, request-scoped identity, and `tool_log_middleware` as the first consumer. The public registration path via `build_pipefy_mcp_server(..., extra_tool_middlewares=...)` is the right shape for a second consumer without re-wrapping private handlers. Tests and CI look solid on this head.

## What worked well

- Builder fold-in with built-ins outer and a real session test that proves consumer middleware runs.
- Composition root owns the list; `core` stays free of `observability`.
- Short-circuit envelope parity with in-tool `tool_error` (decoded JSON, intentional `isError` divergence).
- Clear outcome taxonomy on the logger (`ok` / `error` / `cancelled` / `elicitation`) and stderr-safe emission.

## Also noted

- **`ToolCallContext` and `mcp.types`:** `req` is still public (your Gap 2). Not blocking if the pipefy-owned context lands with the `Plugin` / `build_app` follow-up — worth confirming that plan so consumers don’t treat `req` as stable API.
- **JSON line integrity (inline):** under the default FastMCP/`RichHandler` config, a realistic `tool_call` event wraps in non-TTY logs. Sequencing question for #373’s emitter vs a dedicated handler on this middleware — details on the thread.

packages/mcp/src/pipefy_mcp/observability/tool_log_middleware.py
gbrlcustodio: Fixed in 09a598b4. Confirmed the plan: the #373 structured-log emitter (PR #375) owns handler config when it lands on this seam, so I took the docstring path rather than a private handler here (a hardcoded stderr handler would collide with that emitter). Added a note that line integrity depends on the root handler (RichHandler wraps at 80 in a non-TTY; a root reconfigure can drop INFO lines) and that the hosted logging wiring installs a plain non-wrapping formatter at INFO until then.

adriannoes (approved):
Thanks — the docstring plus #373/#375 ownership on the log-line contract closes the thread. Re-approving on this head.

Issue #354, opened 2026-07-03
Unify the GraphQL execution seam: one result type, one protocol


Danielmoraisg:
## Context

PR #349 introduced GraphQL partial-success handling for `automations.executionMetrics`: a response can carry readable `data` alongside per-node `errors` (e.g. `PERMISSION_DENIED`) in one payload, and we want one denied automation not to drop the rest of the batch.

That PR modelled "partial" as a second capability on the execution seam:

- two execute verbs on the adapter: `execute_query` (all-or-nothing, returns `dict`) and `execute_query_allow_partial` (returns `PartialQueryResult`)
- two protocols: `GraphQLExecutor` and `PartialGraphQLExecutor extends GraphQLExecutor`

A follow-up already landed on that PR (Rung 1): the fatal-vs-partial decision was de-smeared. `execute_query_allow_partial` no longer classifies. It always returns a `PartialQueryResult` (a fully null response yields `data={}`), and the single "no readable data means failure" decision now lives in one pure function, `_normalize_execution_metrics_result`. This issue tracks the remaining structural cleanup.

## Problem

Modelling partial-success as a separate transport capability is a false dichotomy. Every GraphQL response can carry `data` and `errors` together; all-or-nothing is just "get the result, then raise if any errors." Baking that split into the seam means:

- the seam has two verbs and two protocols where one of each would do
- `PartialGraphQLExecutor` is a second Protocol with exactly one implementation and one test fake, i.e. speculative interface segregation
- the next query that needs partial handling must widen its service to `PartialGraphQLExecutor` and reach for the second verb, instead of the pattern being the default

## Proposed change (Rung 2)

Collapse to one result type and one protocol.

- The primitive (e.g. `execute` returning `GraphQLResult{data, errors}`) is the only thing the seam really exposes. It performs the effect and never decides what the errors mean. Genuine transport/network failures still raise.
- `execute_query` stays as the raise-on-error convenience the other ~152 tools use, reimplemented as a thin wrapper: run the primitive, raise if `errors`, return `data`.
- Delete `PartialGraphQLExecutor`. Services depend on the one protocol; a service that wants partials reads `errors` off the result and hands it to its own pure classifier.
- Each domain that needs partial handling owns a pure classifier (as observability already does with `_normalize_execution_metrics_result`).

[...]

PR #382, opened 2026-07-10
refactor(sdk): unify the GraphQL execution seam (one protocol, one result type) [#354 Rung 2]

Danielmoraisg:
## What

Rung 2 of #354: collapse the GraphQL execution seam to one protocol and one result type.

Before, partial-success handling was modelled as a second transport capability: a `PartialGraphQLExecutor` protocol (extending `GraphQLExecutor`) added an `execute_query_allow_partial` verb returning a partial-specific `PartialQueryResult`. Any query that needed partial handling had to widen its service to the fatter protocol and reach for the second verb.

This makes partial handling the default shape of one primitive:

- `PartialQueryResult` becomes `GraphQLResult`; `PartialGraphQLExecutor` is deleted.
- `GraphQLExecutor` exposes `execute` (the primitive that returns `data` and `errors` together and decides nothing about what the errors mean) and `execute_query` (the raise-on-error convenience the query/mutation tools use, reimplemented as a thin wrapper over `execute`). Genuine transport/network failures still raise.
- `ObservabilityService` depends on the one protocol, calls `execute`, and hands the errors to its own pure classifier (`_normalize_execution_metrics_result`), unchanged from a caller's view.

## Behavior preservation

- The ~152 `execute_query` callers are unchanged. On GraphQL errors, formatter-less endpoints (public, interfaces) re-raise a `TransportQueryError` carrying the structured `errors` the MCP tool layer reads; the message is byte-identical to gql's (`str(errors[0])`).
- `_on_graphql_error` is preserved: the internal endpoint's `execute_query` still raises a `ValueError` with the `[code=...] [correlation_id=...]` envelope.

## Acceptance criteria (from #354)

- [x] one execution protocol; `PartialGraphQLExecutor` removed
- [x] one primitive returning data + errors; `execute_query` reimplemented over it, external behavior unchanged
- [x] `_on_graphql_error` behavior preserved for the Internal API endpoint
- [x] observability partial-success path unchanged from a caller's view; full SDK + MCP suites green

## Testing

- Full suite green: `uv run pytest packages -m "not integration"` -> 3067 passed, 5 skipped. Ruff clean.
- Added an endpoint-level test that `execute_query` on a formatter-less endpoint raises a `TransportQueryError` carrying structured `.errors` (and that `str(exc)` still carries the code the tool layer's regex reads).
- E2E against a live org exercised all four paths: `execute_query` happy path, the `execute` primitive plus the observability classifier (real automations + pagination), the `execute_query` error path (reconstructed `TransportQueryError` with `PERMISSION_DENIED` intact), and the classifier's null-connection `ValueError`.

## Follow-up

Rung 3 (separate PR, also tracked under #354) relocates error policy out of the transport seam: the protocol shrinks to the single `execute` method, an owned `PipefyGraphQLError` carries the structured errors, and the `[code=...]` string envelope plus its parsers are deleted.

### DIFF
--- a/packages/sdk/src/pipefy_sdk/graphql_executor.py
+++ b/packages/sdk/src/pipefy_sdk/graphql_executor.py
@@ -14,11 +14,13 @@ from httpx import Auth, Timeout
 
 
 @dataclass(frozen=True)
-class PartialQueryResult:
+class GraphQLResult:
     """GraphQL ``data`` plus the raw per-node error dicts from one response.
 
-    Owned by the executor so services and their tests never touch gql objects.
-    ``data`` is always a dict; a fully null response yields an empty one.
+    A GraphQL response can carry readable ``data`` and per-node ``errors`` at the
+    same time, so the primitive hands both back together. Owned by the executor
+    so services and their tests never touch gql objects. ``data`` is always a
+    dict; a fully null response yields an empty one.
     """
 
     data: dict[str, Any]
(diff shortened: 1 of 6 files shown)


adriannoes (approved):
Local checkout plus `pytest packages -m "not integration"` came back green (3068 passed), CI lint/test are green, and the GraphQL seam collapses cleanly to one protocol and one result type without changing the external `execute_query` contract callers already rely on.

## What worked well

- `execute` stays the policy-free primitive (`GraphQLResult` with data and errors); `execute_query` owns the raise path and `_on_graphql_error`.
- Deleting `PartialGraphQLExecutor` removes the speculative second Protocol instead of keeping a one-caller widen.
- The follow-up test that `execute` ignores a configured formatter pins the one regression CI would otherwise miss.
- Reconstructed formatter-less `TransportQueryError` matches gql on the HTTP dict path (`str_first_element` / `str(errors[0])`), and MCP extractors still read structured `.errors`.

Rung 2 of #354 looks ready to merge, @Danielmoraisg

PR #394, opened 2026-07-10
refactor(sdk): move GraphQL error policy out of the transport seam [#354 Rung 3]

Danielmoraisg:
## What

Rung 3 of #354: move GraphQL error policy out of the transport seam.

Rung 2 collapsed the seam to one protocol and one result type. This rung takes the two decisions the seam still baked in (raise-on-error and how to format the error) out of the transport layer and into the pure core, and deletes the `[code=...] [correlation_id=...]` string envelope round-trip.

- Add `PipefyGraphQLError` (owned, carries the raw structured `errors` list) and the pure `data_or_raise(GraphQLResult) -> dict`.
- `execute_query` is now `data_or_raise` over the `execute` primitive and raises `PipefyGraphQLError` for every endpoint. No gql transport type leaks to services, CLI, or MCP.
- Drop the per-endpoint `on_graphql_error` formatter and delete `internal_api_errors.py`.
- Delete the `[code=...] [correlation_id=...]` envelope and its three parsers in the MCP tool layer. Codes and correlation ids now come off the structured `errors` for both endpoints (`extract_graphql_correlation_id` reads `errors[].extensions`, mirroring `extract_graphql_error_codes`). This removes a serialize-then-reparse round-trip: the envelope only existed to encode codes into a string that the tool layer regexed back out.
- Migrate the `except TransportQueryError` sites (SDK services + CLI) to `except PipefyGraphQLError`, and unify portal_service's two permission mappers into one structured mapper.
- Export `PipefyGraphQLError` from `pipefy_sdk`.

## Why the seam keeps two methods

The protocol still exposes `execute` (the primitive) and `execute_query` (the raise-on-error convenience). Collapsing to a single `execute` was considered and rejected: with no static type checker in this repo, `(await ex.execute(...)).data` would silently drop GraphQL errors, and ~140 of ~142 call sites want raise-on-error. Keeping `execute_query` as the ergonomic default means those callers cannot forget to raise, and it is not a second transport capability (it does no I/O; it is `data_or_raise` over `execute`).

## Behavior

- The ~152 `execute_query` callers are unchanged at the call site. On GraphQL errors they now catch `PipefyGraphQLError` instead of a gql `TransportQueryError` / a `ValueError` envelope.
- Debug output is preserved: MCP tools still surface error codes and correlation ids under `debug=True`, now read from the structured `errors`.
- Genuine transport failures (non-2xx, timeouts) still propagate as gql `TransportError` subclasses, unchanged.
- One deliberate CLI change: a non-permission internal-API GraphQL error now exits 1 (with the clean `message (code)` format) instead of 2. Exit 2 is the usage-error code, so a server-side GraphQL failure exiting 2 was wrong; this aligns internal-API CLI errors with public-API ones. Permission errors are unchanged. No CLI test pinned the old exit 2 (every exit-2 assertion is a client-side validation path that never reaches the API).

## Rebased onto dev, with follow-ups

This branch was stacked on #382 (Rung 2). #382 has merged, so the branch is rebased directly onto `dev` and its diff is the Rung 3 delta plus the reconciliation below. Tracked under #354.

`dev` grew `graphql_problem` (a shared SDK error classifier) and several new tool and service suites while this branch was open. That classifier duck-types on an exception's `.errors`, which `PipefyGraphQLError` carries, so the two fit together. Reconciling with `dev` added three commits on top of the rebased Rung 3 work:

- Remove an obsolete test that constructed `GraphQLEndpoint(on_graphql_error=...)`. This rung deletes that formatter, and the behavior the test pinned (the `execute` primitive returns raw errors instead of raising) is already covered by `test_execute_returns_data_and_errors_together` and `test_execute_null_data_yields_empty_result`.
- Route `portal_service`'s permission mapping through the shared `classify_exception`, the classifier `knowledge_base` and `llm_provider` already use, instead of hand-scanning `exc.errors`. Behavior is unchanged for the single-error responses portal returns. One nuance: `classify_exception` reads the first error's `extensions.code`, where the previous mapper scanned every error. Portal responses carry a single error, so this is not observable, but it is a real difference worth stating.
- Migrate the error-simulation tests off gql `TransportQueryError`. Across 35 tool and service test files, 202 sites that built a `TransportQueryError` to stand in for a GraphQL failure now build `PipefyGraphQLError` (from the structured `errors` list, in `except` clauses, and in `pytest.raises`), and the gql import is dropped per file. After this rung the code path can no longer raise `TransportQueryError` to a caller, so the old simulations described a shape that cannot occur.

## Testing

- Full suite green: `uv run pytest packages -m "not integration"` -> 3579 passed, 5 skipped. Ruff check + format clean. `lint-imports` -> 2 contracts kept.
- Added `data_or_raise` / `PipefyGraphQLError` unit tests; the error-simulation tests across SDK, MCP, and CLI raise `PipefyGraphQLError` with structured errors.


adriannoes (commented):
## Summary

Early draft pass on Rung 3. Checked out the tip (`7f9cbd09`), ran SDK/MCP/CLI unit tests (green) and ruff. The error-policy move looks sound: owned `PipefyGraphQLError`, `data_or_raise` over `execute`, envelope round-trip gone, portal mapper unified, and MCP `debug=True` still gates correlation ids from structured `errors`.

Not approving yet: the PR is still CONFLICTING with current `dev` after #382 landed.

## What worked well

- `execute` stays policy-free; `execute_query` is the raise-on-error convenience, so callers cannot silently drop GraphQL errors.
- Reading `errors[].extensions` in the MCP extractors removes the serialize-then-reparse loop the old `[code=...]` envelope needed.
- The Rung 3-lite follow-up already cleaned the non-dict message path and a vacuous envelope-style test.

## Also noted

- Before leaving draft, rebase onto latest `dev` (Rung 2 from #382 is already there). A local merge probe hit content conflicts in five files: CLI `_common.py`, MCP `ai_tool_helpers.py`, SDK `__init__.py`, `graphql_executor.py`, and `test_graphql_endpoint.py`. Keep the Rung 3 shape on those resolutions (owned `PipefyGraphQLError`, no `on_graphql_error` / string envelope, structured extractors), then re-run `uv run pytest packages -m "not integration"` and ruff. Ready when the PR is mergeable again and the tip suite is green.
- Optional while rebasing: a short Unreleased CHANGELOG bullet for the exception swap / CLI internal GraphQL exit 1, and a note on #354 that `_on_graphql_error` is intentionally gone in Rung 3.

## Review path

- Checkout + package unit tests and ruff on the tip; security pass found no medium+ leak regression on the debug gate.
- Adversarial hunts on architecture/tests/API; refutation kept only the merge-readiness rebase note for this comment.

adriannoes (dismissed):
Re-reviewed the rebased tip after the Jul 20 draft pass. The conflict with `dev` after Rung 2 is gone, CI is green, and the local non-integration suite is green on this SHA.

The error-policy move still looks right: owned `PipefyGraphQLError`, pure `data_or_raise`, Internal API string envelope removed, MCP codes/correlation from `errors[].extensions`, and the CLI Internal-API GraphQL exit 1 alignment is pinned (and the pin fails if that path exits 2 again).

What worked well: first-error portal narrowing and `FORBIDDEN`/`UNAUTHORIZED` synonym widening are documented and tested rather than left as rebase folklore. Default agent-visible messages stay message-only; `debug=True` still gates codes/correlation.

Also noted (non-blocking):

- MCP enrichers still match literal `PERMISSION_DENIED` while the shared classifier treats synonym codes as permission denied (normal portal tools still get guidance via the SDK map).
- The new CLI exit-1 test could also assert the `message (code)` stderr form.
- Before land, tip-of-`dev` has moved a bit; expect possible CHANGELOG/test overlap on rebase.

Review path: detached checkout at the frozen head, package unit tests and ruff, targeted contract suites plus an exit-code litmus, architecture/security/thermo passes. Live MCP/CLI smoke of the error branch was not run (needs a forced GraphQL error).

Danielmoraisg:
Thanks for the re-review. Acting on note 2, pushing back on note 1, and note 3 is done. Also ran the live error-branch smoke you flagged as not covered.

## Note 2: done, and it surfaced a second gap

You were right that the exit-1 test only pinned the exit code. Worse than that: its `errors` payload had no `extensions` at all, so it only ever exercised the formatter's codeless fallback. The `message (code)` form the rung introduced had no coverage.

`test_portal_sub_portal_detach_graphql_error_exits_1` now uses the payload the Internal API actually returns for a nonexistent PortalInterface, captured from a live call, and asserts the rendered line:

```
Couldn't find PortalInterface with (RECORD_NOT_FOUND)
```

`correlation_id` is present in that fixture and deliberately absent from the assertion: the CLI renders code and message only.

Two more tests came out of it:

- The codeless branch is now its own test. Nothing covered it, and there are no unit tests for `_format_transport_query_error` itself.
- A `PortalPermissionError` still exits 2. **No CLI test covered this**, which means the "permission errors are unchanged" half of this rung's behavior claim was unpinned while the half that changed (exit 1) was pinned. Every exit-2 assertion already in that file is a client-side validation path that never reaches the API. `PortalPermissionError` subclasses `ValueError`, not `PipefyError`, so it skips the `PipefyError -> Exit(1)` branch in the command runner and keeps exit 2. That now holds by test rather than by inference, which matters because this rung rerouted portal permission mapping through `classify_exception`.

## Note 1: pushing back, this is out of scope for the rung

I'd rather not change the enrichers here. `_PERMISSION_CODES = {"PERMISSION_DENIED", "FORBIDDEN", "UNAUTHORIZED"}` already exists on `dev`, and the enrichers that match the literal code are untouched by this diff. The inconsistency you spotted is real, but it predates the rung: this PR widened only the SDK/portal side by routing through the shared classifier, and did not narrow anything on the MCP side.

Widening the enrichers would be an agent-visible message change inside a PR whose contract is that behavior is unchanged except for the one documented exit code. That belongs in its own change, where it can be tested and released as a behavior change rather than smuggled in under a refactor. Noted on #354 instead.

Worth adding that I could not get the live API to emit `FORBIDDEN` or `UNAUTHORIZED` in `extensions.code` in any of the errors I forced, so I have no evidence the divergence is reachable in practice. I also can't prove it isn't.

## Note 3: rebased

Rebased onto `dev` (9 new commits: the S3 2xx PUT gate, service-account success-flag gating, the MCP fail-closed fix). No conflicts this time. The `TransportQueryError` migration survived it: the only remaining reference in `packages/*/src` is the single intentional catch inside `GraphQLEndpoint.execute` that converts gql's exception into `GraphQLResult`, and there are zero references left in tests.

Post-rebase: `uv run pytest packages -m "not integration"` -> 3740 passed, 5 skipped. `ruff check packages` and `ruff format --check` clean. `lint-imports` -> 2 contracts kept.

## The live error-branch smoke

You noted this needed a forced GraphQL error and wasn't run. It was worth running: the rung rewrote 202 error simulations to build `PipefyGraphQLError` from hand-authored `errors` lists, so those tests are self-consistent by construction and cannot tell us whether the real payload matches the shape they assume.

Dumping the raw seam output on a forced Internal API error settles it:

```
=== internal_api: execute (primitive, policy-free) ===
data   = {"updateSubPortalElement": null}
errors = [{"message": "Couldn't find PortalInterface with",
           "path": ["updateSubPortalElement"],
           "extensions": {"code": "RECORD_NOT_FOUND",
                          "correlation_id": "a22c1fbab98ae41d-GRU"}}]

=== internal_api: execute_query (raise-on-error) ===
type   = PipefyGraphQLError
errors = [ ...same structured list... ]
```

Real errors carry `extensions.code` and `extensions.correlation_id` (snake_case), which is exactly what `extract_graphql_error_codes` and `extract_graphql_correlation_id` read. `execute` returned data and errors without raising; `execute_query` raised with the full list attached.

Forced errors on all three endpoints, plus the debug gate through a real MCP session:

| Path | Result |
|---|---|
| Public (`card get 999999999999`) | exit 1, `Permission denied (PERMISSION_DENIED)` + card hint |
| Internal (`portal sub-portal unpublish`, all-zeros UUID) | exit **1**, `Couldn't find PortalInterface with (RECORD_NOT_FOUND)` |
| Interfaces (`portal get`, bogus UUID) | exit 1 through the unified mapper |
| MCP `delete_card`, `debug=False` | message only, no code leak |
| MCP `delete_card`, `debug=True` | `(debug: codes=PERMISSION_DENIED; correlation_id=a22c2197fec30ad0-GRU)` |

The MCP calls went through `build_pipefy_mcp_server` and a real client session. `delete_card` on a nonexistent id fails at the `get_card` pre-fetch, before the destructive-confirmation guard, so no mutation ran.

`pytest packages -m integration` against the live API: 33 passed, 12 skipped. Two caveats worth stating rather than burying:

- The live-creds gate ignores the stored-session tier by design, so I promoted a keychain token to `PIPEFY_TOKEN` to make those tests execute. They ran on the static-token tier, not the tier the CLI normally uses.
- Portal `PERMISSION_DENIED` was never forced live. Portal reads succeeded on both orgs I tried, and forcing it needs a write against an org without `manage_portals`. I verified the halves separately (portal errors carry `extensions.code` live; `PERMISSION_DENIED` arrives in `extensions.code` live) but not the combination.

## One pre-existing failure, not from this branch

`test_live_get_pipe_then_get_ai_agents` fails on `assert "agents" in agents_payload`, because the payload is `{"data": {"agents": []}, "success": True}`: the unified envelope nests under `data` and the assertion still reads top level. Same failure at the same line in a detached worktree at `origin/dev`, and this branch touches neither that file nor the envelope. It looks like a live test left behind by the envelope migration, worth its own fix.

adriannoes (approved):
Re-approved after the dismissed tip. The force-push rebased onto current `dev` and added one test-only commit that closes the stderr-format gap from the last pass.

The new CLI pins look right: Internal API GraphQL failures assert both exit 1 and the `message (code)` line (with `correlation_id` kept out of stderr), the codeless formatter branch is covered, and `PortalPermissionError` still exits 2. No production code changed in that follow-up. CI is green on this SHA.

The earlier non-blocking note about MCP enrichers staying on literal `PERMISSION_DENIED` while the shared classifier also treats `FORBIDDEN`/`UNAUTHORIZED` still stands as optional follow-up, not a merge gate.

PR #376, opened 2026-07-08
feat(mcp): add HTTP request log middleware and hosted wiring (#373 stack 2/3)

adriannoes:
Part of pipefy/ai-toolkit#373 (stack 2/3). Depends on #375.

## Summary

- Pure-ASGI `RequestLogMiddleware`: one JSON line per HTTP request, no body buffering (ordering + disconnect tests).
- `wire_hosted_observability`: calls `streamable_http_app()` once, mounts middleware, stores `request_id` in `scope["state"]`.
- HTTP transport serves via manual uvicorn with `access_log=False` (structured lines replace text access logs).
- Configures the stdout emitter only on the HTTP path (stdio stdout stays the JSON-RPC wire).

## What lands

Structured **HTTP request** lines on the hosted profile. Tool-call lines arrive in stack 3/3 (#377).

## Stack status

- Rebased onto updated `dev` (post-#348) and restacked on #375.
- **Draft** — merge after #375; will be marked ready for review once #375 lands.
- Stdio transport unchanged (no middleware, no structured stdout).

## Test plan

- [x] `uv run pytest packages/mcp/tests/observability/test_request_log_middleware.py packages/mcp/tests/observability/test_wiring.py -q`
- [x] `uv run pytest packages/mcp/tests/test_server.py -q`
- [x] `rg BaseHTTPMiddleware packages/mcp/src` (only documented prohibition in AGENTS.md)

PR #396, opened 2026-07-12
feat(mcp): replace bind-host guard with auth-posture interlock

### LINKED ISSUE #379: Replace bind-host guard with auth-posture interlock

## Scope

Remove the bind-interface guard `_assert_safe_http_bind` and replace it with the auth-posture interlock described in #300.

The current guard, in `run_server`, refuses any non-loopback bind host before serving over HTTP. It checks the wrong axis. The property worth protecting is "an unauthenticated profile must not be served to untrusted callers," but bind interface is a poor proxy for that:

- In a container everything binds `0.0.0.0` and is still private (network namespace, no published port), so the guard false-positives on the entire hosted profile.
- Behind a proxy the server must bind non-loopback, and it is safe because the remote profile has inbound auth (resource-server validation, per-request bearer, host allowlist). Every firing of the guard against the hosted profile is a false positive.

It also lives in the run path rather than the app factory, so it protects nothing for callers that build the ASGI app directly and is bypassed simply by calling `streamable_http_app()`. That bypass is the intended path for off-loopback serving today, which trains callers to route around the control and has already led a downstream repo to vendor a guard-less copy of the package.

The replacement is the interlock from #300, keyed on auth posture, not bind host:

- No-inbound-auth profile (local / stdio-style): refuse non-loopback HTTP unless an explicit escape hatch is set. This is the real footgun and the thing worth failing closed on.
- Resource-server profile (remote / hosted, auth on): bind host is irrelevant and is not checked. The operator binds `0.0.0.0` legitimately.

`_assert_safe_http_bind` and the interlock must not coexist: leaving both re-creates the same false positives and the same bypass. Shipping the interlock should delete the old guard.

## Acceptance criteria

- `_assert_safe_http_bind` (the bind-interface check) is removed.
- The remote / resource-server profile serves on a non-loopback host with no bind-host guard and no bypass required.
- The no-auth profile refuses non-loopback HTTP unless an explicit escape hatch is set, and that refusal is exercised by a test.
- No supported serving path requires calling internal factory methods to route around a guard.

## Dependencies

Realizes the safety interlock scoped in #300 (closed) and retires the legacy guard it was meant to supersede. Related to the configurable transport host allowlist (#303).

gbrlcustodio:
Closes #379.

## What

Replaces the bind-interface guard `_assert_safe_http_bind` with an auth-posture interlock, realizing the safety interlock scoped in #300 and retiring the legacy guard it was meant to supersede.

## Why the old guard was wrong

The old guard refused any non-loopback HTTP bind. It checked the wrong axis. The property worth protecting is "an unauthenticated profile must not be served to untrusted callers," but bind interface is a poor proxy:

- In a container everything binds `0.0.0.0` and is still private (network namespace, no published port), so the guard false-positived on the entire hosted profile.
- Behind a proxy the server must bind non-loopback, and the `remote` profile is safe there because it has inbound auth (resource-server validation, per-request bearer). Every firing against `remote` was a false positive.

It also lived in `run_server`, not the app factory, so it protected nothing for callers that build the ASGI app directly via `streamable_http_app()`. That bypass was the intended off-loopback path today, which trains callers to route around the control and has already led a downstream repo to vendor a guard-less copy.

## The interlock

Keyed on auth posture, not bind host, and expressed as a cross-field validator at the settings boundary (`McpSettings._enforce_bind_safety`), so no serving path can route around it: `build_pipefy_mcp_server` and any direct-ASGI-app caller both take a resolved `Settings`.

- No-inbound-auth posture (`profile == "local"`): refuses a non-loopback HTTP bind unless the escape hatch `PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND` is set. This is the real footgun (the full tool surface, no inbound bearer, reachable off the local machine).
- Resource-server posture (`profile == "remote"`): bind host is irrelevant and is not checked. The operator binds `0.0.0.0` legitimately. `remote` always carries inbound auth or fails fast in `McpRuntime.for_profile`, so keying on `profile == "local"` is a faithful proxy for posture at the settings layer.

Loopback detection moves to a new `pipefy_infra.security.is_loopback_host`, a generic security primitive that uses `ipaddress(...).is_loopback` to cover all of `127.0.0.0/8` and `::1`, not the three string literals the old `_LOOPBACK_HOSTS` frozenset hardcoded.

## Acceptance criteria

- `_assert_safe_http_bind` (the bind-interface check) is removed.
- The remote / resource-server profile serves on a non-loopback host with no bind-host guard and no bypass.
- The no-auth profile refuses non-loopback HTTP unless the escape hatch is set, exercised by a test.
- No supported serving path requires calling internal factory methods to route around a guard.

## Testing

- `uv run python -m pytest packages/infra/tests/test_security.py packages/mcp/tests/test_settings.py packages/mcp/tests/test_server.py packages/mcp/tests/test_main.py`: 145 passed.
- Full non-integration `packages/infra/tests packages/mcp/tests`: 1607 passed, 5 skipped.
- `cd packages/mcp && uv run lint-imports`: contract kept.
- `uv run ruff check packages/infra packages/mcp`: clean.

## Not in scope

- The configurable transport host / Origin allowlist for DNS-rebinding protection (#303) stays open and is unaffected.
- Follow-up worth doing: the "auth posture" axis is re-derived from `profile == "remote"`/`"local"` at several sites (settings, server, runtime); a single named posture property on `McpSettings` would consolidate it. Deferred to keep this change focused.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -125,7 +125,8 @@ These variables load into `pipefy_mcp.McpSettings` (`settings.mcp`). TOML keys u
 |----------|---------|--------|
 | `PIPEFY_MCP_PROFILE` | `local` | Launch profile: `local` registers every tool; `remote` exposes only remote-safe tools and validates an inbound bearer per request. |
 | `PIPEFY_MCP_TRANSPORT` | derived from profile | `stdio` or `http`. Unset: `local` → `stdio`, `remote` → `http`. |
-| `PIPEFY_MCP_HOST` | `127.0.0.1` | HTTP bind host (loopback only until hosted OBO lands). |
+| `PIPEFY_MCP_HOST` | `127.0.0.1` | HTTP bind host. The unauthenticated `local` profile refuses a non-loopback bind unless `PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND` is set; the authenticated `remote` profile binds any host. |
 | `PIPEFY_MCP_PORT` | `8000` | HTTP bind port. |
+| `PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND` | `false` | Escape hatch: lets the unauthenticated `local` profile serve HTTP on a non-loopback host, exposing the full tool surface with no inbound bearer. The `remote` profile never needs it. |
 | `PIPEFY_MCP_UNIFIED_ENVELOPE` | `true` | When true, migrated tools return `{success, data, ...}`. |
 | `PIPEFY_MCP_LOG_LEVEL` | `INFO` | Governs **both** hosted structured JSON lines on **stdout** and the FastMCP root logger (RichHandler text on **stderr**). Accepts `DEBUG`, `INFO`, `WARNING`, `ERROR`, or `CRITICAL` (case-insensitive). The server passes the resolved level to `FastMCP(log_level=...)` at construction, so this is the single operator knob for MCP logging. Some MCP tooling documents a separate `FASTMCP_LOG_LEVEL` env var; when using `pipefy-mcp-server`, prefer `PIPEFY_MCP_LOG_LEVEL` — it configures the root logger explicitly and keeps structured events on the same level. |
--- a/packages/infra/src/pipefy_infra/security.py
+++ b/packages/infra/src/pipefy_infra/security.py
@@ -12,6 +12,9 @@ Surface:
   literal-IP check (callers must follow up with the DNS gate).
 * :func:`assert_hostname_is_not_internal`: building block of the above;
   exposed for callers that already have a parsed hostname.
+* :func:`is_loopback_host`: predicate reporting whether a bind host keeps
+  a server reachable only from the local machine (``localhost`` or a
+  literal IP in ``127.0.0.0/8`` or ``::1``).
 * :func:`assert_hostname_resolves_to_public_ips`: asynchronous DNS gate.
   Resolves the hostname and rejects when any resolved IP is in a blocked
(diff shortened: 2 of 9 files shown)


adriannoes (approved):
## Summary

Looks good to merge. 

- Local checkout+tests on `9c818451` (ruff, targeted 145, full infra+mcp 1607, lint-imports kept) and CI lint/test are green. 
- The auth-posture interlock at `McpSettings._enforce_bind_safety` matches #379: local+http refuses non-loopback without the hatch, remote can bind off-loopback, and the old run-path `_assert_safe_http_bind` is gone.

## What worked well

- Moving the control to the settings boundary closes the ASGI/`streamable_http_app` bypass of the old `run_server`-only guard.
- Keying on auth posture (`local` vs `remote`) fixes the false positives on container `0.0.0.0` and authenticated hosted binds.
- `is_loopback_host` via `ipaddress.is_loopback`, with clear escape-hatch docs and tests.

## Also noted

- Merge-order heads-up with the open #373 stack (PR #376 / `feat/373-request-log-middleware`): `server.py` conflicts where this PR removes `_assert_safe_http_bind` and that branch inserts `_serve_streamable_http`, and both rewrite the tail of `run_server`. The new test `test_run_server_remote_serves_off_loopback_without_a_bind_guard` asserts `fake_app.run("streamable-http")`, which that serve path replaces with `anyio.run(_serve_streamable_http, ...)`, so whoever lands second needs to retarget that assertion. The settings-boundary interlock still covers that custom uvicorn path, which is exactly the coverage argument here.

PR #397, opened 2026-07-12
feat(mcp): configurable transport host/Origin allowlist

### LINKED ISSUE #303: Configurable transport host allowlist

gbrlcustodio:
Closes #303.

## What

Makes the HTTP transport's DNS-rebinding host/Origin allowlist configurable. The public host is derived from `resource_server_url`, so a proxied deployment serves without a `421` while DNS-rebinding protection stays on, and the loopback-only default holds when nothing is configured.

## Why

FastMCP auto-enables a loopback-only allowlist when built with a loopback `host` and no explicit `transport_security`. `build_pipefy_mcp_server` always constructs `FastMCP(host=settings.mcp.host, ...)` with the default `127.0.0.1`, so behind a proxy that forwards the public `Host` the transport answers `421 Misdirected Request` before any handler runs. The bind interface is irrelevant; the construction-time `host` arg is what trips it.

## The change

- `McpSettings.allowed_hosts` / `allowed_origins` (env `PIPEFY_MCP_ALLOWED_HOSTS` / `PIPEFY_MCP_ALLOWED_ORIGINS` as JSON), with a light normalization validator (strip, drop empties; `localhost` is a wanted entry, so no SSRF gate).
- `ResourceServer` (moved into the `auth` module) parses `resource_server_url` once into the raw URL plus its wire-form host authorities. The runtime parses it a single time and feeds that one value to both consumers, the allowlist builder and the inbound resource-server auth, so they cannot disagree on the host.
- `core/transport_security.py:build_transport_security(mcp, resource)`: pure resolver. Derives loopback plus the resource host (bare and `host:*`) plus explicit entries; returns `None` only when there is no resource and `allowed_origins` is unset, so FastMCP's own default stands. Origins are derived `http`/`https` per host unless `allowed_origins` overrides; an explicit empty `allowed_origins` is honored verbatim as the strictest posture (reject any request that carries an Origin).
- `build_pipefy_mcp_server` passes the resolved value into `FastMCP(transport_security=...)`.

Placement: this is configuration resolution, not an MCP extension, so it mirrors `build_resource_server_auth` (settings to the SDK's `AuthSettings`) and lives in the composition tier (`core/`, target `wiring/`), keeping the mcp SDK out of `settings.py`. No enable/disable toggle: widen via the allowlist. No `X-Forwarded-Host` handling (the nginx ingress preserves `Host`).

## Acceptance criteria

- Behind a proxy with a public `Host` the server does not return `421`: exercised by `test_http_configured_allowlist_accepts_a_public_host`.
- The loopback-only default is preserved when the allowlist is unset: `build_transport_security` returns `None`, and `test_http_loopback_default_rejects_a_public_host` shows the default still `421`s a public `Host`.

## Testing

- Full non-integration suite (`packages/infra/tests packages/mcp/tests`): 1630 passed, 9 skipped.
- `cd packages/mcp && uv run lint-imports`: contract kept.
- `uv run ruff check packages/mcp && uv run ruff format --check packages/mcp`: clean.

## Migration note

Once `resource_server_url` is set, DNS-rebinding protection is on. This deliberately does not reproduce the wrapper's current "empty `MCP_ALLOWED_HOSTS` disables the check" escape; widening is done by allowlisting, not by turning protection off.


docs/config.md
```diff
@@ -128,5 +132,7 @@ These variables load into `pipefy_mcp.McpSettings` (`settings.mcp`). TOML keys u
 | `PIPEFY_MCP_HOST` | `127.0.0.1` | HTTP bind host. The unauthenticated `local` profile refuses a non-loopback bind unless `PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND` is set; the authenticated `remote` profile binds any host. |
 | `PIPEFY_MCP_PORT` | `8000` | HTTP bind port. |
 | `PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND` | `false` | Escape hatch: lets the unauthenticated `local` profile serve HTTP on a non-loopback host, exposing the full tool surface with no inbound bearer. The `remote` profile never needs it. |
+| `PIPEFY_MCP_ALLOWED_HOSTS` | unset | JSON array of extra `Host` header values the HTTP transport accepts for DNS-rebinding protection, on top of loopback and the `PIPEFY_MCP_RS_RESOURCE_SERVER_URL` host. Unset derives the allowlist from the resource-server URL, so a proxied deployment usually needs none; set it only when a proxy forwards a public `Host` that differs. An entry matches an exact `Host` or, as `host:*`, any port. Empty preserves FastMCP's loopback-only default. |
```
adriannoes: The last sentence of this row ("Empty preserves FastMCP's loopback-only default") reads as if setting `[]` restores the loopback-only default even when `resource_server_url` is configured. The resolver treats an empty array the same as unset (`allowed_hosts or []`, `if explicit_origins:` in `build_transport_security`), so with a resource-server URL the public host is still derived. Same family on origins: an explicit `allowed_origins=[]` is the strictest override the field description advertises, and it is silently replaced by the derived origins.

Two cheap ways to align, pick either: reword this sentence (and the origins description) to "an empty array behaves like unset", or honor explicit-empty for origins with `if explicit_origins is not None:` so `[]` means "reject any Origin present". For hosts, "extra entries on top of" makes `[]` equal to unset legitimately; only the wording needs the fix there.

You will know it is right when the table's promise matches what `build_transport_security` returns for `[]`.

adriannoes (dismissed):
## Summary
Reviewed with checkout + tests (1622 passed, CI green on `f87f3d7d`). The allowlist design matches #303: derive from `resource_server_url`, widen via config, keep protection on, and leave FastMCP's loopback default when nothing is set. Ready to merge from my side.

## What worked well
Pure `build_transport_security` in the composition tier (mcp SDK kept out of settings), dual `host` / `host:*` forms that match the SDK matcher, and both AC directions covered (default still 421s a public Host; configured allowlist clears the gate). The migration note correctly refuses the empty-disable escape.

## Also noted
- **CHANGELOG:** worth an Unreleased bullet for `PIPEFY_MCP_ALLOWED_HOSTS` / `PIPEFY_MCP_ALLOWED_ORIGINS` and the intentional protection-on-when-RS-URL flip (this window also covers the deferred note from #396).
- **Rollout:** on a non-loopback bind, FastMCP previously left Host/Origin validation off; once `resource_server_url` is set this PR turns the gate on. Unauthenticated traffic still 401s before the Host check. Sweep authenticated callers that reach the server by a non-public Host (service DNS); `PIPEFY_MCP_ALLOWED_HOSTS` covers them if any exist.

docs/config.md
gbrlcustodio: Fixed in 6edcb079: honored explicit-empty for origins. `build_transport_security` now gates on `explicit_origins is not None` (and the early-return guard likewise), so `allowed_origins=[]` is kept verbatim as the strictest override (rejects any request that sends an Origin; a request with no Origin still passes). Hosts stay additive, so `[]` there behaves like unset and the resource-server host is still derived; I corrected the config.md wording and the `allowed_origins` field description to match.

adriannoes (dismissed):
## Summary
- Double-checked the commits after the previous APPROVE (`f87f3d7d` → `0c5d09b3`). 
- Checkout + tests: 1634 passed, CI green. The #303 allowlist behavior still holds, and the empty-origins fix from the prior thread is in place (`allowed_origins=[]` honored via `is not None`, docs and unit tests updated). Ready to merge from my side.

## What worked well
Single `ResourceServer` parse in `McpRuntime.for_profile` feeding both the allowlist and inbound auth, blank allowlist entries failing closed at settings, IPv6 authorities bracketed for well-formed URLs, and the composition refactors keeping the mcp SDK out of the settings boundary.

## Also noted
- **CHANGELOG:** still worth an Unreleased bullet for `PIPEFY_MCP_ALLOWED_HOSTS` / `PIPEFY_MCP_ALLOWED_ORIGINS` and the intentional protection-on-when-RS-URL flip (covers the deferred note from #396 as well).
- **Rollout:** on a non-loopback bind, FastMCP previously left Host/Origin validation off; once `resource_server_url` is set this PR turns the gate on. Sweep authenticated callers that reach the server by a non-public Host (service DNS); `PIPEFY_MCP_ALLOWED_HOSTS` covers them if any exist.
- **Unbracketed IPv6:** `https://2001:db8::1/mcp` can still pass `ResourceServerSettings` (urlparse sees host `2001`) and then raise inside `ResourceServer.from_url` when reading `.port`. Bracketed forms work. Not a #303 AC blocker for DNS hostnames; worth a clear config error or settings-boundary reject if you touch that path again.

mocha06 (dismissed):

PR #399, opened 2026-07-13
docs: fix pipefy-login AUTH_URL note and document local-checkout plugin testing

### LINKED ISSUE #235: cli: improve macOS keychain write-failure messaging and first-run UX

mocha06:
## What

Two documentation changes.

### 1. `/pipefy:pipefy-login` command — drop the false `PIPEFY_AUTH_URL` requirement
`commands/pipefy-login.md` claimed `PIPEFY_AUTH_URL` must be set in the shell. It isn't required — it defaults to the prod OIDC issuer (`https://signin.pipefy.com/realms/pipefy`). Only non-prod IdPs need it. Reworded to say so.

### 2. Document testing the plugin from a local checkout
Adds a README section on pointing the plugin marketplace at a local clone (so a working branch loads live), including the `/reload-plugins` step and the `extraKnownMarketplaces` lock gotcha. Linked from CONTRIBUTING's skill quick-start so contributors can try skills before opening a PR.

## Dropped from this PR

An earlier revision also rewrote the macOS keychain troubleshooting entry in `docs/cli/auth.md`. That's been removed — #265 already owns that section (with CLI code, and closes #235), so the keychain guidance is deferred there to avoid a competing narrative. See the review thread for the `-25244` constant details.

## Scope

Docs and command text only. No code or behavior changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,8 @@ Skills are Markdown-only — no Python, no `uv`, no test infrastructure required
 5. Open a PR. CI validates frontmatter, MCP tool names, and `pipefy` CLI
    subcommands referenced in each `SKILL.md`.
 
+> **Try your skill in Claude Code before opening the PR.** Point the plugin marketplace at your local clone so your branch loads live — see [Test the Claude Code plugin from a local checkout](README.md#test-the-claude-code-plugin-from-a-local-checkout).
+
 ### Frontmatter requirements
 
 Every `SKILL.md` must have valid YAML frontmatter with:
--- a/README.md
+++ b/README.md
@@ -219,6 +219,19 @@ npx @modelcontextprotocol/inspector uv --directory . run pipefy-mcp-server
 
 **Adding an MCP tool:** implement under `packages/mcp/src/pipefy_mcp/tools/`, register in `ToolRegistry`, add the name to `PIPEFY_TOOL_NAMES`, and ship the matching CLI command (or document a deferral in `docs/parity.md`). See [`AGENTS.md`](AGENTS.md) for the full TDD workflow.
 
+### Test the Claude Code plugin from a local checkout
+
+The [Claude Code install](#claude-code) adds the marketplace from the `pipefy/ai-toolkit` GitHub repo, which tracks `main`. To run **your local branch** (e.g. `dev`) as the plugin instead, point the marketplace at your clone:
+
+```text
+/plugin marketplace add /absolute/path/to/ai-toolkit
+/plugin install pipefy@pipefy
+```
+
+Whatever is checked out in that clone — any branch — is what loads. Use the `plugin@marketplace` form (`pipefy@pipefy`) since the marketplace and the plugin share the name `pipefy`. After editing plugin files (skills, commands), run `/reload-plugins` to pick up changes without restarting.
+
+> **Already installed the GitHub version?** A marketplace named `pipefy` can be registered only once, and a marketplace declared in `~/.claude/settings.json` under `extraKnownMarketplaces` is locked — `/plugin marketplace add` becomes a no-op (`already on disk — declared in user settings`) and keeps pointing at GitHub. Run `/plugin marketplace remove pipefy` first (or delete that `extraKnownMarketplaces` entry), **then** add the local path.
+
 ---
 
 ## Contributing
--- a/commands/pipefy-login.md
+++ b/commands/pipefy-login.md
@@ -13,4 +13,4 @@ Otherwise prompt the user to confirm running:
 pipefy auth login $ARGUMENTS
 ```
 
-Requires `PIPEFY_AUTH_URL` set in the shell environment (the OIDC issuer URL).
+This targets Pipefy production by default. Only set `PIPEFY_AUTH_URL` (the OIDC issuer URL) in the shell environment when logging in to a non-prod IdP; it defaults to `https://signin.pipefy.com/realms/pipefy`.


docs/cli/auth.md
```diff
@@ -216,7 +216,7 @@ Every `PIPEFY_*` env var is validated against a semantically meaningful regex at
 
 ### `Login succeeded but the session could not be stored in your keychain (<backend>)`
 
-The login worked but `keyring` couldn't write the entry. On macOS / Windows this is rare. On headless Linux it usually means no Secret Service daemon is running — install `gnome-keyring` or `kwallet`, set `PIPEFY_KEYCHAIN_BACKEND=file` to use a plaintext file backend under the Pipefy config directory, or fall back to a static `PIPEFY_TOKEN`.
+The login worked but `keyring` couldn't write the entry. On macOS this happens reliably when `pipefy` runs on uv-managed Python (python-build-standalone), which lacks the keychain entitlements `Security.framework` requires — surfacing as error `-25244`; use a Homebrew `python3` (reinstall the tool so its venv picks it up) or set `PIPEFY_KEYCHAIN_BACKEND=file`. On headless Linux it usually means no Secret Service daemon is running — install `gnome-keyring` or `kwallet`, set `PIPEFY_KEYCHAIN_BACKEND=file` to use a plaintext file backend under the Pipefy config directory, or fall back to a static `PIPEFY_TOKEN`.
```
adriannoes: The new keychain troubleshooting line ties uv-managed Python (PBS) "lacks the keychain entitlements Security.framework requires" to error `-25244`. In Apple's SecBase.h, `-25244` is `errSecInvalidOwnerEdit` (invalid owner edit). Missing entitlement is a different status (`errSecMissingEntitlement`, `-34018`). As written, this entry teaches the wrong constant for the code users see in the CLI.

A small rewrite is enough: keep the remediations (Homebrew Python / `PIPEFY_KEYCHAIN_BACKEND=file`), drop the entitlements↔`-25244` equivalence (or name `errSecInvalidOwnerEdit` if you want a constant), and soften "happens reliably" unless you have a source beyond aligning with `install.sh` (that comment has the same mislabel). If you keep the Homebrew path, note that a bare `uv tool install --force` often stays on PBS unless you pass `--python <system-python>` (as `install.sh` does).

You will know it is right when this paragraph no longer claims that `-25244` means missing entitlements, and the AUTH_URL change remains as-is.

adriannoes (changes requested):
## Summary
Static review of the docs-only changes (checkout not required; CI lint/test are green). The `PIPEFY_AUTH_URL` fix in `commands/pipefy-login.md` matches the prod default in code. The keychain troubleshooting rewrite in `docs/cli/auth.md` needs a correction before merge: it equates `-25244` with missing Security.framework entitlements, which does not match Apple's SecBase mapping.

## What worked well
- Dropping the false `PIPEFY_AUTH_URL` requirement and documenting the prod issuer default.
- Replacing "is rare" with concrete remediations (`PIPEFY_KEYCHAIN_BACKEND=file`, system/Homebrew Python).
- The local Claude Code marketplace checkout section is useful for contributors.

## Required before merge
- **F1:** Rewrite the macOS keychain sentence so `-25244` is not described as missing entitlements (see inline).

## Also noted
- Open PR #265 edits the same troubleshooting section with a different narrative; reconcile when the second of these lands. Updating `packages/mcp/README.md` can stay with that follow-up.

## Verification
- Confirm `DEFAULT_AUTH_URL` remains `https://signin.pipefy.com/realms/pipefy`.
- Confirm the auth.md paragraph no longer ties `-25244` to missing entitlements.

docs/cli/auth.md
adriannoes: Extra context for the "happens reliably" / PBS entitlements framing: in #235, a controlled repro from the same uv-tool-installed Python successfully called `keyring.set_password`, which is why that discussion treated the failure as non-deterministic rather than a hard missing-entitlement gate. SecBase still maps `-25244` to `errSecInvalidOwnerEdit` (not `-34018` / `errSecMissingEntitlement`), so I would keep remediations but drop or soften that causal equation.

mocha06: Good catch, and confirmed. I checked `SecBase.h` in the macOS 26.5 SDK:

- `errSecInvalidOwnerEdit = -25244`
- `errSecMissingEntitlement = -34018`
- `errSecParam = -50`

So `-25244` is an ownership error, not an entitlements one — my framing (inherited from `install.sh`'s comment) was wrong. I've dropped the `docs/cli/auth.md` hunk from this PR entirely rather than ship a corrected-but-competing version, since #265 already owns this section, has the CLI-side hints, and closes #235. This PR is now just the `PIPEFY_AUTH_URL` fix and the local-checkout docs.

One thing for #265 when it lands: it currently labels the same code as `errSecParam (-25244)`, which is also off (`errSecParam` is `-50`). Worth switching both #265 and the existing `packages/mcp/README.md` note to `errSecInvalidOwnerEdit`, and correcting the `install.sh` comment too. Happy to leave this for #265 to ship.

adriannoes (approved):
## Summary
Re-checked the latest head (`b586c6ed`). The contested `docs/cli/auth.md` keychain rewrite is gone, so F1 is addressed by dropping that hunk and deferring the narrative to #265. What remains is the accurate `PIPEFY_AUTH_URL` default note in `commands/pipefy-login.md` plus the local Claude Code marketplace checkout docs.

## What worked well
- Confirming the SecBase mapping in-thread and choosing handoff to #265 instead of shipping a competing keychain story.
- Keeping the AUTH_URL fix and the contributor local-checkout guidance.

## Verification
- Diff no longer touches `docs/cli/auth.md`.
- CI lint/test green on the current head.

Issue #297, opened 2026-06-18
Proposal: hosted (HTTP) on-behalf-of profile for the MCP server


gbrlcustodio:
## Summary

Add a hosted profile to the MCP server: run it as a multi-user HTTP service where each request acts as the calling end user, rather than as today's local stdio process bound to a single identity. The server would act as an OAuth 2.0 resource server, validating an inbound bearer per request and carrying that identity through to the API call.

A local prototype validated the end-to-end flow: resource-server discovery, the unauthenticated challenge, request-scoped identity, concurrency isolation across distinct users, on-behalf-of downstream calls, and reachability behind a proxy. This issue proposes turning that into a supported capability and lists the changes and open questions.

## Motivation

Integrators want browser-based conversational agents that act as their own end users. The current stdio server resolves one identity at startup and runs as a subprocess on the user's machine, so it cannot serve many users from one hosted process. A hosted profile enables that without forking the tool registry or SDK.

## Proposed architecture

```mermaid
flowchart LR
    User([End user])
    Agent[Client / agent]
    MCP[Hosted MCP server]
    API[(API)]
    Auth[[Authorization server]]

    User -. "Browser login (PKCE)" .-> Auth
    Auth -. "access token (bearer)" .-> User
    User --> Agent
    Agent -- "Authorization: Bearer (per request)" --> MCP
    MCP -- "on-behalf-of (request-scoped auth)" --> API
    API -- response --> MCP
```

- Transport: Streamable HTTP alongside the existing stdio transport.
- The server validates the inbound bearer on every request (signature via JWKS, issuer, expiry, and audience) and never resolves a global service identity.
- The API client is shared, but its auth is request-scoped: a callable resolver pulls the current request's validated token from the MCP auth context, so concurrent callers act as their own users.

## Required changes

1. HTTP transport profile. Add Streamable HTTP (bind config, long-lived process). Register tools once on the shared app, not in a per-session lifespan, since the streamable-http lifespan runs per session and re-registering collides.
2. Resource-server role. Serve OAuth protected-resource metadata (RFC 9728), return `401` with a `WWW-Authenticate` challenge on missing or invalid tokens, and validate the bearer (including audience) per request.
3. Request-scoped identity. No global identity at startup. Resolve the caller's token per request via the auth context and a callable resolver on the shared client. The prototype demonstrates this; it needs upstreaming.
[...]

PR #401, opened 2026-07-14
docs(mcp): audit process-global config for the hosted profile (#306)

### LINKED ISSUE #306: Process-global config audit and documentation

## Scope

Audit tools that read process-global configuration for per-user decisions (for example service-account guards) to confirm they are safe for a single-integrator, single-backend hosted deployment. Document the assumption and the rework trigger (one host serving multiple backends with different config).

## Acceptance criteria

The assumption is documented; no tool reads mutable global state for a per-user decision outside the documented assumption; affected tools are marked remote-safe only after review.

## Dependencies

Informs which tools the default-deny allowlist (#304) can mark.

mocha06:
## Motivation
Part of the hosted on-behalf-of MCP server proposal (#297). Under the hosted `remote` profile one process serves many concurrent users, and settings loaded from the environment are shared by all of them. That is safe only while a shared value never decides anything per-user — an assumption the tool surface relied on but never stated, with nothing preventing a future tool from reading global settings for a per-user decision.

## Outcome
- `packages/mcp/AGENTS.md` gains a section documenting the single-integrator, single-backend assumption, the audited inventory of every shared-settings read a tool call can reach, why the Internal-API tools stay unmarked, and the rework trigger (one host serving multiple backends means those reads must become request-scoped, as identity already did in #302).
- A new import-linter `forbidden` contract makes the line mechanical: `pipefy_mcp.tools` cannot *import* `pipefy_mcp.settings`, directly **or transitively**, outside a reviewed, commented exception list. A new settings import in tool-reachable code now fails `uv run lint-imports` and must be reviewed onto `ignore_imports` before it can land.
- No `meta=REMOTE` changes: the existing 19-tool remote seed passes the audit.

## Audit findings
Identity — the one truly per-user input — never comes from settings: in remote mode each caller is resolved from the request bearer (`RequestScopedIdentity`, #302). `settings.auth` is read at startup in both modes, but always for a per-deployment purpose (the outbound credential in local; the default inbound issuer URL in remote, safe under single-realm). Every shared-settings read a tool call can reach:

| Setting | Read from | Verdict |
| --- | --- | --- |
| `PIPEFY_BASE_URL` | engine endpoints, resolved at build | Per-deployment — this read *is* the single-backend assumption |
| `gql_reuse_fetched_graphql_schema` | process-scoped schema cache | Safe only because one backend ⇒ one schema (documented) |
| `permission_denied_enrichment_timeout_seconds` | `tools/graphql_error_helpers.py`, call time | Tuning knob, same for everyone |
| `unified_envelope` | `core/tool_error_envelope.py`, call time | Response-shape flag, applied identically to every caller — **found by the new contract**, missed by the manual pass |
| `default_webhook_name` | SDK webhook service | Cosmetic default; webhook tools are not remote-exposed |

Adjacent note: some tool docstrings (e.g. `delete_card_relation`'s) claimed to "require service-account credentials" for the Internal API. Review flagged this as stale, and the SDK confirms it: `_bind` hands the one session credential to all three endpoints, so nothing is service-account-specific. Those tools stay off the remote seed for the ordinary reason — unreviewed write mutations — and the docstring is fixed here (f1776fb).

## Verification
- `uv run lint-imports`: both contracts **KEPT** (layers + the new one).
- Negative probe: adding `from pipefy_mcp.settings import settings` to a tool module breaks the contract (`pipefy_mcp.tools.member_tools -> pipefy_mcp.settings`); clean again after revert.
- `uv run pytest tests/tools/test_remote_profile.py`: 7 passed — the `REMOTE_SEED` drift guard confirms the exposed set is unchanged.

Known limit (documented, not a gap this PR closes): the contract catches settings *imports*, not value *reads*. Code holding a settings object — `McpRuntime.settings`, reachable from any tool via its request context — can read a per-deployment value without an import edge, so that path stays a manual-review item.

Mapping to #306's acceptance criteria: the assumption is documented (AGENTS.md section); no tool reads mutable global state for a per-user decision outside it (audit above, now import-guarded); affected tools are marked remote-safe only after review (seed untouched, and the contract + drift guard force review on every future change).

Closes #306.


packages/mcp/AGENTS.md
```diff
@@ -181,6 +181,61 @@ than a local `file_path`), enforce that in the tool body gated on
...
+  across users, which is fine because one backend means one schema.
+- `permission_denied_enrichment_timeout_seconds`
+  (`tools/graphql_error_helpers.py`) — a timeout knob. Same for everyone by
+  design.
+- `unified_envelope` (`core/tool_error_envelope.py`) — a flag that changes the
+  shape of tool responses. Applies identically to every caller.
+- `default_webhook_name` (SDK webhook service) — a cosmetic default name; the
+  webhook tools aren't exposed remotely anyway.
+
+A related note: some tools (like `delete_card_relation`) say they "require
```
gbrlcustodio: **issue (blocking):** this note rests on a premise the SDK contradicts.

The internal executor is bound to the same session credential as the public one: `_bind` (`packages/sdk/src/pipefy_sdk/client.py:146-156`) passes the single resolved `auth` to all three executors (public, interfaces, internal). The Internal API endpoint is a different URL, not a different credential, so whatever token the caller resolves (static token, service account, or stored session) is what reaches it, and a caller's own bearer would too.

That makes two statements here inaccurate:
- "require service-account credentials": stale. Every endpoint works with any supported credential; nothing about the Internal API is service-account-only. The `delete_card_relation` docstring (`packages/mcp/src/pipefy_mcp/tools/pipe_tools.py:525,531-533`) carries the same stale claim and is its source.
- "we haven't verified the Internal API accepts one": the client sends the same credential to all three endpoints unconditionally, so there is no separate acceptance to verify at this layer.

If these tools stay off the remote seed, state the real reason (unreviewed write mutations, confirm-gated flow) rather than an Internal-API credential caveat that does not hold. Suggest dropping this paragraph or rewriting it to the actual exclusion reason, and fixing the `pipe_tools.py` docstring in a follow-up.

mocha06: Good catch — verified against the SDK: `_bind` (`client.py`) hands the one session credential to all three executors, so nothing at the client layer is service-account-specific, and the "acceptance to verify" hedge had nothing behind it. Rewrote the note in fe5953c to state the actual exclusion reason: these are write mutations that haven't been reviewed for remote exposure — no credential caveat. Also dropped the stale claim from the `delete_card_relation` docstring here rather than deferring it (f1776fb), since it was the source of the confusion and a two-line fix.

packages/mcp/AGENTS.md
```diff
@@ -181,6 +181,63 @@ than a local `file_path`), enforce that in the tool body gated on
...
+there is one copy for the whole process, not one per user.
+
+That's fine when a setting answers a question about the *deployment* ("which
+Pipefy backend do we talk to?", "how long before a lookup times out?"). It's a
+problem if a setting ever answers a question about a *user* ("who is this call
+acting as?", "which org does this caller belong to?") — because then one shared
+value would silently apply to every user. The one truly per-user thing,
+identity, never comes from settings: in hosted mode each request carries its own
+validated token (`RequestScopedIdentity`), and the startup credential settings
+(`settings.auth`) are only ever read in local mode.
```
gbrlcustodio: issue (non-blocking): this contradicts both the code and the Transport profiles section of this same file.

`McpRuntime.for_profile`'s `remote` branch reads `settings.auth` at startup: `runtime.py:159` passes `default_issuer_url=_login_issuer_url(settings)`, and `_login_issuer_url` (`runtime.py:101`) calls `settings.auth.to_oidc_client()` to derive the default inbound issuer URL.

The Resource-server profile section above states the same behavior (AGENTS.md:55-57): absent a `PIPEFY_JWT_ISSUER_URL` override, the inbound issuer defaults to "the one this process logs into, the `OidcClient` issuer". So the document both relies on the remote-mode `settings.auth` read and denies it here.

The read is per-deployment-safe under the single-realm assumption, so this is a wording fix, not a behavior change. Suggest narrowing the claim: the outbound startup credential is resolved only in local mode, while in remote mode `settings.auth` is still read once at startup for the default inbound issuer, safe under single-realm. That frames it as a sibling of the single-backend read this section is built around.

gbrlcustodio (commented):
One inline note on the process-global configuration section.

packages/mcp/AGENTS.md
```diff
@@ -181,6 +181,63 @@ than a local `file_path`), enforce that in the tool body gated on
...
+is gone: "same for everyone" stops being true, and each of the reads above
+becomes a per-user question. The fix is the same one identity already went
+through in #302 — stop reading the value from shared settings and resolve it
+from the incoming request instead.
+
+**How this is enforced.** An import-linter rule in `pyproject.toml` forbids
+tool code from reaching `pipefy_mcp.settings`, even indirectly through a helper
+(that indirect path is how `unified_envelope` was caught). Every existing read
+is an explicitly listed, commented exception. If someone adds a new settings
+read to a tool, `uv run lint-imports` fails, and the read only gets an
```
gbrlcustodio: issue (non-blocking): this overstates the guard. The contract catches settings imports, not settings reads.

The contract is `forbidden` from `pipefy_mcp.tools` to `pipefy_mcp.settings`, so it detects import edges into the settings module. It cannot see a value read off an object that already holds `Settings`, and that object is in every tool's hand: `McpRuntime.settings` (`runtime.py:181-184`) exposes the whole settings tree, and `tool_context.py:24` binds the runtime straight off the lifespan context.

So a tool that reads `ctx.request_context.lifespan_context.settings.mcp.<anything>` at call time reads a process-global setting with no import of `pipefy_mcp.settings`, and `lint-imports` stays green. The escape hatch is sanctioned rather than exotic: `tool_context` imports `McpRuntime`, and `pipefy_mcp.core.runtime -> pipefy_mcp.settings` is one of the four reviewed `ignore_imports` entries.

Suggest scoping the claim to imports and naming the residual surface: a new settings import in tool-reachable code fails `lint-imports`, but reading a value off `McpRuntime.settings` at call time is the one path the contract cannot see and stays a manual-review item. A stronger fix (code, beyond this PR) is to stop exposing `McpRuntime.settings` as a broad public property, since its only consumer is `server.py:53` reading `unified_envelope`.

gbrlcustodio (commented):
One more inline note, on the enforcement claim.

packages/mcp/AGENTS.md
mocha06: Right on both counts — `for_profile`'s remote branch does read `settings.auth` (via `_login_issuer_url` -> `to_oidc_client`) for the issuer default, and the Resource-server section above already says so. Narrowed the claim in 4c02a26: `settings.auth` is read in both modes but always for a per-deployment purpose — the outbound startup credential in local, the default inbound issuer in remote — safe under the same single-realm assumption this section is built around.

packages/mcp/AGENTS.md
mocha06: Agreed — the contract sees import edges, not value reads, and `McpRuntime.settings` is a sanctioned path around it (`tool_context` -> `McpRuntime` is one of the four reviewed `ignore_imports`). Scoped the doc's claim to imports in 4c02a26 and named the residual surface: reading a value off `McpRuntime.settings` at call time is the one path the contract can't see and stays a manual-review item.

On the stronger fix: narrowing `McpRuntime.settings` (its only consumer is `server.py` reading `unified_envelope`) is code beyond this docs PR — filing it as a follow-up rather than widening scope here. (https://github.com/pipefy/ai-toolkit/issues/405)

adriannoes (dismissed):
## Summary

Closes the #306 process-global audit for the hosted profile: the single-integrator, single-backend assumption is documented, and an import-linter forbidden contract keeps `pipefy_mcp.tools` from reaching `pipefy_mcp.settings` outside a reviewed ignore list. I reviewed on checkout (`b917149c`) with `lint-imports` (both contracts KEPT), a negative probe, and `test_remote_profile` (7 passed); CI lint and test are green. Ready to merge as-is; two small AGENTS.md polish items below are optional.

## What worked well

- The single-backend assumption is stated as an explicit invariant with a clear rework trigger (same shape as the #302 identity move).
- The import-linter guard is mechanical and caught `unified_envelope`, which the manual pass missed; the known limit (imports vs value reads via `McpRuntime.settings`) is called out honestly.
- The stale service-account claim is gone from `delete_card_relation`, and the remote seed is left untouched with the drift guard still green.

## Also noted

- In the new process-global section, the note still says some tool docstrings (e.g. `delete_card_relation`) *claim* to require service-account credentials, but this PR already rewrote that docstring and there are no remaining MCP tool SA-credential claims. Prefer past tense ("previously claimed") or drop the example so the note matches the tree.
- The inventory is labeled a "full list" of shared-settings reads a tool call can hit. It includes `default_webhook_name` from the SDK webhook service but omits the sibling `allow_insecure_urls` read on create/update in the same service. Same remote-exposure caveat applies; one bullet (or softening "full list") would keep the #306 audit list complete.

adriannoes (approved):
## Summary

Follow-up APPROVE on `c74538f`. The AGENTS.md polish landed (Internal API note + `allow_insecure_urls` in the inventory), and CI lint/test are green on this head. Still good to merge for #306.

PR #402, opened 2026-07-14
Add get_ipaas_tools meta tool for iPaaS (Advanced Automations) discovery

mocha06:
## Motivation
Agents need to work with iPaaS (Advanced Automations) flows, but the iPaaS tool
surface is large — dozens of tools, some with very large input schemas. Loading it
into every agent session would crowd out context, so iPaaS support starts with lazy
discovery: expose the catalog on demand, and one tool's schema only right before use.

## Outcome

image

- New `get_ipaas_tools(pipe_id)` MCP tool: a compact `name` + one-line `description`
  catalog of the tools available in the pipe's iPaaS workspace. Passing
  `tool_name=...` expands a single tool's full description and input schema
  (roughly an order of magnitude smaller than fetching every schema).
- The server reaches iPaaS through a fully headless auth chain bound to the
  caller's own identity, stateless per call — no cached tokens or sessions, so any
  replica serves any request.
- Zero configuration: the OAuth client defaults to Pipefy's canonical public PKCE
  client (a publishable identifier, like the `pipefy-cli` OIDC client — the
  caller's pipe-scoped session carries all authorization). `PIPEFY_IPAAS_*`
  settings override the host and client for staging/single-tenant deployments;
  blanking the client id disables the tools with a clear error.
- Requires automation-create permission on the pipe and iPaaS enabled on the
  organization; each pipe maps to its own iPaaS workspace.
- SDK gains `PipefyClient.get_advanced_automations_token`; `docs/ipaas.md` covers
  the meta-tool pattern and flow; the parity matrix lists the tool with the CLI
  twin deferred to the invocation phase.

PR #403, opened 2026-07-15
Invoke iPaaS (Advanced Automations) tools via call_ipaas_tool

mocha06:
## Motivation
#402 gave agents discovery over a pipe's iPaaS (Advanced Automations) tool catalog, but discovery without invocation still sends the user to the product UI for every action. The natural next step of the meta-tool pattern is letting agents call what they discovered — without loading the full catalog into context, and without adding any per-deployment state or configuration.

## Outcome
Agents can now execute any tool from a pipe's iPaaS workspace with `call_ipaas_tool(pipe_id, tool_name, arguments)`, completing the discover → expand → call loop (`get_ipaas_tools` → drill into one schema → invoke). Arguments are forwarded verbatim and validated by the iPaaS host, whose error messages come back unchanged; results are relayed in full, with host-side tool errors mapped onto the standard error envelope. Every invocation runs the same stateless per-call chain as discovery — any replica serves any call — and the invocation hop gets a 120s budget (handshake hops keep 30s) since a called tool may execute a real flow; on timeout the error points agents at the catalog's run-listing tools instead of retrying. The full catalog is exposed, including destructive operations: the caller already holds the identical surface in the product's Advanced Automations UI, so the tool is annotated destructive (client approval flows apply) rather than server-filtered. The registry grows to 155 tools.

## Notes
Stacked on #402 (`rc-dev/feat/ipaas-get-tools`) — merge that first; this PR then rebases onto `dev`. Validated end-to-end against a production pipe: flow listed, created, inspected, and deleted through the new tool.

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
@@ -35,7 +35,7 @@ class IpaasTools:
         # Not marked remote-safe yet: the tool itself is per-user clean (the
         # pipe token is minted with the caller's own session), but the iPaaS
         # chain hasn't been reviewed for hosted exposure.
-        @mcp.tool(annotations=ToolAnnotations(readOnlyHint=True))
+        @mcp.tool(annotations=ToolAnnotations(readOnlyHint=True, openWorldHint=True))
         async def get_ipaas_tools(
             pipe_id: PipefyId,
             ctx: Context,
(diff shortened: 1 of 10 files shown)


packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
```diff
@@ -93,6 +93,84 @@ async def get_ipaas_tools(
...
+                return build_error_payload(_NOT_CONFIGURED_MESSAGE)
+
+            client = get_pipefy_client(ctx)
+            try:
+                token = await client.get_advanced_automations_token(pipe_id)
+                result = await gateway.call_tool(token, tool_name, arguments)
+            except Exception as exc:  # noqa: BLE001
+                return build_error_payload(str(exc))
+
+            return _call_result_payload(pipe_id, tool_name, result)
```
adriannoes, line 145: When the iPaaS host returns JSON-RPC success with `"result": null`, `"result" in payload` is still true, so the gateway treats the call as OK and hands `None` into `_call_result_payload`. That helper does `result.get(...)` outside the tool's try/except that maps `IpaasGatewayError`, so FastMCP blows up with `Error executing tool call_ipaas_tool: 'NoneType' object has no attribute 'get'` instead of the standard error envelope.

Treat a missing or null `result` as a gateway/protocol error (same family as other malformed host bodies), and keep `_call_result_payload` behind that guard (or harden it to reject `None` with `build_error_payload`).

Done when:

- a unit or gateway test with HTTP 200 and `"result": null` returns `success: false` / envelope text (no `AttributeError`)
- the existing happy-path call tests still pass

adriannoes (changes requested):
## Summary

Solid invocation half of the iPaaS meta-tool loop: shared auth seam, 120s only on the call hop, host `isError` mapped to the standard envelope, registry/parity at 155, and intentional full-catalog exposure via `destructiveHint`. Review used checkout+tests (29 ipaas gateway/tool tests green). One IMPORTANT contract hole on the new invoke path should be fixed before merge.

## What worked well

- Completes discover → expand → call without per-deployment state.
- Thin shell over `PipefyClient` token mint + `IpaasGateway.call_tool`; host stays argument authority.
- Timeout split (30s handshake / 120s invoke) is covered by test.
- Host `isError` and unconfigured/gateway errors land in the usual envelope on the happy unhappy paths.

## Also noted

- The shared `_mcp_request` timeout path embeds a U+2014 em dash in the `IpaasGatewayError` string that becomes `error.message`. N1 asks for ASCII-only agent-facing messages, but the same pattern already ships elsewhere on `main` with green tests, so this is consistency cleanup rather than a merge blocker. Swap to `-` or `;` when convenient.
- Docs describe discover → expand → call, but the compact `get_ipaas_tools` success `hint` still only tells agents to expand with `tool_name`. Worth naming `call_ipaas_tool(pipe_id, tool_name=..., arguments={...})` after expand so the live payload matches the docs.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
mocha06: Thanks for the review — pushed fixes in 56a0409 (on top of the merge with the base branch in c32d2e4).

**Null/`result: null` (IMPORTANT)** — fixed at the gateway, treating a missing *or* null/non-object `result` as a method-named protocol error (same family as the other malformed-body checks). That keeps `_call_result_payload` behind the guard, so it only ever sees a real object and can no longer dereference `None`. Covered two ways: a gateway test (`200` + `"result": null` → `IpaasGatewayError`) and a tool test (same scenario → `success: false` envelope, no `AttributeError`).

**Discovery hint** — the compact `get_ipaas_tools` hint now names `call_ipaas_tool(pipe_id, tool_name=..., arguments={...})` after the expand step, so the live payload matches the documented discover → expand → call loop. Asserted in the catalog test.

**Em dash / ASCII-only** — left as-is intentionally. As noted, the pattern already ships on `main`, and em dashes are pervasive in this module's own docstrings, so unless we adopt an actual ASCII-only rule for agent-facing strings I'd rather not diverge on one message. Happy to sweep it if you'd prefer.

33 iPaaS tests green; ruff, format, and import-linter clean.

adriannoes (approved):
## Summary

Rechecked after 56a0409: null/non-object `result` is rejected at the gateway into the standard envelope, the catalog hint names `call_ipaas_tool`, and the new tests cover both. Thanks for addressing the review points so cleanly.

LGTM.

PR #404, opened 2026-07-15
Connect apps to iPaaS workspaces headlessly from the conversation

mocha06:
## Motivation
#403 made iPaaS tools invocable, but any flow step that touches an external app needs a connection — and creating one meant leaving the conversation for the product UI, exactly the interruption the meta-tool surface exists to remove. Connections also carry credentials, so the design had to be explicit about what transits the conversation (and therefore the model vendor) and what never does.

## Outcome
Two tools complete the loop. `create_ipaas_connection` creates (or, on an existing `external_id`, rotates) a connection: token/API-key credentials in one fully headless call, with secrets passable either as literals (the trade-off is documented where agents will read it) or as `{"$env": "PIPEFY_IPAAS_CONNECTION_<NAME>"}` references resolved from the MCP server's own environment — the value never enters the conversation, and the mandatory prefix stops the tool from being steered into shipping unrelated process secrets. For OAuth apps, `get_ipaas_connection_auth_url` builds the third-party consent URL (auto-detecting the deployment's configured OAuth client per piece) plus a stateless completion bundle; the user clicks once, pastes back the redirect URL, and `create_ipaas_connection` finishes the exchange — durable tokens are stored host-side and no lasting secret ever appears in chat. Connection endpoints authenticate with the session from the existing chain's first step, keeping every call stateless across replicas. Credentials are validated by the iPaaS host at creation, create responses relay only non-sensitive fields, and timeout errors now distinguish "handshake failed, retry" from "your tool may still be running". The registry grows to 157 tools.

image

image

image

image

## Notes
Stacked on #403 — merge order #402 → #403 → this. Validated end-to-end in production: an OAuth connection created via the consent-URL → paste-back flow, then used to wire, validate, test-fire (message delivered), and publish a real flow.


adriannoes (dismissed):
## Summary

Reviewed with a local checkout of `537b5cd4` (ruff clean; 51 unit tests on the iPaaS gateway and tools green). The connection tools close the loop cleanly: token/`$env` create, OAuth consent URL plus stateless completion bundle, non-sensitive create responses, and a clearer handshake timeout. No blockers from this pass.

## What worked well

- Secret handling is explicit at the tool boundary: literals transit the conversation by choice; `{"$env": "PIPEFY_IPAAS_CONNECTION_*"}` stays out of model context, with a mandatory prefix.
- OAuth stays two-step and server-stateless; durable tokens stay host-side; create responses filter to non-sensitive fields (covered by the `must-never-leak` test).
- Upsert on `external_id` makes rotation intentional. Registry and `docs/parity.md` both land at 157 with CLI deferred, consistent with the other iPaaS meta-tools.

## Also noted

- README correctly claims 157 tools across eleven domains, but the domain table still has ten rows that sum to 153 and omits iPaaS (now four tools). That same README domain-table sync is already required on [#402](https://github.com/pipefy/ai-toolkit/pull/402) (F1 in my review there). Please fix it once on the tip of the stack (#402 / wherever the README lands after merge). No need to duplicate the edit in this PR if #402 already carries it.

adriannoes (approved):
## Summary

Re-checked after the follow-up. `9cfeb322` adds the iPaaS domain row at **4** tools so the eleven-domain README table sums to **157**, matching the headline and closing the F1 note from the earlier review (and the same sync we called out on #402). Previous approval was dismissed by the rebase onto `dev`; this pass confirms the docs contract is aligned. Approving again.

## What worked well

- You addressed the README sync promptly and in the right place on this tip of the stack, with a clear commit message that states why the row moved from 1 to 4.
- The connection work from the original review still stands: `$env` prefixing, OAuth two-step completion, and non-sensitive create responses.

## Verification

Domain table tool counts now: `40+17+8+17+23+4+11+11+1+20+5 = 157`.

PR #406, opened 2026-07-15
feat(mcp): mark the iPaaS tools remote-safe

mocha06:
## Motivation

The four iPaaS meta tools shipped local-only: the hosted `remote` profile is default-deny, and the iPaaS chain hadn't been reviewed for hosted exposure. That review is this PR. It found the chain already per-user clean — every call mints the pipe-scoped token with the caller's own request-scoped bearer, and the gateway is stateless deployment config — with one real blocker: `{"$env": ...}` credential references resolve from the server process's environment, which on a multi-user deployment is shared, so any caller could resolve any configured variable.

## Outcome

- All four iPaaS tools (`get_ipaas_tools`, `call_ipaas_tool`, `get_ipaas_connection_auth_url`, `create_ipaas_connection`) carry `meta=REMOTE` and register on the remote profile; `REMOTE_SEED` grows 19 → 23.
- `create_ipaas_connection` rejects `$env` references at call time under the remote profile with an error that points to the two modes that remain (literal value, OAuth flow). The gate reads the profile off the runtime via a new `is_remote_profile(ctx)` accessor — the first shipped instance of the "exposure vs input restriction" pattern, which AGENTS.md now documents as the canonical shape.
- Live-validated over Streamable HTTP against a `--profile remote` server with a real inbound bearer: exactly 23 tools exposed (unmarked tools withheld), all four iPaaS tools working end-to-end through the caller-identity token mint, and the `$env` rejection confirmed.

<details>
<summary>Local validation evidence</summary>

Server launched from this branch with `--profile remote --transport http` on loopback, resource-server settings pointed at the production IdP; calls made over Streamable HTTP presenting a real user bearer, JWKS-validated per request (the same identity mechanics a hosted deployment runs).

Unauthenticated probe:

```
GET /mcp (no bearer) -> 401
```

Server startup log:

```
INFO  Remote profile: exposed 23, withheld 134 Pipefy tools.
```

Validation client (all four tools, live against the production iPaaS host):

```
[0] initialize OK (bearer accepted)
[1] tools/list: 23 exposed; ipaas: ['call_ipaas_tool', 'create_ipaas_connection',
    'get_ipaas_connection_auth_url', 'get_ipaas_tools']
[2] get_ipaas_tools: success, catalog count=40
[3] call_ipaas_tool(ap_list_flows): success
[4] get_ipaas_connection_auth_url: success (authorization_url + completion bundle, Slack piece)
[5] create_ipaas_connection($env ref): rejected on hosted profile, nothing resolved
ALL CHECKS PASSED
```

Offline gates: full `packages/mcp` suite 1568 passed / 9 skipped, import-linter contract kept, ruff clean.

</details>

## Notes

For operators fronting the remote profile: `call_ipaas_tool` can hold its response up to 120 s for long-running invocations, so proxy/LB read timeouts must tolerate that, and the deployment needs egress to the iPaaS host.


adriannoes (dismissed):
## Summary

Reviewed with a local checkout of this head (`87246524`) plus the `packages/mcp` suite (1568 passed). The PR does what it claims: the four iPaaS tools carry `meta=REMOTE`, `REMOTE_SEED` moves 19 to 23, and `create_ipaas_connection` rejects `$env` under the remote profile via `is_remote_profile(ctx)` reading the runtime settings rather than the module-global singleton. I am approving.

## What worked well

The shared-env hazard is gated at call time with a clear error that points to literal value or OAuth, and the dedicated remote reject test asserts no secret echo and no `upsert_connection`. AGENTS.md documents the exposure vs input-restriction pattern at the point future tools will copy.

## Also noted

- `get_ipaas_connection_auth_url` still advertises `readOnlyHint=True` while POSTing for an authorization URL and returning a single-use PKCE `completion` (including `code_verifier`). That annotation predates this PR; with `meta=REMOTE` it is worth flipping to `readOnlyHint=False` (or documenting why read-only stays) so hosted clients do not treat step 1 like a cacheable pure read.
- Stack note: this tip sits on `#404` (`rc-dev/feat/ipaas-connections`); landing on `dev` still needs `#402` then `#403` then `#404` first.

mocha06:
> ## Summary
> Reviewed with a local checkout of this head (`87246524`) plus the `packages/mcp` suite (1568 passed). The PR does what it claims: the four iPaaS tools carry `meta=REMOTE`, `REMOTE_SEED` moves 19 to 23, and `create_ipaas_connection` rejects `$env` under the remote profile via `is_remote_profile(ctx)` reading the runtime settings rather than the module-global singleton. I am approving.
> 
> ## What worked well
> The shared-env hazard is gated at call time with a clear error that points to literal value or OAuth, and the dedicated remote reject test asserts no secret echo and no `upsert_connection`. AGENTS.md documents the exposure vs input-restriction pattern at the point future tools will copy.
> 
> ## Also noted
> * `get_ipaas_connection_auth_url` still advertises `readOnlyHint=True` while POSTing for an authorization URL and returning a single-use PKCE `completion` (including `code_verifier`). That annotation predates this PR; with `meta=REMOTE` it is worth flipping to `readOnlyHint=False` (or documenting why read-only stays) so hosted clients do not treat step 1 like a cacheable pure read.
> * Stack note: this tip sits on `#404` (`rc-dev/feat/ipaas-connections`); landing on `dev` still needs `#402` then `#403` then `#404` first.

Thanks for the review — addressed the `readOnlyHint` note in bcf0299.

`get_ipaas_connection_auth_url` now advertises `readOnlyHint=False`. Agreed it's misleading under `meta=REMOTE`: the call POSTs for an authorization URL and returns a fresh, single-use PKCE `completion` (`code_verifier`) on every invocation, so a hosted client treating it as a cacheable pure read could replay a stale verifier and break the consent flow. Left the flag explicit with a comment at the call site rather than relying on the default, so the intent reads there.

Pinned it with a test (`get_ipaas_connection_auth_url` → `readOnlyHint=False`, `get_ipaas_tools` stays `True`) so the distinction can't silently regress. 38 iPaaS + remote-profile tests green; ruff/format clean.

Worth noting the strict "modifies its environment" framing is a touch generous here — the two-step flow keeps no server-side state (the bundle is returned to the caller, nothing persisted). The real reason it isn't read-only is that it's non-idempotent and emits a single-use secret, which is exactly the "don't treat step 1 as a cacheable read" concern.

adriannoes (approved):
## Summary

Re-reviewed after `bcf0299`. The follow-up from the earlier note is done: `get_ipaas_connection_auth_url` now uses `readOnlyHint=False`, with a call-site comment and a pin test that keeps `get_ipaas_tools` read-only. CI is green on this head. Approving again.

## What worked well

The fix matches the real concern (non-idempotent step 1 that returns a fresh PKCE `code_verifier`), and the test locks the distinction so it cannot quietly regress back to `readOnlyHint=True` under `meta=REMOTE`.

## Also noted

- Optional docs polish: `docs/mcp/tools/ipaas.md` still lists `get_ipaas_connection_auth_url` as Read-only Yes in the overview table. The wire annotation is already correct; flipping that cell to No would keep the human doc in sync if you touch the file anyway.

PR #417, opened 2026-07-17
Parse AI behavior actionParams into typed models

### LINKED ISSUE #416: 0. Prep: parse behavior actionParams into typed models
Preparation PR for the milestone — no behavior change, no new tools.

## Problem

`BehaviorInput.action_params` is a raw `dict | None`. Every consumer re-walks it with dual-alias branches (`aiBehaviorParams` / `ai_behavior_params`, `actionsAttributes` / `actions_attributes`): the model validator, the SDK preflight, and MCP helpers such as the pipe-ID extraction in `ai_agent_tools.py`. This contradicts the repo's "parse, don't validate" / self-guaranteeing-types rules (AGENTS.md, docs/architecture.md) and was only tolerable while the dict was pure pass-through.

#410 adds capability shape validation, provider-ID rules, and round-trip selections into that same structure, and #412 walks it again for the `dataSourceIds` union — so the dict-walking multiplies unless it's fixed first.

## Scope

- Introduce nested Pydantic models for the behavior payload (`AiBehaviorParams`, action attributes, capability attributes) with `populate_by_name` + per-field camelCase aliases matching the declared `AiBehaviorParamsInput` / `AiBehaviorActionAttributesInput` schema names, replacing the manual dict-walking in `BehaviorInput`'s validator.
- Reach: both consumer categories. (1) Read/validate consumers — the model validator, SDK preflight/pipe-validation/phase-transition walks, and MCP helpers such as the pipe-ID extraction. (2) The pre-validation transform passes (placeholder/instruction normalization, slug resolution), which normalize inner keys through the typed model at the entry edge instead of preserving the caller's key style.
- Serialization contract: byte-identical output for every payload that was previously **valid end-to-end**. The GraphQL input types declare exactly one wire spelling per field — the AI-behavior subtree is uniformly camelCase — so snake_case inner keys were always rejected by the API even though the toolkit's dual-alias branches accepted them. Normalizing them to the declared names makes that input functional for the first time; it is not a behavior break. The few unit tests pinning snake_case-inner passthrough get updated accordingly.
- Aliasing is strictly per-declared-field, never a blanket snake→camel transform: sibling fields of `aiBehaviorParams` in `AutomationActionParamsInput` (e.g. `card_id`, `to_phase_id`, `field_map`) have snake_case wire names, so generic case conversion would corrupt them. Unknown/extra keys pass through verbatim (`extra="allow"`), and action `metadata` stays a passthrough dict in v1 so GET→update round-trips serialize identically.
- Helpers that run inside error handlers (e.g. the behavior summarizer) keep a lenient parse path — they must never raise on malformed payloads.
- Migrate the SDK preflight and MCP helper dict-walking to the typed models; delete the alias-juggling helpers that become dead.
- No new validation rules in this PR — capability/provider contracts land in #410 on top of the typed base.

## Acceptance

- Existing SDK/MCP/CLI test suites pass unchanged, except tests that pin snake_case inner keys surviving serialization — those assert output the API rejects and are updated to the declared camelCase wire names.
- No remaining `.get("aiBehaviorParams") or .get("ai_behavior_params")`-style branches in SDK or MCP agent code paths.
- `uv run pytest -m "not integration"` + ruff + import-linter green.

Blocks #410.

mocha06:
## Motivation
`BehaviorInput.action_params` was a raw `dict | None`, and every consumer re-walked it with dual-alias branches (`aiBehaviorParams`/`ai_behavior_params`, `actionsAttributes`/`actions_attributes`): the model validator, the SDK preflight and pipe/phase-transition checks, and MCP helpers such as the pipe-ID extraction. That duplicated dict-walking contradicts the repo's "parse, don't validate" / self-guaranteeing-types rules and multiplies as more fields are added to the payload.

## Outcome
`action_params` is now a typed Pydantic model (`AiBehaviorActionParams` → `AiBehaviorParams` → action and capability attributes) with per-declared-field camelCase aliases matching the GraphQL input schema and `extra="allow"` so unknown/sibling keys pass through verbatim. Every consumer parses once and reads typed attributes, so no dual-alias branches remain in SDK or MCP agent code paths.

Behavior is unchanged for every payload that was already valid end-to-end. snake_case inner keys — which the API always rejected while the toolkit's dual-alias branches accepted them — are now normalized to the declared wire names, so that input works for the first time. Aliasing is strictly per declared field (never a blanket snake→camel pass, which would corrupt genuinely snake_case siblings like `card_id`/`to_phase_id`), and action `metadata` stays a pass-through dict so GET→update round-trips serialize identically.

Tested e2e a few times

image

No new validation rules land here — capability and provider hardening build on this typed base.

Closes #416.


adriannoes (approved):
## Summary

Solid prep PR for typed behavior `actionParams`. 

Nested models with per-field aliases replace the dual-alias dict walks without changing valid end-to-end wire payloads, and capability/provider hardening correctly stays out of scope for #424. Review ran checkout plus local tests (2922 passed, ruff green); CI lint and test are green.

## What worked well

Per-field camelCase aliases with `extra="allow"` keep AutomationActionParams siblings (`card_id`, `to_phase_id`) intact while normalizing declared AI-behavior keys. Lenient `BehaviorPayload` reads versus strict `BehaviorInput` writes is a clean split, and the slug-resolve path deep-copies before parse so caller `metadata` is not mutated through pydantic references.

## Also noted

- The public `resolve_and_populate_field_refs` helper early-returns without rewriting keys when no slug fetch runs, while `populate_referenced_field_ids` now only reads declared camelCase names. In-repo MCP/CLI/facade paths already `model_dump(by_alias=True)` first, so they stay correct. If that helper remains a supported entry point for raw snake_case dicts, normalizing on every exit (or documenting dump-first) would match #416's "normalize at the entry edge" wording.

Issue #415, opened 2026-07-16
6. LLM provider management: write tools (BYOM)


mocha06:
Final phase of the milestone. Depends on #411 (read/probe surface).

## Tools

| MCP | CLI |
|-----|-----|
| `create_llm_provider` | `pipefy ai-provider create` |
| `update_llm_provider` | `pipefy ai-provider update` |
| `delete_llm_provider` | `pipefy ai-provider delete` |
| `set_llm_provider_active_status` | `pipefy ai-provider set-active-status` |
| `set_default_llm_provider` | `pipefy ai-provider default set` |
| `reset_default_llm_provider` | `pipefy ai-provider default reset` |

## Scope

- Secret handling: MCP and CLI accept provider configuration only via a local `configuration_file_path` (JSON), never inline arguments; configuration values are never logged, echoed in errors, or returned. Examples use placeholders only. Create/update tools stay excluded from the remote profile.
- Update sends a full replacement configuration (API contract); create requires `name`.
- Verify and document the recommended update flow against a lab organization before publishing docs: whether a fetched (redacted) configuration can be edited and sent back without re-supplying secrets.
- Default set/reset are organization-scoped; setting the default requires exactly one of `provider_id` / `system_provider_id`.
- Docs note that provider management requires organization admin permissions and an eligible plan. Deletes require confirmation; CLI writes gated on the access probe.
- Probe gating contract (raised in the #411 review): `validate_llm_provider_access` can return `ok: true` with a non-null `problem` — when the API returns partial data alongside GraphQL errors, the probe surfaces the classified error rather than discarding it, and deliberately does not flip `ok`. Write-gating must therefore treat a clean gate as `ok: true` **and** `problem` absent; a present `problem` is partial denial and must not be read as full access. As part of this phase, polish the #411 probe docs/messages (`docs/mcp/tools/llm-providers.md`, probe docstrings, CLI `validate-access`) to state this contract explicitly.

## Acceptance

- Full SDK + MCP + CLI parity, `docs/parity.md`, skills updated end-to-end (validate → create/list provider → set behavior provider ID).
- Tests assert no configuration content in logs, errors, or tool responses.
- Write-gating treats a probe with a non-null `problem` as not-clean (partial denial is never read as full access), and the #411 probe docs state the clean-gate contract explicitly.

PR #426, opened 2026-07-17
LLM provider discovery: read tools and access probe

### LINKED ISSUE #411: 2. LLM provider discovery: read tools and access probe
Second phase of the milestone. Depends on #410 (closes its provider-ID discovery gap).

## Tools

| MCP | CLI |
|-----|-----|
| `get_llm_providers` | `pipefy ai-provider list` |
| `get_available_ai_models` | `pipefy ai-provider models` |
| `get_default_llm_provider` | `pipefy ai-provider default get` |
| `get_llm_provider_dependencies` | `pipefy ai-provider dependencies` |
| `validate_llm_provider_access` | `pipefy ai-provider validate-access` |

## Scope

- `get_llm_providers` lists custom and Pipefy-managed providers in one surface (`allLlmProvidersByOrganization`), with pagination (`first` default 50, `after`) and `only_active`, and a normalized `provider_type` (`byom` / `system`) discriminator.
- Include the provider `configuration` in read output and document that the API redacts secret values server-side (placeholders come back instead of secrets).
- Default get is organization-scoped; build on `defaultLlmProvider` (not the deprecated `activeLlmProvider`).
- This PR introduces the shared SDK-level GraphQL error classifier (permission / feature / not-found → structured problems) that all milestone probes and CLI write gating reuse — design it as a small standalone SDK module, not probe-local `except` mapping. (The existing error enrichment lives in the MCP layer and can't be imported by the SDK/CLI.)
- Probe semantics: a green `validate_llm_provider_access` proves read access only, never write entitlement; an empty system-provider list can mean the feature isn't enabled for the organization rather than a permission problem.
- Update the docs/skills shipped in #410 to point provider-ID discovery at these tools.

## Acceptance

- Full SDK + MCP + CLI parity, registry + `docs/parity.md` rows, skills updated, readOnly hints set.
- Tests cover pagination, both union members, and probe error mapping.

mocha06:
## Motivation

AI agent behaviors accept `providerId` / `systemProviderId`, but the only way to discover valid IDs was the organization's AI settings in the Pipefy UI. There was also no way to list the models a vendor exposes, see which provider an owner defaults to, check what depends on a provider, or verify that a credential can read the provider surface at all — and failures on this surface came back as opaque server errors. Closes #411.

## Outcome

- Five read-only tools with full SDK + MCP + CLI parity: `get_llm_providers` (custom + Pipefy-managed system providers in one union surface, paginated, `only_active`), `get_available_ai_models`, `get_default_llm_provider` (general `owner_type`, organization default), `get_llm_provider_dependencies`, and the `validate_llm_provider_access` probe. Registry and parity table go 157 → 162; all carry `readOnlyHint`.
- Providers carry the API's `type` discriminator (`byom` / `system`) and their `configuration`, which the API redacts server-side — placeholders come back instead of secrets.
- New standalone SDK module `pipefy_sdk.graphql_problem` classifies GraphQL errors into structured problems (permission denied / not found / invalid arguments / feature / runtime) with `code` and `correlation_id` carried through. The MCP tools and the probe consume it today; CLI write gating (#415) reuses it next. It lives in the SDK because the existing MCP enrichment cannot be imported by the SDK/CLI.
- Probe semantics are explicit: a green result proves read access only, never write entitlement; an empty system-provider list can mean system models aren't enabled for the organization rather than a permission problem; partial GraphQL errors alongside data stay visible on the result.
- The organization identifier is not uniform across the API and is documented per tool: list/dependencies take the org UUID, the org default takes the numeric org ID, models takes no org argument; `get_organization` returns both.
- CLI: `pipefy ai-provider list | models | default get | dependencies | validate-access`; `validate-access` exits 1 on a red probe.
- Docs and skills that pointed provider-ID discovery at the Pipefy UI now point at `get_llm_providers`; new reference doc `docs/mcp/tools/llm-providers.md`.

## Testing

`uv run pytest -m "not integration"` (3512 passed) + ruff + import-linter green. Exercised end-to-end against a dev org through the MCP server running this branch's code:

|---|---|---|---|---|
| 1 | `get_llm_providers` | Both union members, secrets redacted, pagination | 4 `system` + 2 `byom`; `token` / `clientSecret` / `customHeaders` all `__REDACTED__`; pagination block present | ✅ pass |
| 2 | `validate_llm_provider_access` | Green probe with visibility flags and read-only caveat | `ok: true`, both visibility flags true, note states green proves read access only | ✅ pass |
| 3 | `get_llm_provider_dependencies` (real provider) | Dependents page with total | `dependencies: []`, `total_count: 0` | ✅ pass |
| 4 | `get_llm_provider_dependencies` (bogus id) | Structured not-found + discovery hint | `RESOURCE_NOT_FOUND`, `kind: not_found`, "Use 'get_llm_providers'…" appended | ✅ pass (expected rejection) |
| 5 | `get_available_ai_models` / `get_default_llm_provider` without provider-management rights | Structured permission problem | `PERMISSION_DENIED`, `kind: permission_denied`, `correlation_id` carried | ✅ pass (expected rejection — the documented permission asymmetry) |
| 6 | Blank `organization_uuid` | Rejected before the wire | `'organization_uuid' must be non-empty.` | ✅ pass (expected rejection) |
| 7 | Probe on nonexistent org UUID | Classified failure, no org-existence leak | `PERMISSION_DENIED` rather than not-found | ✅ pass (expected rejection) |

Rows 4–7 are the probe error-mapping acceptance criterion exercised live: each returns the classified problem instead of the opaque error this issue set out to eliminate.

### DIFF
--- a/.github/workflows/scripts/lint_skill_refs.py
+++ b/.github/workflows/scripts/lint_skill_refs.py
@@ -15,6 +15,7 @@ PIPEFY_CLI_ROOT_COMMANDS = frozenset(
     {
         "agent",
         "ai-automation",
+        "ai-provider",
         "attachment",
         "audit",
         "automation",
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Open-source toolkit for **Pipefy** developers: a Model Context Protocol (MCP) se
 
 | Component | Package / path | Purpose |
 |-----------|----------------|---------|
-| **MCP server** | `pipefy-mcp-server` | Exposes **157** tools to MCP clients (Cursor, Claude Desktop, Claude Code, and others). |
+| **MCP server** | `pipefy-mcp-server` | Exposes **162** tools to MCP clients (Cursor, Claude Desktop, Claude Code, and others). |
 | **CLI** | `pipefy-cli` | Terminal commands aligned with MCP capabilities; see [`docs/parity.md`](docs/parity.md). |
 | **SDK** | `pipefy` | Vendor GraphQL client, services, and models shared by MCP and CLI. |
 | **Skills** | [`skills/`](skills/) | Markdown playbooks (Anthropic Skills format) for common Pipefy workflows. |
@@ -141,7 +141,7 @@ Deprecation and semver (post-1.0): [`docs/DEPRECATION.md`](docs/DEPRECATION.md).
 
 ## MCP server
 
-The server registers **157 tools** across eleven domains. Canonical names: `PIPEFY_TOOL_NAMES` in [`packages/mcp/src/pipefy_mcp/tools/registry.py`](packages/mcp/src/pipefy_mcp/tools/registry.py).
+The server registers **162 tools** across twelve domains. Canonical names: `PIPEFY_TOOL_NAMES` in [`packages/mcp/src/pipefy_mcp/tools/registry.py`](packages/mcp/src/pipefy_mcp/tools/registry.py).
 
 Tool descriptions and `Args:` blocks come from Python docstrings (what MCP clients show to models). Per-area reference docs cover parameters, edge cases, and cross-cutting behavior.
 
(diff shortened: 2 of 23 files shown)


adriannoes (approved):
## Summary

Reviewed with local checkout and tests on tip `7cf1aa7d`. This ships the #411 read/probe surface cleanly: five tools with SDK + MCP + CLI parity, `readOnlyHint`, registry/parity updates, and the shared `pipefy_sdk.graphql_problem` classifier ready for #415. CI is green on the tip (including the format + skill-ref allowlist fixup).

## What worked well

- Full parity for the five discovery reads, with org UUID vs numeric ID called out at the tool boundary.
- Probe semantics are explicit and locked in tests: green means list/read access only; empty system list can mean feature off; partial GraphQL errors stay visible on the result.
- Live smoke matrix in the PR body covers the acceptance paths that matter, including expected rejections.

## Also noted

- **F3:** A probe can return `ok: true` with a non-null `problem` when GraphQL returns partial data plus errors. That matches the locked tests and the PR contract. The success path still reads like a clean gate (MCP "confirmed", CLI exit 0), and the docs do not say a clean gate also needs `problem` absent. Worth a small docs (and maybe message) polish before #415 write gating keys off this probe: clean gate = `ok` and no `problem`; classifier `kind` on CLI is probe-shaped today. No need to flip `ok` for this PR.

PR #431, opened 2026-07-18
fix(mcp): scope the not-found hint away from the LLM provider list tool

### LINKED ISSUE #429: LLM provider tools: scope the not-found discovery hint away from the list tool
Follow-up to the review on #427, which fixed this pattern for the knowledge base tools.

## Problem

`_provider_tool_error_from_exception` in `packages/mcp/src/pipefy_mcp/tools/llm_provider_tools.py` appends "Use 'get_llm_providers' to list provider IDs for the organization." to every classified `not_found` error. That includes failures inside `get_llm_providers` itself (e.g. a bad organization UUID), so the error tells an agent to retry the exact tool that just failed — self-referential guidance instead of useful discovery.

## Fix (mirror #427 / 98c3a8d)

- Add a `not_found_hint: bool = True` parameter to `_provider_tool_error_from_exception` and pass `not_found_hint=False` at the `get_llm_providers` call site; per-id tools (`get_llm_provider_dependencies`, etc.) keep the hint.
- Regression tests: `get_llm_providers` + `not_found` asserts the hint is absent; a per-id tool still asserts it present.

## Acceptance

- A `not_found` failure from `get_llm_providers` carries the classified problem (kind/code/correlation_id) without the self-referential hint.
- Per-id provider tools keep the discovery hint.
- Tests cover both paths.

mocha06:
## Motivation
`_provider_tool_error_from_exception` appended "Use 'get_llm_providers' to list provider IDs for the organization." to every classified `not_found` error — including failures inside `get_llm_providers` itself (e.g. a bad organization UUID). The error then told the agent to retry the exact tool that just failed: self-referential guidance instead of useful discovery.

## Outcome
A `not_found` failure from `get_llm_providers` now carries the classified problem (kind/code/correlation_id) without the self-referential hint. The per-id provider tools (`get_llm_provider_dependencies`, etc.) keep the discovery hint. Mirrors the fix already applied to the knowledge-base tools.

Closes #429

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/llm_provider_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/llm_provider_tools.py
@@ -25,18 +25,23 @@ _PROVIDER_ID_DISCOVERY_HINT = (
 )
 
 
-def _provider_tool_error_from_exception(exc: BaseException) -> dict[str, Any]:
+def _provider_tool_error_from_exception(
+    exc: BaseException, *, not_found_hint: bool = True
+) -> dict[str, Any]:
     """Map an SDK/GraphQL exception onto the canonical tool failure envelope.
 
     Uses the shared SDK classifier (``pipefy_sdk.classify_exception``) so the
     kind/code the CLI and probes see is the same one reported here. A
     transport-level failure with no GraphQL errors falls back to ``str(exc)``.
+    ``not_found_hint`` scopes the id-discovery hint to the per-id tools; the
+    list tool passes False so its own failure never tells the caller to retry
+    the call that just failed.
     """
     problem = classify_exception(exc)
     if problem is None:
         return tool_error(str(exc))
     message = problem.message
-    if problem.kind.value == "not_found":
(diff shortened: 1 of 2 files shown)


adriannoes (approved):
## Summary

Surgical fix for the self-referential `not_found` discovery hint on `get_llm_providers`. It mirrors the knowledge-base pattern from #427, meets #429 acceptance, and keeps the classified problem envelope (kind/code/correlation_id). Reviewed with checkout+tests (13 provider tool tests green) and CI lint/test green on tip.

## What worked well

Keyword-only `not_found_hint` with list-tool opt-out, a complementary regression pair (list without hint / dependencies with hint), and a minimal surface area.

## Review path

- Checked out `ff89f460`, ran ruff and `packages/mcp/tests/tools/test_llm_provider_tools.py` (13 passed); CI green on tip.
- Call-site matrix and KB parity: only the inventory list needed `False`; default `True` remains correct for `get_llm_provider_dependencies`.
- Residual wrong-hint UX on non-per-id tools is pre-existing and out of #429 scope; left as a follow-up, not blocking.

PR #432, opened 2026-07-18
fix(sdk): tag raising S3 PUT failures with the s3_upload step

### LINKED ISSUE #430: AttachmentService: raising S3 PUT failures bypass the step-tagged error contract
## Problem

`AttachmentService`'s class docstring promises "Failures across every step surface as `AttachmentUploadError` carrying `step`", but the S3 `put` call in `upload_attachment` is not wrapped: an httpx transport error (connect/timeout) or the uploader's host-allowlist `ValueError` bubbles bare instead of raising `AttachmentUploadError` with `step="s3_upload"`. The `step="s3_upload"` wrap that does exist nearby covers `_extract_storage_path`, not the PUT itself. Only HTTP >= 400 results from a completed PUT are tagged.

Consequence: the MCP attachment tools' step-aware error envelope silently degrades to a generic untagged error for exactly the failures (network, tampered URL host) where the stage tag is most useful.

## Fix

Mirror the pattern `KnowledgeBaseService.create_ai_knowledge_base_document` uses for the same shared uploader (`HttpxS3Uploader`): wrap the `put` call, re-raise as `AttachmentUploadError(..., step="s3_upload")` from the cause, and leave the existing HTTP >= 400 tagging untouched. Add a unit test with a raising uploader double asserting the step tag, mirroring the KB document test.

## Context

Found during the KB documents phase review: the KB service inherited the same gap by mirroring `AttachmentService`, and fixed it on its side; this issue brings the original into parity so the two upload pipelines keep one contract.

mocha06:
## Motivation
`AttachmentService`'s docstring promises every step surfaces as `AttachmentUploadError` carrying `step`, but the S3 `put` call was not wrapped: an httpx transport error or the uploader's host-allowlist `ValueError` bubbled bare instead of being tagged `step="s3_upload"`. Only HTTP >= 400 results from a completed PUT were tagged, so the step-aware error envelope silently degraded to a generic untagged error for exactly the network / tampered-URL failures where the stage tag is most useful.

## Outcome
Transport errors and host-allowlist rejections from the S3 PUT now raise `AttachmentUploadError(step="s3_upload")` from the cause, keeping the attachment and knowledge-base document upload pipelines on one contract. Existing HTTP >= 400 tagging is unchanged.

Closes #430

### DIFF
--- a/packages/sdk/src/pipefy_sdk/services/attachment_service.py
+++ b/packages/sdk/src/pipefy_sdk/services/attachment_service.py
@@ -164,11 +164,18 @@ class AttachmentService:
             )
         upload_url = upload_url.strip()
 
-        put_result = await self._s3_uploader.put(
-            url=upload_url,
-            bytes_=file.bytes,
-            content_type=effective_type,
-        )
+        try:
+            put_result = await self._s3_uploader.put(
+                url=upload_url,
+                bytes_=file.bytes,
+                content_type=effective_type,
+            )
+        except Exception as exc:
+            # Transport errors and the uploader's host-allowlist rejection are
+            # s3_upload-stage failures; tag them so the step contract holds.
+            raise AttachmentUploadError(
+                f"S3 upload failed: {exc}", step="s3_upload"
+            ) from exc
         status = put_result.get("status_code", 0)
         if not isinstance(status, int) or status >= 400:
             body_snippet = put_result.get("body_snippet")
--- a/packages/sdk/tests/services/test_attachment_service.py
+++ b/packages/sdk/tests/services/test_attachment_service.py
@@ -431,6 +431,27 @@ async def test_upload_attachment_s3_http_error_carries_snippet_and_status(tmp_pa
 
 @pytest.mark.unit
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


adriannoes (approved):
## Summary

Surgical fix that closes #430: raising S3 PUT failures now surface as `AttachmentUploadError` with `step="s3_upload"`, matching the KnowledgeBaseService wrap. Reviewed with checkout+tests; CI lint and test are green on the tip.

## What worked well

- Mirrors the KB put wrap from #428 so attachment and knowledge-base uploads stay on one contract.
- Unit test asserts step, message fragment, and `__cause__`; it fails if the wrap is removed.
- Leaves the existing HTTP >= 400 tagging path untouched; no MCP/CLI surface churn needed.

## Review path

- Checked out the PR tip, ran attachment_service unit tests (24 passed) and ruff on touched files; CI green.
- Callers (facade, MCP tools, CLI) already catch `AttachmentUploadError` and benefit from the step tag.
- Coverage polish (allowlist `ValueError` in the service unit test) matches the intentional KB twin gap and was verified manually with `HttpxS3Uploader`; not a merge gate.

PR #436, opened 2026-07-18
fix(cli): consolidate provider + KB probe-gates into one strict shared helper

### LINKED ISSUE #435: CLI: consolidate provider + KB probe-gates into one strict shared helper (fix KB partial-denial gap)
Surfaced in the structural review of the LLM provider write tools (#415 / PR #434).

## Problem

The CLI has two near-identical write-gating helpers whose "clean gate" criteria disagree:

- `pipefy_cli.commands.ai_provider._provider_probe_gate` treats the gate as clean only when the probe is `ok` **and** carries no `problem`.
- `pipefy_cli.commands.knowledge_base._probe_gate` treats it as clean when the probe is `ok` alone.

Both access probes can legitimately return `ok: true` **with** a partial `problem`: when the API returns partial data alongside GraphQL errors, the service surfaces the classified error rather than discarding it, and deliberately does not flip `ok` (see `LlmProviderService.validate_llm_provider_access` and `KnowledgeBaseService.validate_knowledge_base_access`). The provider gate treats that partial denial as not-clean (correct — added in #415); the knowledge-base gate reads it as full access, so a KB write can proceed when it should be blocked. That is a latent partial-denial gap on the KB write path.

## Proposed change

Hoist a single strict `probe_gate(probe)` predicate into `pipefy_cli.commands._common` (clean = `ok` and no `problem`) and have both the provider and knowledge-base CLI write commands use it. This removes the duplication and closes the KB gap in one move.

## Acceptance

- One shared `probe_gate` helper in `_common`; both `ai_provider` and `knowledge_base` CLI write commands call it.
- KB write-gating rejects a probe that is `ok` but carries a `problem` (partial denial is never read as full access).
- A test covers the KB partial-denial case (probe `ok: true` + `problem` blocks the write).

## Notes

Best done after PR #434 (which introduced `_provider_probe_gate`) merges, so the change is a single clean diff off `dev` touching both CLI modules rather than a stacked branch. This changes knowledge-base CLI write-gating to be stricter, which is why it is scoped as its own reviewed follow-up rather than folded into #434.

mocha06:
Closes #435. Stacked on #434 (rc-dev/feat/llm-provider-writes) — retarget to `dev` once #434 merges.

## Motivation

The CLI had two near-identical write-gating helpers with divergent "clean gate" criteria: the provider helper (added in #434) treated a probe as clean only when it is `ok` and carries no `problem`, while the knowledge-base helper checked `ok` alone. Both access probes can return `ok: true` with a partial `problem` when the API returns partial data alongside GraphQL errors, so the knowledge-base gate read that partial denial as full access and let a write proceed.

## Outcome

A single strict `probe_gate(probe)` in `pipefy_cli.commands._common` backs both surfaces. Provider write-gating is unchanged; knowledge-base write-gating is now strict, so a partial-denial probe (`ok: true` with a `problem`) blocks the write instead of passing. The two duplicate helpers are removed.

### DIFF
--- a/docs/mcp/tools/knowledge-bases.md
+++ b/docs/mcp/tools/knowledge-bases.md
@@ -77,6 +77,7 @@ The toolkit validates the definition client-side because the API accepts shapes 
 - A green `validate_knowledge_base_access` proves **read access only** (`read_ai_agents` on the pipe) — never the `manage_ai_agents` entitlement that every knowledge base create / update / delete (plain text, document, data lookup) requires.
 - An **empty knowledge base list** is a valid green result (`knowledge_base_count: 0`), not a failure.
 - The **CLI gates writes** on the probe: `pipefy kb plain-text create` / `update`, `pipefy kb document create` / `update`, and `pipefy kb data-lookup create` / `update` run the read-access probe first and fail with the classified problem if it is denied, before attempting the mutation.
+- **Clean-gate contract.** The probe can return a success that still carries a `problem` — when the API returns partial data alongside GraphQL errors, it surfaces the classified error rather than discarding it, and deliberately does not flip `ok`. The CLI write gate therefore treats the gate as clean only when it is **`ok` and carries no `problem`**; a present `problem` is partial denial and is never read as full access.
 - The **MCP tools stay explicit-validate-first**: create / update do not auto-probe. Call `validate_knowledge_base_access` yourself before writing.
 - **Deletes require confirmation**: the MCP tool needs `confirm=true`; the CLI needs `--yes` (or an interactive prompt).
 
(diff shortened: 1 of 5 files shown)


adriannoes (dismissed):
LGTM for #435: shared strict `probe_gate` closes the KB partial-denial write gap. Non-blocking: mirror the clean-gate wording from `llm-providers.md` onto `knowledge-bases.md` when convenient; merge after #434.

mocha06:
Thanks @adriannoes. Non-blocking doc mirror done in a44f3f1: `knowledge-bases.md` now carries the same clean-gate contract bullet as `llm-providers.md`, so it describes the strict gate this PR introduces (clean only when the probe is `ok` **and** carries no `problem`; a partial denial never reads as full access). Still queued to merge after #434.

adriannoes (approved):
LGTM

PR #453, opened 2026-07-20
Add release orchestrator CLI for the publish flow

mocha06:
## Motivation

Cutting a release meant following a long manual checklist in RELEASE.md — bump the lockstep version, rewrite the CHANGELOG heading, commit, tag, push, wait for the workflow, then run several verification commands by hand (PyPI install, and a curl install.sh sanity-check). The verification steps in particular were easy to forget, and nothing failed loudly if you skipped one.

## Outcome

`scripts/release.py` drives the flow as three subcommands split at the irreversible tag-push boundary. `prepare` bumps the version (via the existing `bump_version.py`), stamps the CHANGELOG, and commits — all local and reversible, after confirming the computed version. `publish` tags, pushes, watches the Release workflow to completion, then verifies. `verify` asserts the GitHub Release ships all five wheels, the version installs from PyPI, and `install.sh` resolves the just-cut tag, then prints the release links. The formerly manual verification steps are now assertions that fail loudly, and the version transform stays owned solely by `bump_version.py`.

## Verification / proof of work

Validated end-to-end against real GitHub infrastructure, including a full `publish` on a throwaway fork ([mocha06/ai-toolkit](https://github.com/mocha06/ai-toolkit)). This surfaced and fixed two eventual-consistency races that unit tests would not have caught, both re-confirmed live afterward: the workflow-watch queried for the run before GitHub had created it (now polls until it appears), and the PyPI-install check ran once with no allowance for propagation lag (now retries with a clear final message).

### 1. Fully-green `verify` against a real published tag

Run against [`v0.3.0-beta.1`](https://github.com/pipefy/ai-toolkit/releases/tag/v0.3.0-beta.1) — all three checks pass and the link summary prints:

```
GitHub Release v0.3.0-beta.1 ships all 5 wheels.
PyPI install resolves: 0.3.0-beta.1
install.sh resolves v0.3.0-beta.1.

All checks passed for v0.3.0-beta.1.
```

Resolved links (each verified live):

- GitHub Release: https://github.com/pipefy/ai-toolkit/releases/tag/v0.3.0-beta.1
- PyPI: [pipefy](https://pypi.org/project/pipefy/0.3.0b1/) · [pipefy-cli](https://pypi.org/project/pipefy-cli/0.3.0b1/) · [pipefy-mcp-server](https://pypi.org/project/pipefy-mcp-server/0.3.0b1/) · [pipefy-auth](https://pypi.org/project/pipefy-auth/0.3.0b1/) · [pipefy-infra](https://pypi.org/project/pipefy-infra/0.3.0b1/)

### 2. Full `publish` end-to-end on the fork

`prepare` → `publish` cut and published `v0.3.0-beta.4` on the fork:

- Release workflow run (green, built and attached all five wheels): https://github.com/mocha06/ai-toolkit/actions/runs/29769309126
- Resulting GitHub Release with all five wheels: https://github.com/mocha06/ai-toolkit/releases/tag/v0.3.0-beta.4

The fork cannot publish to the real PyPI projects (Trusted Publishing is bound to `pipefy/ai-toolkit`, so PyPI rejects a fork's OIDC), which is exactly why it is safe — the PyPI upload step is skipped and the PyPI-install check correctly exhausts its retry window instead of touching the real packages.

### 3. Both fixes, observed live

```
Waiting for the Release workflow run for v0.3.0-beta.4 to appear...
GitHub Release v0.3.0-beta.4 ships all 5 wheels.
Verifying PyPI install of pipefy-cli==0.3.0b4 ...
  not on PyPI yet; retrying in 10s...
release: pipefy-cli==0.3.0b4 did not become installable from PyPI within 120s ...
```

### 4. `prepare` confirms the computed target before any mutation

```
release: Will bump 0.3.0-beta.4 -> 0.3.0-beta.5 and cut tag v0.3.0-beta.5. Proceed?
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ RALPH_TASK.md
 # Tracked release tooling only (ignore ad-hoc local scripts and smoke artifacts)
 scripts/*
 !scripts/bump_version.py
+!scripts/release.py
 
 # macOS
 .DS_Store
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,49 +12,29 @@ The next **GitHub pre-release** after the standalone repo’s [`v0.1.0-beta.1`](
 
 The Release workflow requires the git tag (without leading `v`) to **exactly match** `__version__` in `packages/sdk/src/pipefy_sdk/__init__.py` (and the MCP/CLI/Auth/Infra copies). For example tag **`v0.2.0-beta.1`** implies **`__version__ = "0.2.0-beta.1"`** in all five packages before you push the tag (set via step 2 below using `version=0.2.0-beta.1`, or edit the five `__init__.py` files together).
 
+`scripts/release.py` drives the whole flow as two subcommands split at the irreversible boundary — everything before the tag push is local and reversible, so you review the commit before anything leaves your machine.
+
 1. Merge work to `main` and ensure `CHANGELOG.md` has everything under `## [Unreleased]`.
-2. Bump the shared version (updates all five `__init__.py` files):
+2. Prepare the release. This bumps the shared version across every version-bearing file (via `scripts/bump_version.py`), stamps the `## [Unreleased]` CHANGELOG heading with the new version and today's date, re-seeds an empty `## [Unreleased]`, and commits — all locally:
 
    ```bash
-   uv run python scripts/bump_version.py patch
+   uv run python scripts/release.py prepare patch
    ```
 
-   Supported arguments: `major`, `minor`, `patch`, `prerelease`, or `version=X.Y.Z` (optional `v` prefix on `X.Y.Z`).
+   The bump argument accepts `major`, `minor`, `patch`, `prerelease`, or `version=X.Y.Z` (optional `v` prefix on `X.Y.Z`). `prepare` prints the computed target (`Will bump 0.3.0-beta.1 -> 0.3.0-beta.2 and cut tag v0.3.0-beta.2. Proceed?`) and waits for confirmation before touching any file, so a wrong bump costs nothing to walk away from (`--yes` skips the prompt for automation). Note `prerelease` only increments the current pre-release track (`beta.1 -> beta.2`) and never promotes across tracks; to go `alpha.N -> beta.1`, or to a **`v0.2.0-beta.*`** tag, or to any exact string, pass `version=X.Y.Z` with the PEP-440 form (for example `version=0.2.0-beta.1`) so it matches `GITHUB_REF_NAME` without the leading `v`.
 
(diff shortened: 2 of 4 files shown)


adriannoes (changes requested):
## Summary

Checked out the PR tip, ran `tests/test_release.py` (13 passed) and ruff on the new script; CI lint and test are green. The prepare/publish split and the bump_version ownership look right, and the workflow-appear plus PyPI retry fixes match real eventual-consistency failures. I am requesting changes for two recovery-path gaps that can leave a maintainer stuck or watching the wrong Release run after a failed or repeated cut.

## What worked well

- Clear irreversible boundary between local `prepare` and `publish`.
- Version transform stays solely in `bump_version.py`.
- RELEASE.md and the `.gitignore` allowlist updated with the new script.
- Live E2E on a fork informed the poll/retry behavior unit tests would miss.

## Also noted

- After a failed `git push` mid-`publish`, a local tag can remain while origin has none. The next attempt passes the remote-only existence check and then fails on `git tag` with a generic non-zero exit. Prefer refusing publish when the local tag already exists, tagging only after `main` is pushed, or recovering with a clear `ReleaseError` that names the local tag.
- On same-tag recovery (delete remote tag, fix, re-push), `watch_release_workflow` can attach to a prior completed Release run because the appear-loop stops as soon as any matching id exists. The comment that re-push resolves to the latest run only holds after the new run is already listed. Filter by push time or SHA, or keep polling until a run newer than the push appears.

## Review path

- Worktree at `a5ed385`; unit tests and ruff green; GitHub CI green on tip.
- Security pass on subprocess/`gh` usage; adversarial hunt plus refutation kept the two recovery findings and dropped several overreads (including installer latest-only as intentional).

mocha06:
Thanks @adriannoes — both recovery-path findings were spot on. Addressed in 3cfed35:

**Local tag after a failed push**
- `publish` now checks for a local tag up front and refuses with an actionable message (`Delete it with 'git tag -d <tag>' and re-run`) instead of a generic non-zero exit from `git tag`.
- Reordered so `main` is pushed *before* the tag is created — a rejected `main` push can no longer leave a dangling local tag behind.

**Attaching to a prior completed run on same-tag recovery**
- `publish` snapshots the tag's existing run IDs *before* pushing the tag, and `watch_release_workflow(exclude=…)` now polls until a run **not** in that snapshot appears. So a re-cut of the same tag waits for its genuinely new run rather than the stale completed one.
- Dropped the misleading "resolves to the latest run" comment and added a unit test (`test_newest_run_id_skips_excluded`) covering the exclusion.

Tests green (14) and ruff clean. Re-requesting review.

adriannoes (approved):
Both recovery items from the prior review are addressed: `publish` pushes `main` before creating the local tag and refuses a leftover local tag with a clear `ReleaseError`, and the workflow watcher snapshots prior run ids then waits for a non-excluded run. Unit tests (14) and ruff are green on the tip.

LGTM.

PR #454, opened 2026-07-20
Add release-pr subcommand for dev→main release PRs

mocha06:
## Motivation

The release CLI (added in #453) covers a release once `main` already carries the work, but getting there — branching, bumping the version, and opening the `dev → main` release PR — was still manual. Automating that prep removes the most repetitive part of a release while keeping the one irreversible step (the tag push) a deliberate human action.

## Outcome

Adds a `release-pr <bump>` subcommand. It branches off the latest `origin/dev`, runs the same bump + CHANGELOG stamp + commit as `prepare`, pushes, and opens a PR into `main`. It reads the current version and `## [Unreleased]` from `origin/dev` — not whatever branch you invoke it from — so the computed target is always correct, and it confirms that target before creating anything.

The generated PR description follows the established release-PR shape: a one-line summary, the mechanical bump it performs, the released content inlined (so a reviewer sees what ships without opening the CHANGELOG diff), and the post-merge step. Pre-release versus stable PyPI wording is derived from the version.

It deliberately does not tag or publish. After the release PR is approved and merged into `main`, a maintainer runs `release.py publish` from `main`; keeping the tag push human-driven is what lets it trigger the Release workflow (a tag pushed by CI's default token would not). The `prepare`/`publish`/`verify` behavior is unchanged; the shared bump-and-stamp core is factored out so both `prepare` and `release-pr` use it.

## Usage

```bash
# Recommended: open a dev -> main release PR (branches off latest origin/dev)
uv run python scripts/release.py release-pr patch

# After it's approved and merged into main, cut the release from main:
uv run python scripts/release.py publish

# On main directly (skip the PR flow): bump/stamp/commit, then publish
uv run python scripts/release.py prepare patch
uv run python scripts/release.py publish

# Re-run the post-publish checks for a tag
uv run python scripts/release.py verify v0.3.0-beta.5
```

| Command | What it does |
| --- | --- |
| `release-pr <bump>` | Branch off `origin/dev`, bump + stamp + commit, push, open a PR into `main`. Reversible; the PR review is the gate. |
| `prepare <bump>` | Bump + stamp CHANGELOG + commit on `main`. Local only. |
| `publish` | Tag, push, watch the Release workflow, then verify. The one irreversible step; confirms before the tag push. |
| `verify <tag>` | Assert the GitHub Release ships all wheels, the version installs from PyPI, and `install.sh` resolves the tag; print release links. |

`<bump>` is one of `major`, `minor`, `patch`, `prerelease`, or `version=X.Y.Z`. `prepare`/`release-pr` print the computed target and confirm before touching anything (`--yes` skips the prompt). Note `prerelease` only increments the current track (`beta.1 -> beta.2`); use `version=` to cross tracks (e.g. `version=0.3.0-beta.1`).

## Verification

Exercised end-to-end on a throwaway fork ([mocha06/ai-toolkit](https://github.com/mocha06/ai-toolkit)). Run from a clean `main`, `release-pr prerelease` correctly read `dev`'s version (`0.3.0-beta.4`), branched off `origin/dev`, bumped in lockstep to `0.3.0-beta.5`, stamped the CHANGELOG, pushed, and opened a `dev → main` release PR ([mocha06 PR #2](https://github.com/mocha06/ai-toolkit/pull/2)) whose generated description inlines the released content in the release-PR shape.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -245,16 +245,24 @@ def _show_at(ref: str, path: str) -> str:
     return capture(["git", "show", f"{ref}:{path}"])
 
 
-def _dev_current_version() -> str:
-    """The lockstep version on ``origin/dev`` (read from its root pyproject).
+def _version_at(ref: str) -> str:
+    """The lockstep version at ``ref`` (read from its root pyproject).
 
-    ``release_pr`` computes the target from dev's version, not the working
-    tree's — the branch it is invoked from may carry a different version.
+    ``release_pr`` derives the target from ``origin/dev``'s version, not the
+    working tree's — the branch it is invoked from may carry a different one —
+    and compares against ``origin/main`` to refuse a downgrade.
     """
     import tomllib
 
-    data = tomllib.loads(_show_at(f"origin/{DEV_BRANCH}", "pyproject.toml"))
+    data = tomllib.loads(_show_at(ref, "pyproject.toml"))
     return data["project"]["version"]
+
+
+def _target_ahead_of_main(target: str, main_version: str) -> bool:
+    """Whether ``target`` is a strictly newer version than ``main_version``."""
+    from packaging.version import Version
+
+    return Version(target) > Version(main_version)
 
 
 def _dev_unreleased_body() -> str:
@@ -316,15 +324,29 @@ def release_pr(bump_arg: str, *, assume_yes: bool = False) -> None:
             "add release notes first"
         )
 
-    current = _dev_current_version()
+    current = _version_at(f"origin/{DEV_BRANCH}")
     target = target_for(bump_arg, current)
+
+    # Guard against a downgrade-shaped release: if origin/dev is behind
+    # origin/main (a release bump on main not back-merged into dev), the bump
+    # computed from dev can land below main's already-released version.
+    main_version = _version_at(f"origin/{MAIN_BRANCH}")
+    if not _target_ahead_of_main(target, main_version):
+        raise ReleaseError(
+            f"Computed target v{target} is not ahead of origin/{MAIN_BRANCH} "
+            f"(v{main_version}); origin/{DEV_BRANCH} (v{current}) is behind "
+            f"{MAIN_BRANCH}. Back-merge {MAIN_BRANCH} into {DEV_BRANCH} before "
+            "cutting a release."
+        )
+
     branch = f"rc-{MAIN_BRANCH}/release/v{target}"
     if branch_exists(branch):
         raise ReleaseError(f"Branch {branch} already exists")
 
     confirm(
-        f"Will branch off origin/{DEV_BRANCH}, bump {current} -> {target}, and "
-        f"open a release PR into {MAIN_BRANCH}. Proceed?",
+        f"Will branch off origin/{DEV_BRANCH}, bump {current} -> {target} "
+        f"(origin/{MAIN_BRANCH} is at {main_version}), and open a release PR "
+        f"into {MAIN_BRANCH}. Proceed?",
         assume_yes=assume_yes,
     )
 
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -3,13 +3,18 @@
 
 The release has a single natural fault line: ``git push origin <tag>`` triggers
 PyPI publishing, which cannot be undone. Everything before it is local and
-reversible; everything after is read-only verification. So this tool is two
-subcommands with a review gate between them:
+reversible; everything after is read-only verification. So this tool splits
+into subcommands with a review gate between the reversible and irreversible
+halves:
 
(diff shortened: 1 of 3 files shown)


Danielmoraisg (dismissed):

scripts/release.py
```diff
@@ -209,6 +227,134 @@ def prepare(bump_arg: str, *, assume_yes: bool = False) -> str:
...
+
+    released = _dev_unreleased_body()
+    if not released:
+        raise ReleaseError(
+            f"'## [Unreleased]' in CHANGELOG.md on origin/{DEV_BRANCH} is empty; "
+            "add release notes first"
+        )
+
+    current = _dev_current_version()
+    target = target_for(bump_arg, current)
```
adriannoes: `release_pr` picks the bump target only from `origin/dev` and never compares it to `origin/main` before confirm. On the current remotes, `dev` is still on the alpha track while `main` is already on beta, so `release-pr prerelease` would compute a target below main's lockstep and still open the PR. The confirm prompt also never shows main's version, so it is easy to miss.

Please fail before confirm when the computed target is not strictly ahead of main's lockstep (and/or when `origin/main` is not an ancestor of the release tip), and cover that with a unit test on divergent fixtures. Confirm plus a later human `publish` still help, but they should not be the only line of defense against opening a downgrade-shaped release PR.

adriannoes (changes requested):
## Summary

Solid addition of `release-pr` on top of the #453 orchestrator: shared bump/stamp core, reads version and Unreleased from `origin/dev`, keeps tag/publish human-gated, and CI is green. Local checkout ran ruff and `tests/test_release.py` (16 passed). One merge-delaying gap remains on the new command's correctness claim.

## What worked well

- Clear split between reversible `release-pr` / `prepare` and irreversible `publish`.
- Target preview is computed before mutating, with a shared `_apply_prepare` path.
- Generated PR body matches the established release-PR shape (inlined released content).

## Required before merge

- `release_pr` must refuse a computed target that is not strictly ahead of `origin/main`'s lockstep (and/or when `origin/main` is not an ancestor of the release tip) before confirm, with a unit test on divergent fixtures. On the live remotes today, `release-pr prerelease` would still open a PR targeting `0.3.0-alpha.2` while `main` is already on `0.3.0-beta.1`.

## Review path

- Checked out PR tip `4070bfe`, ran ruff and release unit tests; CI lint+test green.
- Adversarial pass against live `origin/dev` vs `origin/main` versions; refutation kept the gap and demoted it from blocker to important because merge conflicts plus human `publish` limit silent ship.

mocha06:
Thanks @adriannoes — confirmed against the live remotes: `origin/dev` is `0.3.0-alpha.1` while `origin/main` is `0.3.0-beta.1`, so `release-pr prerelease` did compute `0.3.0-alpha.2` (a downgrade below main). Good catch. Fixed in 897a925.

**Applied — version guard.** `release_pr` now reads `origin/main`'s lockstep version and refuses before confirm when the computed target is not strictly ahead of it:

```
Computed target v0.3.0-alpha.2 is not ahead of origin/main (v0.3.0-beta.1);
origin/dev (v0.3.0-alpha.1) is behind main. Back-merge main into dev before
cutting a release.
```

The confirm prompt now also shows main's version, and there's a unit test on divergent fixtures (`test_target_ahead_of_main`, including the `alpha.2` vs `beta.1` case). Verified live: with today's remotes the guard fires and refuses.

**Not applied — the strict ancestry check (`origin/main` must be an ancestor of the release tip).** I looked at the actual topology and it would over-block: `origin/main` currently carries 10 commits `origin/dev` lacks (the beta.1 release bump + a couple of doc/fix commits), so main is *not* an ancestor of dev right now — an ancestry gate would refuse *every* `release-pr` until a back-merge, even the non-downgrade cases. The version guard already prevents the concrete harm (opening a downgrade-shaped PR), and its error message points at the same remedy (back-merge main→dev), which also brings dev up to main's tip. Happy to add the stricter gate if you'd rather enforce back-merge-before-release as hard policy, but it seemed like more constraint than the bug requires.

adriannoes (approved):
## Summary

Thanks for picking up the review feedback so quickly. The downgrade guard (`_target_ahead_of_main`), the clearer confirm prompt that surfaces `origin/main`'s version, and the divergent-fixture test close the gap from the earlier request for changes. CI is green on the tip.

## What worked well

- Guard fires before confirm, with an actionable back-merge message.
- Confirm now shows main's lockstep so the operator can see the relative bump.
- Unit coverage includes the live alpha-behind-beta case that motivated the finding.

## Review path

- Re-checked tip `897a925` against the prior finding: refuse when target is not strictly ahead of `origin/main`, plus tests.
- lint and test checks SUCCESS on the new commit.

Approved.

PR #455, opened 2026-07-21
fix(release): keep marketplace.json version in lockstep

mocha06:
## What

`scripts/bump_version.py` stamped `.claude-plugin/plugin.json` but never `.claude-plugin/marketplace.json`. The marketplace catalog entry's `version` is what the Claude Code plugin marketplace UI actually displays, so it silently drifted behind every other version-bearing file — the plugin showed `0.2.0-beta.1` while the released line was `0.3.0-beta.1`. The `verify` gate did not read it either, so nothing caught the drift.

image

## Changes

- Route both `.claude-plugin/*.json` manifests through a shared `JSON_MANIFESTS` list in the writer (`write_version_to_all_files`) and the `verify_lockstep` loop, so every bump moves both and `verify` fails if either drifts.
- Rename the now-shared regex `PLUGIN_MANIFEST_VERSION_RE` → `JSON_VERSION_RE` and correct its comment (it was documented as "what the marketplace shows", but the marketplace UI reads `marketplace.json`, not `plugin.json`).
- Correct the drifted catalog version to the current lockstep value (`0.3.0-beta.1`).
- A second `"version"` key in a manifest now fails loudly (via `_sub_exactly_one` / `_sole_match`) rather than being silently half-rewritten.
- Extend `RELEASE.md`'s automation reference to list `marketplace.json` under both `bump_version.py` and the CI `verify` gate.

## Notes

- Naming is tool-neutral (`JSON_VERSION_RE`, `JSON_MANIFESTS`, `_write_json_manifest_version`); adding another editor's plugin manifest later is one append to the tuple, covered by both writer and verify automatically.
- The same drift also exists on `dev`; it resolves there on the next `main`→`dev` sync, and future bumps keep it in lockstep.

## Verification

- `uv run python scripts/bump_version.py verify` → passes
- `uv run pytest tests/test_bump_version.py` → 45 passed (adds marketplace writer coverage: nested-entry rewrite + loud failure on a second `"version"`, and a real-manifest lockstep check across both manifests)
- `ruff check` + `ruff format --check` → clean


adriannoes (dismissed):
## Summary

Solid release fix. `marketplace.json` is now on the same writer/verify path as `plugin.json`, the catalog drift is corrected, and CI plus local checkout tests are green. Approving as-is. Please treat the follow-up below soon so the wiring cannot regress quietly.

## What worked well

- Shared `JSON_MANIFESTS` drives both write and verify, so another manifest later is one append.
- Catalog catch-up to `0.3.0-beta.1`, with a loud failure on a second `"version"` key.
- `RELEASE.md` automation table now lists marketplace under bump and CI verify.

## Also noted

- The new tests cover the regex helper and current-tree lockstep, but they still pass if `marketplace.json` is removed from `JSON_MANIFESTS` while versions happen to match. I recommend a small contract assert that `MARKETPLACE_MANIFEST` stays in that tuple (or an E2E write/verify case) so this PR's wiring stays locked. Not a merge blocker; worth doing in a follow-up before the next bump cycle.

## Review path

- Checked out the PR tip, ran `tests/test_bump_version.py` and ruff; CI lint+test green on tip; `verify` fails on forced marketplace drift.
- Coverage hunt confirmed the tuple-membership gap; calibrated to a summary-only note. Unrelated Bugbot noise from a wrong tree was discarded.

mocha06:
Thanks @adriannoes — applied in a27aa7c.

You were right that nothing locked the wiring: I unwired `marketplace.json` from `JSON_MANIFESTS` locally and the whole suite still passed, so the exact regression that caused the original drift had no guard.

Rather than `assert MARKETPLACE_MANIFEST in JSON_MANIFESTS` (tautological, and only covers the manifest we already know about), I matched this file's existing derived-contract pattern (`test_package_pyprojects_derived_from_workspace` / `test_init_paths_derived_from_hatch_config`): the new test derives the expected set from the `.claude-plugin` tree — every versioned `*.json` there must appear in `JSON_MANIFESTS`. That catches both dropping a manifest and a future third manifest someone forgets to wire.

Verified it bites: passes on the current tree, and fails naming the culprit (`Extra items in the left set: .../marketplace.json`) when the manifest is unwired. `verify` + ruff green.

I skipped a full E2E write/verify case on purpose — `verify_lockstep` reads every version-bearing file via absolute-path constants, so a real end-to-end run needs a whole fake repo scaffolded, which felt disproportionate for a non-blocker when the derived test already locks the specific wiring at risk. Happy to add it later if you'd prefer the behavioural coverage.

adriannoes (approved):
LGTM

PR #456, opened 2026-07-21
docs(skills): add pipefy-ipaas skill and harden AI agents guidance

adriannoes:
## Summary

Agent-facing skills update for Advanced Automations (iPaaS) and AI agents, plus a small test rename so fixtures do not bake vendor catalog entry names.

### New: `skills/ipaas/pipefy-ipaas/`

MCP-only skill for Pipefy Advanced Automations (integration flows). Teaches agents to:

- Use the **4 meta-tools** (`get_ipaas_tools`, `call_ipaas_tool`, `get_ipaas_connection_auth_url`, `create_ipaas_connection`) instead of a fixed verb list
- Follow **discover → expand one schema → call** (compact catalog first; expand only the entry about to run)
- Treat the catalog as **per-pipe and discovered at runtime** (capability groups: author / inspect / pieces / connections / publish / test / tables / AI — no hardcoded vendor entry names)
- Run the happy path **build → validate → test → (optional) inspect run → publish**, with connections / OAuth / `$env` (local-only) notes
- Handle common failure modes (org iPaaS disabled, wrong arg names, test side effects, run-detail not-found after a SUCCEEDED test)

Also registers the `ipaas` domain in `skills/README.md` and `skills/AGENTS.md`.

### Refresh: `skills/ai-agents/pipefy-ai-agents/`

Brings the skill in line with the current MCP/CLI surface:

- Documents all **6** known `actionType`s, including `create_table_record` and `send_email_template` (with metadata examples)
- Clarifies that **`inputMode` is required** on every `fieldsAttributes` entry (`fill_with_ai` / `fixed_value` / `copy_from`)
- Notes **remote-safe** reads vs local-only writes
- Documents CLI **auto pre-flight** on `agent create` / `agent update` vs MCP (call `validate_ai_agent_behaviors` yourself)
- Documents `strict_unknown_action_types` / `--strict` / `--no-strict`
- Fixes **partial-create recovery**: `update_ai_agent` needs the full required payload (`uuid`, `repo_uuid`, `name`, `instruction`, complete `behaviors`), not uuid alone

### Tests

- `packages/mcp/tests/core/test_ipaas_gateway.py`
- `packages/mcp/tests/tools/test_ipaas_tools.py`

Fixture catalog names renamed `ap_*` → `demo_*` so unit tests match the discover-at-runtime posture (no vendor-prefixed entry names in fixtures).

## Out of scope

- No MCP/CLI tool behavior changes (skills + test fixture naming only)
[...]

PR #458, opened 2026-07-21
docs(ai-agents): clarify fill_with_ai input vs output field refs

adriannoes:
## Summary
- Documents that `fill_with_ai` marks **output** fields, while `%{field:…}` / `referencedFieldIds` are **input** refs needed only when the agent must read card field values.
- Clarifies that not every `fill_with_ai` requires input refs (instruction-only, OCR/attachment, knowledge-base context); missing refs when card inputs are needed leaves `card.fields` empty at trigger time (hallucination risk).
- Updates MCP tool docstrings (`create_ai_agent` / `update_ai_agent`) and `skills/ai-agents/pipefy-ai-agents/SKILL.md` at the point of use.

## Test plan
- [x] Local CI: `ruff check` / `ruff format --check` on package src trees
- [x] Local CI: `lint-imports` (mcp)
- [x] Local CI: `uv run pytest -m "not integration"` (3738 passed)
- [x] Skills frontmatter check for `pipefy-ai-agents`
- [x] Live MCP smoke: `validate_ai_agent_behaviors` with and without `%{field:…}` both `valid: true` (no forced refs)
- [ ] Reviewer: skim skill + tool docstring for clarity with agent authors


adriannoes:
**Merge order (stacked PR).** #456 is stacked on this branch (its base points at `fix/ai-agent-fill-with-ai-referenced-fields`). Both edit the same step-6 block in `skills/ai-agents/pipefy-ai-agents/SKILL.md` (the `inputMode` bullet and the `validate` checklist), so this one merges first. Once it lands in `dev`, GitHub retargets #456 back to `dev` automatically and its diff stays clean.

mocha06:
## Summary of review

Approve. Docs/docstring-only change clarifying `fill_with_ai` (output) vs `%{field:…}` input refs. I verified the technical claims against the actual code rather than just reading the prose, and they hold up. No functional risk.

## What worked well

- The runtime behavior is described accurately: `referencedFieldIds` is populated only from bare-numeric `%{field:<digits>}` tokens, and dotted connected-pipe refs (`%{field:<parent>.<child>}`) genuinely are **not** forwarded at runtime — this matches the extraction regex and its own code comment.
- The split of responsibilities is correct: `validate_ai_agent_behaviors` does not existence-check instruction tokens and does not rewrite slugs; slug→numeric rewrite and `referencedFieldIds` population happen only on create/update.
- Pointing readers at `get_start_form_fields` / `get_phase_fields` to confirm ids is the right remediation.

## Also noted (optional, non-blocking)

- Precision nit: "A wrong input id/slug … becomes a dead `referencedFieldId`" is strictly accurate only for a wrong **numeric** id. A wrong **slug** never resolves and is dropped by the digits-only extractor, so it never becomes a `referencedFieldId` at all — it stays an unresolved token. The end result you describe (empty `card.fields` → hallucination risk) is identical either way, so the guidance stands; only the mechanism is slightly conflated. Since precision on exactly this behavior is the point of the PR, a one-word tweak (e.g. distinguishing the numeric-id case) might be worth it.

## Review path

Checked the SDK behavior-placeholder extraction/population logic, the pipe-validation module, the create/update client path, and the action model validators — not just the docstring text. Confirmed create/update resolve-then-populate ordering and that validate stays a pure check.

adriannoes:
Addressed the precision nit in `c4e55466`: wrong **numeric** ids become dead `referencedFieldId`s; wrong **slugs** are dropped by the digits-only extractor and stay unresolved tokens. Same empty-`card.fields` outcome either way.

mocha06 (approved):

PR #460, opened 2026-07-22
Add service-account management tools (create, delete, add-to-pipe)

### LINKED ISSUE #459: iPaaS: service account needs pipe membership but no add-to-pipe setup path
## Problem
When agents set up **iPaaS (Advanced Automations)** flows that need a **service account**, creating the SA is not enough: the SA must also be a **member of the target pipe** or pipe-scoped calls fail with permission errors.

Today the toolkit has generic membership tools (`invite_members`, `set_role`, `get_pipe_members`) but **no dedicated / discoverable path** for “take this org service account → grant it a role on this pipe” as part of the iPaaS / integration setup loop. iPaaS docs and skills do not mention this step.

## Evidence
From product testing + live GraphQL introspection + current MCP surface:

- **SA create ≠ pipe access.** Live `createServiceAccount` takes `organizationUuid`, `name`, and an **organization** `role` only. The returned `ServiceAccount` exposes `id` / `uuid` / `email` (and OAuth client fields), but does **not** attach the SA to any pipe.
- **Pipe membership is a second call** on the public API:
  - `inviteMembers` — `pipe_id` + `emails: [{ email, role_name }]` (email-only; no user/SA id).
  - `setRole` / `setRoles` — `pipe_id` + `member: { user_id, role_name }` (by user id).
- **MCP today:** `invite_members` (email), `set_role` (member_id), `get_pipe_members`, plus iPaaS tools (`get_ipaas_tools`, `call_ipaas_tool`, connection helpers). There is **no** MCP tool to create/list org service accounts, and **no** SA-oriented “add to pipe” tool or skill step.
- **Docs gap:** `docs/ipaas.md` / iPaaS MCP docs cover catalog + connections only. README notes that a toolkit SA must be added to each pipe, but does not give an agent-callable recipe. Membership skill/docs describe human email invites, not the iPaaS SA setup path.
- **Workaround is fragile:** agents must somehow learn the SA email (or numeric user id), then call `invite_members` / `set_role`, then verify with `get_pipe_members`. That chain is easy to miss mid-iPaaS setup, which surfaces later as `PERMISSION_DENIED` on pipe operations.

## Impact
iPaaS / integration setup appears to succeed (SA created, connection/flow configured) but pipe actions under that identity fail until someone remembers membership. Extra round-trips, Admin UI fallback, and silent incomplete setups.

## Status on `dev`
**MISSING** for the dedicated flow. Generic membership tools exist; SA create/list MCP tools and iPaaS↔membership guidance do not.

- **Docs / skill only:** document the required sequence in iPaaS + members guidance — resolve SA `email` or `user_id` → `invite_members` or `set_role` with pipe role (`admin` / `member` as appropriate) → `get_pipe_members` verify.
- **Dedicated tool:** e.g. `add_service_account_to_pipe(pipe_id, service_account_…, role_name)` wrapping public `setRole` / `inviteMembers`, with a docstring aimed at iPaaS setup.
- **Broader SA surface:** expose create/list service-account tools (live schema already has `createServiceAccount` / `deleteServiceAccount` + `ServiceAccount` type), then compose membership as the next step — larger scope.
- **Clarify tool contracts:** call out in `invite_members` / `set_role` docs that pipe-scoped membership applies to service-account users (org-level SA invite is a different, restricted path).

Related areas: iPaaS tools, members tools, auth/service-account provisioning.

mocha06:
## Motivation

Closes #459 
Setting up an integration that runs under a Pipefy service account — for example an iPaaS (Advanced Automations) flow — was a multi-step job the toolkit could not complete on its own. There was no tool to create a service account, and even with one in hand, a freshly created account has no pipe access: pipe-scoped calls under its identity fail with a permission error until it is added as a member of each target pipe. Agents had to drop to raw GraphQL to provision an account and could silently skip the membership step, which surfaced later as `PERMISSION_DENIED` on pipe operations.

## Outcome

The full service-account lifecycle is now first-class across MCP, CLI, and the SDK:

- `create_service_account` provisions an organization service account and returns its OAuth2 client id/secret and token endpoint — available only once, at creation, never logged, and excluded from the remote profile. An optional `pipe_ids` argument adds the new account to those pipes immediately (`pipe_role` defaults to `admin`) and returns a per-pipe `pipe_memberships` summary.
- `delete_service_account` removes an account (two-step destructive confirmation).
- `add_service_account_to_pipe` grants an existing account pipe membership by email, defaulting the role to `admin`, and verifies membership afterwards so an incomplete setup is never reported as success.

CLI twins ship alongside: `pipefy service-account create|delete` and `pipefy member add-service-account`. A new "Service accounts" tool area, the parity matrix, and the members and iPaaS skills document the create → add-to-pipe → use → delete flow.

## Proven contract

- `createServiceAccount(organizationUuid, name ≤20, role, [description], [expirationTime {unit,value}])` → `ServiceAccount { id, uuid, email, name, role, client { id, secret }, token { endpoint } }`. The client secret is returned only at creation — there is no read-back query, no list query, and no update mutation. Name length (≤20) is backend-enforced and validated client-side.
- `deleteServiceAccount(organizationUuid, serviceAccountUuid)` → `success`.
- `add_service_account_to_pipe` wraps `inviteMembers(emails: [{ email, role_name }], pipe_id)`. Org roles: `admin`, `normal`, `company_guest`, `external_guest`. Pipe roles: `admin`, `member`, `creator`, `my_cards_only`, `read_and_comment`.
- Service-account credentials are returned to the caller (needed to authenticate the account) but never logged. All three tools are remote-safe (each acts with the request-scoped bearer and is governed by API permissions; the returned secret is never logged, as the hosted logging layer excludes response bodies).

## Testing

Verified live against a test organization and pipe:

| Scenario | Observed |
|----------|----------|
| Create + delete service account (SDK) | client id/secret + token endpoint returned; delete `success: true` |
| `create_service_account(pipe_ids=[…])` chaining | account added to the pipe; `granted_role_on_pipe: admin` (default confirmed) |
| Service account acts on the pipe | pipe-card reads went `0 → 2` once membership was granted |
| iPaaS flow as the service account | flow runs end-to-end; Slack message delivered with the card template resolved |

MCP tool logic, validation, the secrets contract, and the two-step delete are covered by unit tests at the SDK, MCP, and CLI layers.


packages/cli/src/pipefy_cli/commands/member.py
```diff
@@ -77,6 +77,33 @@ async def factory(client: PipefyClient):
...
+    if not rn:
+        typer.echo("--role must be non-empty.", err=True)
+        raise typer.Exit(2)
+    addr = email.strip()
+    if not addr:
+        typer.echo("--email must be non-empty.", err=True)
+        raise typer.Exit(2)
+
+    async def factory(client: PipefyClient):
+        return await client.add_service_account_to_pipe(pipe_id, addr, rn)
```
adriannoes, line 102: `pipefy member add-service-account` only calls the SDK invite wrapper and exits 0 on any non-raising GraphQL response. Soft failures inside `inviteMembers` (empty `users` plus row `errors`) do not raise, so the CLI can report success when the service account was never added.

MCP hardens this with `service_account_is_member` and returns an error when membership is confirmed absent. `docs/parity.md` and the members skill currently claim verification for both surfaces.

Please either share the same post-invite check on the CLI path (prefer behind the SDK so both stay aligned) or narrow parity/skills to MCP-only until CLI catches up. Done when a soft-failed invite exits non-zero (or docs no longer claim CLI verifies).

packages/mcp/src/pipefy_mcp/tools/member_tool_helpers.py
```diff
@@ -12,6 +13,36 @@
...
+    pipe_id_str = str(pipe_id).strip()
+    if not pipe_id_str.isdigit():
+        return None
+
+    try:
+        members_data = await client.get_pipe_members(pipe_id_str)
+    except Exception:  # noqa: BLE001
+        return None
+
+    members = (members_data.get("pipe") or {}).get("members", [])
```
adriannoes: In `service_account_is_member`, `pipe.members` can be GraphQL `null`. `.get("members", [])` then yields `None`, and the membership loop raises `TypeError` outside the try/except (which only covers the network call).

On `create_service_account` with `pipe_ids`, that runs after create. An unhandled parse error can fail the tool without returning the one-time client secret that cannot be read back.

Please coerce null to an empty list (or treat it like other verify-unknown paths) and add a regression test. Done when `members: null` does not raise and create+`pipe_ids` still returns the secret.

```suggestion
    members = (members_data.get("pipe") or {}).get("members") or []
```

adriannoes (changes requested):
## Summary

Solid end-to-end surface for the #459 service-account lifecycle (SDK, MCP, CLI, parity, skills), with CI green and a full local unit run on the tip. Requesting changes on two membership-verification gaps that undercut the "incomplete setup is never reported as success" / one-time-secret contract.

## What worked well

- MCP `add_service_account_to_pipe` hard-fails when membership is confirmed absent (numeric pipe id), with good unit coverage.
- Secrets contract is explicit; tool logging logs arg keys only, not response bodies.
- `REMOTE_SEED` and `meta=REMOTE` stay in lockstep after #462; destructive delete is two-step on MCP and CLI.

## Required before merge

- Fix null `pipe.members` handling in `service_account_is_member` so create+`pipe_ids` cannot crash after create and strand the one-time client secret.
- Align CLI `member add-service-account` with MCP membership verification, or narrow parity/skills so they no longer claim the CLI verifies.

## Also noted

- Lifecycle docs tell agents to capture `client.secret`; the MCP success payload nests credentials under `data.serviceAccount.client.secret`. A one-line docs fix would match the unit test path.
- PR description still says create/delete are not remote-safe; refresh after #462 (they are on the remote seed).

## Review path

- Checked out PR tip `9a23d78`, ran package unit tests and ruff; CI green on tip.
- Subagent passes plus refutation kept CLI verify gap and `members: null` crash; demoted the docs secret-path mismatch to summary; dropped the intentional None-verify fail-open path.

packages/mcp/src/pipefy_mcp/tools/member_tool_helpers.py
mocha06: Fixed in c3ed88c. Confirmed against the live schema that `Pipe.members` is a nullable `LIST`, so `members: null` is possible and `.get("members", [])` returned `None` (the default only applies to an absent key), raising `TypeError` in the loop outside the try/except. Coerced with `.get("members") or []` in both `service_account_is_member` and `_verify_removal` (same latent pattern). Added regression tests: `members: null` no longer raises, and `create_service_account(pipe_ids=…)` still returns the one-time client secret in that case.

packages/cli/src/pipefy_cli/commands/member.py
mocha06: Fixed in c3ed88c by narrowing the docs (your second suggested resolution) rather than adding a CLI-side check. Rationale: the CLI surface is uniformly a thin raw-payload printer — `member invite`, `set-role`, and `remove` all exit 0 on soft `inviteMembers` failures — and membership verification is an MCP-layer enrichment (`_verify_removal` also lives in MCP, not the SDK). Pushing verification behind the SDK would make `add-service-account` the one CLI command that verifies and would move app-level logic into the thin SDK. So `docs/parity.md` and the members skill now state that post-invite verification is the MCP tool's behavior; the CLI returns the raw invite result (inspect `inviteMembers.errors` or run `pipefy member list`). Happy to add a CLI-layer soft-failure exit code instead if you'd prefer the stricter behavior.

adriannoes (approved):
## Summary

Follow-up approve on tip `148f01b`. The membership-verification gaps from the earlier request-changes pass are addressed: null `pipe.members` is coerced safely with regression coverage, and parity/skills now state that post-invite verification is MCP-only while the CLI returns the raw invite payload. CI is green.

## What worked well

- Fix commit documents the schema nullability and covers both `service_account_is_member` and `_verify_removal`.
- Choosing the docs/parity narrowing for CLI keeps the CLI as a thin raw-payload surface, consistent with sibling member commands.

## Review path

- Re-checked tip against the two required items and the secret response-shape note; no remaining blockers.

Issue #442, opened 2026-07-18
Policy: exposing write tools on the remote profile


mocha06:
Companion to the read-tool remote migrations (#437–#441). Those are mechanical (mark pure reads `meta=REMOTE`). This issue is the **policy decision** for the ~100 write tools, which the remote profile deliberately withholds today ("write mutations that haven't been reviewed for remote exposure yet", `packages/mcp/CLAUDE.md`).

## Decision needed

Exposing mutations on a hosted, multi-user server is a security/product posture change, not a marker flip. Before marking any write remote-safe, decide the policy:

- **Governance**: do remote writes need per-user quotas / rate limiting / cost weighting (the tool-call middleware seam exists) before exposure?
- **Destructive ops**: how does the two-step `confirm` flow behave for a hosted caller, and is that sufficient for deletes?
- **Input restriction**: which writes need per-call input restriction under the remote profile (via `is_remote_profile(ctx)`), like `create_ipaas_connection` rejecting `{"$env": ...}` refs?
- **Internal-API writes** (e.g. `delete_card_relation`): any additional review beyond the ordinary per-permission governance?
- **Per-deployment vs per-user settings**: confirm no write reads a process-global setting for a per-user decision (e.g. `default_webhook_name`, `allow_insecure_urls` on webhook create/update).

## Hard-excluded (not candidates, ever)

- Local-filesystem inputs: `upload_attachment_to_card`, `upload_attachment_to_table_record`, `create_ai_knowledge_base_document`, `create_llm_provider`, `update_llm_provider`. (A `file_url`-style remote variant would be separate follow-up work.)
- `execute_graphql` — arbitrary GraphQL escape hatch.

## Scope of the ~100 writes (for staging once policy is set)

Creates, updates, deletes across pipes/cards/phases/fields/tables/records/relations/labels/reports/portals/webhooks/members/automations/AI agents/AI automations/knowledge bases/LLM providers, plus action-style writes (`move_card_to_phase`, `set_role`, `toggle_ai_agent_status`, `fill_card_phase_fields`, `set_table_record_field_value`, `invite_members`, `send_*`, `export_*`, `publish_*`/`unpublish_*`, `clone_pipe`, `simulate_automation`, `duplicate_portal_element`, `sort_portal_pages`, the LLM provider owner ops, …).

## Acceptance

- A written policy for remote write exposure (governance + input-restriction + destructive-op stance).
[...]

PR #463, opened 2026-07-22
docs(remote): policy for exposing write tools on the remote profile

mocha06:
## Motivation
The remote profile gained its first public-GraphQL write tools — the organization service-account tools (the iPaaS meta-tools already on the seed are writes too, but reach the iPaaS host) — without a written standard for what makes a *write* (as opposed to a read) safe to expose on a hosted, multi-user server. #442 asks for that policy so future write categories are staged against one shared test instead of being decided case by case, and so the reasoning behind the service-account exposure is captured rather than implicit.

## Outcome
`packages/mcp/AGENTS.md` gains a "Write tools on the remote profile" section stating the test a write must pass to earn `meta=REMOTE`: authorization is enforced downstream by the API and never client-side or inferred from reachability; a returned secret is kept out of both hosted logging layers (argument values and response bodies are never logged); and the two-step `confirm` flow is a deliberateness guard, not an authorization boundary. The section records the service-account tools as the worked example, notes when per-call input restriction is and isn't required, and states that per-user write governance (quotas, rate limiting, cost weighting) is deliberately deferred to the scaling milestone — tracked on #307 — rather than being a precondition of exposure.

Refs #442.


adriannoes (approved):
## Summary

Clear, greppable write-exposure policy for the remote profile. The three write-specific criteria (API-only authorization, returned-secret logging, `confirm` as deliberateness rather than authz), the input-restriction rule, and the deliberate deferral of per-user governance to #307 match the code and the stated outcome of #442's policy half. Static review only (docs-only diff); logging, `confirm`, and `$env` claims were checked against the implementation.

## What worked well

- Concrete pointers to `tool_log_middleware`, `RequestLogMiddleware`, and `check_destructive_confirmation` instead of abstract security language.
- Service-account tools as a worked example of an already-shipped remote write, with governance explicitly not a precondition of exposure.

## Also noted

- Nit: under this section's own definition of write, `create_ipaas_connection` and `call_ipaas_tool` were already on `REMOTE_SEED`. Calling the service-account tools "the first writes on the remote seed" (and the same ordinal in the PR body) is inaccurate. Dropping "first" or narrowing to first public-GraphQL org/pipe writes would keep the history honest; the same phrasing already appears in #460's `REMOTE_SEED` comment.

## Review path

- Docs-only PR (+14 in `packages/mcp/AGENTS.md`); claims verified against middleware and tools on the stack tip.
- Ticket #442 AC1 (written policy) met; category follow-ups stay on #442 rather than blocking this foundation PR.
- One wording nit only; no blockers.

mocha06:
Thanks — the "first writes on the seed" nit is correct: \`call_ipaas_tool\` and \`create_ipaas_connection\` are already writes on \`REMOTE_SEED\` (they reach the iPaaS host rather than public GraphQL). Fixed the wording in `9b5033f` to "the first **public-GraphQL** writes on the remote seed," with a parenthetical noting the iPaaS meta-tools are writes too. Applied the same correction to the PR body and to the source of the phrasing — the `REMOTE_SEED` comment on #460 (`87879f9`). No behavior change; docs-only.

PR #464, opened 2026-07-22
Type eventParams into AutomationEventParamsInput

### LINKED ISSUE #418: Type eventParams (behaviors + automations) — fill AutomationEventParamsInput
Part of the "Typed payloads: parse, don't validate" milestone. Follows #416 / PR #417.

## Problem

`eventParams` is walked with dual-alias branches in several places instead of a typed model:
- `ai_pipe_validation.py`: `ep.get("to_phase_id") or ep.get("toPhaseId")`
- `ai_phase_transition_validation.py`: `src = ep.get("to_phase_id") or ep.get("toPhaseId")`
- `automation_preflight.py`: `extra.get("event_params") or extra.get("eventParams")`, then `ev.get("to_phase_id") or ev.get("toPhaseId")`

`AutomationEventParamsInput` already exists as an **empty-shell** pydantic model (`models/ai_automation.py`, `extra="allow"`, no fields), so `event_params` on behaviors/automations is effectively untyped.

## Wire shape (introspected, dev)

`AutomationEventParamsInput` inputFields: `kindOfSla`, `fromPhaseId`, `inPhaseId`, **`to_phase_id`** (snake!), `triggerAutomationId`, `triggerFieldIds`.

Note the mixed casing in one type: `to_phase_id` is snake while its siblings are camel — **per-declared-field aliases only, never a blanket converter.**

## Scope

- Fill `AutomationEventParamsInput` with the six declared fields (`populate_by_name=True`, per-field aliases, keep `extra="allow"`).
- Type `event_params` on the behavior read/normalize path (`BehaviorPayload`, `BehaviorInput`) and the automation input, and read typed attributes at the consumers above.
- Delete the `to_phase_id`/`toPhaseId` and `event_params`/`eventParams` dual-alias branches.

## Sequencing

Touches `models/ai_agent.py` and `models/ai_automation.py`. Sequence **after #410** (capability/provider hardening) to avoid conflicts on `models/ai_agent.py`.

## Acceptance

- No `to_phase_id`/`toPhaseId` or `event_params`/`eventParams` dual-alias branches remain in the migrated paths.
- `uv run pytest -m "not integration"` + ruff + import-linter green.

mocha06:
## Motivation

`eventParams` was walked with dual-alias `.get("to_phase_id") or .get("toPhaseId")` branches in the AI pipe/phase-transition validators and the traditional-automation preflight, and `AutomationEventParamsInput` existed only as an empty-shell model — so `event_params` on behaviors and automations was effectively untyped. This continues the "parse, don't validate" approach proven in #417.

## What ships

- `AutomationEventParamsInput` filled with its six declared inputFields (`populate_by_name=True`, per-field aliases, `extra="allow"`).
- `event_params` typed on `BehaviorPayload` / `BehaviorInput` (alias `eventParams`) and read as typed attributes in `ai_pipe_validation` and `ai_phase_transition_validation`; the classic-automation move preflight parses `event_params` leniently (`_parse_event_params`).
- `_automation_event_params_for_api` now dumps `by_alias=True` — it was correct only by accident while the model was an empty shell with no aliased fields.

## Proven contract (introspected, live)

`AutomationEventParamsInput` inputFields: `kindOfSla`, `fromPhaseId`, `inPhaseId`, **`to_phase_id`** (snake), `triggerAutomationId`, `triggerFieldIds` — mixed casing in one type, so aliases are strictly per declared field; `to_phase_id` keeps its snake wire name (no alias). Behaviors carry event params under the camel wrapper `eventParams` (`AutomationInput`); classic automations under the snake wrapper `event_params` (`CreateAutomationInput`). Both nest the same inner type.

## Behavior boundaries

Behavior is unchanged for every payload already valid end-to-end. Camel inner spellings (`toPhaseId`) that GraphQL always rejected at coercion are no longer silently accepted; tests pinning that broken passthrough are rewritten to the declared name.

## Testing

Offline: `uv run pytest -m "not integration"` + ruff + import-linter + `bump_version.py verify` green. Live E2E on pipe `303088927`:

| 1 | `create_ai_automation` with `event_params.triggerFieldIds` → GET | `triggerFieldIds` round-trips as the camel wire name (by_alias) | GET returned `triggerFieldIds: ["360646190"]` | ✅ |
| 2 | Create AI agent behavior with `eventParams.to_phase_id` → GET | `to_phase_id` round-trips (snake) | GET returned `to_phase_id: "319036222"` | ✅ |
| 3 | Classic `card_moved`→`move_single_card` with `event_params.to_phase_id` → GET | `to_phase_id` round-trips | GET returned `to_phase_id: "319036222"` | ✅ |

Closes #418.

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
@@ -627,7 +627,7 @@ class AiAgentTools:
             Known ``actionType`` values for pipe-context checks are:
             ``move_card``, ``update_card``, ``create_card``, ``create_connected_card``,
             ``create_table_record``, and ``send_email_template``.
-            For ``move_card`` with trigger ``card_moved`` and ``eventParams.to_phase_id``/``toPhaseId``,
+            For ``move_card`` with trigger ``card_moved`` and ``eventParams.to_phase_id``,
             the tool also checks that ``destinationPhaseId`` is allowed from that phase
             (``cards_can_be_moved_to_phases``).
             For ``create_table_record``, ``fieldsAttributes`` hold **table** field IDs — they are not
--- a/packages/mcp/src/pipefy_mcp/tools/automation_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/automation_tools.py
@@ -387,7 +387,7 @@ class AutomationTools:
             ``active: false`` to disable a rule after creation.
 
             For ``card_moved`` rules with action ``move_single_card``, when ``extra_input`` includes
-            ``event_params.to_phase_id`` (or ``toPhaseId``) and ``action_params.to_phase_id`` (or
+            ``event_params.to_phase_id`` and ``action_params.to_phase_id`` (or
             ``phase.id``), the tool rejects impossible transitions before calling the API, using the
             same read-only transition data as ``move_card_to_phase``.
 
--- a/packages/mcp/tests/tools/test_ai_tool_helpers.py
+++ b/packages/mcp/tests/tools/test_ai_tool_helpers.py
@@ -447,13 +447,16 @@ def test_validate_invalid_event_params_to_phase_id():
         related_pipe_ids=RELATED_PIPES,
     )
     assert warnings == []
-    assert any(
-        "ph-ghost" in p and "eventParams" in p and "toPhaseId" in p for p in problems
-    )
+    assert any("ph-ghost" in p and "eventParams.to_phase_id" in p for p in problems)
 
 
 @pytest.mark.unit
-def test_validate_invalid_event_params_to_phase_id_camel_case():
+def test_validate_event_params_camel_to_phase_id_is_not_honored():
+    """``toPhaseId`` (camel) is not the declared wire spelling (``to_phase_id``).
+
+    GraphQL rejects it at coercion, so the toolkit no longer reads it as the
+    destination phase: it rides ``extra`` untouched and raises no phase problem.
+    """
     behavior = _move_card_behavior()
     del behavior["eventParams"]["to_phase_id"]
     behavior["eventParams"]["toPhaseId"] = "ph-ghost"
@@ -465,7 +468,7 @@ def test_validate_invalid_event_params_to_phase_id_camel_case():
         related_pipe_ids=RELATED_PIPES,
     )
     assert warnings == []
-    assert any("ph-ghost" in p for p in problems)
+    assert not any("ph-ghost" in p for p in problems)
 
 
 @pytest.mark.unit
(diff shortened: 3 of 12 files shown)


adriannoes (changes requested):
The typed `AutomationEventParamsInput` work looks solid and matches #418: declared fields with per-field aliases, `to_phase_id` kept snake, consumers reading typed attributes, and `by_alias=True` on the API dump. Ignoring camel `toPhaseId` inside event params is the right wire alignment.

Before merge, the MCP tool contracts still need to catch up. `validate_ai_agent_behaviors` still documents `eventParams.to_phase_id`/`toPhaseId`, and `create_automation` still says preflight runs on `event_params.to_phase_id` (or `toPhaseId`). After this PR those camel spellings are ignored client-side, so agents that follow the live tool text get a green dry-run (or skip preflight) and then a GraphQL coercion failure, or they never apply the filter they thought they set.

Please drop `/toPhaseId` and `(or toPhaseId)` from the **event_params** side of those two docstrings so they only name snake `to_phase_id` (skills/docs already do). Prefer fixing that here on #464, or once at the stack tip; #466 will ask for the same `create_automation` sentence to also drop `phase.id` on the action side.

Also noted (non-blocking): advisory preflight still fail-opens on a whole-model `ValidationError`, so a bad sibling can skip a valid `to_phase_id`; the camel-ignore preflight unit test is stronger with an `action_params` dest present.

Merge this before #465–#469. `Closes #418` may need a manual close if GitHub does not auto-close into non-default `dev`.

mocha06:
Thanks — applied the event-params half here (`7b2cc8a`):

- `validate_ai_agent_behaviors` docstring: dropped `/toPhaseId`, so it names only snake `eventParams.to_phase_id`.
- `create_automation` docstring: dropped the event `(or toPhaseId)`. Left `(or phase.id)` on the action side untouched here on purpose — at this point in the stack the `action_params.phase.id` fallback still exists (it's removed in #466), so naming it is still accurate. #466 drops that half, so the tip names only the declared snake keys.
- Strengthened `test_validate_ignores_camel_to_phase_id_in_event_params` with a valid snake `action_params.to_phase_id` dest as a control, so it proves the skip is due to the camel event src being ignored, not a missing dest.

On the non-blocking fail-open: I looked at it and kept it deliberately. `validate_traditional_automation_move_transition` parses `event_params` and `action_params` as separate typed models and *fails open* (skips the advisory check, never raises) on a coercion error; Core still rejects the real write. Reading `to_phase_id` independently of its siblings to avoid the skip would mean abandoning the typed parse this PR introduces, to defend an already-malformed advisory input — so I left it as the intended trade-off.

Good call on `Closes #418` — since these merge into `dev` (non-default), I'll close the issues manually as the stack lands.

adriannoes (approved):
Thanks for the follow-up. The event-params docstring cleanup on `validate_ai_agent_behaviors` and `create_automation`, plus the stronger camel-ignore preflight test with a snake dest control, address what we asked for. Leaving `(or phase.id)` on this PR until #466 removes that fallback is a clean stack split, and the fail-open trade-off is clear.

Approve. Appreciate the careful replies and the test hardening.

Issue #422, opened 2026-07-17
Resolve one-off casing dual-reads (portal source, attachment URL)


mocha06:
Part of the "Typed payloads: parse, don't validate" milestone. Follows #416 / PR #417.

## Problem — small, batched one-offs

Remaining casing dual-reads that don't belong to a larger domain:
- `services/portal_service.py`: `source.get("repoId") or source.get("repo_uuid") or source.get("repoUuid")` — a **three-way** alias on a portal element data-lookup source. Introspect the relevant portal element input/output type in-issue and canonicalize.
- `cli/commands/_common.py`: `node.get("fileURL") or node.get("fileUrl")` — response-side attachment URL. Confirm the query's selected field name and read one.

## Explicitly out of scope (not casing aliases — leave as-is)

- `auth/resource_server.py`: `claims.get("azp") or claims.get("client_id") or claims.get("sub")` — JWT claim **precedence**, deliberate.
- `cli/commands/agent.py`: `payload.get("error") or payload.get("message")` — error-message **fallback**, deliberate.
- `mcp/tools/ai_tool_helpers.py`: `b.get("eventId") or b.get("event_id")` — sanctioned lenient fallback in the `_summarize_behaviors` error handler (see #416).

## Acceptance

- The two casing dual-reads resolved to a single declared name each (introspect to confirm which).
- `uv run pytest -m "not integration"` + ruff + import-linter green.

PR #468, opened 2026-07-22
Resolve one-off casing dual-reads (portal source, export URL)

mocha06:
## Motivation

Two remaining one-off casing dual-reads, resolved to the single declared name each after confirming the wire truth. Follows #421.

## What ships

- **Portal `data_sources`**: read the repo id only from the declared Interfaces field `repoId`. The backend `DataSourceInput` declares `repo_id` (wire `repoId`) and `field_keys` (wire `fieldKeys`), both required. The `repo_uuid` / `repoUuid` sugar the normalizer also accepted is dropped (an entry keyed only under those is skipped with the existing SDK warning). Docs, the tool docstring, a stale error hint, and tests are updated to the declared key.
- **Export `fileURL`**: `poll_export_until_done` serves only report exports, whose types declare `fileURL` (introspected: `PipeReportExport.fileURL`). The `or fileUrl` fallback matched a spelling those queries never return (the automation-jobs export, which uses `fileUrl`, has its own path). Read only `fileURL`.

## Behavior boundaries

This is the one user-facing change in the milestone (noted in the CHANGELOG under Changed): portal `data_sources` now accepts only `repoId`. The emitted wire payload (`{repoId, fieldKeys}`) is byte-identical before and after — only the client-side input aliases are narrowed. The `field_keys` / `fieldKeys` sugar the issue did not list is left as-is. Out of scope, deliberately left: the JWT claim-precedence and error-message fallbacks, and the sanctioned lenient summarizer fallback.

## Testing

Offline gates green, incl. a new unit test asserting `repo_uuid` / `repoUuid` entries are skipped and only `repoId` is sent, and a test asserting a `fileUrl`-only export node has no url. This is a client-side alias narrowing with no wire-behavior change, so it is proven by unit tests rather than a live create that would exercise unchanged behavior.

Closes #422.

PR #469, opened 2026-07-22
Survey response-side read-models (parse vs TypedDict)

### LINKED ISSUE #423: Survey response-side read-models (parse vs TypedDict)
Part of the "Typed payloads: parse, don't validate" milestone.

## Goal — survey, not a blanket migration

The response-side `*GraphPayload` / `TypedDict` payloads (`services/types.py`, `services/automation_graphql_types.py`, and the many `*_tool_helpers.py` TypedDicts) describe shapes but validate nothing — consumers still `.get()` into them and re-derive types.

This is a **bounded survey**, not a migration: decide, per boundary, where parsing into a validating model actually pays off (invariants that consumers currently re-check, dual-reads that mask query drift) versus where a TypedDict is fine (pure pass-through to a serializer).

## Deliverable

- A short findings doc: for each response boundary, "parse / leave as TypedDict / align query" with a one-line rationale.
- Spin off follow-up issues only where parsing removes real downstream re-checking. Do **not** convert TypedDicts wholesale — response models that guard nothing are the over-engineering this milestone is meant to avoid.

## Note

Input-side typing (issues in this milestone) is the higher-value work — the API enforces input shapes, so a wrong wire name is a hard failure. Response typing only pays when *our* code re-derives invariants the payload already guarantees.

mocha06:
## Motivation

The response-side `*GraphPayload` / `TypedDict` payloads describe shapes but validate nothing. This is a bounded survey — decide, per boundary, where parsing into a validating model actually pays off versus where a `TypedDict` is fine. Follows #422.

## Deliverable

`docs/response-typing.md` (linked from the docs index) records the decision rule and the per-boundary verdict.

Findings:

- The MCP `*_tool_helpers.py` envelopes are uniformly pass-through serializer shells — they store the GraphQL payload verbatim inside a `{success, message, data}` shell and never walk it. **Leave as `TypedDict`.**
- The SDK read-models in `services/types.py` and `services/automation_graphql_types.py` are consumed pass-through into those envelopes or normalized write-result dicts. **Leave as `TypedDict`.**
- **No response-side casing dual-reads remain** and the query documents agree on key style, so no query alignment is outstanding.
- The single genuine "parse, don't validate" candidate is `AutomationRuleRecord`, re-walked with an `isinstance` ladder in `pipe_config_tool_helpers._automation_mentions_phase` — recommended as an **optional** follow-up (modest payoff; feeds only the delete-preview dependents heuristic).
- The request-side `_behavior_data_source_ids` dual-read (out of a response-side survey's scope) is noted as a separate small follow-up.

Response typing pays off in exactly one place, and only modestly; the higher-value typing work was on the input side, where the API enforces the contract.

## Testing

Doc-only change; offline gates green (no code touched).

Closes #423.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,6 +20,7 @@ Human-facing guides for the **[pipefy/ai-toolkit](https://github.com/pipefy/ai-t
 | [`MIGRATION.md`](MIGRATION.md) | Notes for existing MCP users across packaging changes |
 | [`dependencies.md`](dependencies.md) | Why each runtime dependency exists |
 | [`architecture.md`](architecture.md) | Intra-package layering, type ownership at boundaries, ports, and alternative constructors |
+| [`response-typing.md`](response-typing.md) | When to parse a GraphQL response into a validating model vs. leave it a `TypedDict` |
 | [`ipaas.md`](ipaas.md) | iPaaS (Advanced Automations) tools: meta-tool pattern, flow overview, vocabulary |
 | [`../CONTRIBUTING.md`](../CONTRIBUTING.md) | Contributing skills (Markdown playbooks) |
 | [`../RELEASE.md`](../RELEASE.md) | Versioning and GitHub Releases |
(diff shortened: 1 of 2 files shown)


adriannoes (approved):
The response-typing survey matches the head tree: MCP success/error envelopes stay TypedDict pass-through shells, SDK read-models in `types.py` / `automation_graphql_types.py` are consumed pass-through, and the single modest parse candidate (`AutomationRuleRecord` for the delete-preview heuristic) is correctly framed as optional follow-up. Linking from the docs index is enough.

Approve. Merge last in the stack after #468.

adriannoes (dismissed):
Re-approve after stack rebase. Docs survey still checks out. Thanks for landing the docstring fixes on #464/#466 so the tip is agent-safe.

adriannoes (approved):

PR #495, opened 2026-07-23
refactor(mcp): narrow McpRuntime to per-deployment flags, not the Settings tree

### LINKED ISSUE #405: Narrow McpRuntime.settings so tools can't read process-global config off the request context

## Problem

## Proposed change

## Acceptance criteria

mocha06:
## Motivation
Closes #405.

The #306 import-linter contract ("Tools read no process-global settings") catches settings *imports* from tool-reachable code, but not value *reads* off an object that already holds `Settings`. `McpRuntime` exposed the whole settings tree as a broad public property, and every tool can reach the runtime through its request context (`ctx.request_context.lifespan_context`), so a tool could read `runtime.settings.mcp.<anything>` at call time with no import edge — and `lint-imports` stayed green. This was documented as a known manual-review limit in the package `AGENTS.md`.

The issue described this as a surface that "could" be read; it was already live: `is_remote_profile()` read `runtime.settings.mcp.profile` at request time — a real process-global read off the request context.

## Outcome
`McpRuntime` no longer holds the `Settings` tree. It resolves two narrow per-deployment booleans at construction — `is_remote` and `unified_envelope` — and exposes only those. The startup `unified_envelope` log and `is_remote_profile()` read the booleans instead of the tree.

Tool code reaching the runtime off the lifespan context can no longer read any other process-global value, so the import-linter contract is now a complete guard by construction rather than a partial one backed by manual review. The "known limit" caveat in the `AGENTS.md` process-global section is removed accordingly.

Behavior is unchanged: both booleans are per-deployment facts, identical for every caller, resolved once at startup.


packages/mcp/AGENTS.md
```diff
@@ -262,7 +262,7 @@ A related note: tools that call Pipefy's Internal API (like `delete_card_relatio
...
+**How this is enforced.** An import-linter rule in `pyproject.toml` forbids tool code from *importing* `pipefy_mcp.settings`, even indirectly through a helper (that indirect path is how `unified_envelope` was caught). Every existing import is an explicitly listed, commented exception. If someone adds a new settings import to tool-reachable code, `uv run lint-imports` fails, and it only gets an exception after review confirms it reads a per-deployment value — never a per-user one. The rule sees import edges, not value reads off an object that already holds a `Settings`; that gap is closed by construction rather than review, because the one object a tool can reach through its request context — the `McpRuntime` on the lifespan context — holds no `Settings` tree. It resolves only narrow per-deployment booleans at startup (`is_remote`, `unified_envelope`; see `core/runtime.py`), so there is no settings tree to read off it at call time. Per-user values must come from the request.
```
adriannoes, line 265: This paragraph says the import-linter gap is closed by construction because the lifespan `McpRuntime` holds no Settings tree, then leans into absolute language that tools cannot read process-global settings off that path. The tree removal is right and closes #405. Absolute "any process-global" wording still overstates what remains: other public runtime attrs stay on the object, and reviewed settings imports (`tool_error_envelope`, `graphql_error_helpers`) still read per-deployment knobs at call time.

I would keep the closed-by-construction claim scoped to "no MCP Settings tree on the lifespan runtime" and leave the reviewed-exception inventory explicit.

Worth aligning:

- AGENTS (and the matching `pyproject.toml` / `is_remote_profile` docstring wording) so they no longer claim tools cannot reach any process-global setting or value via the request context
- The accurate claim is "no Settings tree on `McpRuntime`"

packages/mcp/src/pipefy_mcp/core/runtime.py
```diff
@@ -59,10 +59,15 @@ def __init__(
...
-        self._settings = settings
         self._identity = identity
         self.inbound_auth = inbound_auth
         self.transport_security = transport_security
+        # Narrow per-deployment facts resolved at startup. The runtime deliberately
+        # holds no Settings tree: tools reach it off the request context, so
+        # exposing the tree would let tool code read any process-global value at
+        # call time (#405; see the "Process-global configuration" section of AGENTS.md).
+        self.is_remote = settings.mcp.profile == "remote"
+        self.unified_envelope = settings.mcp.unified_envelope
```
adriannoes, line 70: `is_remote` and `unified_envelope` are assigned as plain public attributes. They are per-deployment facts resolved once at startup, and `is_remote` backs `is_remote_profile` (hosted `$env` restriction). Any in-process holder of the lifespan runtime can reassign them after construction.

I would expose them as read-only properties over private attrs (same public names) so the startup snapshot stays frozen.

- Assigning `runtime.is_remote` / `runtime.unified_envelope` from outside `__init__` fails or is otherwise non-writable
- Existing unit asserts on the public names still pass

packages/mcp/tests/core/test_runtime.py
```diff
@@ -70,6 +70,30 @@ def test_construction_needs_no_configured_credential(self, clear_auth_env):
...
+
+        Tools reach the runtime off the request context, so exposing the whole
+        settings tree would let tool code read any process-global value at call
+        time (#405). Only the narrow deployment flags are surfaced.
+        """
+        runtime = McpRuntime(_settings(), RequestScopedIdentity())
+
+        assert runtime.is_remote is False
+        assert runtime.unified_envelope is True
+        assert not hasattr(runtime, "settings")
```
adriannoes, line 87: `assert not hasattr(runtime, "settings")` catches a public `.settings` property coming back, which is the #405 regression that matters most. A check that only restores `self._settings = settings` (no public property) leaves this assert green while tools could still read `lifespan_context._settings`.

I would also assert `not hasattr(runtime, "_settings")` (or pin the intended public surface) so the private store cannot sneak back without a failing test.

```suggestion
        assert not hasattr(runtime, "settings")
        assert not hasattr(runtime, "_settings")
```

You will know it is right when re-adding only `self._settings = settings` fails the test, and a public `.settings` reintroduction still fails.

adriannoes (changes requested):
This closes the #405 lifespan gap in the right place: `McpRuntime` no longer carries a `Settings` tree that tools can poke through the request context, and the two former call sites now read startup-resolved booleans. CI and the local MCP suite are green, and the remote `$env` gate still matches `profile == "remote"`.

A few tightenings would make the contract and the docs match what the code actually guarantees before merge.

### Required before merge

- Soften the absolute process-global wording in `AGENTS.md`, the `pyproject.toml` runtime exception comment, and the `is_remote_profile` docstring. The Settings tree is gone from the lifespan runtime, which closes the import-linter blind spot. Tools can still reach other runtime attrs (for example `ipaas_gateway`) and can still read reviewed per-deployment settings via the existing import exceptions. Prefer "no MCP Settings tree on the lifespan runtime" over "cannot read any process-global setting/value."

- Make `is_remote` and `unified_envelope` read-only (`@property` over private attrs, or equivalent). They are startup-resolved deployment facts that feed remote input restriction; mutable public fields invite accidental reassignment from any code that already holds the lifespan runtime.

- Strengthen `test_exposes_narrow_deployment_flags_not_the_settings_tree` so it also rejects a private `self._settings = settings` reintroduction (for example `assert not hasattr(runtime, "_settings")`). Today `not hasattr(..., "settings")` stays green if only the private store returns.

- Add a short CHANGELOG or `docs/MIGRATION.md` note that `McpRuntime.settings` was removed and embedders should use `is_remote` / `unified_envelope`. The package is beta and the property was not advertised, but the lifespan object is still reachable for anyone embedding `build_pipefy_mcp_server`.

### What worked well

Surgical six-file change, both production callers migrated, and the new unit tests lock the public surface against bringing `.settings` back.

mocha06:
Thanks @adriannoes — addressed in 1200a3f. Three applied, one declined with reasoning:

**✅ Softened the process-global wording.** You're right that "cannot read any process-global setting/value" overstates it — `is_remote` and `unified_envelope` *are* curated per-deployment settings values, and `ipaas_gateway` stays reachable on the runtime. The import-linter comment and the `is_remote_profile` docstring now scope the claim to "no Settings *tree* / no import-free path to the Settings tree," and the comment explicitly notes reviewed per-deployment reads still go through the listed import exceptions. (The `AGENTS.md` paragraph already scoped to "no Settings tree" with the reviewed-exception inventory in the bullet list right above it, so I left it.)

**✅ Test rejects `_settings` reintroduction.** Added `assert not hasattr(runtime, "_settings")` alongside the public-property check — re-adding only `self._settings = settings` now fails.

**✅ CHANGELOG.** Added an `[Unreleased] → Changed` entry noting `McpRuntime` no longer exposes the `Settings` tree. I skipped `docs/MIGRATION.md`: it documents user-facing v0.1 migration, and `McpRuntime.settings` was an internal composition-root attribute — `build_pipefy_mcp_server` returns the `FastMCP` app, not the runtime, so no embedder reached it through the public surface.

**❌ Read-only properties for `is_remote` / `unified_envelope` — declined, happy to revisit.** The class already exposes its other startup-resolved facts as plain public attributes — `inbound_auth` and `transport_security` — and `inbound_auth` (the inbound token verifier pair) is *more* security-load-bearing than `is_remote`. Freezing only the two new booleans would be inconsistent with the class's own convention, and the repo norm (`AGENTS.md` "Type validation at boundaries, not inside") is to trust in-process holders rather than defend against a same-process reassignment no code performs. A consistent "freeze the startup snapshot" would be a class-wide change (all five members) and reads as a separate refactor. If you'd prefer that hardening, I'm glad to do it class-wide in a follow-up — but I don't think singling out these two is the right shape. Let me know if you disagree.

adriannoes (approved):
Thanks for the follow-up in 1200a3f. The wording tightenings, the `_settings` guard, and the CHANGELOG entry land cleanly, and skipping `docs/MIGRATION.md` for an internal composition-root attribute makes sense.

On the read-only properties for `is_remote` / `unified_envelope`: agreed to leave them as plain attrs. That was only a suggestion, and freezing just those two while `inbound_auth` and `transport_security` stay mutable would be inconsistent with the class. A class-wide startup-snapshot freeze can stay a separate follow-up if anyone wants it later.

Looks good to merge.

PR #497, opened 2026-07-23
feat: add list_organizations tool and `pipefy org list`

mocha06:
## Motivation
Closes #383.

The most common onboarding question — "which organizations do I have access to?" — had no matching tool. `get_organization` requires an id you don't yet have, so agents fell back to raw `execute_graphql` and guessed the fields wrong twice (`plan` doesn't exist on `Organization`; `OrganizationMember` has no `id`). Discovery should not require schema spelunking.

## Outcome
Adds `list_organizations` (MCP) and `pipefy org list` (CLI): a read that needs no id and returns the organizations the caller can access — the zero-knowledge entry point for discovery. Each entry carries `id`, `uuid`, `name`, `planName`, `role`, `membersCount`, `pipesCount`, `createdAt`. An empty result is a valid answer (no raise), unlike `get_organization`.

## Proven contract
- Root `organizations` query takes no required args and is scoped to the caller by the API (`accessible_by(current_ability, :show)`) — a per-request identity concern the hosted profile already handles, so the tool is on the remote seed (`meta=REMOTE` + `REMOTE_SEED`).
- Field forms match the schema and the issue's reported gotchas: `planName` (not `plan`), and members carry no `id` (so none is selected). `id`/`uuid` are numeric-id/uuid strings.
- Read-only; governed entirely by API permissions.

## Docs & skills
Parity matrix (+row, count 185→186), README counts, `organization.md` reframed around discovery, introspection skill "Recipe 5" now leads with `list_organizations` (was the exact GraphQL workaround this issue documents), onboarding first-read.

## Testing
Offline: full non-integration suite (3810 passed), `lint-imports`, skills ref lint, version lockstep, ruff/format, CLI help golden regenerated.

Live (CLI, authenticated as a real user against the live API):

| # | Scenario | Expected | Observed | Verdict |
|---|----------|----------|----------|---------|
| 1 | `pipefy org list --json` | accessible orgs with id/uuid/name/planName/role/membersCount/pipesCount/createdAt | 45 orgs, every field populated | ✅ |
| 2 | Test org present | org `300514213` with its uuid/role/counts | uuid `341c1327-…c3970d`, role `super_admin`, 1440 members, 951 pipes | ✅ |
| 3 | Non-JSON render | Rich table, all columns | rendered | ✅ |
| 4 | `planName` (not `plan`); no member `id` selected | matches schema | confirmed in payload | ✅ |
[...]

PR #499, opened 2026-07-23
docs(readme): restructure Installation with a client decision guide

mocha06:
## What

Rewrites the README **Installation** section around the *five ways to use the toolkit*, so a first-time reader can quickly tell which path they want and how the paths differ.

- **Decision guide** — five one-liners mapping client/need → path.
- **Comparison table** — the columns that actually drive the choice: where the MCP server runs, tool surface, auth, what else it installs, and best-for.
- **Reordered subsections** (Hosted MCP → Claude Code plugin → Quick-install → CLI-only → Skills-only), each opening with a **"Pick this when:"** line.
- **Recommendation note** — Claude Code is the recommended, best-tested client; other clients (Cursor, Claude Desktop, Codex) are supported but still maturing.

## Why

The prior section listed install methods as a flat set of subsections with no "which one do I pick?" guidance, and the key differentiator — hosted exposes a remote-safe subset of tools vs. all 185 locally — wasn't surfaced. This was confusing to follow as a first-time user.

## Note

Renaming the install subsection headings changed their GitHub anchors, so this also updates the cross-file links in `packages/mcp/README.md` and `skills/onboarding/pipefy-toolkit-setup/SKILL.md`. Verified no dangling references to the old anchors remain.


README.md
```diff
@@ -47,53 +47,48 @@ Feedback and issues: [GitHub Issues](https://github.com/pipefy/ai-toolkit/issues
...
+- **Terminal, scripting, or CI, with no agent?** → **CLI only**.
+- **Just want the workflow playbooks in any agent?** → **Skills only**.
 
-- **Human OAuth (interactive)**: `pipefy auth login` runs the browser flow and stores a session in your OS keychain. Works anywhere the `pipefy` CLI is on PATH (`uv tool install` once, any client can invoke it). Claude Code additionally exposes it as the `/pipefy:pipefy-login` slash command via the plugin marketplace. Pipe membership is whatever the signed-in user already has.
-- **Service account (unattended / CI)**: provision a Service Account in [Pipefy Admin](https://app.pipefy.com/) (Admin → Service Accounts) and add that account to every pipe the tools should touch. Wire `PIPEFY_SERVICE_ACCOUNT_CLIENT_ID` and `PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET` into the client config below.
+| Install path | MCP server runs on | Tools available | Auth | Also installs | Best for |
+|---|---|---|---|---|---|
+| **[Hosted MCP](#1-hosted-mcp-claude-code)** | Pipefy cloud (HTTPS) | Remote-safe subset (~76 of 185; reads & discovery) | In-client OAuth | nothing else | Fastest start in Claude Code; zero local Python |
```
adriannoes: The Hosted "Tools available" cell says a remote-safe subset of about 76 of 185 tools, framed as reads and discovery. That steers the decision guide above it: read/analysis work to Hosted, writes to the Claude Code plugin.

At this branch tip the remote allowlist is larger and not read-only. `PIPEFY_TOOL_NAMES` has 185 names; `REMOTE_SEED` in `packages/mcp/tests/tools/test_remote_profile.py` has 179. The six withheld tools are local-bound (`create_ai_knowledge_base_document`, `create_llm_provider`, `execute_graphql`, `update_llm_provider`, `upload_attachment_to_card`, `upload_attachment_to_table_record`). The seed already includes create/update/delete tools such as `create_card` and `delete_pipe`.

The prior README only said Hosted exposes the remote-safe surface. The count and the reads-only framing are new in this rewrite, so the comparison table currently teaches the wrong Hosted vs plugin fork.

Done when:

- Hosted tools cell matches the remote allowlist (about **179 of 185**, or "all but the local-bound tools").
- "Reads & discovery" / "read & analysis only" wording is dropped or rewritten so Hosted is not sold as read-only.
- Decision bullets and the Hosted "Pick this when" line differentiate on **no local Python / no CLI or slash commands** (and the six local-bound tools), not on "need writes → plugin".
- Plugin / quick-install rows can still say full **185** locally; if a sibling lands another tool, sync the totals in the same follow-up.

adriannoes (changes requested):
This install restructure is the right shape: five paths, a comparison table, and "Pick this when" subsections make the choice greppable. Anchors and cross-links look consistent, CI is green, and there is nothing security-sensitive in the diff.

Please hold merge on one accuracy issue in the new Hosted row. The table and decision guide claim a remote-safe subset of roughly 76 of 185 tools limited to reads and discovery, and they send write needs to the Claude Code plugin. On this tip the remote seed is 179 of 185 and already includes create/update/delete; only six local-bound tools are withheld. That misstates the main Hosted vs plugin differentiator this PR is trying to teach. Requesting changes until the Hosted cell and the matching bullets match the allowlist (and until "writes require plugin" is reframed around local-only capabilities and CLI/slash ergonomics).

Separate note: if PR #497 merges first and the registry moves 185→186, refresh the totals when you land the wording fix so the guide does not churn twice.

mocha06:
Good catch — confirmed and fixed in `716c17f`. Verified on the branch tip: `REMOTE_SEED` is **179 of 185** (includes `create_card`, `delete_pipe`, etc.); only the 6 local-bound tools are withheld. Updated the Hosted cell to "179 of 185 (all but 6 local-bound tools)", dropped the reads-only framing, and reframed the decision bullets + both "Pick this when" lines around no-local-Python / no-CLI-or-slash-commands / the 6 local-only tools rather than reads-vs-writes.

README.md
adriannoes: Confirmed on `716c17f`: Hosted cell is **179 of 185**, reads-only framing is gone, and the Hosted vs plugin fork is now no-local-Python / CLI-slash / the six local-bound tools. That closes the accuracy issue from this thread.

Still holding approve for the count sync against current `dev` (see the follow-up comment on the PR).

adriannoes:
The Hosted surface fix in `716c17f` looks good and the review thread on that point is resolved.

Before I approve: please rebase (or merge) onto current `dev` and bump the Installation numbers to match post-#497 allowlists. On `dev` today that is **180 of 186** remote-safe (and **186** for the full local rows), not **179 of 185**. The branch is still behind `dev`, so landing as-is would reintroduce a stale count right after merge.

Once the tip shows **180 of 186** (and local rows say **186**) against an up-to-date `dev` base, I am happy to approve.

adriannoes (approved):
Re-reviewed on the current tip after the Hosted surface fix, the merge from `dev`, and the follow-up that describes tool surfaces by rule instead of hard-coded counts.

The decision guide no longer sells Hosted as a small reads-only subset, local rows point at the canonical MCP tool surface, and the withheld set is described by the local-file rule that matches today's remote allowlist. Happy to approve.

Issue #308, opened 2026-06-18
Tool-surface progressive discovery


gbrlcustodio:
Part of the hosted on-behalf-of MCP server proposal (#297).

## Scope

Keep the model-facing working set small. Ship curated default toolsets with on-demand dynamic discovery as the primary mechanism; expose the introspection-plus-execute tools as a search-plus-execute power profile; rely on native client tool-search. The remote allowlist is the floor; this manages the runtime working set within it.

## Acceptance criteria

A hosted deployment exposes a working set under the model's effective selection limit without losing reach; any interim client-side allowlist is removed once server-side discovery lands.

## Dependencies

Builds on the default-deny allowlist (#304).

## Design (settled)

Runtime `notifications/tools/list_changed` discovery is unreliable for our target clients (Claude Code, Claude.ai, ChatGPT do not honor it; the init handshake gives no way to detect support), so we ship two client-independent mechanisms, both applied to the local and remote profiles and layered on the #304 remote floor:

1. A unified `power` discovery profile (about 9 tools): catalog meta-tools (`search_tools`, `describe_tool`, `execute_tool`, `get_tool_categories`) that discover and dispatch to the real, validated tool bodies kept hidden from `tools/list`, plus the raw-GraphQL escape hatch (`search_schema`, `introspect_*`, `execute_graphql`). Keyword search, no new dependencies.
2. Curated named selection at startup (`--toolsets` / `PIPEFY_MCP_TOOLSETS`) across two orthogonal namespaces: 8 subject domains (a complete, disjoint partition of the 153) and overlapping persona profiles, plus the `all`/`default`/`power` keywords. See the taxonomy design comment below for the locked membership and rationale.

Default stays no curation (backward compatible); curation is opt-in. `execute_graphql` becomes remote-safe (`REMOTE_SEED` 19 to 20) so remote `power` has full reach. Per-session `X-MCP-Toolsets` header selection and semantic search are follow-ups.

## Tasks

[...]

PR #501, opened 2026-07-23
Partition the tool surface into subject domains with a drift-guard

mocha06:
## Motivation

The MCP server exposes 187 tools with no machine-checked grouping and no completeness gate — nothing catches a newly registered tool that never gets classified, and there is no vocabulary for selecting a coherent slice of the surface. The hosted profile also lacked full raw-GraphQL reach: `execute_graphql` was withheld from the remote allowlist, so a remote caller could introspect the schema but not run what they found.

## Outcome

A disjoint subject-domain partition of all 187 tools (`workflow`, `database`, `interfaces`, `automation`, `intelligence`, `analytics`, `governance`, `integration`) now lives in `tools/toolsets.py`, enforced by a drift-guard that fails the build if any registered tool has no domain — the repo's first tool-surface completeness gate, keyed to `PIPEFY_TOOL_NAMES` with no hardcoded count. `execute_graphql` is now remote-safe, giving the remote profile full raw-GraphQL reach under the request-scoped bearer's own API permissions — the same trust boundary as its introspection siblings, and its write reach bounded entirely by that caller's API authorization.

Domains are the disjoint *subject* axis; overlapping persona *profiles* are a separate axis, and named `--toolsets` selection plus the meta-tool `power` profile follow in stacked PRs. See the "Subject-domain taxonomy" section in `packages/mcp/AGENTS.md`.

Part of #308.

## Live validation

Real in-process MCP client sessions confirm the default profile exposes all **187** tools — the partition covers the full surface — and `execute_graphql` is present in the exposed set. The partition itself (complete, disjoint, keyed to `PIPEFY_TOOL_NAMES`) is enforced at build time by the drift-guard. The full end-to-end matrix across the stack is on #308.


adriannoes (dismissed):
The subject-domain partition and drift-guard look solid: eight locked domains, completeness keyed to `PIPEFY_TOOL_NAMES` (187/187, no magic count), and tests that actually fail when inverted. Marking `execute_graphql` remote-safe is a minimal, lockstep change (decorator, seed, server assertion) and matches the documented remote-write criteria (API permission is the auth boundary; hosted logging stays key-only).

A few soft notes before or after you merge:

- Update the PR description counts to 187 tools and `REMOTE_SEED` 183 to 184 (the code already matches that).
- Retarget the `CONTEXT.md` / ADR pointers to `packages/mcp/AGENTS.md`, where the taxonomy actually landed.
- Optional: tighten the new remote GraphQL paragraph so it says public GraphQL only, and that filesystem / `$env` / iPaaS paths stay unreachable (public GraphQL equivalents of some withheld tools remain callable).

Stack order stays #501 then #502 then #504 then #505. Live MCP smoke for the remote exposure was blocked here because the connected server was not proven to be this worktree at the frozen head; composition-root tests cover the exposure contract.

What worked well: the registry-keyed completeness gate, the Domain vs Profile split before selection code, and the coordinated REMOTE marker/seed/tests.

adriannoes (approved):
Re-approve after the force-push to `820d6c95`.

The prior review on `1bcae695` was dismissed by the new head. The delta since then is the AGENTS remote-GraphQL wording (public executor only, with the public-equivalent nuance) plus PR description fixes for the 187-tool count and the `packages/mcp/AGENTS.md` pointer. CI is green on this commit.

Same call as before: the subject-domain partition, drift-guard, and `execute_graphql` remote lockstep are good to land.

PR #502, opened 2026-07-23
Add startup toolset selection (--toolsets / PIPEFY_MCP_TOOLSETS)

mocha06:
## Motivation

The subject-domain partition landed as data, but nothing let an operator or a hosted deployment actually narrow the exposed tool surface — every launch advertised all 187 tools (or the full remote floor). Large tool sets degrade model selection accuracy and can exceed client tool caps, which is the whole reason the taxonomy exists.

## Outcome

`--toolsets` / `PIPEFY_MCP_TOOLSETS` now takes a comma-separated list of subject domains (`workflow`, `database`, …) and narrows the exposed surface at startup. Selection runs after the remote floor — floor then selection — so it only ever narrows and can never widen past the remote-safe set; on the remote profile the survivors are the floor intersected with the selection. `all` / `default` or leaving it unset keep every tool, so existing launches are unchanged. An unknown name is a startup usage error that lists the known toolsets.

Part of #308. The overlapping persona profiles and the `power` discovery profile extend this same selection vocabulary in later stacked PRs.

Stacked on #501 (base `rc-dev/feat/tool-surface-taxonomy`); CI fires once that merges and this retargets to `dev`.

## Live validation (in-process MCP sessions → live API)

|---|---|---|
| default (no toolsets) | 187 tools | ✅ |
| `--toolsets database,governance` (resolve + union) | 27 | ✅ |
| `--toolsets bogus` | rejected (unknown toolset) | ✅ |

Floor-then-selection (the remote floor ∩ the selection) is covered by unit tests — booting a `remote`-profile server needs hosted resource-server config, so it's not in the live harness.

---
**Applied the optional follow-up (env exit-2).** `main` now validates the resolved `mcp.toolsets` value (flag merged with `PIPEFY_MCP_TOOLSETS`) and drops the argv-only pre-check, so an env-only bad name gets the same argparse usage error (exit 2) as `--toolsets`, not a build traceback. The server still re-validates at build for embedders that bypass the entry point. New test: `test_unknown_toolset_from_env_is_a_usage_error`.


adriannoes (commented):
This slice does what Task 2 needs. `resolve_selection` plus `apply_toolset_selection` after `apply_remote_profile` keeps selection subtractive, the default stays uncured, and the unit coverage for floor intersection and non-widen is the right kind of proof. Local MCP unit suite is green on the frozen head (full GitHub lint/test still waits on #501 merge and retarget to `dev`, as you noted).

Also noted:

- The live validation row `--toolsets database,admin` → 54 cannot succeed on this head: `admin` is not a domain here (it lands as a persona later in the stack). On this PR that spec is an unknown-toolset error. `database,governance` is 27 tools; default surface is 187, not 186. Worth fixing the table so copy-paste matches this PR.
- Unknown names via `PIPEFY_MCP_TOOLSETS` alone still fail closed with the same message, but as a traceback (exit 1) rather than argparse exit 2. Documented and layer-legal. Prefer validating `resolved.mcp.toolsets` in `main` after `resolve_mcp_settings` and dropping the argv-only pre-check, so MCP client env configs get the same usage-error shape as `--toolsets`.
- Specs that mix `all`/`default` with a domain intentionally mean no curation; a one-line note in `docs/config.md` and `--help` would prevent "default,workflow means narrow to workflow" misreads.

What worked well: floor-then-selection as a composition invariant, settings keeping an opaque string without importing `DOMAINS`, and tests that pin both the intersection and the "cannot re-add floored tools" property.

adriannoes (dismissed):
Re-checked after the rebase onto the updated taxonomy base. The live validation table now matches this head (187 default, `database,governance` → 27, unknown rejected), and `docs/config.md` spells out that mixing `all`/`default` with a domain still means no curation.

The startup selection path itself still looks right: floor then selection, subtractive only, backward-compatible default, and solid unit coverage for the intersection and non-widen cases. Full GitHub lint/test will still land after #501 merges and this retargets to `dev`.

Optional follow-up only: validating the resolved `PIPEFY_MCP_TOOLSETS` value in `main` so env-only bad names get the same argparse usage-error shape as `--toolsets`. Not blocking.

mocha06:
Applied the optional env exit-2 follow-up on `edfcb82`: `main` now validates the resolved `mcp.toolsets` value (the flag merged with `PIPEFY_MCP_TOOLSETS`) and drops the argv-only pre-check, so an env-only bad name gets the same argparse usage error (exit 2) as `--toolsets`, rather than a build traceback. The server still re-validates at build for embedders that bypass this entry point. New test: `test_unknown_toolset_from_env_is_a_usage_error`.

Live check:
```
$ PIPEFY_MCP_TOOLSETS=bogus pipefy-mcp-server
pipefy-mcp-server: error: unknown toolset(s): bogus. Known toolsets: all, analytics, automation, database, default, governance, integration, intelligence, interfaces, workflow.
```

adriannoes (approved):
Thanks for addressing the earlier feedback. The live validation table now matches this head, `docs/config.md` documents the `all`/`default` mix rule, and validating the resolved `mcp.toolsets` value in `main` (with the env-only usage-error test) gives `--toolsets` and `PIPEFY_MCP_TOOLSETS` the same exit-2 shape.

Re-checked on the retarget to `dev`: lint and test are green, floor-then-selection still looks right, and the remaining surface is ready to merge.

PR #503, opened 2026-07-23
feat: hosted-safe attachment upload via file_url

### LINKED ISSUE #305: Hosted-safe attachment path

mocha06:
## Motivation
Phase 2 of the hosted on-behalf-of MCP profile (#297): the attachment upload tools were the last local-file exclusions from the remote surface (#442). `upload_attachment_to_card` / `upload_attachment_to_table_record` accepted only a local `file_path`, which the server reads off its own disk — meaningless on a multi-user hosted server, so both stayed withheld. A field report also showed the same gap bites the local stdio server when the MCP client isn't co-located with the files (claude.ai + connector). This gives the tools a hosted-safe input and marks them remote-safe.

## What ships
A `file_url` input on both attachment tools, alongside the existing `file_path`. The SDK downloads the URL and feeds the bytes into the existing presigned → S3 PUT → store-key pipeline, so the stored value and read-back are unchanged. The download runs behind a layered SSRF guard: HTTPS + public-host only (unless the deployment sets `PIPEFY_ALLOW_INSECURE_URLS`), a standard-port allow-list (80/443), a connect-time re-validation of the resolved host on every request (DNS-rebinding defense), manual redirect following (max 3, re-validated per hop), a 60s timeout, and the same 100 MiB cap as `file_path`, enforced client-side. The fetch and its policy live in the SDK `AttachmentService` (an injected `UrlDownloader`), with `allow_insecure_urls` threaded from settings exactly as the webhook service does; the MCP tool stays free of settings imports.

Both tools now carry `meta=REMOTE` (remote seed 180 → 182). Under the hosted profile `file_path` is rejected per call via `is_remote_profile(ctx)` — the exposure-vs-input-restriction pattern, mirroring `create_ipaas_connection`'s `$env` case — so the hosted surface accepts only `file_url`. Two field-report fixes ride along: the `file_read` error now states the path is read on the machine running the server, and the attachments skill is rewritten from "there is no separate sandbox / no file_url mode" into a topology decision guide.

## Proven contract
`createPresignedUrl`, corroborated by a live run): required `fileName` + `organizationId` (numeric or uuid accepted); the caller needs access to the organization. The value stored on the field is the raw S3 object key (`orgs/<uuid>/uploads/<uuid>/<file>`), never the PUT url or download url — storing the wrong one fails at `updateCardField`, not silently; the download url is minted on read. The 100 MiB cap is enforced client-side (fail-fast, before the presigned request). `file_name` is required-or-inferred from the source basename (URL path or file path); supply it explicitly when a URL has none.

## Behavior boundaries
- local profile: exactly one of `file_path` or `file_url`.
- remote profile: `file_url` only (`file_path` rejected per call).
- CLI: unchanged — local `--file` only (it always runs as the user with a real filesystem).
- No change to the stored value, the read-back download url, or the existing `file_path` behavior and its 100 MiB cap.

## Docs & skills
`docs/mcp/tools/pipes-and-cards.md`, `docs/mcp/tools/database-tables.md`, `docs/parity.md`, the `pipefy-attachments` SKILL rewrite, `packages/mcp/AGENTS.md` (the two "remote-safe file inputs are follow-up work" notes are now resolved), and the CHANGELOG.

## Testing
Offline: full non-integration suite green (3857 passed), ruff, ruff format, `lint-imports` (incl. "Tools read no process-global settings" — the tool imports no settings), `bump_version verify`, skill-ref + frontmatter linters, and the parity/remote drift guards. New unit coverage for the URL source, the SSRF guard (non-HTTPS, private-IP, oversize, redirect re-validation, redirect loop, port allow-list, connect-time rebinding transport), the exactly-one-of validation, and the remote-profile `file_path` rejection.

Live (this branch, production Pipefy, org 300514213 / pipe 303088927):

| 1 | `file_url` upload (HTTPS public) | download → presigned → PUT → attachment on card | 4657 B stored at `…/Python.gitignore`, download_url minted | ✅ |
| 2 | verify via `get_card` | attachment present | field shows the signed URL under the org uuid | ✅ |
| 3 | `file_path` upload (regression) | success | 25 B stored at `…/hosted_safe_e2e.txt` | ✅ |
| 4 | `http://` URL | rejected | `step=download`, "file_url: must use HTTPS (got http://)." | ✅ |
| 5 | non-standard port | rejected pre-fetch | `port 8080 is not allowed (only 80 and 443)` | ✅ |
| 6 | throwaway card upload → delete | delete succeeds | card deleted | ✅ |

Remote-profile exposure + `file_path` rejection + `401` posture are covered by the automated resource-server and remote-profile tests; a real production-OAuth remote upload was deemed unnecessary since the upload call is profile-identical and that plumbing is independently tested.

Closes #305.

## Diagrams

<details>
<summary>One pipeline, two entry points — <code>file_path</code> and <code>file_url</code> converge into the same unchanged flow</summary>

```mermaid
flowchart LR
  A([Agent has a file]):::start
  FP["file_path — read local file<br/>(local profile only)"]:::src
  FU["file_url — download URL<br/>SSRF-guarded, 100 MiB cap"]:::src
  A -->|local only| FP
  A -->|any profile| FU
  FP --> P
  FU --> P
  P["createPresignedUrl<br/>fileName + orgId"]:::step
  S["S3 PUT bytes<br/>presigned url, 300s"]:::step
  K["extract object key<br/>orgs/uuid/uploads/uuid/file"]:::step
  U["updateCardField /<br/>setTableRecordFieldValue — store the key"]:::step
  V(["on read: signed download URL minted"]):::done
  P --> S --> K --> U --> V
  classDef start fill:#e0f2f6,stroke:#0e7490,color:#0f1720;
  classDef src fill:#dcfce7,stroke:#15803d,color:#0f1720;
  classDef step fill:#ffffff,stroke:#0e7490,color:#0f1720;
  classDef done fill:#e0f2f6,stroke:#0e7490,color:#0f1720;
```

On the hosted profile the `file_path` branch is rejected before it runs; `file_url` feeds the same steps.

</details>

<details>
<summary>What guards the URL fetch — five SSRF gates, any failure surfaces as <code>step=download</code></summary>

```mermaid
flowchart TD
  U([file_url]) --> G1{"HTTPS and public host?"}
  G1 -->|no| X["reject · step=download"]:::deny
  G1 -->|yes| G2{"port 80 or 443?"}
  G2 -->|no| X
  G2 -->|yes| G3{"DNS resolves to a public IP?"}
  G3 -->|no| X
  G3 -->|yes| G4{"connect-time recheck<br/>host and port — DNS-rebinding guard"}
  G4 -->|rebound internal| X
  G4 -->|ok| G5{"within 100 MiB?<br/>at most 3 redirects, each re-validated"}
  G5 -->|no| X
  G5 -->|yes| OK([bytes to pipeline]):::ok
  classDef ok fill:#dcfce7,stroke:#15803d,color:#0f1720;
  classDef deny fill:#fee4e2,stroke:#b42318,color:#0f1720;
```

</details>


packages/sdk/src/pipefy_sdk/models/attachment.py
```diff
@@ -87,23 +92,37 @@ class Attachment:
...
     @property
     def name(self) -> str:
-        """Explicit name if provided, else the path's basename."""
-        return self._explicit_name or self._path.name
+        """Explicit name if provided, else the path or URL basename (may be empty)."""
+        if self._explicit_name:
+            return self._explicit_name
+        if self._path is not None:
+            return self._path.name
+        return Path(urlparse(self._url or "").path).name
```
adriannoes: A URL with no path basename (for example `https://cdn.example.com/?id=1`) resolves to `name == ""`. The docstring already says callers must pass `name` in that case, but construction does not enforce it, so `AttachmentService.upload_attachment` can still request a presigned URL with an empty `fileName` and complete the field update. MCP catches this only in `_prepare_attachment`.

I would reject an empty resolved name in `Attachment.__init__` (or a dedicated parse step that returns only a valid `Attachment`), then map `ValueError` to `step=validation` in the MCP shell and drop the `Attachment | dict` special case.

- `Attachment(url="https://cdn.example.com/?id=1")` raises with a clear expected-shape message
- the same URL through MCP still returns soft `step=validation` without calling the client
- an explicit `name="report.pdf"` still uploads successfully

adriannoes (changes requested):
The hosted-safe `file_url` path is in good shape: SSRF-guarded download in the SDK, `meta=REMOTE` with seed lockstep, per-call `file_path` rejection on remote, and soft `step=download` failures without false success on the paths we probed. CI and the narrow attachment suite are green. I would still hold merge on three enablement and contract gaps below.

### What worked well
- Layered SSRF (HTTPS, ports 80/443, public DNS, connect-time revalidation, manual redirects, 100 MiB) with injectable `UrlDownloader` and no settings import in the MCP tools.
- Clear exposure-vs-input-restriction pattern for hosted `file_path`, documented next to the existing `$env` case.
- Primary attachments skill, area MCP docs, and CHANGELOG match the shipped dual-source contract, including intentional CLI `--file` only.

### Required before merge
- Sibling skills `pipefy-pipes-and-cards` and `pipefy-database-tables` still describe attachment upload as local-file / `--file` only. Agents that enter via those inventories will miss `file_url` and the hosted rejection. Update both rows to exactly-one-of `file_path`|`file_url` (hosted: `file_url`) and link to `skills/attachments/pipefy-attachments/SKILL.md`.
- `packages/mcp/AGENTS.md` Distribution model still says SSRF guards, redirect loops, and download size caps are not appropriate here, while this PR ships exactly those controls for `file_url`. Rewrite so local `file_path` stays user-trust and any server-side URL fetch keeps the SDK defenses.
- Empty URL basename (`https://cdn.example.com/?id=1`) still builds `Attachment.name == ""`. MCP rejects that in `_prepare_attachment`, but the SDK upload path can still call `createPresignedUrl` with an empty `fileName`. Enforce a non-empty resolved name on `Attachment` (or a shared parse step) so MCP and SDK share one fail-fast. See the inline on `attachment.py`.

adriannoes (changes requested):
### Also noted
- Tool Args and some docs say HTTP(S) while the default guard is HTTPS-only unless PIPEFY_ALLOW_INSECURE_URLS. Prefer HTTPS-by-default wording at the point of use.
- New remote / file_url MCP tests cover the card tool; the table-record twin duplicates the remote gate without matching tests.
- ctx.debug and some download HTTPStatusError strings can echo a full signed file_url (query tokens). Prefer status-only agent messages and redacted debug.
- Bare int(Content-Length) turns malformed headers into cryptic invalid literal download errors; the export downloader optional_int pattern would be a better fit.
- Stacked #507 should merge after this one (shared attachment surface).

Merge order note: review and land this PR before #507 retargets to dev.

packages/sdk/src/pipefy_sdk/models/attachment.py
mocha06: Done in `3d2cd5b` — enforced at construction. `Attachment.__init__` now resolves the name (explicit → path/URL basename) and raises `ValueError` when it's empty, so the SDK `upload_attachment` path can no longer reach `createPresignedUrl` with an empty `fileName`. The MCP shell drops the `Attachment | dict` special case: it builds the `Attachment` in a `try/except ValueError` and maps to `step=validation`; the CLI wraps construction into a `BadParameter`.

Verified against your checklist:
- `Attachment(url="https://cdn.example.com/?id=1")` raises with an expected-shape message ("no file name … pass name explicitly").
- the same URL through MCP returns soft `step=validation` without calling the client.
- an explicit `name="report.pdf"` still uploads successfully.

mocha06:
Thanks for the thorough review — all points addressed in `3d2cd5b` (force-pushed; #507 rebased on top).

**Required**
- **Sibling skills**: `pipefy-pipes-and-cards` and `pipefy-database-tables` attachment rows now state exactly-one-of `file_path`|`file_url` (hosted: `file_url`) and link to `skills/attachments/pipefy-attachments/SKILL.md`.
- **AGENTS.md Distribution model**: rewritten so a local `file_path` stays user-trust (no guards needed), while a server-side URL fetch keeps the SDK defenses (`HttpxUrlDownloader`) — no longer contradicts the shipped `file_url` controls.
- **Empty URL basename**: enforced at construction (see inline reply); the SDK path can't reach `createPresignedUrl` with an empty `fileName`, and the `Attachment | dict` shell hack is gone.

**Also noted**
- HTTPS-by-default wording at the tool args and docs (http only when the deployment enables insecure URLs).
- The table-record twin now has matching `file_url`-success and remote `file_path`-rejection tests.
- Debug logs strip the URL query, and download failures surface status-only, so a signed source URL's tokens are never echoed.
- `Content-Length` now parses via the shared `optional_int` — a malformed header falls through to the streamed byte-count cap instead of raising a cryptic `int()` error.
- Merge order: agreed — land this before #507 retargets to `dev`.

adriannoes:
Thanks for the follow-up on `3d2cd5b`. The required review items look addressed in the squashed tip (sibling skills, AGENTS Distribution model, empty basename enforced on `Attachment`, plus the also-noted HTTPS wording, table-record twins, URL redaction in debug, and `optional_int` for Content-Length).

One regression remains before I can approve.

In `HttpxUrlDownloader.download`, `raise_for_status()` was replaced with `if response.status_code >= 400` so download failures stay status-only and do not echo a signed `file_url`. That intent is right, but the new gate is too loose: after handling the followed redirect set, any non-error status below 400 is treated as file bytes. Non-followed 3xx responses such as 300 or 304 therefore reach `_read_capped` and can continue into presign / S3 PUT / field update with a successful agent-visible result.

Please require a successful response before reading the body (prefer `status_code == 200`, or at least a 2xx allow-list), and keep the status-only error message for everything else, for example `URL download failed: HTTP {code}.` with no URL in the text.

Once that lands, I am happy to approve.

mocha06:
Good catch — fixed in `7f44c3e`.

You're right that `>= 400` was too loose: `_REDIRECT_STATUS` only follows `{301,302,303,307,308}`, so a non-followed `300`/`304`, or a bodyless/partial `2xx` (`204`/`206`), fell through to `_read_capped` and would upload as file bytes with an agent-visible success.

`HttpxUrlDownloader.download` now requires a plain `200` before reading the body; everything else (after the redirect set is handled) fails status-only:

```python
if response.status_code != 200:
    # Only a plain 200 carries a body to treat as the file:
    # an unfollowed 3xx (300/304), a bodyless 2xx (204/206),
    # or a 4xx/5xx must not upload as file bytes.
    # Status only — never echo the (possibly signed) source URL.
    raise ValueError(f"URL download failed: HTTP {response.status_code}.")
return await self._read_capped(response)
```

The error text is unchanged (`URL download failed: HTTP {code}.`, no URL echoed). New parametrized coverage asserts `300, 304, 204, 206, 404, 500` all raise status-only with no URL in the message. Redirects still `continue` above this gate, so following behavior is unaffected.

#507 has been rebased onto the new tip.

adriannoes (approved):
Thanks for the follow-ups. The earlier required items and the download status gate are in good shape on this tip: `file_url` is hosted-safe with SSRF controls, empty basenames fail at construction, and only a plain HTTP 200 is treated as file bytes (status-only errors, no signed URL echoed). CI is green.

Approved.

adriannoes:
Thanks for working through the review feedback so carefully, @mocha06. The sibling-skill / AGENTS updates, the empty-name enforcement, and the HTTP 200 download gate in particular were clear, well-tested fixes. Appreciate the thorough follow-through.

PR #505, opened 2026-07-24
Add persona profiles for toolset selection

mocha06:
## Motivation

Subject domains group tools by *what they are about*, but a user's real working set is shaped by their role and journey, which cut across domains: running cases spans cards, reads, and reports; building spans pipe config and automation. A persona-shaped selection keeps that working set small without the user naming domains one by one.

## Outcome

Six overlapping persona profiles join the `--toolsets` / `PIPEFY_MCP_TOOLSETS` vocabulary — `requester`, `operator`, `manager`, `builder`, `admin`, `auditor` — each a curated, journey-sized selection grounded in a Pipefy role scope (external guest, pipe member, pipe admin, org super admin, read-only). They resolve and compose exactly like domains (`--toolsets operator`, or `--toolsets database,admin` unioned), narrow after the remote floor like any selection, and are guarded as subsets of the registered tools with names kept disjoint from the domains and keywords. `power` stays a distinct branch.

This completes the #308 tool surface: two orthogonal selection axes — disjoint subject domains and overlapping persona profiles — plus the `power` discovery profile.

The profile set corrects the original 5-persona sketch against Pipefy's actual role model: there is no canonical "manager" role (operator and manager are both org members — the real split is pipe member vs pipe admin), "guest" is three distinct roles (requester maps to external guest), and building is a pipe-admin capability distinct from org super-admin governance. The final six map to those real boundaries; `manager` is kept as a pipe-admin oversight superset of `operator`, and `admin` covers super-admin governance.

Stacked on #504; CI fires once the stack merges down to `dev`.

## Live validation (in-process MCP sessions)

|---|---|---|
| `--toolsets requester` | 8 (== `PROFILES["requester"]`) | ✅ |
| `--toolsets operator` | 39 (== `PROFILES["operator"]`) | ✅ |
| `--toolsets database,admin` (domain + profile union) | 54 | ✅ |

Each profile's subset validity (⊆ registered tools, names disjoint from domains/keywords) is enforced by the drift-guard. The full end-to-end matrix across the stack is on #308.


adriannoes (commented):
This lands the persona axis cleanly: six overlapping PROFILES join the same `--toolsets` / `PIPEFY_MCP_TOOLSETS` resolve path as domains, with subset drift-guards, `operator` contained in `manager`, and `power` kept as a distinct branch. On the current tip, profile sizes are requester 8, operator 39, manager 73, builder 73, admin 36, auditor 76, and `database,admin` still unions to 54. Floor-then-narrow still holds.

What worked well: the planner change stays in `resolve_selection`, the new drift-guards can fail under mutation, and `docs/config.md` was updated in the same change as the vocabulary.

Also noted: `pipefy-mcp-server --help` and the `McpSettings.toolsets` Field still say subject domains only. Runtime and the config table already accept the persona names, so this is discoverability lag rather than a selection bug. Aligning those two strings with `docs/config.md` would close the loop for operators who never open the env docs. Optional polish only; not a merge gate for this PR.

adriannoes (approved):
Approving on the current tip.

The persona axis looks solid: six overlapping PROFILES join the same `--toolsets` / `PIPEFY_MCP_TOOLSETS` path as domains, with subset drift-guards and `power` kept as a distinct branch. The earlier discoverability gap is closed too: `--help` and the `McpSettings.toolsets` Field now name persona profiles (and `power`), matching `docs/config.md`.

Optional follow-ups like encoding `manager` as `operator | oversight` can land later; nothing here needs to block merge of this PR.

mocha06:
The discoverability items are done — `--help` and the `McpSettings.toolsets` Field now name the persona profiles and `power`, matching `docs/config.md`.

On the optional `manager = operator | oversight` suggestion: I'm keeping `manager` as an explicit materialized set for now. It's consistent with how every other domain and profile is defined (explicit and eyeball-reviewable), the `operator ⊆ manager` invariant is already enforced by `test_operator_is_contained_in_manager`, and a compositional definition would couple one profile to another's internals for little gain. Happy to revisit if you feel strongly.

adriannoes (dismissed):
Re-approving on the current tip after the latest rebase.

The persona axis remains in good shape: six overlapping PROFILES on the shared `--toolsets` / `PIPEFY_MCP_TOOLSETS` path, subset drift-guards, discoverability fixed in `--help` and settings, env-resolved unknown names failing as usage errors, and changelog coverage for the vocabulary. Nothing in this PR's own delta needs to block merge.

adriannoes (approved):
LGTM

PR #507, opened 2026-07-24
feat: create_attachment_presigned_url handshake tool

### LINKED ISSUE #506: Hosted local-file attachment upload: create_attachment_presigned_url handshake (+ base64 companion)

## The gap #305 leaves

## Why a presigned handshake

## Scope

## Acceptance criteria

## Dependencies

mocha06:
> Stacked on #503 (`file_url` attachment upload). Base branch is `rc-dev/feat/hosted-safe-attachment-file-url`; review/merge #503 first. CI fires once this retargets to `dev`.

## Motivation
#503 made attachment upload hosted-safe via `file_url`, but that only works when the file is already reachable at a URL the server can fetch. A file that lives only on the caller's disk — or bytes too large to inline — still can't be attached on the hosted profile, because the server never sees those bytes. #506 closes that with a presigned handshake: the server mints an upload target and the client does the upload, so the bytes never pass through the server or the model's context.

## What ships
`create_attachment_presigned_url` on all three surfaces — SDK (`PipefyClient.create_attachment_presigned_url`), MCP tool, and CLI (`pipefy attachment presign`). It calls `createPresignedUrl` and returns `upload_url` (the S3 PUT url), `storage_path` (the object key to store), and `expires_in_seconds` (parsed off the url's `X-Amz-Expires`), transferring no bytes. The caller then PUTs the file to `upload_url` within the expiry and sets an attachment field to `[storage_path]` via `update_card_field` / `set_table_record_field_value`.

The tool is remote-safe (`meta=REMOTE` + `REMOTE_SEED`): it reaches the public API with the request-scoped bearer, reads no filesystem, and moves no bytes through the server. It reuses the SDK's existing presigned request and object-key extraction (the private helper is renamed `_request_presigned_url` to distinguish it from the new public method). Tool count 186 → 187.

The attachments skill gains the handshake's three-step recipe and per-client guidance: code-execution clients PUT the bytes from their own code; clients that support subagents can run the PUT inside one to keep the main context clean; and the claude.ai code-execution sandbox cannot reach the S3 upload host by default (network egress allow-list), so that path needs an org to allow-list the host or falls back to a `file_url` the server can fetch.

## Proven contract
`createPresignedUrl`: required `organizationId` (numeric or uuid) + `fileName`; the caller needs access to the organization. `storage_path` is the raw S3 object key (`orgs/<uuid>/uploads/<uuid>/<file>`) taken from the url's path — store that, never `upload_url` or the download url; the signed download url is minted on read. The presigned url expires in 300s (`X-Amz-Expires`), surfaced as `expires_in_seconds`. No bytes traverse the server; the client performs the PUT.

## Behavior boundaries
- Remote-safe and profile-agnostic: the handshake itself transfers no bytes, so it is exposed on both profiles.
- Whether the client can complete step 2 (the PUT) depends on its environment reaching the upload host — documented per client, including the claude.ai egress caveat.
- `file_name` is required; a blank value is rejected at the tool/CLI boundary. Optional `content_type` / `content_length` are signed into the upload when provided.
- No change to the existing `file_path` / `file_url` upload tools or their 100 MiB cap.

## Docs & skills
`docs/mcp/tools/pipes-and-cards.md`, `docs/parity.md` (new row + 187 count), the `pipefy-attachments` skill (handshake recipe + per-client guidance), README counts, and the CHANGELOG.

## Testing
Offline: full non-integration suite green (3868 passed), ruff, ruff format, `lint-imports`, `bump_version verify`, skill-ref + frontmatter linters, parity/remote drift guards, and the regenerated CLI help golden. New unit coverage across SDK (target shape, missing/pathless url → `presigned_url` step, `X-Amz-Expires` parsing), MCP (success payload, blank `file_name`, presigned failure), and CLI (`presign` success + argument validation).

Live (this branch, production Pipefy, org 300514213 / pipe 303088927):

| 1 | mint presigned target | `upload_url` + `storage_path` + expiry, no bytes | key `…/handshake.txt`, `expires_in_seconds`=300 | ✅ |
| 2 | client PUT to `upload_url` | 200 | HTTP 200 | ✅ |
| 3 | store `storage_path` → verify | attachment on card | present (signed URL under the org uuid) | ✅ |
| 4 | `expires_in_seconds` parsed | 300 | 300 | ✅ |
| 5 | throwaway card upload → delete | delete succeeds | card deleted | ✅ |

## Notes

**base64 companion deferred (not dropped).** #506 framed a decision on whether to also ship a `file_content_base64` in-body mode alongside the handshake. This PR intentionally ships the handshake only: it's the portable, no-size-limit, context-clean path for a file the server can't read. base64 remains the option for the small-file case on clients that can neither reach an upload host nor run code (e.g. claude.ai default egress), and is left as a scoped follow-up to open when that demand is concrete — so `Closes #506` reflects the handshake deliverable, with base64 explicitly deferred rather than silently dropped.

Closes #506.


skills/attachments/pipefy-attachments/SKILL.md
```diff
@@ -124,16 +126,40 @@ To store the attachment under a different name than the source basename, pass
...
+  so even transient handling stays out of the main conversation's context.
+- **claude.ai caveat**: the claude.ai code-execution sandbox cannot reach the S3
+  upload host by default (network egress allow-list). An org owner must allow-list
+  the upload host for the `PUT` to succeed there; until then, this path is blocked
+  on claude.ai and a `file_url` the server can fetch is the alternative.
+
+The CLI equivalent is `pipefy attachment presign` (prints `upload_url` /
+`storage_path` / `expires_in_seconds`); you run the `PUT` and the field update.
 
 ## Success criteria
```
adriannoes, line 169: The new handshake section correctly walks mint → client PUT → store `storage_path`, but the global Success criteria right below still describes a finished `upload_attachment_*` outcome (`download_url`, `content_type`, `file_size`, and the field already listing the file).

Mint returns `success: true` and often a `download_url` without updating any field, so an agent can read this section as "done" and skip steps 2–3.

Please split Success criteria (and adjust Failure modes for client-side PUT) so handshake done-ness is "after steps 1–3 the field lists the file", and one-shot upload keeps the current payload shape.

Done when:

- Success criteria distinguishes `upload_attachment_*` vs `create_attachment_presigned_url`
- Handshake path cannot be satisfied by mint alone

adriannoes (changes requested):
The handshake capability itself looks solid across SDK, MCP, and CLI: remote seed lockstep, soft GraphQL mapped to errors, and the three-step recipe in the skill and tool docstring are clear. Offline attachment tests on this head were green.

One enablement gap should land before merge. The skill elevates `create_attachment_presigned_url` into a first-class path, but the global Success criteria still describes a finished one-shot upload. An agent can treat mint `success` plus `download_url` as done and skip the client PUT and field update. Details are on the inline on that section.

Also noted (non-blocking):

- Mint success payloads also include `download_url` while most one-liners list only `upload_url`, `storage_path`, and expiry. Naming it as not storeable (or omitting it from the handshake surface) would reduce vocab drift.
- When to use / Prerequisites still steer hosted-local bytes toward `file_path` / `file_url` only. A handshake bullet would help skimming agents; the tools table and dedicated section already make the path discoverable.
- This PR is stacked on #503. Merge that first, then retarget here to `dev` before treating CI as full coverage (Skills Lint alone is not the monorepo suite).
- `Closes #506` without a recorded base64 companion decision. Prefer an explicit deferral note or a follow-up issue before that close fires.
- README domain table still has Pipes and cards at 40 (sum 186) while the headline and registry say 187.
- Optional cleanup later: have `upload_attachment` call the new `create_presigned_url` instead of duplicating the mint validate/extract block.

What worked well: thin shells through `PipefyClient`, REMOTE + REMOTE_SEED lockstep, and honest stack/CI framing in the PR body.

skills/attachments/pipefy-attachments/SKILL.md
mocha06: Done in `59ea711` — split the section. Success criteria now distinguishes the one-shot `upload_attachment_*` (payload + the field lists the file = done in one call) from `create_attachment_presigned_url`, where `success: true` only means a target was minted (no field touched) and done-ness is "after steps 1–3 the field lists the file", with an explicit "do not treat the mint response (or its premature signed URL) as a finished upload." Failure modes now notes the client PUT and field update happen outside the tool, so a non-2xx PUT / failed field update surfaces there, not as a `step` on the mint response.

mocha06:
Thanks — addressed in `59ea711` (rebased on the updated #503).

**Enablement gap**: Success criteria split (see inline reply) so the handshake can't be read as done at mint.

**Also noted**
- `download_url` removed from the handshake surface (SDK `PresignedUploadTarget`, MCP payload, CLI output) — it was premature pre-upload and invited storing the wrong value; the signed download URL is minted on read once the field is set. The one-shot `upload_attachment_*` still returns it.
- "When to use" now has a handshake bullet for the "server can't read the file / larger than the in-tool cap" case.
- README domain table fixed: **Pipes & cards 40 → 41**, so the per-domain sum matches the 187 headline/registry.
- `Closes #506` kept, with an explicit **base64-deferral note** added to the PR body (deferred pending concrete demand, not silently dropped).
- Optional `upload_attachment` → `create_presigned_url` consolidation: skipped for now — it would move the extract-failure step from `s3_upload` to `presigned_url` on the core upload path for little gain; flagged as a deliberate defer.
- Stack/CI framing: agreed — CI is full coverage only after this retargets to `dev`.

adriannoes (dismissed):
Approving the handshake work on this head. The enablement gap from the earlier review is addressed (Success criteria split, handshake surface without download_url, When to use / README / base64 deferral).

Please wait to merge until #503 lands first. This PR is still stacked on `rc-dev/feat/hosted-safe-attachment-file-url`. After #503 merges to `dev`, retarget this branch to `dev` so full CI runs, then merge.

adriannoes (approved):
Re-approving after the rebase onto `dev` (#503 merged). Previous approval was dismissed by the force push.

Spot-checked this tip: Success criteria still distinguish one-shot vs handshake (mint is not done), and the handshake surface still omits `download_url`. Same handshake scope as before; full CI should finish on this `dev` base.

Issue #246, opened 2026-05-27
docs: broaden setup.md and audit for honesty before next release


gbrlcustodio:
## Problem

`docs/setup.md` is indexed under "Shared (all packages)" in `docs/README.md`, but the body is ~90% MCP-server-flavored:

- Header reads "first-time install ... and **MCP client** wiring" (line 3).
- "How configuration is loaded" anchors on `pipefy_mcp.settings.Settings` (line 50). The TOML source landed in PR #244 actually lives on `pipefy_auth.AuthSettings` and `pipefy_sdk.PipefySettings`; `pipefy_mcp.settings.Settings` just composes them.
- "Smoke test" runs `uv run pipefy-mcp-server` (line 34).
- The back half is Cursor / Claude Desktop / Claude Code JSON.

The genuinely shared content is Quick start steps 1-3 (install `uv`, clone, copy `.env`). Everything else is surface-specific (MCP).

This surfaced during the PR #244 review (adriannoes' feedback, https://github.com/gbrlcustodio/pipefy-mcp-server/pull/244#pullrequestreview-4374617557): "`docs/setup.md` is where first-time installers land, but the precedence section still reads as *env overrides .env* only." The right fix is not a small note pointing at `docs/config.md`. It is acknowledging that setup.md is currently mis-scoped and broadening it to match its index slot.

## Goals (broadening)

- Rewrite the header and table-of-contents to be surface-neutral.
- "How configuration is loaded" gets rewritten around `AuthSettings` / `PipefySettings` (where the loading stack actually lives), with the full precedence chain `init kwargs > env > .env > config.toml > defaults` and a pointer to `docs/config.md`.
- Demote "MCP client setup" to one of N surface-specific sections. Add equivalent short pointers for CLI (`docs/cli/`) and SDK (`docs/sdk/`).
- `docs/README.md` shared-docs table gets a `config.md` row next to `setup.md` (adriannoes' second suggestion, folded in here).

## Release gate: audit for honesty

`docs/setup.md` must be walked end-to-end against the current code state and confirmed accurate before the next release tag ships. Drift accumulates fast in this doc because it touches env vars, file paths, CLI behavior, MCP client JSON shapes, and the keychain flow simultaneously. The audit checklist:

- [ ] Every `PIPEFY_*` env var listed exists in `.env.example` with the documented default; deprecated aliases match the current code path.
[...]

PR #509, opened 2026-07-28
chore: release v0.4.0-beta.1

### LINKED ISSUE #305: Hosted-safe attachment path

### LINKED ISSUE #308: Tool-surface progressive discovery

### LINKED ISSUE #383: Discovery: cannot list organizations I can access without an org ID
## Problem
There is no dedicated way to answer “which organizations do I have access to?” without already knowing an organization ID. `get_organization` requires an ID.

## Evidence
From live product testing against `main` (Claude Code):
- First question in the session (“quais orgs tenho acesso”) had no matching tool.
- Fell back to `execute_graphql` `{ organizations { ... } }`.
- Wrong fields twice before succeeding: `plan` does not exist on `Organization`; `id` does not exist on `OrganizationMember`. Third attempt used the correct set (e.g. `planName`, member role fields).

## Impact
The most common onboarding / discovery question forces raw GraphQL plus schema guessing.

## Status on `dev`
**MISSING** — registry has `get_organization` only (ID required). Partial workaround: `search_pipes` returns `organizations[].id/name` as a side effect of listing pipes, not a whoami/list-orgs contract.

- New tool (e.g. `list_organizations` / `whoami`) returning id, uuid, name, role, plan, counts with correct field mapping encapsulated
- Enrich an existing surface (`search_pipes` / org docs) so org discovery is explicit without a new tool
- Skill playbook for “how to discover orgs” using current tools + introspection fallback
- CLI parity (`pipefy org list`) if MCP gets a list surface

### LINKED ISSUE #387: Automations: condition shape and operations only discoverable via introspection
## Problem
Automation conditions are not a clear first-class contract on the main create path. Agents must introspect GraphQL types to learn expression shape, AND/OR structure, and allowed operations — and `field_address` semantics (slug vs `internal_id`) stay ambiguous.

## Evidence
From live product testing against `main`:
- Introspected `ConditionInput` and `ConditionExpressionInput` to learn `expressions` + `expressions_structure` (array-of-arrays AND/OR).
- Discovered operations only via schema (`equals`, `not_equals`, `present`, `blank`, `string_contains`, `number_greater_than`, `date_is_after`, …).
- Unclear whether `field_address` wants slug or `internal_id` (slug was used by trial).

## Impact
High friction and error risk for any conditional automation; pairs with missing `condition` on read (#386).

## Status on `dev`
**PARTIAL** — `condition` is first-class on some helpers (`create_send_task_automation`, AI automation create/update, simulate). Generic `create_automation` still pushes condition through `extra_input`. No documented operations enum in automations MCP docs/skill.

- First-class typed `condition` on create/update traditional automations
- Document operations list and `field_address` semantics in tool docstring / docs / skill
- Shared parser/model for condition at the MCP/CLI boundary
- Cross-link with #386 so write and read stay consistent

Related: #386

### LINKED ISSUE #390: Identifiers: slug vs internal_id vs uuid vs numeric id are inconsistent across tools
## Problem
Pipefy identifiers (slug, `internal_id`, uuid, numeric id) are used inconsistently across tools. Agents must learn by trial which ID each tool expects.

## Evidence
From live product testing against `main` (recurring):
- `create_phase_field` returns `id` (slug) and `internal_id`.
- `update_card_field` uses slug.
- AI prompts use `%{internal_id}`.
- Automation `condition.field_address` is ambiguous (slug vs `internal_id`).
- `get_ai_agents` expects `repo_uuid` (pipe uuid) while most tools use numeric pipe id — required an extra `get_pipe` just to obtain the uuid.

## Impact
Extra round-trips, wrong-ID errors, and introspection. Especially painful across cards, fields, AI agents, and automations in one session.

## Status on `dev`
**PARTIAL** — identifier notes exist fragmented across docs/skills (pipes-and-cards, automations-and-ai, observability, AI agents). No single canonical map for “which tool expects which id”.

- Canonical reference doc or skill section mapping tool → expected identifier
- Accept multiple id forms where the API allows, with clear normalization
- Stronger per-tool docstring “Args: expects slug | internal_id | uuid | numeric”
- Align AI/automation field references with the same vocabulary

Related: #387

### LINKED ISSUE #405: Narrow McpRuntime.settings so tools can't read process-global config off the request context

### LINKED ISSUE #506: Hosted local-file attachment upload: create_attachment_presigned_url handshake (+ base64 companion)

mocha06:
Cuts `v0.4.0-beta.1`, capturing everything merged into `dev` since `v0.3.0-beta.1`. Tagging publishes a GitHub Release with wheels and a pre-release PyPI upload.

## What this PR contains

- Lockstep version bump `0.3.0-beta.1` -> `0.4.0-beta.1` across every workspace package's `__init__.py`, the root `pyproject.toml`, `.claude-plugin/plugin.json`, and `uv.lock`.
- `CHANGELOG.md`: `## [Unreleased]` finalized to `## [0.4.0-beta.1] - 2026-07-28`; the Release workflow uses this section as the GitHub Release body.

### Added

- **MCP**: startup tool-surface selection — `--toolsets` / `PIPEFY_MCP_TOOLSETS` narrows the exposed tools to named **subject domains** (`workflow`, `database`, `interfaces`, `automation`, `intelligence`, `analytics`, `governance`, `integration`) and/or overlapping **persona profiles** (`requester`, `operator`, `manager`, `builder`, `admin`, `auditor`), unioned and applied after the remote floor so selection only ever narrows (never widens past the floor). The `power` profile (alias `architect`) instead hides the curated tools behind four catalog meta-tools (`search_tools`, `describe_tool`, `execute_tool`, `get_tool_categories`) alongside the raw-GraphQL tools, keeping the model-facing set at nine regardless of catalog size — `execute_tool` dispatches to each hidden tool through its own argument validation and error envelope, and cannot reach a tool the floor withholds. `all` / `default` or unset keep the full surface (backward-compatible). A build-time drift-guard enforces the disjoint domain partition: every registered tool has exactly one domain, or the build fails. Closes #308.
- **MCP / SDK / CLI**: `create_attachment_presigned_url` (`pipefy attachment presign`) mints a presigned S3 upload target — `upload_url` (the PUT url), `storage_path` (the object key to store), and `expires_in_seconds` — without transferring any bytes, so a client can attach a file the server cannot read (a local file on the hosted profile, or bytes too large to inline): the caller PUTs to `upload_url`, then stores `storage_path` on the field. Remote-safe (reaches the API with the request-scoped bearer, no filesystem, no bytes through the server). Closes #506.
- **MCP / SDK**: attachment uploads (`upload_attachment_to_card`, `upload_attachment_to_table_record`) accept a `file_url` source alongside the local `file_path`. The SDK downloads the URL behind an SSRF guard — HTTPS + public hosts only (unless the deployment sets `PIPEFY_ALLOW_INSECURE_URLS`), a standard-port allow-list (80/443), and a connect-time re-validation of the resolved host on every request (DNS-rebinding defense) — following redirects with per-hop re-validation and the same 100 MiB cap as `file_path`. Both tools are now remote-safe: on the hosted profile `file_path` is rejected per call and `file_url` is the upload path (the CLI stays local `--file` only). The `file_read` error now states the path is read on the machine running the server. Closes #305.
- **MCP / CLI**: `list_organizations` (`pipefy org list`) — list the organizations the caller can access with no id required, the zero-knowledge entry point for discovery. Returns `id`, `uuid`, `name`, `planName`, `role`, `membersCount`, `pipesCount`, `createdAt` per org; an empty result is valid (not an error). Read-only and remote-safe. Closes #383.
- **Docs / skills**: hosted MCP (`mcp.pipefy.com`) install snippet on the root README front door; first-time agent checklist [`pipefy-toolkit-setup`](skills/onboarding/pipefy-toolkit-setup/SKILL.md) (path choice / ask-your-agent / verify — commands stay in `README.md#installation`, per #246).
- **Docs / skills**: canonical [Identifiers map](docs/mcp/tools/identifiers.md) — which tool/argument expects slug vs `internal_id` vs uuid vs numeric id (the five argument names for a pipe, the UUID-named-accepts-numeric exceptions, and a per-area quick reference). The per-area doc Identifiers sections and identifier-heavy skills link into its relevant section. Closes #390.

### Changed

- **MCP**: `execute_graphql` is now remote-safe (public GraphQL only), so the hosted `power` profile has full raw-GraphQL reach under the request-scoped bearer's own API permissions — the same trust boundary as the already-remote-safe introspection tools. Part of #308.
- **MCP (internal)**: `McpRuntime` no longer exposes the resolved `Settings` tree; it surfaces only the per-deployment booleans it needs (`is_remote`, `unified_envelope`). This removes the last import-free path from tool-reachable code (the lifespan context) to process-global settings, so the `#306` import-linter contract guards it by construction. The removed `McpRuntime.settings` was an internal composition-root attribute — embedders using `build_pipefy_mcp_server` are unaffected. Closes #405.
- **MCP / CLI / SDK**: automation `condition` is now a first-class, typed input on `create_automation` and `update_automation` (CLI `--condition`), instead of only riding opaquely inside `extra_input`. A new `ConditionExpressionInput` types the expression interior (`field_address` = field `internal_id`, `operation`, `value`, `structure_id`), the documented operations are exported as `CONDITION_OPERATIONS` (soft enum — any value passes, the API validates), and the `expressions_structure` AND-of-ORs grouping is documented. An explicit `condition` wins over any `condition` in `extra_input`; `pipefy automation update` no longer requires `--extra` (pass `--condition` and/or `--extra`). Closes #387.

### Fixed

- **Installer**: Claude Code next-steps print `/pipefy:pipefy-login` (and the full plugin slash sequence) instead of the stale `/pipefy-login` label.

## After merge


packages/mcp/src/pipefy_mcp/tools/service_account_tools.py
```diff
@@ -0,0 +1,258 @@
...
+                    organization_uuid=organization_uuid.strip(),
+                    service_account_uuid=service_account_uuid.strip(),
+                )
+            except Exception as exc:  # noqa: BLE001
+                return handle_tool_graphql_error(
+                    exc,
+                    "Delete service account failed.",
+                    debug=debug,
+                )
+            return tool_success(
```
adriannoes, line 227: After a confirmed delete, this path always returns `tool_success` with "Service account deleted." It never checks `deleteServiceAccount.success`.

With a soft GraphQL payload `{ "deleteServiceAccount": { "success": false } }`, the agent envelope is still `success: true` while `data.success` is false. Peer MCP deletes already gate on the mutation success flag. Agents will treat a failed delete as done.

I would return `tool_error` (or a soft-fail envelope) when the node is missing or `success` is not true, and add a regression for `success: false`.

Done when:

- soft `success: false` is not reported as deleted
- a unit/session test covers that payload

packages/mcp/src/pipefy_mcp/tools/service_account_tools.py
```diff
@@ -0,0 +1,258 @@
...
+            if memberships is not None:
+                email = (account.get("serviceAccount") or {}).get("email")
+                for entry in memberships:
+                    entry["member"] = (
+                        await service_account_is_member(client, entry["pipe_id"], email)
+                        if email
+                        else None
+                    )
+                data["pipe_memberships"] = memberships
+            return tool_success(
```
adriannoes, line 151: `create_service_account` maps any non-raising client return into `tool_success` with "Store the client secret… shown only once", including when `createServiceAccount` is null/absent or `client.secret` is null.

That breaks the one-shot credential contract: the agent believes provisioning finished and will try to store a secret that was never returned (and cannot be read back later). These tools are already on the remote floor, so hosted agents inherit the same envelope.

I would require a non-null `serviceAccount` with a non-empty `client.secret` (and ideally the token endpoint) before `tool_success`. If the account uuid is present without a secret, fail closed in a way that still surfaces the uuid for cleanup.

Done when:

- null create node / null `serviceAccount` / null secret do not claim success or "shown only once"
- regressions cover those three soft payloads

packages/sdk/src/pipefy_sdk/services/attachment_service.py
```diff
@@ -167,7 +314,7 @@ async def upload_attachment(
         try:
             put_result = await self._s3_uploader.put(
                 url=upload_url,
-                bytes_=file.bytes,
+                bytes_=file_bytes,
```
adriannoes, line 317: This `put` path still treats any HTTP status below 400 as success a few lines later (`if not isinstance(status, int) or status >= 400`).

So 301/302/307/308/399 proceed to field update. httpx does not follow redirects by default, and S3 documents 301/307 for mis-routed PUTs. The agent can see a successful upload while bytes never land on the object key. That matters for the release `file_url` / presign work that funnels through this same pipeline.

I would accept only 2xx (or an explicit 200/201/204 allow-list) and reject 3xx the same way the download path requires HTTP 200. Add a regression for status 307.

Done when:

- 3xx raises `AttachmentUploadError` at `step=s3_upload` and does not update the field
- 200/201/204 still succeed

adriannoes (changes requested):
This release cut looks structurally solid (lockstep version, CI green, remote floor lockstep), but three agent-visible false-success paths need to land before merge. Soft GraphQL and S3 PUT thresholds matter especially because the new service-account writes and attachment upload path are already on the remote/hosted floor.

## Required before merge

- Gate `delete_service_account` on GraphQL `deleteServiceAccount.success` (do not claim deleted when it is false).
- Gate `create_service_account` so null create nodes / missing one-shot secrets never return success with the "shown only once" message.
- Treat S3 PUT success as 2xx only (reject 3xx) so field updates do not run when bytes may not have landed. The gate today is still `status >= 400` a few lines under the `put` call called out inline.
- Add service-account coverage to the `0.4.0-beta.1` CHANGELOG section (that section becomes the GitHub Release body; these tools are first-class remote writes in this cut).

## Also noted

- CLI `automation update --extra '{}'` is accepted while MCP rejects empty `extra_input` (`not extra_input` vs `is None`). Worth aligning so agents get the same "nothing to update" outcome.
- Hosted `file_path` / `file_read` recovery strings only steer to `file_url` and omit `create_attachment_presigned_url`, which this release adds for client-local files on the hosted profile.
- README still advertises Current line `v0.3.0-beta.*` while packages are `0.4.0-beta.1`.
- `PipefyClient.create_service_account` orchestrates create + per-pipe invites + partial `pipe_memberships` on the facade (architecture.mdc expects pure delegation). Fine as follow-up if the soft-success gates land first.
- Typed `condition` rejects `expressions: []` but accepts `expressions: [{}]` (empty object still wins over `extra_input.condition`).
- Point-of-use copy that says "local file on the hosted server" inverts topology (agent/client file under the hosted profile).
- `scripts/release.py publish` does not preflight a stamped `## [version]` CHANGELOG section before the irreversible tag; `prepare` confirm text says it will "cut tag" though tagging is only in `publish`.
- `skills/README.md` tool/domain counts look stale vs the 187-tool / expanded surface.
- Local QA smoke corpus under `.cursor/qa-tests-and-smoke` is still pinned to 182 / `v0.3.0-beta.1`; worth a `v0.4.0-beta.1` delta brief before full release smoke sign-off.

## What worked well

Version lockstep and marketplace/plugin pins are clean. Attachment download SSRF defense and REMOTE_SEED lockstep (including the new writes) look intentional. In-process MCP inventory is 187 local / 184 remote floor. CLI live delta from the review worktree passed for `org list`, attachment presign mint, and automation `condition` reads on PipeClaw. Cursor MCP in this review session was still on a stale 177-tool surface, so MCP-live proof for the new tools was blocked until the client points at this head.

mocha06:
Thanks @adriannoes — all three hold up. I verified each against the code rather than the diff, since they land on `dev` work rather than on this PR's version bump.

Rather than fold fixes into this PR, I've filed them as standalone issues so each stays independently reviewable and this PR stays a pure version stamp — a reviewer should be able to approve it by confirming the diff is nothing but version strings. I'm picking these up right away.

**#510 — service-account tools must not report success on soft GraphQL payloads** (`bug`, `hosted-mcp`)

Covers your first two findings together. They're the same file, adjacent regions, and one test module, so splitting them would guarantee a rebase conflict between the two PRs.

Both confirmed. On the delete: the mutation does select `success` (`queries/service_account_queries.py:46`), so the value is present and simply unread — and you're right that this is an outlier, every peer delete gates on it (`ai_agent_tools.py:607`, `ai_automation_tools.py:293`, `automation_tools.py:759`, `field_condition_tools.py:487`, `knowledge_base_tools.py:311` and `:497`).

On the create: agreed, and there's supporting precedent you didn't cite — `client.py:682-684` already defends against this exact shape on the memberships path (`"Service account email missing from create payload."`), so the payload is already known to come back incomplete. Your "surface the uuid for cleanup" nuance is in the issue as a requirement; it's the part that makes the failure recoverable.

One narrowing, captured as explicit out-of-scope: I'm not gating success on `token { endpoint }`. The client secret is the irreplaceable one-shot value; the token endpoint is a static per-deployment URL, so requiring it adds a failure mode without protecting anything.

**#511 — S3 attachment upload accepts 3xx as success and updates the field** (`bug`)

Confirmed, including the mechanism: `HttpxS3Uploader.put` builds its client without `follow_redirects` (`attachment_service.py:126-129`) and httpx defaults it to `False`, so a 301/307 really does surface as a 3xx and slip through the `>= 400` gate. What convinced me this is an oversight rather than policy is the contrast within the same module — the download path sets `follow_redirects=False` explicitly (`:185`) and `HttpxUrlDownloader` follows redirects manually with per-hop re-validation. Only the PUT leg leaves the outcome unexamined.

One scope correction: `create_attachment_presigned_url` (#506) isn't affected. It mints and returns the presigned target without transferring bytes — the caller does its own PUT, so it never reaches `_s3_uploader.put`. The `file_url` path (#305) does funnel through it, so that half of the concern stands.

**Sequencing**

Both land on `dev` first. Once they're in I'll close this PR and re-cut `v0.4.0-beta.1` off the updated `dev` — the tag was never pushed, so nothing is burned, and the re-cut release PR will be the bare version bump again with both fixes already in the changelog.

Danielmoraisg (approved):

mocha06:
Correction to my earlier summary on this PR: I stated the create mutation has no `success` field. That was wrong — I read it off our query selection rather than the schema. Introspecting the live API shows `CreateServiceAccountPayload.success` exists as a `Boolean!`; `CREATE_SERVICE_ACCOUNT_MUTATION` simply never requested it.

@adriannoes — this strengthens your original finding rather than weakening it. Inferring success from the secret would have left the defect in place for a `success: false` payload that still carried a secret. #512 now selects the field and gates on it, keeping the secret check as a secondary guard. Full detail on #510.

adriannoes:
Thanks for landing [#512](https://github.com/pipefy/ai-toolkit/pull/512) and [#513](https://github.com/pipefy/ai-toolkit/pull/513) on `dev` so quickly. Those close the release blockers from my earlier review, and the sequencing (standalone fixes, keep this PR as a version stamp, then re-cut) is the right call.

I am **not** approving **this** tip. Head is still `9593ea61`, which predates those merges (for example the attachment S3 gate here is still `status >= 400`, while `dev` already requires 2xx). Approving and merging this PR would ship `v0.4.0-beta.1` without those fixes.

Please close or refresh this branch and **re-cut** `v0.4.0-beta.1` from current `dev` (with CHANGELOG notes for the SA soft-payload and S3 2xx fixes). Happy to approve the new cut promptly.

mocha06:
Agreed on all counts @adriannoes — closing this in favour of a fresh cut, per your second option.

You are right that this tip is stale: head `9593ea6` branched from `dev` before any of the fixes landed, so the attachment gate here is still `status >= 400`. Merging it would have shipped `v0.4.0-beta.1` without them.

Four fixes have since landed on `dev`, all traceable to your original review:

- #512 — service-account tools fail closed on soft payloads (your findings 1 and 2)
- #513 — both S3 upload legs require a 2xx (your finding 3, plus the identical bug in the knowledge-base upload that the attachment fix surfaced)
- #515 — the CLI half of the same soft-payload class, which you flagged on #512

Re-cutting `v0.4.0-beta.1` from current `dev` now. The tag was never pushed, so the version is not burned and the new cut reuses it. The new release PR will again be a bare version stamp, with all four fixes already stamped into the changelog section.
(#527)

PR #512, opened 2026-07-28
Fail closed on soft service-account GraphQL payloads

### LINKED ISSUE #510: fix(mcp): service-account tools must not report success on soft GraphQL payloads
## Summary

Both organization service-account write tools (`packages/mcp/src/pipefy_mcp/tools/service_account_tools.py`) report `tool_success` for any GraphQL response that did not raise, including soft payloads that say the operation did not happen. An agent is told the work succeeded when it did not.

Two distinct defects, same file and same defect class, so they are tracked together — they touch adjacent regions and share one test module.

## Defect 1 — `delete_service_account` ignores the `success` flag

`service_account_tools.py:227-230` returns success unconditionally:

```python
return tool_success(
    data=(raw or {}).get("deleteServiceAccount") or {},
    message="Service account deleted.",
)
```

The mutation selects the flag, so the value is present and simply unread (`packages/sdk/src/pipefy_sdk/queries/service_account_queries.py:42-50`):

```graphql
mutation DeleteServiceAccount($input: DeleteServiceAccountInput!) {
    deleteServiceAccount(input: $input) {
        success
    }
}
```

Given `{"deleteServiceAccount": {"success": false}}`, the envelope is `success: true` with the message "Service account deleted." while `data.success` is `false`.

This is an outlier against an otherwise consistent repo convention — every peer delete gates on the flag:

- `tools/ai_agent_tools.py:607`
- `tools/ai_automation_tools.py:293`
- `tools/automation_tools.py:759`
- `tools/field_condition_tools.py:487`
- `tools/knowledge_base_tools.py:311` and `:497`

## Defect 2 — `create_service_account` claims a secret was returned when it was not

`service_account_tools.py:139-163` maps any non-raising return into success:

```python
account = (raw or {}).get("createServiceAccount") or {}
data: dict[str, Any] = dict(account)
...
return tool_success(
    data=data,
    message=(
        "Service account created. Store the client secret and token "
        "endpoint now — they are shown only once."
    ...
```

With a null `createServiceAccount`, a null `serviceAccount`, or a null `client.secret`, `account` collapses to `{}` and the agent is still instructed to store a one-shot secret it never received.

This breaks the one-shot credential contract stated in the query module's own docstring (`service_account_queries.py:7-9`): the create payload returns the OAuth2 client credentials once, and **there is no query to read them back later**. So a false success here is unrecoverable — the account may exist with credentials nobody holds.

The codebase already concedes this payload can come back incomplete. `packages/sdk/src/pipefy_sdk/client.py:682-684` defends against exactly this shape on the memberships path:

```python
entry["error"] = "Service account email missing from create payload."
```

## Why it matters

Both tools carry `meta=REMOTE` and are on the hosted remote-safe floor — they are the first public-GraphQL writes on the remote seed (see `packages/mcp/CLAUDE.md`, "Write tools on the remote profile"). Hosted agents inherit the same envelope, and `create_service_account` is the specific tool whose returned secret is deliberately never logged, so a caller that believes it captured the secret has no second source.

## Proposed fix

**Delete**: return a failure envelope when the payload node is missing or `success` is not `true`. Match the peer convention above rather than inventing a new shape.

**Create**: require a non-null `serviceAccount` with a non-empty `client.secret` before emitting `tool_success` and the "shown only once" message.

Fail **closed but informative**: when the account uuid is present without a usable secret, the account may exist and is otherwise unreachable, so the error must still surface the uuid so the caller can delete it. Do not swallow it into a bare error string.

Explicitly **out of scope**: do not gate success on `token { endpoint }`. The client secret is the irreplaceable one-shot value; the token endpoint is a static per-deployment URL, and requiring it adds a failure mode without protecting anything.

## Done when

- `{"deleteServiceAccount": {"success": false}}` is not reported as deleted
- a missing/null `deleteServiceAccount` node is not reported as deleted
- null `createServiceAccount`, null `serviceAccount`, and null `client.secret` each fail closed and do not emit the "shown only once" message
- the uuid-present-but-no-secret case surfaces the uuid for cleanup
- regressions for each of the payloads above land in `packages/mcp/tests/tools/test_service_account_tools.py`
- the happy paths for both tools still pass unchanged

## References

- Convention reference: `packages/mcp/CLAUDE.md`, "Write tools on the remote profile"

mocha06:
## Motivation

Both organization service-account write tools reported success for any GraphQL response that did not raise, including soft payloads stating the operation did not happen.

`create_service_account` is the sharper of the two. It returns the OAuth2 client secret once and there is no query to read it back, so a false success is unrecoverable: the agent is told to store a secret it never received, for an account that may exist with credentials nobody holds. Both tools sit on the hosted remote-safe floor, so hosted agents inherit the same envelope.

## Outcome

Both tools now gate on the API's own `success` flag, which each payload carries as a non-null `Boolean!`.

`delete_service_account` gates on `deleteServiceAccount.success` — a missing node, a null node, or `success: false` is a failure, matching `delete_card_relation` and every other peer delete.

`create_service_account` gates on `createServiceAccount.success`. That field was never selected by `CREATE_SERVICE_ACCOUNT_MUTATION`, so the tool previously had no way to see it; the mutation now requests it. A soft failure is an error even when a secret rode along in the payload.

A secondary guard still requires a non-null `serviceAccount` carrying a non-empty `client.secret` before the tool claims the one-shot credentials were returned, since the message promises a secret and its absence must fail closed regardless of what the flag says. That failure is closed but informative: when the payload carried an account uuid, the error names it and repeats it at `error.details.service_account_uuid`, so the caller can delete an account that may exist and is otherwise unreachable.

Closes #510.

## Notes

The schema was verified by introspecting the live API rather than inferred from the query. `CreateServiceAccountPayload.success`, `.serviceAccount`, `ServiceAccount.client`, `ServiceAccountClient.secret`, and `DeleteServiceAccountPayload.success` are all non-null, so neither gate can reject a genuine success — there is no happy-path false-negative. Both mutation fields are themselves nullable, so a null payload node is reachable and the defensive null handling is not dead code.

Success is deliberately not gated on `token { endpoint }`. The client secret is the irreplaceable one-shot value; the endpoint is a static per-deployment URL, so requiring it would add a failure mode without protecting anything.

No secret can reach an error: the failure path returns `tool_error`, which has no `data` parameter, so no payload subtree is echoed. A dedicated test pins that, and the soft-failure regression asserts a secret present in a failed payload never appears in the envelope.

Happy paths are unchanged. No tool added or removed — the remote seed, domain partition, parity, and CLI help golden all pass unedited.


packages/mcp/src/pipefy_mcp/tools/service_account_tools.py
```diff
@@ -224,10 +227,35 @@ async def delete_service_account(
...
+    uuid_text = (
+        account_uuid.strip()
+        if isinstance(account_uuid, str) and account_uuid.strip()
+        else None
+    )
+    if uuid_text is None:
+        return tool_error(
+            "Create service account returned no client secret and no account UUID. "
+            "The secret cannot be read back; check the organization for a stray "
+            "account before retrying."
```
adriannoes: The no-uuid soft-fail path tells the agent to "check the organization for a stray account before retrying." In the same doc area we already say there is no list tool and no list API query (`docs/mcp/tools/service-accounts.md` Listing note), so that step is not executable on this surface.

I would replace that sentence with an honest dead-end: the secret cannot be read back, there is no way to inventory org service accounts via MCP or API, and the caller should not retry create until a human confirms in org settings whether a stray account appeared.

- the no-uuid `tool_error` message no longer implies an agent-callable inventory step
- `test_create_service_account_without_secret_fails_closed` (or a sibling) still fails closed without claiming "once"

```suggestion
            "Create service account returned no client secret and no account UUID. "
            "The secret cannot be read back, and there is no API or tool to list "
            "organization service accounts. Do not retry create until a human "
            "confirms in org settings whether a stray account was created."
```

adriannoes (changes requested):
The MCP fail-closed work for soft create and delete looks right, and the regression suite pins the #510 shapes (usable `client.secret` required, delete gated on `success: true`, UUID cleanup details, secret never echoed into errors). Happy paths and the remote seed stay intact.

One agent-facing recovery string needs a quick fix before merge: the no-uuid soft-fail path invents an inventory step this surface cannot perform.

### What worked well
Fail-closed create and delete close the false-success class on the hosted remote floor, with structured `error.details.service_account_uuid` when cleanup is possible, without dumping payload secrets.

### Required before merge
See the inline on `_missing_secret_error` for the no-uuid recovery copy.

### Also noted
The same soft GraphQL create and delete fixtures still exit 0 on `pipefy service-account create|delete` (raw nested payload). That matches this PR's MCP-only #510 scope; worth a follow-up so CLI scripts do not keep the false-success class.

If this lands on `dev` ahead of release PR #509, the beta cut avoids shipping the old soft-success envelopes. `Closes #510` will not auto-close while the merge target is `dev` (default branch is `main`); close the issue when the fix reaches the release line, or do it manually.

mocha06:
Added a commit after live schema introspection: `CreateServiceAccountPayload` carries a non-null `success` field that `CREATE_SERVICE_ACCOUNT_MUTATION` never selected, so the original fix inferred the outcome from the presence of a client secret instead of reading the API's own signal.

That left the reported defect in place in a narrower form — a payload with `success: false` that still carried a secret would have been reported as created. The mutation now selects `success` and the tool gates on it first, with the secret check kept as the secondary guard.

Introspection also confirmed every field the gates read is non-null, so neither gate can reject a genuine success, and that both mutation fields are nullable, so the defensive null handling is reachable rather than dead code. Description updated; the incorrect premise is corrected on #510.

mocha06:
Thanks — applied the required change, and filed the two follow-ups.

**No-uuid recovery copy — applied as suggested.** I verified the premise against the live schema rather than the doc: there is no root `Query` field and no `Organization` field matching service/account, so "check the organization for a stray account" genuinely is not executable on this surface. Your wording is in as-is.

Worth recording for whoever reads this later: that branch is close to unreachable. `ServiceAccount.uuid` is `NON_NULL`, and a null `serviceAccount` null-propagates up to a null `createServiceAccount`, which the `success` gate now catches first. So reaching the no-uuid path needs the API to violate its own non-null contract twice. It stays as defense in depth, and the copy is now honest if it ever does fire.

**CLI blind spot — filed as #514.** Confirmed: `packages/cli/src/pipefy_cli/commands/service_account.py` hands the client result straight to `run_cli_command` for both commands, and every `typer.Exit(2)` in the module is argument validation reached before the call. Nothing inspects the payload, so exit status is 0 either way. Agreed it is out of scope here.

**Auto-close — you are right, and I have left `Closes #510` in place deliberately.** GitHub only auto-closes on merge to the default branch, which is `main`, so merging to `dev` will not fire it. The keyword still rides the commits into `main` at release time via #509, which is when it should close. I will close manually if the release slips.

Also rebased on the merged #513 — the `CHANGELOG` conflict was a both-added `### Fixed` bullet, resolved by keeping both. Full suite green on the merged tree at 3985 passed, with `lint-imports` 2 kept / 0 broken.

adriannoes (approved):
The required no-uuid recovery copy is in place (honest dead-end; no invented inventory step). The create path also gates on the API `success` flag before the secret check, which closes the soft-failure-with-secret hole.

Happy paths, remote seed, and the secret non-echo contract look good. CLI soft-fail exit status is tracked in #514 as a follow-up; leaving `Closes #510` for the `main` release path is fine.

Approving.

PR #513, opened 2026-07-28
Require a 2xx S3 PUT before recording the upload

### LINKED ISSUE #511: fix(sdk): S3 attachment upload accepts 3xx as success and updates the field
## Summary

The S3 upload leg of the attachment pipeline treats every HTTP status below 400 as success, so a 3xx redirect response is read as a completed upload. The field is then updated with a storage path the bytes never reached, and the caller sees success.

## Current behaviour

`packages/sdk/src/pipefy_sdk/services/attachment_service.py:326-334` gates only on `>= 400`:

```python
status = put_result.get("status_code", 0)
if not isinstance(status, int) or status >= 400:
    ...
    raise AttachmentUploadError(...)
```

Redirects are not followed, so a 3xx genuinely surfaces here as a 3xx. `HttpxS3Uploader.put` (`attachment_service.py:126-129`) builds its client without `follow_redirects`, and httpx defaults it to `False`:

```python
async with httpx.AsyncClient(
    timeout=Timeout(timeout=self._timeout_seconds),
) as client:
    response = await client.put(url, content=bytes_, headers=headers)
```

So on a 301/302/307/308 the body is never transferred, `put` returns `{"status_code": 307}`, the `>= 400` check passes, `_extract_storage_path` runs, and the pipeline proceeds to update the card or table-record field.

S3 documents 301 (`PermanentRedirect`) and 307 for requests reaching the wrong regional endpoint, so this is a reachable status on a real misconfiguration, not a hypothetical.

## Why it matters

The failure is silent and lossy: the agent reports a successful attachment, the field holds a storage path, and the object does not exist at that key. Nothing downstream re-checks it.

The rest of this module is deliberate about redirects, which is what makes this leg an oversight rather than a policy:

- the download path sets `follow_redirects=False` explicitly (`attachment_service.py:185`)
- `HttpxUrlDownloader` follows redirects **manually** with per-hop re-validation, precisely so an intermediate URL cannot bypass the SSRF gate (`attachment_service.py:146-162`)

Only the PUT leg leaves the outcome unexamined.

## Scope note

This affects the `file_path` and `file_url` upload sources, which both funnel through `AttachmentService` → `_s3_uploader.put`.

It does **not** affect `create_attachment_presigned_url` (#506). That tool mints and returns a presigned target without transferring bytes; the caller performs its own PUT, so it never reaches this code path.

## Proposed fix

Accept only 2xx (or an explicit `200/201/204` allow-list) and reject 3xx the same way the download path already requires HTTP 200.

Keep the existing `AttachmentUploadError` contract: `step="s3_upload"`, and populate `status_code`. A 3xx has no meaningful error body, so `body_snippet` may stay `None` — do not widen the `>= 400` body-capture branch just to fill it.

## Done when

- a 3xx status raises `AttachmentUploadError` with `step="s3_upload"` and does not update the field
- 200, 201, and 204 still succeed
- 4xx/5xx behaviour is unchanged, including the existing `body_snippet` capture
- regressions land in `packages/sdk/tests/services/test_attachment_service.py`, alongside `test_httpx_s3_uploader_success` and `test_httpx_s3_uploader_forbidden_includes_body_snippet`

## References

- Related: #305 (`file_url` upload source), #506 (presigned-URL handshake — not affected, see scope note)

mocha06:
## Motivation

Both S3 upload legs in the SDK treated every status below 400 as a completed upload, so a 3xx passed the check and the pipeline proceeded on bytes that were never transferred.

On the attachment path the card or table-record field was updated with a storage path pointing at an object that does not exist. On the knowledge-base path `createAiKnowledgeBaseDocument` registered a document against that same never-written object. Both failures are silent and lossy: the caller sees success, the record holds a key, and nothing downstream re-checks.

The status is reachable rather than hypothetical — S3 answers 301 (`PermanentRedirect`) and 307 when a request reaches the wrong regional endpoint. Neither PUT follows redirects, since both go through `HttpxS3Uploader` and `httpx.AsyncClient` defaults `follow_redirects=False`, so the 3xx genuinely surfaces at the gate rather than being resolved beneath it.

## Outcome

Both gates now require a 2xx instead of "not 4xx/5xx". A 3xx fails at `step="s3_upload"` with `status_code` populated, before the field update or the create mutation runs.

Rejecting rather than following is the point: a presigned PUT carries the file bytes, and chasing the `Location` would re-send them to a host the uploader's allow-list only ever checked on the original URL.

Closes #511.

## Notes

Strict 2xx over a `200/201/204` allow-list — a presigned single-object PUT succeeds with 200, and the allow-list's other members belong to different operations. Two independent one-line gates rather than a shared helper, since the modules raise different exception types and two call sites do not pay for the abstraction.

4xx/5xx are unchanged on both paths, including the `body_snippet` capture. A 3xx carries `status_code` with `body_snippet` left `None`, since a redirect has no meaningful error body. Both error contracts are otherwise unchanged, so the MCP and CLI envelopes need no adjustment.

Does not affect `create_attachment_presigned_url` (#506), which mints and returns a target without transferring bytes. SDK-internal throughout — no tool added or removed, no CLI surface change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Fixed
 
 - **Installer**: Claude Code next-steps print `/pipefy:pipefy-login` (and the full plugin slash sequence) instead of the stale `/pipefy-login` label.
+- **SDK**: the attachment upload's S3 PUT now requires a 2xx status instead of merely "not 4xx/5xx", so a 3xx no longer reads as a completed upload. The PUT does not follow redirects (re-sending the body to the `Location` host would ship file bytes to an unvalidated destination), so a wrong-region 301/307 used to pass the check and the card or table-record field was then updated with a storage path the bytes never reached. It now fails as `step=s3_upload` with the status attached, leaving the field untouched. Closes #511.
 
 ## [0.3.0-beta.1] - 2026-07-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Fixed
 
 - **Installer**: Claude Code next-steps print `/pipefy:pipefy-login` (and the full plugin slash sequence) instead of the stale `/pipefy-login` label.
-- **SDK**: the attachment upload's S3 PUT now requires a 2xx status instead of merely "not 4xx/5xx", so a 3xx no longer reads as a completed upload. The PUT does not follow redirects (re-sending the body to the `Location` host would ship file bytes to an unvalidated destination), so a wrong-region 301/307 used to pass the check and the card or table-record field was then updated with a storage path the bytes never reached. It now fails as `step=s3_upload` with the status attached, leaving the field untouched. Closes #511.
+- **SDK**: both S3 upload legs — the attachment pipeline and the knowledge-base document upload — now require a 2xx PUT status instead of merely "not 4xx/5xx", so a 3xx no longer reads as a completed upload. Neither PUT follows redirects (re-sending the body to the `Location` host would ship file bytes to an unvalidated destination), so a wrong-region 301/307 used to pass the check: the card or table-record field was then updated with a storage path the bytes never reached, and the knowledge base registered a document URL for an object that was never written. Both now fail at `step=s3_upload` with the status attached, before the field update or the create mutation runs. Closes #511.
 
 ## [0.3.0-beta.1] - 2026-07-20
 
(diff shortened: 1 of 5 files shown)


adriannoes (approved):
This closes the silent false-success on S3 PUT correctly. Both upload legs now require a 2xx before the attachment field update or the knowledge-base document create, and a 3xx fails at `step=s3_upload` with the field or mutation left untouched.

Rejecting rather than following the redirect is the right call here: the allow-list only ever checked the original URL, and the new uploader test locks that the PUT does not chase `Location`.

What worked well:
- Parametrized 2xx success and non-2xx reject coverage on card, table-record, and KB paths, including assertions that downstream side effects do not run
- Keeping distinct exception types at the two call sites instead of forcing a shared helper
- Leaving MCP and CLI envelopes alone, since the agent-visible `step=s3_upload` contract did not need to change

Local verification on the frozen head was green (unit suite for the touched services, plus a cheap gate-revert litmus that failed the new 3xx cases and passed again after restore). CI lint and test were green as well.

Merge-ready from my side.

PR #515, opened 2026-07-28
Exit non-zero on soft service-account CLI payloads

### LINKED ISSUE #514: fix(cli): service-account create/delete exit 0 on soft GraphQL failure payloads
## Summary

`pipefy service-account create` and `pipefy service-account delete` print the raw GraphQL payload and exit 0 regardless of what that payload says. A soft failure — `success: false`, or a create that returned no client secret — is reported to the caller as a completed operation.

This is the CLI half of #510. That issue and its fix (#512) were deliberately scoped to the MCP tools; the same defect class remains on the CLI surface.

## Current behaviour

Both commands hand the client result straight to `run_cli_command` with no inspection of the payload (`packages/cli/src/pipefy_cli/commands/service_account.py`):

```python
    return await client.create_service_account(...)

```

```python
    return await client.delete_service_account(...)

```

Every `typer.Exit(2)` in the module is argument validation reached before the call. Nothing examines the response.

So `{"deleteServiceAccount": {"success": false}}` prints as a normal result with exit status 0, and a create whose payload carries no `client.secret` prints a payload the caller is expected to harvest credentials from.

## Why it matters

Exit status is the whole contract for a CLI in a script. A shell pipeline that provisions an account and then configures an integration cannot tell a successful create from a soft failure, so it proceeds with no credentials. `create` is the sharper case for the same reason it was on the MCP side: the OAuth2 client secret is returned once and there is no query to read it back, so a false success is unrecoverable.

The MCP tools now gate both operations on the API's own non-null `success` flag (#512). Leaving the CLI ungated means the two surfaces disagree about whether the same payload was a success.

## Proposed fix

Gate both commands on the payload the way the MCP tools do, and exit non-zero when the operation did not happen:

- `delete`: require `deleteServiceAccount.success` to be `true`.
- `create`: require `createServiceAccount.success` to be `true`, then require a non-null `serviceAccount` carrying a non-empty `client.secret` before reporting success.

Worth confirming against the CLI's existing conventions rather than inventing one: check how other commands surface a soft-failure payload and reuse that exit-code and message shape. If no convention exists yet, that choice is the substance of this issue.

Keep the secret out of any error path, as on the MCP side — on failure the account uuid is the useful thing to surface, not the payload.

## Done when

- a `success: false` delete payload exits non-zero and does not read as deleted
- a create payload with no usable `client.secret` exits non-zero and does not present itself as carrying credentials
- the happy paths for both commands are unchanged, including `--json` output shape
- regressions cover both soft payloads

## References

- Raised in review on #512 by @adriannoes
- MCP-side fix: #512, closing #510
- Schema note: both `CreateServiceAccountPayload.success` and `DeleteServiceAccountPayload.success` are non-null `Boolean!`, so the flag is always present to gate on

mocha06:
## Motivation

`pipefy service-account create` and `pipefy service-account delete` printed whatever the mutation returned and exited 0, so a payload stating the operation did not happen read as a completed one.

Exit status is the whole contract for a CLI in a script: a pipeline that provisions an account and then configures an integration could not tell a success from a soft failure, and proceeded with no credentials. `create` is the sharper case — the OAuth2 client secret is returned once with no query to read it back, so a false success is unrecoverable. The MCP tools already gate both operations, so leaving the CLI ungated meant the two surfaces disagreed about the same payload.

## Outcome

Both commands gate on the mutation's non-null `success` flag and exit 1 when it is not true, through the CLI's existing `exit_1_on_unsuccessful` convention rather than a new mechanism.

`create` additionally fails closed when a reported success carries no non-empty `client.secret`, since handing that secret over once is the command's purpose. When the payload still carried an account uuid the failure names it, so an account that may exist with credentials nobody holds can be deleted.

No failure path renders the payload. A soft-failed create can still carry the one-shot secret, and it is never printed — on success it still is.

Closes #514.

## Notes

The `--json` shape gains a top-level `success`, matching the other commands on this convention. That was unavoidable rather than chosen: gating on the create flag requires selecting it, which changes the payload regardless of exit mechanism. The credential path callers read, `createServiceAccount.serviceAccount.client.secret`, is unmoved.

Three existing `create` fixtures asserted exit 0 on payloads the API cannot produce — no `success`, no `client.secret`, both non-null in the schema. They were corrected to realistic payloads; no assertion was relaxed, and one gained an assertion that the secret is printed.

A soft-failed `create` deliberately surfaces no uuid: if the API says the account was not created, naming one invites deleting something that should not exist. `delete` echoes back the uuid the caller passed in, which is not payload-derived.

Success is deliberately not gated on `token { endpoint }` — the secret is the irreplaceable one-shot value; the endpoint is a static per-deployment URL.

No tool added or removed. Tool counts, parity, the remote seed, the domain partition, and the CLI help golden all pass unedited.


adriannoes (approved):
This lands the CLI half of the soft GraphQL false-success gap that #512 closed on MCP. Both `service-account create` and `delete` now gate on the mutation `success` flag through `exit_1_on_unsuccessful`, and create fails closed when a reported success has no usable `client.secret`. Failure paths return a sanitized envelope, so a soft-failed create never prints a one-shot secret that rode along in the payload.

Local contract coverage looks solid: the soft-fail / missing-secret / uuid-cleanup / happy-path cases in `test_service_account_commands.py` pin the acceptance criteria, and a cheap revert litmus on create (dropping `exit_1_on_unsuccessful`) turns those soft-fail asserts red as expected. The joint CLI+MCP suite is green, and the docs/CHANGELOG updates match the shipped behavior.

Also noted:
- `Closes #514` will not auto-close the issue while this PR targets `dev` and the default branch is `main` (same pattern as #512 / #510). Close #514 manually after merge, or rely on a later `main` release PR.
- The feature commit and Notes still read as if this PR newly selected GraphQL `success` on the mutations. That selection already landed in #512; this change only consumes it and splices a top-level CLI `success`. Optional wording cleanup if you touch the Notes again, not a code blocker.

Approve. Merge-ready for `dev`.

PR #527, opened 2026-07-29
chore: release v0.4.0-beta.1

### LINKED ISSUE #305: Hosted-safe attachment path

### LINKED ISSUE #308: Tool-surface progressive discovery

### LINKED ISSUE #383: Discovery: cannot list organizations I can access without an org ID

### LINKED ISSUE #387: Automations: condition shape and operations only discoverable via introspection

### LINKED ISSUE #390: Identifiers: slug vs internal_id vs uuid vs numeric id are inconsistent across tools

### LINKED ISSUE #405: Narrow McpRuntime.settings so tools can't read process-global config off the request context

### LINKED ISSUE #506: Hosted local-file attachment upload: create_attachment_presigned_url handshake (+ base64 companion)

### LINKED ISSUE #510: fix(mcp): service-account tools must not report success on soft GraphQL payloads

### LINKED ISSUE #511: fix(sdk): S3 attachment upload accepts 3xx as success and updates the field

### LINKED ISSUE #514: fix(cli): service-account create/delete exit 0 on soft GraphQL failure payloads

mocha06:
Cuts `v0.4.0-beta.1`, capturing everything merged into `dev` since `v0.3.0-beta.1`. Tagging publishes a GitHub Release with wheels and a pre-release PyPI upload.

## What this PR contains

- Lockstep version bump `0.3.0-beta.1` -> `0.4.0-beta.1` across every workspace package's `__init__.py`, the root `pyproject.toml`, `.claude-plugin/plugin.json`, and `uv.lock`.
- `CHANGELOG.md`: `## [Unreleased]` finalized to `## [0.4.0-beta.1] - 2026-07-29`; the Release workflow uses this section as the GitHub Release body.

### Added

- **MCP**: startup tool-surface selection — `--toolsets` / `PIPEFY_MCP_TOOLSETS` narrows the exposed tools to named **subject domains** (`workflow`, `database`, `interfaces`, `automation`, `intelligence`, `analytics`, `governance`, `integration`) and/or overlapping **persona profiles** (`requester`, `operator`, `manager`, `builder`, `admin`, `auditor`), unioned and applied after the remote floor so selection only ever narrows (never widens past the floor). The `power` profile (alias `architect`) instead hides the curated tools behind four catalog meta-tools (`search_tools`, `describe_tool`, `execute_tool`, `get_tool_categories`) alongside the raw-GraphQL tools, keeping the model-facing set at nine regardless of catalog size — `execute_tool` dispatches to each hidden tool through its own argument validation and error envelope, and cannot reach a tool the floor withholds. `all` / `default` or unset keep the full surface (backward-compatible). A build-time drift-guard enforces the disjoint domain partition: every registered tool has exactly one domain, or the build fails. Closes #308.
- **MCP / SDK / CLI**: `create_attachment_presigned_url` (`pipefy attachment presign`) mints a presigned S3 upload target — `upload_url` (the PUT url), `storage_path` (the object key to store), and `expires_in_seconds` — without transferring any bytes, so a client can attach a file the server cannot read (a local file on the hosted profile, or bytes too large to inline): the caller PUTs to `upload_url`, then stores `storage_path` on the field. Remote-safe (reaches the API with the request-scoped bearer, no filesystem, no bytes through the server). Closes #506.
- **MCP / SDK**: attachment uploads (`upload_attachment_to_card`, `upload_attachment_to_table_record`) accept a `file_url` source alongside the local `file_path`. The SDK downloads the URL behind an SSRF guard — HTTPS + public hosts only (unless the deployment sets `PIPEFY_ALLOW_INSECURE_URLS`), a standard-port allow-list (80/443), and a connect-time re-validation of the resolved host on every request (DNS-rebinding defense) — following redirects with per-hop re-validation and the same 100 MiB cap as `file_path`. Both tools are now remote-safe: on the hosted profile `file_path` is rejected per call and `file_url` is the upload path (the CLI stays local `--file` only). The `file_read` error now states the path is read on the machine running the server. Closes #305.
- **MCP / CLI**: `list_organizations` (`pipefy org list`) — list the organizations the caller can access with no id required, the zero-knowledge entry point for discovery. Returns `id`, `uuid`, `name`, `planName`, `role`, `membersCount`, `pipesCount`, `createdAt` per org; an empty result is valid (not an error). Read-only and remote-safe. Closes #383.
- **Docs / skills**: hosted MCP (`mcp.pipefy.com`) install snippet on the root README front door; first-time agent checklist [`pipefy-toolkit-setup`](skills/onboarding/pipefy-toolkit-setup/SKILL.md) (path choice / ask-your-agent / verify — commands stay in `README.md#installation`, per #246).
- **Docs / skills**: canonical [Identifiers map](docs/mcp/tools/identifiers.md) — which tool/argument expects slug vs `internal_id` vs uuid vs numeric id (the five argument names for a pipe, the UUID-named-accepts-numeric exceptions, and a per-area quick reference). The per-area doc Identifiers sections and identifier-heavy skills link into its relevant section. Closes #390.

### Changed

- **MCP**: `execute_graphql` is now remote-safe (public GraphQL only), so the hosted `power` profile has full raw-GraphQL reach under the request-scoped bearer's own API permissions — the same trust boundary as the already-remote-safe introspection tools. Part of #308.
- **MCP (internal)**: `McpRuntime` no longer exposes the resolved `Settings` tree; it surfaces only the per-deployment booleans it needs (`is_remote`, `unified_envelope`). This removes the last import-free path from tool-reachable code (the lifespan context) to process-global settings, so the `#306` import-linter contract guards it by construction. The removed `McpRuntime.settings` was an internal composition-root attribute — embedders using `build_pipefy_mcp_server` are unaffected. Closes #405.
- **MCP / CLI / SDK**: automation `condition` is now a first-class, typed input on `create_automation` and `update_automation` (CLI `--condition`), instead of only riding opaquely inside `extra_input`. A new `ConditionExpressionInput` types the expression interior (`field_address` = field `internal_id`, `operation`, `value`, `structure_id`), the documented operations are exported as `CONDITION_OPERATIONS` (soft enum — any value passes, the API validates), and the `expressions_structure` AND-of-ORs grouping is documented. An explicit `condition` wins over any `condition` in `extra_input`; `pipefy automation update` no longer requires `--extra` (pass `--condition` and/or `--extra`). Closes #387.
- **SDK**: a GraphQL response carrying `errors` now raises `PipefyGraphQLError` (exported from `pipefy_sdk`) for every endpoint, replacing the gql `TransportQueryError` the public and Interfaces endpoints leaked and the `ValueError` carrying a `[code=…] [correlation_id=…]` string that the Internal API endpoint raised. The exception carries the raw structured `errors` list, so codes and correlation ids are read off `extensions` instead of parsed back out of a formatted message. Embedders catching `gql.transport.exceptions.TransportQueryError` around SDK calls must catch `PipefyGraphQLError` instead; genuine transport failures (non-2xx, timeouts) still raise gql `TransportError` subclasses unchanged. Part of #354.
- **CLI**: a non-permission GraphQL error from an Internal API command (the `portal sub-portal` group) exits 1 with a single-line `message (code)`, instead of exiting 2. Exit 2 is the usage-error code, so a server-side failure claiming it was wrong; this aligns Internal API command failures with public API ones. Permission errors are unchanged. Part of #354.

### Fixed

- **Installer**: Claude Code next-steps print `/pipefy:pipefy-login` (and the full plugin slash sequence) instead of the stale `/pipefy-login` label.
- **SDK**: both S3 upload legs — the attachment pipeline and the knowledge-base document upload — now require a 2xx PUT status instead of merely "not 4xx/5xx", so a 3xx no longer reads as a completed upload. Neither PUT follows redirects (re-sending the body to the `Location` host would ship file bytes to an unvalidated destination), so a wrong-region 301/307 used to pass the check: the card or table-record field was then updated with a storage path the bytes never reached, and the knowledge base registered a document URL for an object that was never written. Both now fail at `step=s3_upload` with the status attached, before the field update or the create mutation runs. Closes #511.
- **MCP**: the service-account tools no longer report success for a GraphQL payload that says the operation did not happen. `delete_service_account` gates on `deleteServiceAccount.success` (a missing node or `success: false` is now a failure, matching every peer delete), and `create_service_account` now selects `createServiceAccount.success` (previously unrequested) and gates on it, then additionally requires a non-null `serviceAccount` carrying a non-empty `client.secret` before it claims the one-shot credentials were returned — when the account UUID arrives without a usable secret the failure surfaces that UUID (in the message and in `error.details.service_account_uuid`) so the caller can delete the unreachable account, and no payload value is echoed into the error. Closes #510.
- **CLI**: `pipefy service-account create` and `pipefy service-account delete` no longer exit 0 on a GraphQL payload that says the operation did not happen. Both now gate on the mutation's non-null `success` flag and exit 1 when it is not true, joining the `exit_1_on_unsuccessful` convention that splices a top-level `success` onto the rendered result. `create` additionally fails closed when a reported success carries no non-empty `client.secret`: the one-shot secret cannot be read back, so a create that returns without it is unusable, and the failure surfaces the account UUID (when the payload carried one) so the stray account can be deleted. No failure path renders the payload, so a secret riding along in a failed create is never printed. Closes #514.

## After merge


mocha06:
Re-cut of #509 from current `dev` (`4d5180c`), per @adriannoes's review there.

The previous cut branched from `dev` before the fixes landed and would have shipped `v0.4.0-beta.1` without them. That branch is closed and deleted; the tag was never pushed, so the version is reused rather than burned.

The stamped `## [0.4.0-beta.1]` section now carries the four fixes that came out of the review on #509:

- `Closes #510` — service-account tools fail closed on soft GraphQL payloads (#512)
- `Closes #511` — both S3 upload legs require a 2xx before recording the upload (#513)
- `Closes #514` — the CLI half of the same soft-payload class (#515)

plus everything else that reached `dev` since, for 15 entries in the section.

The diff is 14 version-bearing files and the two-line changelog stamp — no source change. `bump_version.py verify` passes, so the five `__version__` strings, the root `pyproject.toml`, `uv.lock`, both plugin manifests, and the sibling `==` pins all agree at `0.4.0-beta.1`.

Note the closing keywords fire when this reaches `main`, not on the `dev` merges — `dev` is not the default branch.

Danielmoraisg (approved):
LGTM

adriannoes (approved):
This re-cut is the right follow-through. Closing the earlier tip and shipping `v0.4.0-beta.1` from current `dev` means the release stamp now includes the soft GraphQL fail-closed paths on the service-account tools and the 2xx S3 PUT gates on both upload legs, instead of publishing the beta without them.

Credit to @mocha06 for treating the review feedback as independent, reviewable fixes (#512 / #513, plus the CLI twin), landing them on `dev` first, and re-cutting a clean version bump rather than folding hotfixes into the stamp PR. That sequencing kept the release review honest and got the blockers addressed without burning a tag.

Verified on this tip (`7575dc58`): create/delete service-account tools gate on mutation `success` and the one-shot secret, attachment and knowledge-base uploads require `200 <= status < 300`, CHANGELOG Fixed covers #510 / #511 / #514, and the lockstep bump to `0.4.0-beta.1` is in place. CI is green.

Merge-ready from my side. After merge, the human-gated `scripts/release.py publish` from `main` remains the irreversible step.

Issue #545, opened 2026-07-30
Phase 0: de-risk the mcp 2.x migration while staying on 1.x


Danielmoraisg:
Execution issue for the phase 0 checklist decided in #543. Phase 0 ships on its own and is
useful whether or not we ever cross the 2.x boundary: it retires a version lag, and it funnels
the two mechanical patterns that would otherwise dominate the phase 3 diff into single files.

Phases 1 through 4 stay out of scope here. The revisit trigger for those is in #543 (`2.0.2` or
later published, or a concrete product need for a `2026-07-28` capability).

## Tasks

- [x] Move the pin from `>=1.25.0,<2` to `>=1.29.0,<2` and re-lock. Also corrects the stale
      rationale comment at `packages/mcp/pyproject.toml`, which justified the upper bound by our
      middleware wrapping a FastMCP internal. 2.x offers a supported `ServerMiddleware` seam that
      would replace that wrap, so the bound holds for a different reason: 2.0 removes the
      `mcp.server.fastmcp` module the server imports at startup. Done in #546.
- [ ] Add `packages/mcp/tests/_mcp_compat.py` re-exporting
      `create_connected_server_and_client_session`, and route the 29 test files (44 call sites)
      through it. On 1.x this is a one-line re-export. At phase 3 it becomes the shim for the
      removed helper, so the 44 sites collapse to one file to edit.
- [ ] Normalize the repeated `assert result.isError is False` sites onto a conftest helper. The
      wire fields move camelCase to snake_case in 2.0, and this is test-only churn (zero src
      sites are affected, since all the src uses are alias-compatible kwargs). Doing it now keeps
      the phase 3 diff free of mechanical noise.

The last two are held back from #546 deliberately: each touches roughly 30 test files, and the
AGENTS.md guidance (split PRs over 10 files or 300 changed lines) argues against bundling them with
a pin bump. Open question on #546 is whether to keep that split.

## Verification for each task

`uv run pytest -m "not integration"` plus `cd packages/mcp && uv run lint-imports`. No behavior
change is expected from any of the three, so a green suite is the whole contract.

## Task 1 result

4004 passed, 5 skipped, 34 deselected at `mcp==1.29.0`, and both `packages/mcp` import-linter
contracts kept. The internals this package patches, verified present at 1.29.0:

| Internal | Used by | State at 1.29.0 |
| --- | --- | --- |
| `mcp.server.fastmcp.FastMCP` | `server.py` composition root | present |
[...]

Issue #543, opened 2026-07-30
Decision: stay on mcp 1.x for now, with a scoped 2.x migration plan


mocha06:
## Decision

**Stay on `mcp` 1.x for now. Do phase 0 (below) immediately. Revisit the 2.x migration when `2.0.2`
or later is published, or when a concrete product need for a `2026-07-28` capability appears.**

The migration is fully scoped across five phases, with **no blockers**.
This issue records the scoping, the decision, and the phase 0 work that is worth doing regardless.

## Background

#297 proposed the hosted on-behalf-of profile, and #308 shipped toolset selection under it
(`--toolsets` / `PIPEFY_MCP_TOOLSETS`, subject domains and persona profiles, plus the `power`
discovery profile). The MCP `2026-07-28` spec revision then published, which raised a direct question:
does the new revision, or our pinned SDK version, change any of that? Scoping that question produced
this decision.

The corrections below are relative to a first-pass reading of the 2.0.0 release notes and the spec
announcement. They are recorded because that first pass is the reading anyone would arrive at from the
same public sources, and several parts of it are wrong.

## Why this was a decision and not a pin

We pin `mcp[cli]>=1.25.0,<2`, with `uv.lock` at `1.25.0`. The cap was added because 2.0.0 drops the
`mcp.server.fastmcp` module the server imports at startup, so an uncapped install failed with
`ModuleNotFoundError` before serving anything.

Two facts turned that into a fork:

- The `2026-07-28` spec revision exists **only** in 2.0.0. The string appears nowhere in the 1.25.0 or
  1.29.0 trees, so no 1.x release will ever carry it.
- The 1.x line is frozen at `2025-11-25`.

## Finding that shaped everything: the two revisions are separable

One 2.0.0 install serves both revisions concurrently, routing on the `mcp-protocol-version` header
(`streamable_http_manager.py:180-190`): present and not in `HANDSHAKE_PROTOCOL_VERSIONS` goes to the
modern handler, otherwise the legacy leg.

Two sharpeners, both verified by running a client against a real 2.0.0 install:

- **`2025-11-25` is the default, not a fallback.** Upgrading the SDK moves no client onto the new
  revision.
- **No single release has both `FastMCP` and `MCPServer`.** 1.29.0 has one, 2.0.0 the other, and 1.29.0
  tops out at `2025-11-25`. There is no intermediate release where both APIs coexist, so the swap
  itself is atomic.

[...]

PR #546, opened 2026-07-30
chore(mcp): move the mcp SDK floor to 1.29.0 [#543 Phase 0]

Danielmoraisg:
Completes task 1 of #545 (phase 0 of the #543 decision). #545 stays open for tasks 2 and 3.
Depends on nothing; revert is a pin bump.

## Summary

Moves `mcp[cli]` from `>=1.25.0,<2` to `>=1.29.0,<2` and re-locks (`uv.lock` 1.25.0 to 1.29.0), so
the package tracks the head of the 1.x maintenance line instead of trailing it by four minor
versions. No logic changes: the only source edits are the docstrings recording the tested-against
version. Two caller-visible SDK behaviors do arrive with the bump, a 4 MiB `413` and a moved
`serverInfo.version`; both are recorded under "Two behavior changes this bump introduces" below and
in the CHANGELOG.

The value is in what it re-tests. This package patches three SDK internals, and until now they had
only ever been exercised against 1.25.0 (`tools/validation_envelope.py` said so in its docstring).
Four minor versions of unverified drift is the wrong thing to discover inside the 2.x port, where
every other symptom will also be a suspect. All of them are intact at 1.29.0:

| Internal | Used by | State at 1.29.0 |
| --- | --- | --- |
| `mcp.server.fastmcp.FastMCP` | `server.py` composition root | present |
| `app._mcp_server.request_handlers[CallToolRequest]` | `core/tool_middleware.py` install | present |
| `app._tool_manager` | `core/fastmcp_tool_lifecycle.py`, `tools/registry.py` | present |
| `Tool.run(arguments, context, convert_result)` | `PipefyValidationTool.run` override | signature unchanged |
| `mcp.server.lowlevel.server.request_ctx` | `tests/core/test_tool_middleware.py` | present |
| `mcp.shared.memory.create_connected_server_and_client_session` | 29 test modules | present |

The three places that recorded the tested-against version (`core/tool_middleware.py`,
`tools/validation_envelope.py`, `packages/mcp/AGENTS.md`) move to 1.29.0 with it, since that claim
is the reason for the bump.

Also corrects the rationale comment on the upper bound (the fourth phase 0 task, same lines). It
justified `<2` by our tool-call middleware wrapping a FastMCP internal, which reads backwards now
that 2.x ships a supported `ServerMiddleware` seam that would *replace* that wrap. The bound holds
for the reason #529 added it: 2.0 removes the `mcp.server.fastmcp` module the server imports at
startup, so an uncapped install fails before it serves anything.

## What is in the 1.26 to 1.29 range

Relevant to this package, from the upstream v1.x release notes:

- 1.26.0: unknown Streamable HTTP session ids return 404 instead of 400.
- 1.27.2: `AccessToken` gains `subject` and `claims`, and transport sessions bind to the
  authenticated principal. Worth noting for the hosted profile, where the tool-call middleware's
  `ctx.identity` deliberately omits the end-user subject until a consumer exists. Nothing changes
  here; it means the field is available when per-user quotas want it.
- 1.28.0: `DeprecationWarning` on the WebSocket transport and the experimental tasks API. This
  package uses neither, so no warning filters were needed.
- 1.29.0: Streamable HTTP request body limits, and tool-name validation rejects a trailing newline.
  Every registered tool name is a Python identifier, so the stricter validation is inert.

## Testing performed

```
4004 passed, 5 skipped, 34 deselected in 148.25s

cd packages/mcp && uv run lint-imports
Contracts: 2 kept, 0 broken.

uv run ruff check packages/mcp/src && uv run ruff format --check .
All checks passed. 404 files already formatted.
```

Runtime check that the patched internals resolve, rather than only that the imports succeed:

```
uv run python -c "from mcp.server.fastmcp import FastMCP, Context; \
from mcp.server.lowlevel.server import request_ctx; \
from mcp.shared.memory import create_connected_server_and_client_session; \
from mcp.server.fastmcp.tools.base import Tool; import inspect; print(inspect.signature(Tool.run))"
(self, arguments: dict[str, Any], context: Context[ServerSessionT, LifespanContextT, RequestT] | None = None, convert_result: bool = False) -> Any
```

Resolved versions on Python 3.11: `mcp 1.29.0`, `starlette 0.46.1`, `httpx 0.28.1`.

## Live verification

The suite above runs against mocked transports, which cannot show whether the patched SDK internals
still take effect end to end. Three parallel runs covered that, targeted at the SDK-version risk
rather than touring the tool surface. Every seam this package patches was confirmed live at 1.29.0,
each with a control that would have caught a silent no-op:

- **`Tool.run` validation patch.** A malformed-argument call returns the canonical
  `{"success": false, "error": {"code": "INVALID_ARGUMENTS", ...}}` envelope with no `pydantic.dev`
  URL or internal class name. Negative control: in a process without the patch installed, vanilla
  FastMCP 1.29.0 returns `isError: True` carrying
  `1 validation error for needs_intArguments ... https://errors.pydantic.dev/2.13/v/int_parsing`.
  So the check passes because the patch ran, not because the SDK changed.
- **`request_handlers[CallToolRequest]` wrap.** The composed handler carries
  `__pipefy_tool_call_chain__`, and a real HTTP tool call emitted the `tool_call` structured line
  (`{"event":"tool_call","tool":"get_pipe","outcome":"ok","arg_keys":["pipe_id"],...}`) whose
  `request_id` matches its `http_request` line. This is the claim `packages/mcp/AGENTS.md` pins to
  the version, verified over the wire.
- **`_tool_manager` removal.** Count deltas are dispositive: 187 registered, 9 under
  `--toolsets power`, 184 under the remote floor (set-equal to `REMOTE_SEED`, withholding exactly
  `create_ai_knowledge_base_document`, `create_llm_provider`, `update_llm_provider`). If
  `_tool_manager` had moved, the removal helper logs a warning and returns, leaving 187.

Transport and surface: stdio via the real console entry point and Streamable HTTP both negotiate
`2025-11-25` and expose the same 187 tools. Structured logging holds its privacy bound under
planted sentinels (a query-string token, an `Authorization` header, and an argument value are all
absent from the emitted lines; `arg_keys` carries key names only). The 421 allowlist, the
`PIPEFY_MCP_ALLOWED_HOSTS` widening, and the non-loopback bind refusal all behave as documented
(the bind refusal was probed with TEST-NET-3, so a broken guard could not have produced a reachable
bind).

Write path against a real org: a scratch pipe, card, phase field, field update, phase move, and
comment, each mutation read back rather than trusted from its own response, then the two-step
`confirm=False` / `confirm=True` destructive guard (with a read between the steps proving the
preview mutated nothing), then deletion of everything created. Nothing was left behind.

## Two behavior changes this bump introduces

Neither is a defect, and both are caller-visible, so they belong in review rather than in a
changelog footnote:

1. **`serverInfo.version` now advertises `1.29.0`** where it advertised `1.25.0`. The lowlevel SDK
   resolves `server_version=self.version if self.version else pkg_version("mcp")`, and FastMCP at
   1.x neither passes a version nor accepts one, so the advertised string tracks the SDK and never
   reflects `pipefy_mcp.__version__` (`0.4.0-beta.1`). Not fixable at 1.x through the FastMCP API;
   `MCPServer.__init__` in 2.0 does take `version`, so it becomes fixable during the port.
2. **A request body over 4 MiB now returns a plain-text `413`** instead of a JSON-RPC error. 1.29.0
   added `DEFAULT_MAX_REQUEST_BODY_SIZE`; 1.25.0 had no limit. Verified by bracketing: 4194304 bytes
   passes, 4194305 returns `413 Request body too large`. Attachments cannot reach it, since none of
   the 187 schemas has a bytes or base64 parameter and uploads go through `file_path`, `file_url`,
   or a presigned URL. The exposed paths are large free-text arguments:
   `create_ai_knowledge_base_plain_text.content`, `update_ai_knowledge_base_plain_text.content`,
   `send_inbox_email.body`, and `execute_graphql`. `FastMCP(max_request_body_size=...)` is settable
   at 1.29.0 if we would rather preserve the old behavior; mcp 2.0 ships the same 4 MiB default.

## Pre-existing issues found while testing, not caused by this bump

Recorded here because the live runs surfaced them and they would otherwise be lost:

- Envelope coverage is not uniform. `move_card_to_phase` and `get_phase_fields` return
  `isError=True` with a raw string (`Error executing tool move_card_to_phase: Couldn't find Card
  with ID=...`). That is deliberate at `tools/pipe_tools.py:912` (`raise exc` after enrichment
  fails) and asserted by `test_move_card_to_phase_surfaces_original_error_when_destination_allowed`.
- `update_card_field` against a nonexistent card reports `"Phase field not found"`. The envelope
  shape is right, the diagnosis points at the wrong object.
- `ArgModelBase.model_config` sets no `extra="forbid"`, so unknown tool arguments are silently
  accepted and the `extra_forbidden` branch in `_format_validation_errors` is unreachable.
- An immediate read-back after a successful `delete_pipe` returned a full, live-looking (and
  mutated) pipe body; the same read minutes later returned `INTERNAL_SERVER_ERROR`. Worth knowing
  before writing any live test that asserts deletion by reading back.

## Notes for the reviewer

The `uv.lock` diff is large for a single-package upgrade. Only one package entry is added,
`starlette 1.3.1` gated on `python_full_version >= '3.14'`, which arrives with the Python 3.14
support added in mcp 1.28.0. The rest of the churn is uv rewriting resolution markers around that
new fork. This project runs 3.11 (`.python-version`, which CI reads), so the resolved starlette
stays 0.46.1 and no runtime dependency actually moves except `mcp` itself.

The remaining two phase 0 tasks (funnelling the test helper, normalizing the `isError` assertions)
are deliberately not here. Each touches roughly 30 test files, and the AGENTS.md guidance to split
PRs over 10 files or 300 changed lines applies. They land as their own PRs against #545, where a
mechanical diff can be reviewed as mechanical instead of buried under a lock file.

Out of scope: a pre-existing ruff `I001` in `tests/test_parity.py` on `dev`. CI lints changed files
plus the `src` trees, so it does not surface, and fixing it here would be unrelated churn.


adriannoes (approved):
Task 1 of the phase 0 pin bump lands cleanly. Locked mcp 1.29.0 keeps the FastMCP seams this package patches (validation envelope, CallToolRequest middleware wrap, ToolManager registration) behaving as the existing contract suite expects, and CI plus a full non-integration run on the frozen head both came back green. The CHANGELOG correctly names the two caller-visible SDK side effects (4 MiB Streamable HTTP 413, and serverInfo.version tracking the SDK). Keeping the mechanical test-helper chores in separate PRs is the right split.

### What worked well
- Version claims in AGENTS.md and the two patch module docs moved in lockstep with the floor and the lock.
- The follow-up changelog commit owned the earlier "no behavior change" mistake instead of leaving it in Unreleased.

### Also noted
- The CHANGELOG and the "Two behavior changes" section correctly record the 4 MiB 413 limit and serverInfo.version tracking the SDK. The Summary still opens with "no behavior change," which undercuts that honesty. A one-line Summary edit to match would help future readers. Non-blocking.

Review path: local checkout at e0dcf389 against base 1727cd48 (trusted, env-isolated unit run). Reviewer live MCP from that worktree was not exercised; author live notes were treated as corroboration only.

mocha06 (approved):

adriannoes:
@Danielmoraisg Follow-up after the approve: we ran an extra wave pass against a detached worktree at `e0dcf389` (locked `mcp==1.29.0`), so the reviewer-side live evidence matches your pin bump rather than ambient Cursor MCP.

What we exercised (read-only / loopback or in-process, no Pipefy mutations):

- Validation envelope still returns `INVALID_ARGUMENTS` with `isError=False` and no `pydantic.dev` leak; vanilla FastMCP without the patch still takes the error path with a pydantic URL.
- Tool-call middleware over Streamable HTTP still installs the chain sentinel, emits a correlated `tool_call` / `http_request` pair, and keeps argument values out of structured logs.
- Tool surface counts stay dispositive: 187 default, 9 under `--toolsets power`, 184 on the remote floor; deleting `_tool_manager` makes removal a no-op.
- `initialize` advertises `serverInfo.version` `1.29.0` (SDK), not `pipefy_mcp` `0.4.0-beta.1`.
- Request body bracket: 4194304 accepted, 4194305 returns HTTP `413` (`Request body too large`).

Result: all five matrix rows pass under worktree provenance. 

That closes the earlier gap where S5 (413) and live seam checks were only author-reported. No new findings; the approve stands. 

Cursor MCP was not restarted from that worktree, so we did not treat ambient Cursor calls as PR-head proof. Write-path smoke against a real org was not re-run on our side (no mutation consent in this pass).

PR #547, opened 2026-07-30
refactor(mcp)!: port to the mcp SDK 2.x [#543 Phases 1-4]

Danielmoraisg:
Ports `packages/mcp` to `mcp[cli]==2.0.*`. Phases 1 through 4 of #543, on top of the Phase 0 floor bump merged in #546.

## Read this first: this reverses a recorded decision

#543 decided to **stay on 1.x**, with the revisit trigger set at "`2.0.2` or later published, or a
concrete product need for a `2026-07-28` capability." Neither has happened: 2.0.0 is still the only
2.x release, published 2026-07-28, with no patch releases. I raised that before starting and was
asked to proceed, so this PR is a deliberate override rather than a triggered one. @mocha06 should
weigh in, since #543 is theirs.

#543's shape held up (the test suite dominated, the middleware was the only real design work). Its
scale did not, because both mechanical passes were sed-able and the two chokepoints it identified did
their job.

## What a caller sees: nothing

Verified rather than assumed. 187 tools over both transports, same `--toolsets` selections, same
remote-safe floor, same tool-error envelope, same structured log shapes. The server still negotiates
`2025-11-25`, which confirms #543's separability finding empirically: **upgrading the SDK moves no
client onto `2026-07-28`.** Adopting that revision stays separate work.

## What an embedder has to change

- `install_tool_call_middleware` → `build_tool_call_middleware`, which returns one `ServerMiddleware`
  for the constructor instead of patching a handler. `extra_tool_middlewares` on
  `build_pipefy_mcp_server` is unchanged, so an embedder registering through the builder needs no
  edit; one that called the install function directly does.
- For a middleware author: `ToolCallContext.req` is gone (read `tool_name` and `arguments`, now the
  raw pre-validation values off the params), and a middleware returns a `CallToolResult` rather than
  a `types.ServerResult`.
- `wire_hosted_observability` takes `host`, `transport_security`, and `json_response`, because 2.0
  moved all three off the server constructor onto the per-transport call.

## Commits, each reviewable on its own

| Commit | What |
| --- | --- |
| Phase 1 | `mcp.server.fastmcp` → `mcp.server.mcpserver`, `FastMCP` → `MCPServer`. Mechanical. |
| Phase 2 | The middleware rewrite, plus the registry, tool-lifecycle, and transport wiring it forces. |
| Phase 3 | The test suite: the compat shim, the field renames, the middleware test rewrite. |
| Phase 4 | Docs that argued from removed internals, and the CHANGELOG. |
| Review 1 | Byte-for-byte asserts on the advertised OAuth URLs, plus an SSE-framed observability case. |
| Review 2 | The pin narrowed from `<3` to the 2.0 patch line. |
| Review 3 | The mount lifespan contract, the body limit, and the `legacy` pin's prerequisites. |

Phase 1 does not import on its own (`core/tool_middleware.py` still reaches the removed
`request_ctx`). The SDK cannot be half-installed, so only the branch tip is expected to be green;
Phase 1 is split out because #543 asked for it as a separate reviewable commit.

## Three judgement calls worth your attention

**1. `ServerMiddleware` over `Extension.intercept_tool_call`.** Both could host the chain.
`intercept_tool_call` (SEP-2133) is semantically narrower, but registering an `Extension` advertises a
reverse-DNS capability in the `initialize` response, and internal logging should not change what
clients see on the wire. Upstream also calls `middleware` the designated replacement for the private
`_handle_*` overrides, which is exactly this seam.

The adapter in front of it is deliberate: upstream marks the `middleware` list **provisional**, so
keeping `ToolCallContext` / `ToolCallMiddleware` as the registration surface confines churn there to
one module. A `ServerMiddleware` also sees every method, so the `ctx.method` filter lives in one place
rather than in each consumer.

**2. The registry keeps one private read.** `_live_tools` still calls `mcp._tool_manager.list_tools()`.
The public `MCPServer.list_tools()` is async and returns wire `mcp.types.Tool` models, which carry no
`fn` and no `run`; the remote floor needs the tier object's `meta` marker, and `apply_power_profile`
keeps the objects themselves in the catalog `execute_tool` dispatches through. Registration also runs
synchronously at build time, so awaiting would push `async` up through the composition root into both
transports. Removal moved to the public `remove_tool`, so five private touchpoints became one, with
the reason on the method.

**3. The test shim pins `mode="legacy"`.** `Client` otherwise negotiates `2026-07-28`, where the
server cannot call the client, and the `pipe_config` and `service_account` tests drive exactly that
back channel through elicitation. Legacy also matches the revision the server negotiates in
production, so the suite tests what ships. Flipping it is one line, and it comes last: the
elicitation tests have to move to `InputRequiredResult` first, and the `ctx.debug` / `ctx.info`
calls stop reaching clients that do not opt in. Recorded on the shim.

## Testing performed

```
4008 passed, 5 skipped, 34 deselected in 128.02s

cd packages/mcp && uv run lint-imports
Contracts: 2 kept, 0 broken.

uv run ruff check packages/*/src packages/mcp/tests
All checks passed!
```

Beyond the suite, since mocked transports cannot show these:

- **Middleware semantics** through a real client: chain order outer-to-inner, a short-circuit
  returning `is_error=True` with the envelope in `structured_content`, and `is_error=False` preserved
  for a business error. The argument-validation envelope still reshapes a pydantic `ValidationError`
  with no `pydantic.dev` leakage, so the `Tool.run` patch survives on 2.0.
- **Live API**, org `300917585`: `get_organization` returned `Event Doggo` with `is_error=False` and
  `success=True`.
- **Real stdio subprocess**: `initialize` advertises `serverInfo` `pipefy 0.4.0-beta.1`, negotiates
  `2025-11-25`, and lists 187 tools.

The 34 integration tests remain deselected; the stored session deliberately does not satisfy
live-creds detection (`oidc_client=None`), so the live checks above go through the running server
instead.

## Second live round, after review

Both surfaces re-driven against org `300917585`, read-only, no writes attempted.

**MCP**, as real processes rather than in-process imports:

- **The separability claim, falsification-tested.** A client offering `protocolVersion: 2026-07-28`
  still negotiates `2025-11-25`. The earlier check only observed the default.
- **Wire casing, on raw bytes before parsing.** In 285 KB of `tools/list`: `"inputSchema"` 187 times,
  `"input_schema"` zero; `"outputSchema"` 143 times, snake_case zero.
- **The middleware chain sits where the old handler patch sat.** An unknown-tool call still logged
  `arg_keys`, which only happens if middleware reads params before SDK dispatch. `outcome: "error"`
  with `isError: true` for a raising tool confirms the failing-call fix on a real transport.
- **Toolsets**: `governance` 9 (a strict subset, no name outside the 187), `power` 9. `execute_tool`
  returned a payload identical to the direct call.
- **HTTP**: SSE-framed, one `http_request` line per request, and the org id appears nowhere in the
  stderr capture.

**CLI**: 48 API-invoking commands (34 exit 0, 9 exit 1, 5 exit 2, every non-zero intentional) and 47
`--help` targets. The exit-1-vs-exit-2 split holds, and the unified `PipefyGraphQLError` is confirmed
on all three endpoint families, including the two former leak sites: `portal get` on Interfaces
(previously `TransportQueryError`) and `ai-automation get` on the Internal API (previously a bare
`ValueError`). Zero hits for `Traceback`, `gql.transport`, or `pydantic.dev` across 89 captured
streams, with a positive control confirming the detector fires. This branch touches no CLI or SDK
source, so a clean CLI run is the expected result rather than a new finding.

One correction to an earlier claim: `get_organization` with an unreachable id returns `isError:
false` with `success: false`, not `isError: true`. That is the documented contract (`isError=True`
means the tool never ran), so the tool is right. Worth knowing that the path is a bare
`except Exception`, a tool-layer convention with 186 occurrences across 24 modules, which is part of
why the port did not break callers and also bounds what "nothing changed" can mean at the tool
boundary. Pre-existing, not touched here.

## Two things I did not do

- **`serverInfo.version` now reports this package's version** (`0.4.0-beta.1`) instead of the SDK's.
  This is a behavior change I chose to make rather than defer: it fixes the misreporting found while
  testing #546, and `MCPServer` accepting `version` is exactly what made it fixable. Say the word if
  you would rather it land separately.
- **The 68 `ctx.debug`/`ctx.info` calls across 11 tool modules are untouched**, so the suite emits 997
  `MCPDeprecationWarning` lines (SEP-2577 deprecates the logging capability). Warnings only, since no
  `filterwarnings = ["error"]` is configured. Replacing them stops the client receiving those lines,
  which is a behavior change deserving its own PR.

## Dependency changes worth a look before deploy

The SDK now uses `httpx2` instead of `httpx`, which is a non-event here (this package, `sdk`, and
`auth` all declare `httpx` themselves, and nothing hands an httpx object into the SDK). It also pulls
`starlette` from 0.46 to **1.x** and adds `mcp-types`, `opentelemetry-api`, and `truststore`.
`truststore` validates TLS against the OS trust store rather than certifi, so a minimal container with
no system CA store needs `SSL_CERT_FILE`.

Phase 4 in #543's table (the downstream consumer of `extra_tool_middlewares` and
`streamable_http_app()`) is tracked outside this repository and must land in lockstep.


mocha06:
We synced regarding the decision to hold off till a smaller minor bump for the MCP package gets added (as mentioned on the 543 originally), but ultimately we decided it's ok to move forward with this migration now, given our current needs. LGTM

adriannoes (commented):
The MCP 2.x port looks solid on the product contracts that matter: suite green locally and in CI, 187 tools with the remote floor intact, middleware short-circuit vs soft business errors still map the way agents expect, and the server still negotiates `2025-11-25`. The three judgement calls in the PR body (ServerMiddleware adapter, residual private `_live_tools`, Client `mode="legacy"`) hold up under review.

**Required before merge**

- This PR deliberately overrides #543's stay-on-1.x decision while the revisit triggers (`2.0.2+` or a concrete `2026-07-28` need) are unmet. Please get an explicit supersede from @mocha06 on #543 or here so tracker truth matches the ship.
- Confirm the out-of-repo consumer of `extra_tool_middlewares` / `streamable_http_app()` (and `wire_hosted_observability`'s new kwargs) lands in the same window, or waive that lockstep explicitly. Embedders that still call `install_tool_call_middleware` will break.

**Also noted**

- After moving the allowlist onto `streamable_http_app()`, `run_server` re-derives via `transport_security_for(settings)` while `McpRuntime.transport_security` is still built and unused in production, and `for_profile`'s "parse once" docstring no longer matches the code. Settings are immutable so behavior matches today; worth collapsing to one ownership path.
- Soft `INVALID_ARGUMENTS` still breaks if a tool has a strict Pydantic return model (`convert_result` validates the error dict against the success output). Same on mcp 1.29 and 2.0; today's DictOutput catalog is fine. A typed-return regression would lock the claim that the envelope was re-verified on 2.0.
- Optional hygiene: pin `protocolVersion` / `serverInfo.version` in tests; retire the deleted `fastmcp_tool_lifecycle` path in `docs/architecture.md`; CHANGELOG says `wire_hosted_observability` takes `port`, but port stays on uvicorn.

**What worked well**

Clear four-phase commits, honest disclosure of the decision override and embedder breaks, concentrated private registry read with rationale, and the `_mcp_compat` harness so the suite tests the revision that production actually serves.

adriannoes (commented):
Re-reviewed at the new head `0a10a839` (my earlier review's snapshot was `3c8da051`), including the follow-up commit, with the suite, lints, and a live stdio boot of the built artifact all green: 1752 passed, handshake negotiates `2025-11-25`, serverInfo `pipefy 0.4.0-beta.1`, 187 tools local and the 184-tool remote floor confirmed live.

On the follow-up commit: good catch, and the fix holds up under adversarial checks. The SDK does serialize inside the middleware chain (`runner.py`, `by_alias=True`), `result_is_error()` covers every result shape a middleware sees, and the new tests discriminate exactly as the commit message claims: restoring the old read fails precisely the wire-dict and end-to-end tests while the short-circuit test stays green (mutation-tested). This also resolves the `transport_security` ownership note from my previous review.

Thanks @mocha06 for the explicit supersede of #543; that closes the first gate from the previous review. Suggest annotating or closing #543 so the tracker record matches.

**One gate left before merge**

- Downstream lockstep, unchanged from the previous review: no sibling PR is linked and no waiver is recorded. Note the follow-up commit adds to what the embedder must absorb: middleware should read the error flag via the new `result_is_error()` (porting the 1.x attribute read compiles and silently logs every failure as ok), and `McpRuntime` no longer carries `transport_security`.

**CHANGELOG completeness, worth fixing in this PR since embedders will migrate from it**

- Missing: the `context_from_request` rename to `context_from_server_request` (ImportError at head, input type changed).
- Missing: the `McpRuntime.transport_security` removal from the follow-up commit (TypeError/AttributeError for an embedder; replacement is `transport_security_for(settings)`).
- Worth adding: what `call_next` returns is now polymorphic (wire dict keyed `isError` vs `CallToolResult.is_error`) and `result_is_error()` is the supported read; that exact trap bit this PR's own middleware.
- Still says `wire_hosted_observability` takes `port`; port only reaches the uvicorn bind.

**Tracking, not blocking**

- The pinned mcp 2.0.0 has an open upstream session-leak issue (modelcontextprotocol/python-sdk#3228) in default stateful streamable HTTP; this deployment runs that mode with 421 rejections enabled, so each rejected request strands ~22.6 KiB. Availability rather than access; worth watching for 2.0.1.
- Hygiene from the previous review, still open: protocol/serverInfo handshake pins in tests; stale FastMCP references now in three places (`docs/architecture.md:27`, `docs/config.md:139` including the nonexistent `FASTMCP_LOG_LEVEL`, root `AGENTS.md:86`); `runtime.py` docstrings now overstate the runtime's allowlist role in the other direction after the field removal; a dead `runtime.transport_security` mock line at `tests/test_server.py:80`.

adriannoes:
@Danielmoraisg We reviewed the MCP Python SDK [Migration Guide: v1 to v2](https://py.sdk.modelcontextprotocol.io/migration/) and have a few additional suggestions, ordered by priority for you to evaluate:

### P1 — please address or explicitly waive

1. **Observability E2E on the real SSE path** — tests force `json_response=True`, but production leaves the default `False`. Please add an ASGI case on SSE framing (`initialize → tools/call`) so we are not only validating the test-mode path. (*Transport-specific parameters moved…*)

2. **CI matrix across `mcp>=2.0,<3`** — the product depends on provisional `ServerMiddleware` and the `ToolManager.add_tool` / `Tool.run` patch. Running CI on the floor (`2.0.0`) and the latest resolvable 2.x under `<3` avoids silent breakage from a transitive bump.

3. **OAuth issuer literal equality** — the guide treats issuer / trailing slash as a literal match. Asserts that `.rstrip("/")` can hide drift that a v2 client rejects. Prefer exact equality plus a trailing-slash case.

### P2 — embedder / ops contract

4. **`wire_hosted_observability` mount contract** — serving directly with uvicorn is fine; if the returned app is mounted under another Starlette app, the host lifespan must enter `mcp.session_manager.run()`. Document that in CHANGELOG/AGENTS/the helper docstring to close the embedder contract.

5. **4 MiB Streamable HTTP body limit** — SDK default; presigned uploads are safe, but please note this for consumers in CHANGELOG/AGENTS (or set `max_request_body_size` if hosted needs more).

### P3 — tracking / follow-up

6. **`mode="legacy"` follow-up** — the pin in `_mcp_compat` is correct given `NoBackChannelError` on 2026-07-28. When that revision is adopted, flipping this one line (plus elicitation migration) should be explicit in the tracking.

None of this reopens the gates already discussed (embedder lockstep, CHANGELOG, #543 supersede). These are reinforcements from the official guide.

Danielmoraisg:
@adriannoes Went through all six against the guide. Five landed as changes, one is a documentation cross-reference plus a correction to what it actually buys. Nothing waived outright.

### P1

**1. Observability E2E on the real SSE path.** Added `test_sse_framed_responses_still_produce_correlated_log_lines`, which runs `wire_hosted_observability(app)` with `json_response` at its default: initialize, initialized, one `tools/call`, asserting `content-type: text/event-stream` on both responses and then the same `http_request` / `tool_call` line assertions plus per-POST correlation.

One correction to the framing, because the test proves less than it might look like it does: a POST's SSE stream closes after its single response message, so this does not exercise a stream the server holds open. `BaseHTTPMiddleware` would pass it too. The genuinely long-lived stream is the standalone `GET /mcp`, which I did not open (a different, harder fixture). What the case does close is your actual gap: the mode a deployment answers with is now covered, not only the test-mode path. Both facts are on the test docstring so nobody reads more into it later.

**2. CI matrix across `mcp>=2.0,<3`.** Solved by narrowing the range instead of testing across it: the pin is now **`mcp[cli]==2.0.*`**.

I built the matrix job first and it was the wrong shape. Two reasons. It adds zero coverage today: `uv lock --upgrade-package mcp --dry-run` reports "No lockfile changes detected", because 2.0.0 is still the only 2.x release, so the lock already holds the newest resolvable version. And it would only ever tell us after the fact that a release we already allow has broken us. Since the exposure is exactly the two surfaces you named (the provisional `MCPServer(middleware=[...])` list and the `Tool.run` patch), keeping a new minor out entirely closes the gap rather than reporting on it. Patch releases are still accepted; widening to a new minor becomes a deliberate change that re-tests both surfaces, with the reasoning recorded on the dependency and in the CHANGELOG.

Worth noting the existing "Smoke install CLI and MCP wheels" step already resolves from PyPI rather than the lock, so a release inside the accepted patch range that breaks import still fails CI at `--help` level.

**3. OAuth issuer literal equality.** Dropped `.rstrip("/")` from all seven asserts (`test_server.py`, `auth/test_resource_server.py`, `core/test_runtime.py`) and added trailing-slash cases at two layers, because the builder-level one alone would not have been enough: `RS_RESOURCE` and `RS_ISSUER` carry no slash, so with slash-free input a normalizer anywhere in the chain stays invisible.

- `test_build_advertises_urls_byte_for_byte_including_a_trailing_slash` pins `build_resource_server_auth`.
- `test_metadata_keeps_a_configured_trailing_slash_end_to_end` runs a slashed URL through `ResourceServerSettings` and out of the served RFC 9728 document.

The end-to-end one turned up something worth knowing before a hosted config change: a trailing slash moves the well-known route, because the SDK derives that path from the resource URL's path. With `https://mcp.example.com/mcp/` the document is served at `/.well-known/oauth-protected-resource/mcp/` and the unslashed path answers `307`. Both are now asserted. `ResourceServerSettings` strips whitespace and deliberately not slashes, which is the behavior that makes the URL literal end to end.

Two things I checked and found inert here, so no change: the guide's "client rejects authorization server metadata with a mismatched issuer" applies to the SDK's own OAuth client provider, which this repo does not use anywhere. The `.rstrip("/")` in `pipefy_auth/discovery.py` is our own OIDC discovery client on the CLI login path, comparing a document's `issuer` claim to the URL it was fetched from, and its leniency predates this PR. The literal-match surface that does matter is inbound: `JwtValidator` passes `issuer=` to `jwt.decode`, which is an exact comparison, so the advertised issuer and the configured one have to be the same string. That is what the new tests protect.

### P2

**4. `wire_hosted_observability` mount contract.** Documented in three places: the helper docstring, the `AGENTS.md` observability section, and the CHANGELOG entry. The app the helper returns owns the lifespan that enters `session_manager.run()` (`lifespan=lambda app: session_manager.run()` in the SDK's `streamable_http_app`). Serving it directly needs nothing, since uvicorn runs it. Mounting it under a host Starlette or FastAPI app means the host's own lifespan has to enter `app.session_manager.run()`, because Starlette does not run a mounted app's lifespan; without it the session manager never starts and every request fails.

**5. 4 MiB Streamable HTTP body limit.** This was already documented, in the 1.29.0 floor-bump entry in the same CHANGELOG cycle: the limit arrived with 1.29, not with 2.0. Confirmed `DEFAULT_MAX_REQUEST_BODY_SIZE` is still 4 MiB on 2.0.0 and that the wiring forwards no `max_request_body_size`, so the default stands unchanged. Added the forward-reference in the 2.x entry, plus `AGENTS.md` and the helper docstring, naming the reachable case (a long free-text argument: knowledge-base `content`, `send_inbox_email.body`, `execute_graphql`) and where a hosted deployment raises it. Not raising it, since nothing has hit it.

### P3

**6. `mode="legacy"` follow-up.** Expanded the `_mcp_compat` docstring from "one line to flip" to the flip plus its two prerequisites, which is the part that was missing: under 2026-07-28 a server-initiated elicitation raises `NoBackChannelError`, so `pipe_config` and `service_account` have to move to `InputRequiredResult` first, and the `ctx.debug` / `ctx.info` calls stop reaching clients that do not set `logging.optInLogMessages`.

On "explicit in the tracking": there is no follow-up issue yet. #543 is the decision issue this PR supersedes, so it is the wrong home for it. Happy to open one carrying those two prerequisites, say the word.

### Sweep of the rest of the guide

Ran the remaining items against the tree. All clear, listed so the negatives are on record rather than assumed:

- No `model_dump()` on an MCP wire type anywhere, so the snake_case field rename cannot have moved a caller-visible key. The `by_alias=True` hits are all on our own SDK input models. `describe_tool` builds its `input_schema` key as a literal over `tool.parameters`, so the `power` catalog shape is unchanged.
- No `McpError` / `MCPError` raised anywhere, so the "raising it in a handler now becomes a top-level JSON-RPC error instead of `isError=True`" change cannot reach us.
- No reads of `LATEST_PROTOCOL_VERSION` or `SUPPORTED_PROTOCOL_VERSIONS`, and no `mcp.shared.version` import. `LATEST_PROTOCOL_VERSION` changing to `2026-07-28` silently changes the meaning of any comparison against it; we have none.
- No `isinstance` against `ElicitationResult` (now a `TypeAliasType`, so that raises `TypeError`), no `requestedSchema`, no `progressToken`, no `ctx.client_id`, no camelCase field reads left (`.isError`, `.inputSchema`, `.mimeType`, `.structuredContent`).
- No registered resources or resource templates, so the RFC 6570 matching changes, the path-safety checks, and `FileResource.is_binary` → `encoding` are all inert.
- No synchronous `def` tool handlers, so the move to `anyio.to_thread.run_sync` (where `asyncio.get_running_loop()` raises and thread-affine state stops being safe) does not reach production tools.

### Verification

```
4008 passed, 5 skipped, 34 deselected in 128.02s

cd packages/mcp && uv run lint-imports
Contracts: 2 kept, 0 broken.

uv run ruff check packages/*/src packages/mcp/tests
All checks passed!

uv run python scripts/bump_version.py verify
exit 0
```

The count moves from the 4003 in the body to 4008: two from the failing-tool-call fix that landed
after I wrote it, three from this round. Body updated, including the pin and a new live-round section.

Also re-drove both surfaces against org `300917585`, read-only, as real processes. Two results are
worth pulling out here because they bear on your P1 items:

- **The separability claim now has a falsification test, not just an observation.** A client offering
  `protocolVersion: 2026-07-28` still negotiates `2025-11-25`.
- **Wire casing checked on raw bytes before parsing**, which is the one way the snake_case rename
  could have moved a caller-visible key while the suite stayed green: in 285 KB of `tools/list`,
  `"inputSchema"` 187 times and `"input_schema"` zero; `"outputSchema"` 143 and snake_case zero.

On the CLI side, 48 API-invoking commands with every non-zero exit intentional, and the unified
`PipefyGraphQLError` confirmed on both former leak sites (`portal get` on Interfaces, `ai-automation
get` on the Internal API). That branch touches no CLI or SDK source, so a clean run there is the
expected result rather than a finding.

One correction to something I claimed earlier in this PR: `get_organization` with an unreachable id
returns `isError: false` with `success: false`, not `isError: true`. That is the documented contract
(`isError=True` means the tool never ran), so the tool is correct and my earlier description was not.
The failing-call fix is still verified, via a raising tool returning `outcome: "error"` with
`isError: true` on a real transport.

adriannoes:
@Danielmoraisg Thanks — went through the six replies and the three follow-up commits (`812b99b2`, `2d30d91d`, `f5b28ca0`). Accepting all of them.

- **P1.1** SSE case is enough for the gap we asked about; the docstring caveat on long-lived `GET /mcp` is the right honesty.
- **P1.2** Pinning `mcp[cli]==2.0.*` is the better shape than an empty matrix while 2.0.0 is the only release. Widening the minor becomes a deliberate, re-tested change.
- **P1.3** Literal issuer asserts + the well-known trailing-slash discovery are solid.
- **P2.4 / P2.5** Mount contract and 4 MiB cross-refs look good.
- **P3.6** Docstring prerequisites are what we needed. Please **open the follow-up issue** you offered (flip `_mcp_compat` off `mode="legacy"` only after `pipe_config` / `service_account` move to `InputRequiredResult`, and after the `ctx.debug`/`ctx.info` client-visibility story is settled). Link it here when it exists.

### Still required before APPROVE / merge

**Downstream lockstep** — unchanged from the earlier review, and still the only open gate. The PR body says the out-of-repo consumer of `extra_tool_middlewares` / `streamable_http_app()` (and now the mount / `result_is_error` / pin surface) must land in lockstep. Please either:

1. link the sibling PR/change that absorbs the embedder breaks, or
2. post an explicit waiver on this thread (who owns the consumer, and why merge without lockstep is safe).

I am closing #543 as superseded by this PR + @mocha06's go-ahead, so tracker truth matches the ship.

Once lockstep is linked or waived (and the 2026-07-28 follow-up issue is opened), I will APPROVE.

mocha06 (commented):
Reviewed independently across four lenses (behavior parity, the request-scoped plumbing rewrite, test-suite integrity, and runtime validation), driving both branches and diffing measured output rather than reading the diff. Re-checked everything against the new head `f5b28ca`.

@adriannoes's open items and your own guide sweep I am not restating. The raw-wire-byte casing check (`"inputSchema"` 187, `"input_schema"` zero) is better evidence than what I had, and `mcp[cli]==2.0.*` is a stronger answer to the provisional-API exposure than testing across the range would have been, because it closes the gap rather than reporting on it after the fact.

Three things remain, and the first one is the reason I am posting.

## The separability claim is still false, and the new check does not reach it

The falsification check drives a client **offering** `protocolVersion: 2026-07-28` in `initialize`. That is the handshake leg, and I measured the same result independently: it correctly negotiates `2025-11-25`. The exposure is a different code path. `Client` defaults to `mode="auto"`, which uses the modern discovery handshake rather than offering a version:

```
PR head, Client(url, mode="auto")   -> CONNECTED, 187 tools, negotiated 2026-07-28
PR head, Client(url, mode="legacy") -> CONNECTED, 187 tools, negotiated 2025-11-25
dev,     Client(url, mode="auto")   -> CONNECTED, 187 tools, negotiated 2025-11-25
```

New exposure, not inherited. And `_mcp_compat.py`'s expanded docstring already names what breaks there: a server-initiated elicitation raises `NoBackChannelError`. That reaches production, because `create_card` and `fill_card_phase_fields` gate on `supports_elicitation(ctx)`, which reads the client's **advertised capability** rather than whether the transport has a back channel:

```
mode=legacy  2025-11-25  guard=True -> elicit succeeded
mode=auto    2026-07-28  guard=True -> NoBackChannelError
```

`create_card` catches `MalformedFieldDefinitionError` and `UserCancelledError`. Neither matches, so the caller receives a JSON-RPC protocol error rather than a tool envelope. Blast radius is limited today and I confirmed it: Claude Code and other 1.x-era clients still land on `2025-11-25`. It grows as clients adopt the 2.0 SDK, and `create_card` is a headline tool.

Suggested fix, all three parts: pin the server to `2025-11-25` so the claim is true by construction rather than by accident; change `supports_elicitation` to test back-channel availability rather than the advertised capability; and catch `NoBackChannelError`, degrading to the non-elicit path.

**On the test gap, which is the same gap.** There is still no committed assertion on a negotiated revision. `2026-07-28` appears only in a docstring, and every `protocolVersion` in the suite is an input (a synthetic `"2025-06-18"`, or echoed back into a header), never an assertion on output. Worse, removing the shim's `mode="legacy"` pin breaks **7 elicitation tests in `test_pipe_tools.py` and nothing else**, so revision drift surfaces as confusing elicitation failures rather than as "the negotiated revision changed." `Client` exposes `.protocol_version` and `.server_info` directly, so this is two lines, and it is the assertion that would have caught this.

## `resultType: "complete"` still reaches the wire on short-circuit results

`core/tool_middleware.py`, unchanged at the new head. The follow-up commit is right that the SDK serializes inside the chain with `by_alias=True`, but that path is `_inner` then `_serialize`. A middleware returning without awaiting `call_next` is dumped by `_dump_result` only, skipping `_serialize`, which is where `serialize_server_result(method, version, dumped)` sieves 2026-era fields off a legacy connection. Measured, one server, one session, negotiated `2025-11-25`:

```
real tool error : {"content":[...],"isError":true}
short-circuit   : {"content":[...],"structuredContent":{...},"isError":true,"resultType":"complete"}
```

And against the sieve directly: dropped at `2025-11-25`, kept at `2026-07-28`. The SDK's own `_serialize` docstring warns that a middleware short-circuiting without `call_next` owns its result, envelope included.

A Python SDK 2.0 legacy-mode client tolerates it, since `result_type` is a declared field, so this bites strict validators, other-language clients, and wire-shape snapshot tests. `short_circuit_error` has no non-test caller in this repo, so it can only affect the downstream embedder, and `tests/core/test_tool_middleware.py` compares short-circuit to in-tool error on the **decoded envelope only**, never returning one through a client. The suite structurally cannot see it.

## The `_ok_result()` sibling of the bug the follow-up commit fixed

`tests/observability/test_tool_log_middleware.py:36-37` still builds a `CallToolResult` model. I instrumented the live chain: **every** result is a dict, including success. That is the same class of fixture that caused the outcome-logging regression, on the path that was not fixed. Latent only because the middleware reads nothing off the result except through `result_is_error`. The moment any log field derives from the result (content length, `structured_content`, `result_type`), these tests pass against a model while production hands a dict, which is the same bug in the same file.

## Concretes for the lockstep gate

@adriannoes's gate is right and still open. Three call sites break in the downstream consumer, and the CHANGELOG will lead a migrator past the second:

1. `app.settings.transport_security = ...` raises `ValueError: "Settings" object has no field "transport_security"`. v2's `Settings` carries only `{auth, debug, dependencies, lifespan, log_level, warn_on_duplicate_*}`.
2. `app.settings.json_response = True` raises the same `ValueError`. **The CHANGELOG lists the moved arguments as "`host`, `port`, and `transport_security`", omitting `json_response` and wrongly including `port`**, which only reaches the uvicorn bind. That consumer deliberately serves plain JSON and calls `streamable_http_app()` directly, so fixing only `transport_security` silently flips every response to SSE framing.
3. `instrument_tool_calls(app)` reads `app._mcp_server.request_handlers[CallToolRequest]`. Both are gone, so this is a rewrite onto `extra_tool_middlewares`, not an import swap.

As written the failure is **loud**: an import-time crash, so a rollout self-reverts. The danger is the obvious fix. Delete the two assignments, leave `streamable_http_app()` argument-less, and `host` defaults to loopback with the SDK's own allowlist auto-enabled:

```
GET  /health  Host: <public>  ->  200 'ok'
POST /mcp     Host: <public>  ->  421 'Invalid Host header'
POST /mcp     Host: 127.0.0.1 ->  200
```

Readiness green, all proxied MCP traffic rejected. Combined with the upstream session leak @adriannoes found, that state strands roughly 22.6 KiB per rejected request on a single-replica deployment, so a mis-migration would leak until OOM rather than merely failing. Worth naming explicitly in the migration note so nobody reaches for that fix.

## Smaller, still open

**Session binding collapses to `client_id`.** 2.0 rejects a request whose authorization context differs from the session's creator, keyed on `(client_id, claims["iss"], subject)`. `JwtTokenVerifier._to_access_token` sets its own `sub` field rather than the SDK's `subject` and passes no `claims`, so two of the three components are `None` and the check reduces to `azp`. Under the hosted profile `azp` is the OAuth client, plausibly shared across every end user of one client app, so B can present A's `mcp-session-id`. Tool identity stays correct per message, so there is no outbound-token leak, but the back channel is unproven and `pipe_config` and `service_account` drive elicitation through it. Not a regression, since v1 had no ownership check at all, but this port adopts the mechanism without wiring the two fields that would make it per-user. Cheap hardening: move `PipefyAccessToken.sub` onto the base `subject`, and pass `claims={"iss": ...}`.

**`sse-starlette` 2.2.1 to 3.4.6** is a second major bump, still absent from the CHANGELOG. Separately, `starlette` is imported directly in four modules and declared in no `pyproject.toml`; it arrives transitively through the SDK, so a release that drops or vendors it breaks those imports with nothing to catch it. The starlette major itself is well insulated, and I checked why: only its two most stable APIs are used, and `RequestLogMiddleware` is raw ASGI rather than a `BaseHTTPMiddleware` subclass, which is where 1.x churn concentrates.

**Request-layer-invalid `tools/call`** (`params={}`, no `name`) now enters the middleware chain where dev fired zero times, emitting `{"event":"tool_call","tool":"","outcome":"error"}`. Narrower than it first looks: a missing required argument or an unknown tool name fires the chain on **both** branches, and only the request-layer case differs. The effect is on operator dashboards grouping by tool name, and on any governance middleware that would now count a call that never reached a tool.

**Suite warnings went from 0 on dev to 997 at head**, with no `filterwarnings` configured anywhere. Deferring the `ctx.debug` migration is reasonable; losing a clean baseline where a new warning was conspicuous is the cost worth naming.

## Live verification, and one corroboration

Built the branch's own wheels (`uv build`, then `uv tool install` into an isolated tool dir) and ran them against **production**, org `300514213`, read-only. Real data through `mcp` 2.0.0: `list_organizations`, `get_organization`, `search_pipes`, `get_pipe`, `get_start_form_fields`, `get_phase_fields`, `get_phase_cards`, `get_labels`, plus the `power` profile's four meta-tools and, the item every static reviewer flagged as unreachable, **`execute_tool` dispatching to a successful live read**, which returned the real org payload.

Your correction is confirmed independently: an unreachable id returns `isError: false` with `success: false`, and the envelope is genuinely helpful with no internal leakage. I also scanned 13 disclosure patterns (`pydantic.dev`, `Traceback`, `/site-packages/`, internal class and path shapes, and others) across every rejection envelope on both builds, with zero hits.

Not verified: the write round-trip, the remote profile's inbound bearer validation, and the `mode="auto"` elicitation break against a live pipe. The first two are environmental. The third needs a client-library run against a live pipe that I have not done, so treat the elicitation break as measured at the protocol layer and inferred at the tool layer.

mocha06:
regarding my review, I implemented them myself in #552, and merged the changes here fyi

adriannoes (approved):
Approving at `fb2ffcb4` so this can land on `dev` and keep the stack moving.

The port tip looks solid after #552: elicitation gated on the back channel, short-circuit results shaped per negotiated revision, session ownership wired with `subject`/`iss`, handshake/`serverInfo` pinned in tests, and the embedder CHANGELOG contract corrected. CI is green.

**Lockstep waiver (accepted for this merge).** The out-of-repo hosted wrapper that still calls the 1.x construction path (`settings.transport_security` / `json_response`, `install_tool_call_middleware` / `request_handlers`, bare `streamable_http_app()`) is not linked here. Merging to `dev` does not break pinned consumers. Anyone who floats to this package version without updating that glue will hit import/serve failures, and the tempting bare-`streamable_http_app()` fix can leave readiness green while proxied MCP traffic returns `421` (and can amplify the upstream session-leak memory growth). Owner of that wrapper needs to absorb the CHANGELOG break list in the same release window as the package bump. Not blocking `dev`.

PR #555, opened 2026-07-31
build: migrate uv.lock to revision 2

mocha06:
## Motivation

`uv.lock` is committed at `revision = 1`. Any current uv upgrades it to `revision = 2` — adding an `upload-time` field per artifact — the moment it has to write the file. That never surfaces during normal work because `uv lock` is a no-op while nothing has changed, but a release bump always forces a write: `bump_version.py` runs `uv lock` right after rewriting the version. So the next release commit, whoever cut it, would have carried an unrelated ~1900-line format migration alongside the version bump.

This came up while reviewing #554, where exactly that happened by accident.

## Outcome

The migration lands on its own, decoupled from any version bump, so the next release commit is only a version bump.

The change is purely the format: stripping the added `upload-time` fields and the revision line makes the new lockfile byte-identical to the old one, so no dependency resolution moved and no package version changed. `uv sync --locked` exits 0, `bump_version.py verify` passes, and the suite is green.

It was produced by bumping the workspace version away and straight back, which forces uv to rewrite the lockfile while leaving every other version-bearing file untouched — hence the single-file diff.

## Notes

Worth considering as a follow-up: the repo pins no uv version (`setup-uv@v7` with no version, no `[tool.uv] required-version`), which is why the committed lockfile could drift behind contributors' local uv in the first place. A `required-version` pin would keep this from recurring, but it needs agreement on a version and would block anyone on an older uv, so it is not folded in here.


adriannoes (approved):
LGTM

PR #557, opened 2026-07-31
docs(release): drop the last cross-refs to the removed alpha subcommand

mocha06:
## Motivation

The non-blocking item from the #556 approval: `_apply_prepare`'s stamp note ended with a `(see `alpha`)` cross-reference to a subcommand that no longer exists. It was the one pointer a reader would follow to understand why `stamp=False` exists, and it led nowhere.

Sweeping for the same pattern turned up two more that the review did not catch, both in `tests/test_release.py` — comments crediting the ahead-of-main guard and the "only one flow cuts alphas" contract to `alpha` rather than `alpha-pr`. Those guards now live in the shared `release_pr` path.

## Outcome

No comment, docstring, or doc in the release tooling names a subcommand that argparse does not register. The stamp note states its reason directly rather than pointing at a symbol, which is what made it rot in the first place.

`prerelease_track`'s docstring deliberately keeps its `alpha` mentions: there the word names the PEP 440 track, not the subcommand.

Comment-only, so there is no CHANGELOG entry — no behavior changed, and the existing `## [Unreleased]` bullet already describes the shipped contract. Adding one would only put noise in the next alpha's release body, since that body is `## [Unreleased]`.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -253,9 +253,9 @@ def _apply_prepare(bump_arg: str, target: str, *, stamp: bool = True) -> str:
     ``alpha_pr`` delegates to); callers own the preflight and confirmation.
     Returns the version bump_version actually wrote.
 
-    ``stamp=False`` leaves ``## [Unreleased]`` intact — what the alpha flow
-    needs, since consecutive alphas share one accumulating section (see
-    ``alpha``).
+    ``stamp=False`` leaves ``## [Unreleased]`` intact — what an alpha needs,
+    since consecutive alphas share one accumulating section that the eventual
+    beta promotion stamps in full.
     """
     # bump_version.py owns the transform; run its CLI so the file set and lock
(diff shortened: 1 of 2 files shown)


adriannoes (approved):
The leftover from the #556 approval is gone: `_apply_prepare`'s `stamp=False` note no longer points at a removed `alpha` subcommand, and the two test comments now credit `release_pr` / `alpha-pr` instead of the deleted one-step path. The rewrite states why Unreleased stays intact (beta promotion stamps the whole section) rather than naming a symbol that can rot again.

Checked clean against argparse (`prepare`, `release-pr`, `alpha-pr`, `publish`, `verify`), a leftover sweep of scripts/tests/RELEASE, and `tests/test_release.py` (49 passed). No behavior change, so skipping CHANGELOG is right. PEP 440 track `alpha` in `prerelease_track` correctly stays.

PR #571, opened 2026-08-04
fix(ai-agents): active lifecycle — inactive create + update preserve (#520)

### LINKED ISSUE #520: AI agents: create always active; update re-enables a previously disabled agent
## Problem
AI agents created via the toolkit start active, and updating an agent can clear disabled state so a previously stopped agent starts running again without an explicit activate request.

## Evidence
- Appeared in **3 builds**.
- Report detail: create has no `active=false` option; `update_ai_agent` clears `disabledAt`.
- Testers flagged this as dangerous in production (AI credits + actions on live cards).

## Impact
A client who disabled an agent can have it revived by a routine edit. Unexpected credit spend and card-side effects without operator intent.

## Status on `dev`
**PARTIAL** — `toggle_ai_agent_status(active=…)` exists as a separate tool; `create_ai_agent` still does not take an inactive-on-create flag. Whether update preserves `disabledAt` needs verification when picking this up.

- Allow `active=false` (or equivalent) on create
- Preserve disabled state on update unless `toggle_ai_agent_status` / explicit `active` is passed
- Skill rule: after create/update, confirm status; never assume inactive
- Return current `active`/`disabledAt` clearly on create/update responses

adriannoes:
## Summary

- **Closes #520.** Path **A** (schema + live probe 2026-08-04): `AiAgentInput.disabledAt` is writable; update without it clears/revives; update with it preserves; create can send it for inactive-on-create.
- SDK: inactive create sends `disabledAt`; routine update re-sends current `disabledAt` (`preserve_disabled_at=True`); create-active chains use `preserve_disabled_at=False` so the configure update clears the API’s default disabled shell. Results expose `disabled_at` / `active`.
- MCP: `create_ai_agent(active=False)`; CLI: `pipefy agent create --inactive`. Success payloads and skill/docs document the contract; activation remains `toggle_ai_agent_status` / `pipefy agent toggle`.

## Research note (Path A)

| Finding | Evidence |
|---------|----------|
| `AiAgentInput.disabledAt: DateTime` writable | introspect 2026-08-04 |
| Empty `createAiAgent` often returns `disabledAt` set | live probe |
| Update **without** `disabledAt` clears it (revive / also activates configure step) | behavioral probe |
| Update **with** `disabledAt` preserves | behavioral probe |
| Path B (warning-only) skipped | API supports preserve/write |

## Test plan

- [x] `uv run pytest -m "not integration"` (4117+ passed after C.7)
- [x] Live smoke (CLI on this branch): create active → `active: true`; create `--inactive` → stays disabled after prompt update; toggle off → update → still disabled
- [ ] Optional: re-smoke via Cursor MCP after restarting the local server so `active` appears on `create_ai_agent`

## Docs / skills

- [x] `docs/parity.md` updated
- [x] `skills/ai-agents/pipefy-ai-agents/SKILL.md` Active lifecycle section
- [x] `docs/mcp/tools/automations-and-ai.md` AI agents table
- [x] `CHANGELOG.md` [Unreleased]

## Legal / contributions

### DIFF
--- a/packages/cli/src/pipefy_cli/commands/agent.py
+++ b/packages/cli/src/pipefy_cli/commands/agent.py
@@ -177,7 +177,7 @@ def agent_create(
             behaviors=validated.behaviors,
             data_source_ids=validated.data_source_ids,
             disabled_at=validated.disabled_at,
-            preserve_disabled_at=not active,
+            preserve_disabled_at=False,
         )
         update_result = await client.update_ai_agent(update_input)
         result_disabled_at = update_result.get("disabled_at")
@@ -217,6 +217,14 @@ def agent_update(
         "--data-sources",
         help="Optional JSON array of knowledge-source id strings.",
     ),
+    disabled_at: str | None = typer.Option(
+        None,
+        "--disabled-at",
+        help=(
+            "Optional ISO-8601 disabledAt from agent get. Pass through to skip the "
+            "preserve re-read; omit to let the SDK preserve."
+        ),
+    ),
     strict_unknown: bool = typer.Option(
         True,
         "--strict/--no-strict",
@@ -229,9 +237,10 @@ def agent_update(
 ) -> None:
     """Replace AI agent configuration (``update_ai_agent``).
 
-    Full-replace of behaviors; does not reactivate a disabled agent. Use
-    ``pipefy agent toggle`` to change active status. Confirm status from the
-    response ``disabled_at`` / ``active`` fields. To re-read via ``agent get``,
+    Full-replace of behaviors; does not intentionally reactivate a disabled agent.
+    Pass ``--disabled-at`` from ``agent get`` (``disabledAt``) to preserve without an
+    extra read. Use ``pipefy agent toggle`` to change active status. Confirm status
+    from the response ``disabled_at`` / ``active`` fields. To re-read via ``agent get``,
     use agent ``disabledAt`` (null means active).
     """
     behavior_list = _parse_behaviors_json(behaviors)
@@ -252,6 +261,7 @@ def agent_update(
             instruction=inst,
             behaviors=expanded,
             data_source_ids=data_source_ids,
+            disabled_at=disabled_at.strip() if disabled_at else None,
         )
     except ValidationError as exc:
         raise typer.BadParameter(str(exc)) from exc
--- a/packages/cli/src/pipefy_cli/commands/agent.py
+++ b/packages/cli/src/pipefy_cli/commands/agent.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from datetime import datetime, timezone
 from typing import Any
 
 import typer
@@ -109,6 +110,14 @@ def agent_create(
         "--data-sources",
         help="Optional JSON array of knowledge-source id strings.",
     ),
+    active: bool = typer.Option(
+        True,
+        "--active/--inactive",
+        help=(
+            "Create enabled (default) or inactive. --inactive sets disabled_at on "
+            "create and the chained update so the second step does not revive."
+        ),
+    ),
     strict_unknown: bool = typer.Option(
         True,
         "--strict/--no-strict",
@@ -121,7 +130,12 @@ def agent_create(
         False, "--json", "-j", help="Print machine-readable JSON to stdout."
     ),
 ) -> None:
-    """Create and configure an AI agent (``create_ai_agent`` + ``update_ai_agent`` chain)."""
+    """Create and configure an AI agent (``create_ai_agent`` + ``update_ai_agent`` chain).
+
+    Agents are active by default. Pass ``--inactive`` to start disabled. Confirm
+    status from the response ``disabled_at`` / ``active`` fields. To change status
+    later, use ``pipefy agent toggle`` (routine update preserves disabled state).
+    """
     behavior_list = _parse_behaviors_json(behaviors)
     ds_raw = parse_json_value(data_sources, "--data-sources") if data_sources else None
     data_source_ids: list[str] = []
@@ -131,6 +145,7 @@ def agent_create(
         data_source_ids = list(ds_raw)
 
     inst = normalize_pipefy_ai_instruction_tokens(instruction.strip())
+    disabled_at = None if active else datetime.now(timezone.utc).isoformat()
     try:
         expanded = expand_behaviors_placeholders(behavior_list)
         validated = CreateAiAgentInput(
@@ -139,6 +154,7 @@ def agent_create(
             instruction=inst,
             behaviors=expanded,
             data_source_ids=data_source_ids,
+            disabled_at=disabled_at,
         )
     except ValidationError as exc:
         raise typer.BadParameter(str(exc)) from exc
@@ -160,12 +176,17 @@ def agent_create(
             instruction=validated.instruction,
             behaviors=validated.behaviors,
             data_source_ids=validated.data_source_ids,
+            disabled_at=validated.disabled_at,
+            preserve_disabled_at=not active,
         )
-        await client.update_ai_agent(update_input)
+        update_result = await client.update_ai_agent(update_input)
+        result_disabled_at = update_result.get("disabled_at")
         out: dict[str, Any] = {
             "success": True,
             "agent_uuid": agent_uuid,
             "message": f"Created agent {agent_uuid}",
+            "disabled_at": result_disabled_at,
+            "active": update_result.get("active", result_disabled_at is None),
         }
         if pre.get("warnings"):
             out["preflight"] = pre
@@ -206,7 +227,12 @@ def agent_update(
     ),
     json_out: bool = typer.Option(False, "--json", "-j"),
 ) -> None:
-    """Replace AI agent configuration (``update_ai_agent``)."""
+    """Replace AI agent configuration (``update_ai_agent``).
+
+    Full-replace of behaviors; does not reactivate a disabled agent. Use
+    ``pipefy agent toggle`` to change active status. Confirm status from the
+    response ``disabled_at`` / ``active`` fields (or re-read via ``agent get``).
+    """
     behavior_list = _parse_behaviors_json(behaviors)
     ds_raw = parse_json_value(data_sources, "--data-sources") if data_sources else None
     data_source_ids: list[str] = []
--- a/packages/cli/src/pipefy_cli/commands/agent.py
+++ b/packages/cli/src/pipefy_cli/commands/agent.py
@@ -231,7 +231,8 @@ def agent_update(
 
     Full-replace of behaviors; does not reactivate a disabled agent. Use
     ``pipefy agent toggle`` to change active status. Confirm status from the
-    response ``disabled_at`` / ``active`` fields (or re-read via ``agent get``).
+    response ``disabled_at`` / ``active`` fields. To re-read via ``agent get``,
+    use agent ``disabledAt`` (null means active).
     """
     behavior_list = _parse_behaviors_json(behaviors)
     ds_raw = parse_json_value(data_sources, "--data-sources") if data_sources else None
--- a/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
@@ -182,7 +182,9 @@ class AiAgentTools:
             Pass ``active=False`` to create inactive (sets ``disabled_at`` on create and the
             chained update so the second step does not revive). Confirm status from the
             response ``disabled_at`` / ``active`` fields - no extra get required. To change
-            status later, use ``toggle_ai_agent_status`` (not ``update_ai_agent``).
+            status later, use ``toggle_ai_agent_status`` (not ``update_ai_agent``). An agent
+            with no active behavior is disabled by the API regardless of this flag
+            (``BehaviorInput.active`` defaults to true).
 
             Discovery workflow (call these tools first):
               1. ``get_pipe(pipe_id)`` → obtain ``uuid`` (use as ``repo_uuid``) and phase IDs.
@@ -350,7 +352,7 @@ class AiAgentTools:
                 behaviors=validated.behaviors,
                 data_source_ids=validated.data_source_ids,
                 disabled_at=validated.disabled_at,
-                preserve_disabled_at=not active,
+                preserve_disabled_at=False,
             )
             try:
                 update_result = await client.update_ai_agent(update_input)
@@ -372,6 +374,7 @@ class AiAgentTools:
                 return build_create_agent_partial_failure(
                     agent_uuid=agent_uuid,
                     error=error_text,
+                    disabled_at=create_result.get("disabled_at"),
                 )
 
             msg = f"AI Agent created and configured successfully. UUID: {agent_uuid}"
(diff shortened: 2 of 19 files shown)


adriannoes:
## Follow-up from internal code review

We ran an internal code-review pass on this PR (checkout at `4bc63e49`, local gates + hunts + CLI live smoke on internal pipe). Nothing blocked merge; we chose to land the optional docs nits here instead of posting review threads.

### What changed (2 commits)

1. **Skill** (`skills/ai-agents/pipefy-ai-agents/SKILL.md`)
   - Create/update success still uses `disabled_at` / `active`.
   - Get / re-read now correctly uses agent `disabledAt` (null means active).
   - Explicit warning: `behaviors[].active` is **not** agent enablement (live smoke showed a disabled agent with `behaviors[0].active: true`).
   - CLI flags `--active` / `--inactive` scoped to **create** only; update status stays on `agent toggle`.
   - Partial-failure recovery aligned with preserve + toggle.

2. **CLI help** (`pipefy agent update` + help golden)
   - Same get vs write vocabulary bridge so help matches the skill.

### Why

Agents following the earlier skill wording could look for write-envelope keys on `get_ai_agent` / `agent get`, or trust `behaviors[].active`, and mis-report a disabled agent as enabled. Create/update lifecycle behavior was already correct; this is enablement/docs hygiene from the review.

### Live smoke (CLI, this branch)

- create active → `active: true`; create `--inactive` → stays disabled after update; toggle off + update → still disabled; agents deleted afterward. 
- Cursor MCP was not used for `active=` (connected server schema still lacks that arg until restart against this branch).

No further review threads needed from that pass for these items.

packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
```diff
@@ -335,9 +349,11 @@ async def create_ai_agent(
                 instruction=validated.instruction,
                 behaviors=validated.behaviors,
                 data_source_ids=validated.data_source_ids,
+                disabled_at=validated.disabled_at,
+                preserve_disabled_at=not active,
             )
             try:
-                await client.update_ai_agent(update_input)
+                update_result = await client.update_ai_agent(update_input)
```
mocha06, line 358: Partial-failure recovery now silently yields a **disabled** agent, which is a behavior regression on this branch.

If this update raises, the `except` returns `build_create_agent_partial_failure` with the uuid and no status — but the shell created above is disabled (the API stamped `disabledAt`). The documented recovery is a plain `update_ai_agent(...)`, which now takes the preserve path, re-reads the shell's `disabledAt`, re-sends it, and leaves the agent disabled forever unless someone also calls `toggle_ai_agent_status`. Before this PR that same recovery update cleared `disabledAt` and honored `active=True`.

`SKILL.md` does mention the toggle, but the partial-failure envelope is the only text the calling model sees on this branch, and it does not.

Cheap fix: `create_result["disabled_at"]` is already in hand and currently discarded — pass it into `build_create_agent_partial_failure` and add "the agent is disabled until you toggle it" to the recovery hint.

packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
```diff
@@ -107,15 +120,26 @@ async def update_agent(self, agent_input: UpdateAiAgentInput) -> AgentServiceRes
         ]
         behaviors_with_refs = inject_reference_ids(behaviors_raw)
 
+        disabled_at = agent_input.disabled_at
+        if agent_input.disabled_at is None and agent_input.preserve_disabled_at:
+            current = await self.get_agent(agent_input.uuid)
+            raw = current.get("disabledAt")
+            current_disabled_at = raw if isinstance(raw, str) else None
+            disabled_at = resolve_update_disabled_at(None, current_disabled_at)
```
mocha06: This makes every routine `update_ai_agent` a read-modify-write, and `GET_AI_AGENT_QUERY` pulls the full agent with all behaviors just to recover one timestamp. Three consequences:

1. Two round trips per update for every MCP and CLI caller.
2. A read failure now aborts a write that previously would have gone through — `execute_query` raises, so a transient error on the GET fails the update.
3. A genuine race: if a `toggle` lands between the GET and the mutation, the update re-sends the stale timestamp and silently re-disables a just-activated agent. The mirror case reactivates one that was just disabled.

The documented workflow already tells callers to `get_ai_agent` first, so they usually hold `disabledAt` already. Exposing `disabled_at` as a pass-through on the `update_ai_agent` tool would skip the second read on the common path and close the race for callers that use it.

packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
```diff
@@ -381,6 +401,11 @@ async def update_ai_agent(
             one action (same constraint and shape as ``create_ai_agent`` — see its docstring for the
             full behavior dict example, discovery workflow, and constraints).
 
+            Active lifecycle: a routine update never reactivates (or deactivates) the agent.
+            ``disabled_at`` is preserved by the SDK when omitted. Use ``toggle_ai_agent_status``
+            to activate or deactivate. Confirm status from the response ``disabled_at`` /
+            ``active`` fields.
```
mocha06: "never reactivates (or deactivates)" is too absolute.

When the agent is currently active the SDK omits `disabledAt`, and the API then derives status from the payload: an update whose behaviors all carry `active: false` **deactivates** the agent. Same mechanism undercuts the create docstring at line 180 — `active=True` is not honored if every behavior in the payload is `active: false`, and the tool reports the resulting `disabled_at` without explaining why.

`BehaviorInput.active` defaults to true so this is not a bug, but since these docstrings are the model's only contract, suggest qualifying both with something like "an agent with no active behavior is disabled by the API regardless of this flag".

packages/cli/src/pipefy_cli/commands/agent.py
```diff
@@ -160,12 +176,17 @@ async def factory(client: PipefyClient):
             instruction=validated.instruction,
             behaviors=validated.behaviors,
             data_source_ids=validated.data_source_ids,
+            disabled_at=validated.disabled_at,
+            preserve_disabled_at=not active,
```
mocha06: `not active` reads as if it matters here, but it never does. When `active=False`, `disabled_at` is explicitly provided, so `update_agent`'s guard (`disabled_at is None and preserve_disabled_at`) short-circuits and the flag is never consulted. `False` is the honest value in **both** create cases — the create chain never wants a re-read.

Worth fixing because the new tests lock the confusing value in (`assert update_arg.preserve_disabled_at is True` in both lifecycle suites). If the two conditions in `update_agent` are ever reordered to check `preserve_disabled_at` first, the create chain starts issuing an extra read and re-preserving the shell timestamp, and these tests still pass.

Same at `packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py:353`.

packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
```diff
@@ -54,6 +54,15 @@ def inject_reference_ids(behaviors: list[dict[str, Any]]) -> list[dict[str, Any]
     return result
 
 
+def resolve_update_disabled_at(provided: str | None, current: str | None) -> str | None:
+    """Choose ``disabledAt`` for updateAiAgent; never clear (activation is toggle-only)."""
+    if provided is not None:
+        return provided
+    if current:
+        return current
+    return None
```
mocha06: The `provided` parameter is dead. The only call site is `resolve_update_disabled_at(None, current_disabled_at)`, reached *after* the caller has already checked `disabled_at is None` — so `if provided is not None` is unreachable in production and the helper reduces to `current or None`.

Either inline it, or move the whole `provided` / `preserve` decision into the helper and call it unconditionally, so the branch it exists to express is actually reachable from the real path.

mocha06 (commented):
Checked the lifecycle model against how the API actually treats `disabledAt` on these mutations, not just against the diff. The core model is right, and worth stating explicitly because it is not obvious from reading the code:

- The API branches on **key presence** for `disabledAt`, not on null. So conditionally omitting the key in `create_agent` / `update_agent` is exactly the correct shape — those `if ... is not None` guards are load-bearing, not defensive.
- An update that omits `disabledAt` has it recomputed, so re-sending really is required to preserve it. The extra read is not gratuitous.
- `updateAiAgentStatus` is the only path that clears the timestamp, so routing activation through `toggle` is correct.
- `datetime.now(timezone.utc).isoformat()` round-trips through the `DateTime` scalar cleanly (microseconds and `+00:00` both survive), so the preserve path is safe.

One nuance the PR does not account for: the API clears `disabledAt` on update only when at least one behavior in the payload carries `active: true` — otherwise it *sets* it. `BehaviorInput.active` defaults to true, so every happy path here works, but it makes two of the new docstring claims non-absolute. Noted inline.

Two of the five comments are worth acting on before merge — the partial-failure envelope and the read-modify-write on update. The other three are wording and cleanup.

packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
adriannoes: Agreed — this was a real recovery regression. Fixed in b02d6a8b: `create_result["disabled_at"]` is now passed into `build_create_agent_partial_failure`, the envelope exposes `disabled_at` / `active`, and the error message tells the model the shell stays disabled until `toggle_ai_agent_status` after a successful `update_ai_agent` recovery.

packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
adriannoes: Good call. Kept auto-preserve as the safe default (so callers that omit status do not reintroduce #520), and exposed optional `disabled_at` pass-through on MCP `update_ai_agent` / CLI `--disabled-at` so the documented get-then-update path skips the second read and closes the toggle race when the caller already holds the timestamp.

packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
adriannoes: Qualified both create and update (and the toggle) docstrings: preserve is intentional for enablement, but an agent with no active behavior is still disabled by the API (`BehaviorInput.active` defaults to true). Skills/docs updated the same way.

packages/cli/src/pipefy_cli/commands/agent.py
adriannoes: Fixed — create chain now always sets `preserve_disabled_at=False` (MCP + CLI). The inactive path already passes an explicit `disabled_at`, so the flag was never consulted; tests now lock the honest value.

packages/sdk/src/pipefy_sdk/services/ai_agent_service.py
adriannoes: Moved the full `provided` / `preserve` / `current` decision into `resolve_update_disabled_at` and call it unconditionally after the optional get. With the update pass-through, the `provided` branch is reachable again on the real path.

adriannoes:
Really helpful pass, especially the API nuance on key-presence vs null for `disabledAt`, and the two pre-merge items.

Addressed all five threads in **b02d6a8b**:

| Thread | Change |
|--------|--------|
| Partial-failure envelope | Pass shell `disabled_at` + toggle recovery hint |
| RMW on update | Optional `disabled_at` / `--disabled-at` pass-through; keep auto-preserve when omitted |
| Absolute docstring | Qualified create/update/toggle for `BehaviorInput.active` |
| `preserve_disabled_at=not active` | Always `False` on the create chain |
| Dead `provided` helper arg | Decision folded into `resolve_update_disabled_at` |

Could you take another look when you have a moment, @mocha06? 

Thanks again 👍🏻

packages/sdk/src/pipefy_sdk/models/ai_agent.py
```diff
@@ -431,3 +440,5 @@ class UpdateAiAgentInput(BaseModel):
     )
     instruction: str | None = None
     data_source_ids: list[str] = Field(default_factory=list)
+    disabled_at: str | None = None
```
mocha06: Non-blocking nit, and partly a consequence of my own suggestion on the helper.

Moving the decision into `resolve_update_disabled_at` replaced the old `if current:` truthy guard with `if preserve: return current`, so a blank string is no longer filtered — and this field is plain `str | None`, so a blank value survives validation and reaches the payload as `disabledAt: ""`:

```
blank provided -> ''
blank current  -> ''
model accepts blank disabled_at -> ''
```

Newly reachable from caller input rather than only in theory: `--disabled-at "  "` strips to `""`, and the MCP `disabled_at` arg isn't stripped at all. Worst case is an opaque `DateTime` coercion error from the API instead of a clear validation message — no corruption, and never a wrong enablement state, which is why it isn't blocking.

Cheapest fix is to match the convention this model already uses on `uuid` / `name` / `repo_uuid`:

```python
disabled_at: NonBlankStr | None = None
```

That rejects it at the boundary and lets the helper stay as written.

mocha06 (dismissed):
Approving. Checked each fix against the code rather than the summary, fast-forwarded to `b02d6a8b`, and ran the agent suites plus the CLI help golden — 186 passed.

All five hold up:

- Partial failure now passes the shell `disabled_at` through, and the hint fires only when the shell is actually disabled — right condition, not just the right key.
- The pass-through closes both the second round trip and the toggle race for callers that already hold the timestamp.
- You qualified `toggle_ai_agent_status` too, which I hadn't flagged.
- The create-chain flag is honest now, and the test split into a `False` create case and a `True` routine-update case, so the two paths can't drift into each other unnoticed.
- `resolve_update_disabled_at` is reachable on the real path, keyword-only, with unit tests on all three branches.

Two of your calls I want to explicitly endorse rather than leave ambiguous, since both look like open questions from the outside:

**Auto-preserve staying the default is correct.** A caller who omits status can't silently reintroduce #520, which matters more than saving a round trip on a path the model may not take.

**A failed preserve-read aborting the write is correct, not a regression.** I listed it as a consequence, but the only fallback would be omitting `disabledAt`, which *clears* — so failing closed is the right behavior. Nothing to change.

One non-blocking nit inline on the blank-string edge. Fine to fold into a follow-up or ignore; it can't corrupt state, only produce a worse error message.

packages/sdk/src/pipefy_sdk/models/ai_agent.py
adriannoes: Good catch — fixed in the tip: `disabled_at` on both `CreateAiAgentInput` and `UpdateAiAgentInput` is now `NonBlankStr | None`, matching `uuid` / `name` / `repo_uuid`. Blank and whitespace-only values raise at the Pydantic boundary (with strip), so `--disabled-at \"  \"` / MCP blank no longer reach the payload as `disabledAt: \"\"`. Model tests cover reject + strip.

mocha06 (approved):
Re-approving — the earlier approval was auto-dismissed when `0362ffb` landed, not withdrawn.

The nit is fully closed, and on both models rather than only the one I pointed at. `NonBlankStr` carries a `BeforeValidator` that strips, so a padded timestamp is normalized instead of rejected, and the blank cases now fail at the Pydantic boundary — where the CLI already converts a `ValidationError` into a clear `BadParameter` instead of an opaque `DateTime` coercion error from the API:

```
''                              -> ValidationError (rejected at boundary)
'   '                           -> ValidationError (rejected at boundary)
None                            -> None
'  2026-08-04T12:00:00+00:00  ' -> '2026-08-04T12:00:00+00:00'
```

Covering `CreateAiAgentInput` too is the right call even though its value is always generated internally — it removes the asymmetry rather than relying on the caller staying internal.

Gates at `0362ffb`: 3511 passed / 39 skipped, `ruff` clean, import-linter 2 contracts kept. Nothing further from me.

PR #572, opened 2026-08-04
fix(release): refuse a cut when main has commits absent from dev

### LINKED ISSUE #528: fix(release): release-pr must fail when main has commits absent from dev
## Summary

`release.py release-pr` cuts the release branch from `origin/dev` alone. When `main` carries commits that were never back-merged into `dev`, the cut silently omits their **release notes**, and the guard that exists to catch exactly this cannot see them.

## Current behaviour

`release_pr` branches from `dev`:

```python
run(["git", "fetch", "--quiet", "origin", DEV_BRANCH, MAIN_BRANCH])
...
run(["git", "checkout", "-b", branch, f"origin/{DEV_BRANCH}"])
```

`main` is fetched only to compare versions. The one guard against a stale `dev` (`scripts/release.py:334-340`) already states the right remedy:

```python
if not _target_ahead_of_main(target, main_version):
    raise ReleaseError(
        f"Computed target v{target} is not ahead of origin/{MAIN_BRANCH} "
        f"(v{main_version}); origin/{DEV_BRANCH} (v{current}) is behind "
        f"{MAIN_BRANCH}. Back-merge {MAIN_BRANCH} into {DEV_BRANCH} before "
        "cutting a release."
    )
```

But it detects via **version strings** — `_target_ahead_of_main` is `Version(target) > Version(main_version)` (`release.py:261-265`). A commit merged straight to `main` that does not bump the version is invisible to it.

That is not hypothetical. #455 (`fix(release): keep marketplace.json version in lockstep`) and #457 (hosted-MCP onboarding docs) were both merged directly into `main` on 2026-07-21 without touching the version. Had a release been cut from `dev` before they were back-merged, the version guard would have passed and their entries would have been left out of the stamped section.

## Why it matters

The commits themselves are not lost — the release PR merges into `main`, so `main`-only work survives, and GitHub tests the merge result. Two things do go wrong:

- **The changelog stamp is wrong.** `_dev_unreleased_body()` reads `## [Unreleased]` from `origin/dev` only, so `main`-only entries never reach the stamped `## [X.Y.Z]` section. The released notes then under-report what shipped, and the orphaned entries sit under `## [Unreleased]` on `main` afterwards.
- **An untested combination gets tagged.** `publish` tags `main` after the merge, so the tag covers a tree whose `main`-only and `dev` halves were only ever exercised together by the PR merge check.

The failure is quiet in the way that matters: nothing errors, and the gap only shows up when someone reads the release notes later and finds a change missing.

## Proposed fix

Extend the existing guard from a version comparison to a commit comparison, keeping the same error and the same remedy:

```python
behind = int(capture(["git", "rev-list", "--count", f"origin/{DEV_BRANCH}..origin/{MAIN_BRANCH}"]))
if behind:
    raise ReleaseError(
        f"origin/{MAIN_BRANCH} has {behind} commit(s) not in origin/{DEV_BRANCH}; "
        f"a cut from {DEV_BRANCH} would omit them. Back-merge {MAIN_BRANCH} into "
        f"{DEV_BRANCH} before cutting a release."
    )
```

It must run **before** the branch is created, alongside the existing version guard, so a bad state costs nothing to walk away from.

### Explicitly not auto-merging `main` into the release branch

Considered and rejected:

- It would leave `dev` still behind `main`. The merge would live only in the throwaway release branch, so the divergence persists and the next cut hits the same thing — the symptom gets fixed every release instead of once.
- It moves a conflict past the point of no harm. `release-pr` is deliberately split so everything before the tag push is local and reversible; an auto-merge can strand the operator after the branch exists. An assertion fires before anything is created.
- The bump is computed from `_version_at("origin/dev")` before any merge would run, so the target would be derived from a tree that no longer exists.
- It does not fix the changelog. `_dev_unreleased_body()` reads `dev`'s `Unreleased` regardless, so merging the commits would ship the code without its release notes. Forcing the back-merge first is what makes the stamp correct.

## Done when

- `release-pr` fails fast, before creating or pushing anything, when `origin/main` has commits absent from `origin/dev`
- the error names the count and prescribes the back-merge
- a clean `dev` (no `main`-only commits) is unaffected
- coverage sits beside the existing `test_target_ahead_of_main` in `tests/test_release.py`
- `RELEASE.md` notes the precondition in the `release-pr` section

## Notes

Worth deciding while implementing: whether `prepare` needs the same guard. It runs on `main` directly rather than branching from `dev`, so the failure mode differs — confirm rather than assume it is unaffected.

mocha06:
## Motivation

`release-pr` cuts the release branch from `origin/dev` alone, and the released notes come from `dev`'s `## [Unreleased]` section alone. The one guard against a stale `dev` compared version strings, so it only saw divergence that had moved the version. Anything merged straight into `main` without a bump — a doc fix, a CI workflow, a LICENSE — passed it unnoticed.

The result was quiet in the way that matters: nothing errored, the commits themselves survived (the release PR merges into `main`), but their entries were stamped out of the release section and left orphaned under `## [Unreleased]` on `main` afterwards. The gap only surfaced later, when someone read the notes and found a change missing. The tag also covered a combination that only the PR merge check had ever exercised.

This is not hypothetical: `origin/main` currently carries 9 commits absent from `origin/dev` (LICENSE, DCO workflow, security and conduct docs, README banner), none of which touched the version. A release cut today would have dropped all of their notes.

## Outcome

`release-pr` now refuses when `origin/main` has commits absent from `origin/dev`, naming the count and prescribing the back-merge:

```
origin/main has 9 commit(s) not in origin/dev; a cut from dev would omit them.
Back-merge main into dev before cutting a release.
```

The check runs before the branch is created and before the confirmation prompt, so a stale `dev` costs nothing to walk away from — no branch, no push, nothing to unwind. A `dev` that already contains `main` is unaffected.

`alpha-pr` enforces the same precondition, since it delegates to `release_pr`. A staging alpha that is missing what `main` already shipped does not represent what will ship, which matches the existing version guard's stated scope ("Applies to an alpha too"). The practical consequence is that alphas are blocked until someone back-merges.

`prepare` is deliberately left alone. It runs on `main`, `preflight_prepare` requires local `main` to match `origin/main`, and it stamps `main`'s own section — it never consults `dev`, so the failure mode does not exist there. Confirmed by reading the path rather than assumed.

Auto-merging `main` into the release branch was considered and rejected for the reasons in #528: it would leave `dev` still behind, move a conflict past the point of no harm, derive the bump from a tree that no longer exists, and still ship the code without its notes, since the stamp reads `dev`'s `## [Unreleased]` regardless. Forcing the back-merge first is what makes the stamp correct.

## Verification

|---|---|---|---|---|
| 1 | Real repo state, `main`-only commits that never moved the version | Refuse, naming the count and the remedy | `origin/main has 9 commit(s) not in origin/dev; a cut from dev would omit them. Back-merge main into dev before cutting a release.` | Pass |
| 2 | Guard fires before anything exists | No branch created, no push, tree untouched | No `rc-main/release/v0.5.0-beta.1`; working tree unchanged | Pass |
| 3 | `rev-list` range direction | `origin/dev..origin/main`; reversed, it would count `dev`'s own work and block every release | Pinned by unit test on the exact argv | Pass |
| 4 | Clean `dev`, beta promotion | Proceeds, branches off `origin/dev`, stamps the CHANGELOG | Existing suite passes unchanged | Pass |
| 5 | `alpha-pr` with `main`-only commits | Refuse via the same guard | Unit test | Pass |
| 6 | `prepare` | Unaffected — reads `main`'s own tree | Path read; no `dev` reference | Pass |

Scenarios 1 and 2 were run against the live repository, not stubs.

Gates: `pytest -m "not integration"` (4109 passed, 5 skipped), `ruff check`, `ruff format --check`, `lint-imports`, `bump_version.py verify`. Skills-lint is path-filtered to `skills/**` and is not triggered by this diff.

Closes #528.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - **Installer**: `install.sh` resolves the newest GitHub Release whose tag is **not** an alpha, so a staging alpha published from `dev` never becomes what a default `curl … | sh` install hands out; the public install path stays on the beta line. Alphas are also flagged as GitHub pre-releases, keeping the "Latest" badge on the newest beta. Filtering is by tag shape rather than the API's `prerelease` flag, because the whole pre-1.0 line is betas that must stay installable. `install.sh --version vX.Y.Z-alpha.N` still installs an alpha on purpose.
 
+### Fixed
+
+- **Release process**: `release-pr` (and `alpha-pr`, which delegates to it) now refuses to cut when `origin/main` carries commits absent from `origin/dev`, naming the count and prescribing the back-merge. The existing staleness guard compared version strings, so it only saw divergence that moved the version; work merged straight into `main` without a bump — a doc fix, a CI workflow — passed it unnoticed. Since the cut branches off `dev` and the released notes come from `dev`'s `## [Unreleased]` alone, those entries were stamped out of the release section and left orphaned under `## [Unreleased]` on `main`, and the resulting tag covered a combination only ever exercised by the PR merge check. The new guard runs before the branch is created, so a stale `dev` costs nothing to walk away from. `prepare` is unaffected: it runs on `main` and stamps `main`'s own section, never consulting `dev`. Closes #528.
+
 ## [0.4.0-beta.2] - 2026-07-29
 
 ### Added
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -73,6 +73,8 @@ uv run python scripts/release.py release-pr patch
 
 It reads the current version and `## [Unreleased]` from `origin/dev` (not your checked-out branch), confirms the computed target, then opens the PR. After the PR is approved and merged into `main`, cut the release from `main` with `publish` (step 4 below) — deliberately a human step, so the tag push triggers the Release workflow. `release-pr` never tags or publishes.
 
+**Precondition: `origin/dev` must contain every commit on `origin/main`.** Because the cut branches off `dev` and the released notes come from `dev`'s `## [Unreleased]` alone, anything merged straight into `main` — a doc fix, a CI workflow — would be stamped out of the release notes. `release-pr` refuses with the count and the remedy before it creates or pushes anything, so a stale `dev` costs nothing to walk away from: back-merge `main` into `dev`, then re-run. `alpha-pr` enforces the same precondition, since a staging cut that is missing what `main` already shipped does not represent what will ship.
+
 ### The bump, step by step
 
 `release-pr` automates steps 1–2 below off `dev`; they are spelled out here because the bump argument and its guards are what you actually choose. Note that `main` and `dev` both require a pull request, so a locally-prepared commit can never be pushed to either — `prepare` gives you the commit, but it still reaches the branch through a PR.
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -330,6 +330,15 @@ def _target_ahead_of_main(target: str, main_version: str) -> bool:
     return Version(target) > Version(main_version)
 
 
+def _main_only_commit_count() -> int:
+    """How many commits ``origin/main`` carries that ``origin/dev`` does not."""
+    return int(
+        capture(
+            ["git", "rev-list", "--count", f"origin/{DEV_BRANCH}..origin/{MAIN_BRANCH}"]
+        )
+    )
+
+
 def _dev_unreleased_body() -> str:
     """The text under ``## [Unreleased]`` in ``origin/dev``'s CHANGELOG."""
     m = UNRELEASED_RE.search(_show_at(f"origin/{DEV_BRANCH}", "CHANGELOG.md"))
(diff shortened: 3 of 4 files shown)


adriannoes (approved):
Reviewed at `7a9d36dc`: Phase 3 gates on a detached worktree (ruff, 52 release tests), targeted refuse/range tests, H5 scratch litmus, and hunts H1/H4/H5/H10/H13/H14/H17. CI lint/test/DCO green on this head.

### What worked well

- Guard sits after `git fetch` and before `confirm` / `checkout -b`, so a stale `dev` costs nothing to walk away from.
- Unit test pins `origin/dev..origin/main` (reversed would block every healthy cut).
- `alpha_pr` inherits via delegation; `prepare` correctly left alone; RELEASE.md and CHANGELOG match the code.

<details>

Tier `standard`, trust `trusted`, reviewed at `33c75934…7a9d36dc`.

| Gate | Result |
| :--- | :--- |
| CI (lint/test/DCO) | green |
| Local uv sync / ruff / pytest release | pass (52) |
| Targeted main-only refuse tests | 3 passed |
| H5 revert litmus | failed-as-expected |
| Live MCP/CLI smoke | n/a (maintainer script) |

Execution matrix: S1–S3/S5–S7 pass; S4 waived (prepare unchanged).

Not covered: live `release-pr` against real remotes (would mutate branches/PRs; unit + author live scenarios 1–2 in the PR body already cover the refuse path).

</details>

PR #573, opened 2026-08-04
ci: run the packaging smoke install daily, and against published wheels

### LINKED ISSUE #531: ci: run the packaging smoke install on a schedule to catch upstream breakage
## Summary

The packaging smoke install is the only gate that resolves dependencies from PyPI instead of `uv.lock`, and it runs only on push and pull request. A dependency breakage published upstream therefore stays invisible until someone happens to open a PR — or, as happened with `mcp` 2.0.0, until a release is attempted.

Run that smoke install on a schedule so upstream breakage surfaces on its own timeline rather than on the release's.

## Why

`uv.lock` pins every dependency, so the whole test suite passes against known-good versions no matter what is published. That is the point of a lock, but it means **no test in this repository can observe a fresh install**. The packaging smoke step (`.github/workflows/ci.yml`) is the sole exception: it installs the built wheels into a clean virtualenv with `pip`, which resolves from PyPI.

The `mcp` 2.0.0 incident is the worked example. `mcp[cli]>=1.25.0` had no upper bound, 2.0.0 restructured the SDK and removed `mcp.server.fastmcp`, and `pipefy-mcp-server` stopped starting. Nothing in the repository changed. The failure armed itself when the upstream release was published and would have detonated on whatever release came next, regardless of content. It was caught by the release workflow's own smoke step, which is the last possible moment — after the tag is pushed, and with the GitHub Release already created.

A repository between merges can go from green to broken-on-release with zero commits, and nothing reports it.

## Proposed change

Add a scheduled trigger that runs the packaging build plus the smoke install (build wheels, install into a clean venv, launch every published entry point). Daily or weekly are both defensible; weekly is probably enough given releases are not continuous.

- Whether to reuse the existing `test` job via a `schedule` trigger with a conditional, or add a separate lightweight job that only builds and smoke-installs. The latter avoids paying for the full suite on a timer.
- How failure is surfaced. A red scheduled run that nobody watches is not a control. Decide whether it opens an issue, notifies, or is expected to be seen — and say so in the workflow.
- Whether it should also run against the *published* wheels from PyPI rather than locally built ones, which would additionally catch a broken upload.

## Done when

- the packaging smoke install runs on a schedule, independent of pushes and pull requests
- it launches every published console entry point, not only `pipefy`
- a failure is surfaced somewhere a human will see it, and the mechanism is stated in the workflow
- `RELEASE.md` or `CONTRIBUTING.md` notes that this gate exists and what it protects

## Notes

Related: #528 (release-pr must fail when main has commits absent from dev). Both are about a release-time surprise that could have been a routine signal earlier.

A known limitation to record rather than solve here: the smoke install proves process startup only. A dependency release that breaks behaviour without breaking imports — for example the tool-call middleware's use of a FastMCP internal (`request_handlers[CallToolRequest]`) — would pass this gate and would not be seen by the suite either, since the suite runs against the lock. Closing that band means running some portion of the tests unlocked, which is a larger decision than this issue.

mocha06:
## Motivation

Nothing in this repository can see a broken fresh install between merges.

`uv.lock` sets an exact version for each dependency, so the tests always run against known-good versions. One CI step installs from PyPI instead, and that step ran only on a push or a pull request. Between merges, no job resolved a workspace dependency from the index.

A dependency released inside that window therefore broke the next release, whatever that release contained. `mcp` 2.0.0 is the worked example: the constraint had no upper bound, 2.0.0 deleted the module the server imports at startup, and `pipefy-mcp-server` stopped starting with no code change here. The release workflow found it after the tag was public and the GitHub Release existed.

The published wheels had no gate at all. `release.yml` installs the locally built wheels, and `release.py verify` installs from PyPI only after the tag ships.

## Outcome

`packaging-smoke.yml` runs the fresh-install gate once a day, and on demand. Two jobs run, because each one finds a different error:

- **Wheels built from this checkout** — finds an upstream release outside the declared bounds. That error would break the next release.
- **Wheels already on PyPI** — a plain `pip install` of all five distributions by name. Finds a bad upload, and an upstream release that breaks the version already shipped.

A failure opens an issue labelled `packaging-smoke-failure`, or adds a comment to the open one. A break that lasts a week gives one thread, not seven issues. The workflow creates the label itself, so the repository needs no manual setup.

`scripts/smoke_entry_points.py` now holds the single definition of both halves of the smoke. Three workflows each carried a copy of the launch list.

- `--check-wheels <dir>` runs **before** the install and asserts one wheel per published member. An incomplete set must not reach `pip`: the sibling `==` pins let it satisfy the absent member from the index, so the smoke would pass against a wheel the build never produced. It also rejects an unexpected sixth member, and a duplicate wheel left over from an earlier version.
- With no arguments it reads installed distribution metadata and runs every console script. It fails when a member is absent from the install, when a wheel stops shipping its command, or when a command exits non-zero or hangs.

Two of the issue's acceptance criteria needed correction against the tree. Launching every entry point rather than only `pipefy` was already true: two console scripts exist and `ci.yml` already ran both, which landed with #529. That criterion is now enforced by structure instead of restated. The published wheels were the criterion with no coverage anywhere.

## Verification

|---|---|---|---|---|
| 1 | Built wheels, clean venv, dependencies from PyPI | Both commands run | `exit=0` — Every published console entry point launched | Pass |
| 2 | Published PyPI wheels, plain `pip install` | Resolves all five, both commands run | `exit=0` — same | Pass |
| 3 | A published member absent from the install | Fail, naming the member | `exit=1` — `missing from the install: pipefy-infra` | Pass |
| 4 | Wheel installed, console script absent from `bin` | Fail, naming the script and path | `exit=1` — `pipefy is declared in the installed metadata, but …/bin/pipefy does not exist` | Pass |
| 5 | Dependency absent, so the import fails (the `mcp` 2.0.0 shape) | Fail, carrying the exit code | `exit=1` — `pipefy-mcp-server --help exited 1`, `ModuleNotFoundError: No module named 'mcp'` | Pass |
| 6 | Built dist holds 4 of 5 wheels, no guard | — | `pip install *.whl` **succeeded**, pip resolved `pipefy-infra` from PyPI | Defect confirmed |
| 7 | That same directory, with `--check-wheels` | Fail before `pip` runs | `exit=1` — `has no wheel for: pipefy-infra. pip would resolve that member from the index instead` | Pass |
| 8 | Duplicate wheel from an earlier version | Fail | `exit=1` — `more than one wheel for: pipefy-cli` | Pass |
| 9 | Unexpected sixth member wheel | Fail | `exit=1` — `unexpected wheel for: pipefy-extra` | Pass |
| 10 | Reporter, no open issue | Create the issue with the label | `issue create --title … --label packaging-smoke-failure --body-file` | Pass |
| 11 | Reporter, an issue already open | Comment on it, no second issue | `issue comment 999 --body-file` | Pass |
| 12 | Plain install without `--pre` | — | Resolved all five; pip falls back to a pre-release when no stable release exists | Assumption corrected |

Rows 1 to 9 ran against real virtualenvs and real built wheels. Rows 10 and 11 ran against a stubbed `gh` that exercises both branches. Row 12 disproved a `--pre` flag this branch first carried, so the flag is gone: a plain install is what a user gets, and it tracks the stable release after 1.0 with no edit.

Row 6 is the review finding, and it is my regression. Replacing five per-member globs with `*.whl` removed a fail-closed property. Today a stale published `pipefy-infra` masks it by failing the import, so the run happens to go red. It greens whenever the published artifact is compatible, which is the normal state right after a release.

Gates: `pytest -m "not integration"` (4140 passed, 5 skipped), `ruff check`, `ruff format --check`, `lint-imports`, per-package upward-import lints, `bump_version.py verify`, `lint_skill_refs.py`, plus a YAML parse and `bash -n` over every `run:` block in all five workflows. That last one caught an apostrophe inside a heredoc inside `$( )`, which is a parse error. Skills-lint is path-filtered to `skills/**` and this diff does not trigger it.

## Notes

`schedule` and `workflow_dispatch` fire only from the default branch, so neither runs while this sits on `dev`. The `pull_request` trigger is path-filtered and does run, which is how both jobs greened on this pull request. Treat the daily observer as live only once the workflow reaches `main`.

`release.yml` now calls the shared script for the launch half. That touches the publish path. Its own one-wheel-per-member check still runs before the smoke install, so the guard there is unchanged.

`.gitignore` needed `!scripts/smoke_entry_points.py`. The pattern `scripts/*` blocks every file except named ones, so without that line git never adds the script.

The gate proves process startup, not behavior. A dependency release that breaks behavior without breaking imports passes it, and the locked suite does not see it either.

Closes #531.


.github/workflows/ci.yml
```diff
@@ -73,18 +73,14 @@ jobs:
...
       run: |
         python -m venv .venv-cli-smoke
-        .venv-cli-smoke/bin/pip install dist-packaging-smoke/pipefy_infra-*.whl \
-          dist-packaging-smoke/pipefy-*.whl \
-          dist-packaging-smoke/pipefy_auth-*.whl \
-          dist-packaging-smoke/pipefy_cli-*.whl \
-          dist-packaging-smoke/pipefy_mcp_server-*.whl
-        .venv-cli-smoke/bin/pipefy --help
-        .venv-cli-smoke/bin/pipefy-mcp-server --help
+        .venv-cli-smoke/bin/pip install dist-packaging-smoke/*.whl
```
adriannoes, line 88: **Blocking.** This install step used to list one glob per published member. An unmatched glob made `pip` fail before smoke ran. It now installs `dist-packaging-smoke/*.whl`, so a missing workspace wheel is no longer a hard error: sibling `==` pins can pull that member from the index, and with a compatible wheel `scripts/smoke_entry_points.py` exits 0 even though that artifact was not in the checkout-built `dist/` (reproduced with a four-wheel subset plus a find-links infra wheel). The same shape is in `packaging-smoke.yml` `built_wheels`. `release.yml` still checks one wheel per member before smoke.

I would restore membership fail-closed on both built-wheel installs (per-member globs again, or the release-style one-of-each check) before `pip install`.

Done when:

- incomplete `dist-packaging-smoke/` fails the job before or during install without relying on the index for a workspace member
- `ci.yml` and `packaging-smoke.yml` `built_wheels` both have that gate
- happy-path CI packaging smoke still greens

.github/workflows/packaging-smoke.yml
```diff
@@ -0,0 +1,159 @@
...
+            --color b60205 \
+            --description "A scheduled packaging smoke install failed" \
+            --force
+          # Written to a file rather than a $(cat <<EOF) capture: an apostrophe
+          # inside a heredoc inside a command substitution is a parse error on
+          # older bash, and prose that cannot contain one is prose that will be
+          # edited wrong later.
+          body="$RUNNER_TEMP/packaging-smoke-body.md"
+          cat > "$body" <<EOF
+          A scheduled packaging smoke install failed. This gate resolves dependencies from PyPI instead of \`uv.lock\`, so it is the only one that sees a fresh install — a green test suite does not contradict it.
```
adriannoes: **Optional.** The failure issue body always says a "scheduled" packaging smoke failed, including when the run was `workflow_dispatch`. I would say "packaging smoke" without "scheduled", or branch on `github.event_name`. Your call.

tests/test_smoke_entry_points.py
```diff
@@ -0,0 +1,245 @@
...
+        tmp_path: Path,
+    ) -> None:
+        for name in ("pipefy", "pipefy-mcp-server"):
+            (tmp_path / name).touch()
+        monkeypatch.setattr(_smoke, "distributions", _full_install)
+        monkeypatch.setattr(_smoke.sysconfig, "get_path", lambda _name: str(tmp_path))
+        monkeypatch.setattr(_smoke.subprocess, "run", FakeRunner())
+        assert _smoke.main() == 0
+        assert (
+            "Every published console entry point launched." in capsys.readouterr().out
```
adriannoes, line 289: **Optional.** `test_succeeds_when_every_entry_point_launches` checks exit 0 and the success banner, but never that `FakeRunner` was called. Stripping `launch(...)` from `main` still leaves this unit test green (CI packaging smoke would catch it). I would assert `runner.calls` (or keep a handle to the FakeRunner) so the `main` to `launch` wire cannot regress in the unit suite alone. Your call whether to take this.

adriannoes (changes requested):
Reviewed at `c8529fb`: local smoke unit suite (21), workflow YAML/`bash -n`, GitHub CI green on the happy path, plus a targeted incomplete-`dist/` repro of the built-wheel install seam. H16 n/a.

### Required before merge

- Built-wheel installs must fail closed when a published member wheel is missing from the dist directory. See the thread on `.github/workflows/ci.yml:83` (same fix needed in `packaging-smoke.yml` `built_wheels`).

### Decisions I would make

- After the install-seam fix, treat #531 Done-when as satisfied for landing on `dev`, but do not treat the daily observer as live until `packaging-smoke.yml` exists on default branch `main`. Note that on #531 at close, or keep a promotion checkbox.
- Published-wheels and daily cadence are intentional expansions of #531, not a Done-when miss.

### Optional

Your call on all of these.

- Pointers on the main-success unit assert (`runner.calls`) and the report body "scheduled" wording.
- Soft docs: narrow "inert until promoted" to schedule/dispatch, and avoid "the one gate" once both workflows resolve from the index.

### What worked well

- Shared `smoke_entry_points.py` (metadata + `REQUIRED_SCRIPTS` floor) removes triplicated launch lists and fails closed on missing members/scripts.
- Dual jobs match distinct failure modes; report skips PRs and scopes `issues: write`.
- Strong unit coverage of the failure contracts the PR verification table claims.

<details>

Tier `standard`, trust `trusted`, reviewed at `482d5bda...c8529fb5`.

| Gate | Result |
| :--- | :--- |
| Phase 3 sync/ruff/pytest/YAML/bash -n | pass |
| GitHub CI + packaging-smoke both jobs | pass (happy path) |
| H15 security | no medium+ |
| Bugbot | no bugs |
| H1 built-wheel `*.whl` origin blindness | IMPORTANT (reproduced) |
| H16 live MCP/CLI | n/a |

Not covered: live `report` issue create (only on schedule/dispatch failure).

</details>

.github/workflows/ci.yml
mocha06: Confirmed and fixed in `ebd77a2`. This was my regression, and I reproduced the mechanism before accepting it: with four of five wheels in `dist-packaging-smoke/`, `pip install *.whl` **succeeded** and resolved `pipefy-infra` from PyPI. The install no longer fails closed, exactly as you describe.

One correction to my own repro, so the record is accurate: I could not make the run go green today. The published `pipefy-infra` 0.5.0a1 predates `ClientDeployment` (added by #551 after that alpha was cut), so the stale artifact fails the import and masks the hole. The fail-open install is the defect either way, and it greens whenever the published artifact is compatible — the normal state right after any release.

Fix: `scripts/smoke_entry_points.py --check-wheels <dir>` runs before `pip` in both `ci.yml` and `packaging-smoke.yml` `built_wheels`. I chose this over restoring per-member globs so the member list stays in one place instead of being copied back into two workflows, and so the error names the cause. It also rejects an unexpected sixth member and a duplicate wheel from an earlier version.

Against your Done-when:

- Incomplete `dist-packaging-smoke/` fails before install: `exit=1`, `has no wheel for: pipefy-infra. pip would resolve that member from the index instead, so the smoke would test an artifact this build did not produce.` Verified against the exact 4-of-5 directory from the repro.
- Both `ci.yml` and `packaging-smoke.yml` `built_wheels` have the gate.
- Happy path still greens: both jobs pass on this PR (28s and 26s), and the full suite is 4140 passed.

`release.yml` is unchanged, since its own one-of-each check already runs before its smoke install. 13 new unit tests cover the guard, including a docstring on the missing-member test recording why the check cannot run after the install.

.github/workflows/packaging-smoke.yml
mocha06: Taken in `ebd77a2`. Dropped "scheduled" from the issue body, the title, and the label description, rather than branching on `github.event_name` — `workflow_dispatch` reaches the same reporter and the wording is accurate for both without a conditional.

tests/test_smoke_entry_points.py
mocha06: Taken in `ebd77a2`. The test now keeps the `FakeRunner` handle and asserts `runner.calls` against both expected argv lists, so removing the `launch(...)` call from `main` fails in the unit suite alone.

Asserting it surfaced a second defect: `main()` defaulted to `sys.argv`, so under pytest it read pytest's own arguments and every launch-mode test hit the usage error once `--check-wheels` existed. `sys.argv` is now read only at the `__main__` entry point.

mocha06:
Review addressed in `ebd77a2`. Description rewritten too, so the Verification table now records the defect and the guard rather than only the happy path.

**Required — done.** The built-wheel install fails closed on an incomplete `dist/`. Details and the repro are in the `ci.yml` thread. Both `ci.yml` and `packaging-smoke.yml` `built_wheels` carry the gate, and the happy path still greens.

**All three optional items — taken.** The "scheduled" wording, the `runner.calls` assert, and both soft-docs points ("inert until promoted" narrowed to schedule and dispatch, "the one gate" replaced now that two workflows resolve from the index).

**On your decisions.** Agreed on both. The promotion caveat is now stated in three places rather than left to memory: the workflow header, the `RELEASE.md` gate section, and a line in the PR Notes — "Treat the daily observer as live only once the workflow reaches `main`." Say the word and I will put the same note on #531 at close, or open a follow-up issue for the promotion so it is tracked outside this PR.

Also worth recording: asserting `runner.calls` surfaced a second defect that no review comment named. `main()` defaulted to `sys.argv`, so once `--check-wheels` existed, every launch-mode test read pytest's arguments and hit the usage error. `sys.argv` is now read only at the entry point.

Suite 4127 → 4140. Not covered, unchanged from your note: the live `report` path, which needs a real failure on schedule or dispatch.

@adriannoes

adriannoes (approved):
Re-checked at `ebd77a2` after the follow-up fix.

`--check-wheels` before `pip install` in both built-wheel paths closes the incomplete-`dist/` hole. I re-ran the incomplete vs complete check locally (exit 1 / 0), and the unit suite plus CI look good on this tip. The optional notes (`runner.calls`, dropping "scheduled" from the reporter copy) landed too.

Thank you for taking the blocking point seriously, reproducing the fail-open install yourself, and fixing it with a shared guard instead of copying member lists back into the workflows. Clear write-up on the thread as well.

Issue #534, opened 2026-07-29
fix(release): publish tags without ever building or installing the wheels


mocha06:
## Summary

`scripts/release.py publish` pushes the tag — the one irreversible step in the flow — without ever having built or installed the artifacts it is about to release. Every check that touches a wheel happens afterwards, inside the tag-triggered workflow. A break that a one-minute local build would have caught instead costs a tag, a GitHub Release, and a manual cleanup.

## Current behaviour

`publish` (`scripts/release.py:466-502`) runs, in order:

```python
if current_branch() != "main": ...                # 468
if not working_tree_clean(): ...                  # 470
if tag_exists(tag): ...                           # 475
if local_tag_exists(tag): ...                     # 477

confirm(                                          # 484
    f"About to tag {tag}, push to origin, and publish to PyPI. "
    "This cannot be undone.",
    assume_yes=assume_yes,
)

run(["git", "push", "origin", "main"])            # 492
run(["git", "tag", tag])                          # 493
run(["git", "push", "origin", tag])               # 497
watch_release_workflow(tag, exclude=prior_runs)   # 499
verify(tag)                                       # 500
```

Every guard before `confirm` is about repository state — branch, working tree, tag collision. None of them touches an artifact. The first time a wheel is built or installed is on the CI runner, after the point of no return, and `verify` (`release.py:500`) inspects the *published* result rather than a candidate.

## Why it matters

`RELEASE.md:15` states the principle the command is built around:

> `scripts/release.py` drives the flow, split at the irreversible boundary — everything before the tag push is local and reversible, so you review before anything leaves your machine.

What is actually reviewable before that boundary is the release commit: the version bump and the stamped changelog (`RELEASE.md:40`, "Review the release commit (`git show HEAD`). Nothing has been pushed yet."). The wheels are not reviewable, because they do not exist yet. The guarantee holds for the parts that can be diffed and lapses for the parts that are actually shipped.

[...]

PR #578, opened 2026-08-04
fix(release): validate the wheels before publishing anything

### LINKED ISSUE #535: fix(release): the release workflow creates the GitHub Release before validating the wheels
## Summary

`.github/workflows/release.yml` creates the GitHub Release before it validates that the wheels install and run. A smoke-install failure therefore leaves a published Release with broken wheels attached to it, which has to be deleted by hand.

## Current behaviour

The job's step order (`.github/workflows/release.yml`):

| Line | Step | Publishes? |
| --- | --- | --- |
| 21 | Verify tag matches package version | |
| 42 | Build release notes body for this tag | |
| 67 | Build wheels and sdist (all workspace packages) | |
| 69 | Verify expected wheels in dist/ | |
| **97** | **Create GitHub Release** | **yes** |
| 102 | Stage PyPI wheels (SDK + CLI + MCP + Auth + Infra) | |
| 118 | Smoke install staged wheels (clean venv) | |
| **125** | **Upload to PyPI** | **yes** |

The only step that runs the artifacts is at 118, three steps after they were published:

```yaml
      - name: Smoke install staged wheels (clean venv)
        run: |
          python -m venv .venv-pypi-smoke
          .venv-pypi-smoke/bin/python -m pip install --upgrade pip
          .venv-pypi-smoke/bin/pip install dist-pypi/*.whl
          .venv-pypi-smoke/bin/pipefy --help
          .venv-pypi-smoke/bin/pipefy-mcp-server --help
```

The step at 69-96 that precedes `Create GitHub Release` checks only that `dist/` holds exactly five wheels with the expected name prefixes. It never installs one or launches an entry point.

## Why it matters

The ordering is right with respect to PyPI — the upload is irreversible, and validation precedes it — and wrong with respect to GitHub, where the same reasoning was not applied.

The `v0.4.0-beta.1` run is precisely this shape: `Create GitHub Release` succeeded, `Smoke install staged wheels (clean venv)` failed with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, and `Upload to PyPI` was skipped. PyPI was protected by the ordering; GitHub was not. The result was a published Release advertising five wheels, one of which could not start, and the cleanup — deleting the Release, deleting the tag, reconciling `main` back into `dev` — was manual (#530).

A failed release should leave nothing behind but a tag and a red run. The tag is unavoidable: the workflow is tag-triggered, so it exists before the job starts. Everything downstream of it should be atomic — either every published artifact exists, or none does.

## Proposed fix

Move the validation above `Create GitHub Release`. `Stage PyPI wheels` (102) is a prerequisite of the smoke install, so it moves with it, giving:

```
Build wheels and sdist  ->  Verify expected wheels in dist/  ->  Stage PyPI wheels
  ->  Smoke install staged wheels  ->  Create GitHub Release  ->  Upload to PyPI
```

Any future validation belongs in the same band: no step that publishes anything runs until every check has passed.

- **Which directory the smoke install reads.** `dist-pypi/` and `dist/` hold the same five wheels today; once staging is no longer sandwiched between the publish steps, installing straight from `dist/` may be simpler, but keeping `dist-pypi/` has the virtue of testing exactly the set that gets uploaded.
- **Relative order of the two publish steps.** After the move, `Create GitHub Release` still precedes `Upload to PyPI`, so a failing upload leaves a Release without matching PyPI artifacts. That gap is narrower than today's and arguably acceptable (PyPI is the irreversible one and should go last), but it should be a decision rather than a leftover.
- **A comment stating the invariant**, so the order is not reshuffled later by someone optimising step layout. The reason this band exists is not obvious from reading the steps.

This stays worth doing even with a local pre-tag check in `publish` (#534): the runner's environment is not the operator's, and a re-pushed tag or a cut driven from anywhere but a maintainer's machine still lands here first.

## Done when

- no step that publishes an artifact runs before the built wheels have been installed into a clean virtualenv and both console entry points launched
- a failing smoke install leaves no GitHub Release for the tag
- the validate-then-publish boundary is stated in a comment in the workflow
- the `.github/workflows/release.yml` row in the `RELEASE.md` automation reference describes the order, including where validation sits

## References

- Failed run: https://github.com/pipefy/ai-toolkit/actions/runs/30482434562
- #529 — the dependency cap that unblocked the release
- #530 — the manual withdrawal of the tag and Release
- #534 — the same gap in `scripts/release.py publish`, one layer earlier
- #531 — running the same smoke install on a schedule

mocha06:
Closes #535.

## Motivation

The release workflow created the GitHub Release before anything installed the wheels, so a failing smoke install left a published Release carrying a wheel that cannot start.

The ordering was right for PyPI and wrong for GitHub. The PyPI upload is irreversible, and validation ran before it. The same reasoning never reached the Release. `v0.4.0-beta.1` is that shape: `Create GitHub Release` succeeded, the smoke install then failed with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, and `Upload to PyPI` was skipped. PyPI was protected by the ordering; GitHub was not. Deleting the Release, deleting the tag, and reconciling `main` back into `dev` was manual work (#530).

A failed release must leave nothing behind but the tag and a red run. The tag is unavoidable, because the tag is what triggers the workflow.

## Outcome

The job runs in two bands:

```
validate:  build -> verify the wheel set -> stage the PyPI set -> smoke install
publish:   Create GitHub Release -> Upload to PyPI
```

A failing smoke install now leaves no GitHub Release for the tag. Staging moved up with the smoke install, because the smoke install reads what staging writes. It reads `dist-pypi/`, which is exactly the set that goes to PyPI.

`Create GitHub Release` still runs before `Upload to PyPI`. An upload cannot be taken back and a Release can be deleted, so the irreversible step goes last. A failing upload therefore leaves a Release with no matching PyPI artifacts. That gap is narrower than the one this fixes, and the comment in the file names it as a choice rather than a leftover.

A comment marks the boundary and states where new validation belongs. A comment is documentation rather than a guard, so `tests/test_release_workflow.py` asserts the ordering as well: both publish steps come after the smoke install, and no step opts out of fail-fast, which would let a publish run after a failed check by another route.

### One definition of the published-member list

The job carried two hand-written copies of "exactly one wheel per published member": an inline Python check on `dist/`, and a five-glob bash check inside the staging step. `scripts/smoke_entry_points.py --check-wheels` already is that check, and `ci.yml` and `packaging-smoke.yml` already call it. Both copies now call it too. Staging drops to `cp dist/*.whl dist-pypi/`, and adding a workspace member means editing one list.

## Docs

`RELEASE.md`: the `release.yml` row in the automation reference describes both bands and where validation sits. One sentence under "The scheduled packaging gate" implied the old order and no longer does.

## Testing

I ran the validate band locally against this checkout, step by step, including the failure shapes. I did not cut a tag, so the reordered job has not run on a runner.

|---|---|---|---|---|
| 1 | `uv build --all-packages -o dist --wheel --sdist` | 5 wheels + 5 sdists | 5 + 5 | PASS |
| 2 | `--check-wheels dist` on a complete `dist/` | exit 0, sdists ignored | exit 0, five wheels listed | PASS |
| 3 | `cp dist/*.whl dist-pypi/` | the 5 wheels, no sdists | 5 wheels | PASS |
| 4 | Smoke install `dist-pypi/*.whl` into a clean venv, launch entry points | both launch | `pipefy --help -> ok`, `pipefy-mcp-server --help -> ok` | PASS |
| 5 | `--check-wheels` with 4 of 5 wheels | exit 1, before any publish step | exit 1, `has no wheel for: pipefy-infra` | PASS |
| 6 | `--check-wheels` with an unexpected sixth wheel | exit 1 | exit 1, `unexpected wheel for: pipefy-newthing` | PASS |
| 7 | `--check-wheels` with two wheels for one member | exit 1 | exit 1, `more than one wheel for: pipefy-cli` | PASS |
| 8 | `--check-wheels` on a missing directory | exit 1 | exit 1, all five reported missing | PASS |
| 9 | The `v0.4.0-beta.1` shape: the wheel installs, the MCP entry point cannot start | smoke exits 1, both publish steps skipped | exit 1, `ModuleNotFoundError: No module named 'mcp'` at `pipefy-mcp-server --help` | PASS |
| 10 | Parsed step order | publish indices > smoke index | publish `[9, 10]` > smoke `8` | PASS |
| 11 | Fail-fast not opted out | no `continue-on-error` | none | PASS |
| 12 | Drift guard against a publish step moved ahead of the smoke install | guard fails | fails | PASS |
| 13 | Drift guard against a renamed step, and against a `continue-on-error` | guard fails loudly | fails, naming the step | PASS |

Full gates: 4174 passed, 5 skipped — the 4172 baseline plus the two drift-guard tests. `ruff check`, `ruff format --check`, `lint-imports`, `bump_version.py verify`, and the skill-ref linter all pass.

## Notes

`scripts/release.py` holds its own copy of the member list (`EXPECTED_WHEEL_STEMS`), with no test tying it to the workspace. It checks the Release after publishing, one layer away from this change, and `release.py` is what #534 covers. Worth folding in there.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
...Release was created, the smoke install failed with `ModuleNotFoundError`, the PyPI upload was skipped, and withdrawing the Release and the tag was manual work. The job now runs in two bands: build, bound the wheel set, stage the PyPI set, and install it into a clean virtualenv launching every console entry point — and only then create the GitHub Release and upload to PyPI. A validation failure now leaves nothing behind but the tag, which is unavoidable because the tag is what triggers the workflow. PyPI stays last within the publish band, since an upload cannot be taken back while a Release can be deleted — which leaves one narrower gap by choice, a failing upload leaving a Release with no matching PyPI artifacts — and a comment in the file states the boundary so later validation is not added between the two publish steps. The two hand-rolled wheel-set assertions the job carried — an inline Python check on `dist/`, and a five-glob bash check while staging — collapse onto `smoke_entry_points.py --check-wheels`, already used by the CI and scheduled-smoke workflows, so every workflow reads one definition of the published-member list rather than three that can drift apart. Closes #535.
```
adriannoes: **Optional.** The Unreleased Fixed bullet says the old job created the GitHub Release "three steps before" the smoke install. On base, after build the order was Verify, Create GitHub Release, Stage, then Smoke, so Release sat one step (Stage) before install, not three. The orphan-Release story for `v0.4.0-beta.1` is still right. Your call whether to drop or correct the step count.

adriannoes (dismissed):
Reviewed at `977c79e`: local validate band (build, `--check-wheels`, stage, clean-venv smoke), CI green on `dev`, and ops/security/docs hunts on the frozen three-dot diff.

### Decisions I would make

- Merge this before #579, then retarget #579 to `dev` (it is stacked on this head).
- Add `Closes #535` to the PR body if you want the auto-close graph to register (it is only in the commit message today).

### Optional

Your call on all of these.

- CHANGELOG "three steps before" smoke overstates the base step distance (see the pointer on that line).
- No automated assert of YAML publish-after-smoke indices; the in-file band comment is the living guard.

### What worked well

- Validate-then-publish band with an explicit boundary comment, and a single `--check-wheels` definition shared with CI.

<details>

Tier `standard`, trust `trusted`, reviewed at `05ed348e...977c79e2`.

| Gate | Result |
| :--- | :--- |
| CI (DCO, lint, test) | green |
| Local sync/ruff/build/check-wheels/smoke | pass |
| H14 / H15 / H17 / bugbot | clean |
| H16 live MCP/CLI | n/a |

</details>

adriannoes (approved):
Since `977c79e`, `4fef935` drops the wrong CHANGELOG step count and adds `tests/test_release_workflow.py` so both publish steps stay after the smoke install and no step opts out of fail-fast.

Re-checked on this tip: CI green (DCO, lint, test); the two new tests pass; moving Create GitHub Release above the smoke step fails the order guard as expected.

Still optional process notes only: merge before retargeting #579, and add `Closes #535` to the PR body if you want the auto-close graph to register.

PR #579, opened 2026-08-04
fix(release): build and install the wheels before the tag push

### LINKED ISSUE #534: fix(release): publish tags without ever building or installing the wheels

mocha06:
Closes #534.

## Motivation

`release.py publish` pushed the tag without ever having built or installed the wheels it was about to release.

Every guard before the confirmation prompt checked repository state: branch, working tree, tag collision. None touched an artifact. The first wheel existed on the CI runner, past the point of no return. `verify` then inspected the published result rather than a candidate.

`RELEASE.md` states the principle the command is built around: everything before the tag push is local and reversible, so you review before anything leaves your machine. What was reviewable was the release commit. The wheels were not, because they did not exist yet.

`v0.4.0-beta.1` is the failure this catches. `mcp` 2.0.0 removed a module the MCP server imports at startup, so a fresh resolve produced a server that could not start. It was detectable on the operator's machine in about a minute. Nothing in the repository changed to cause it, so the release commit's diff could not show it either. It cost a tag, a GitHub Release, and manual cleanup (#530).

## Outcome

`publish` now checks the artifacts before its confirmation prompt. It builds all five wheels with `uv build --all-packages --wheel` into a temporary directory, asserts one wheel per published member, installs them into a clean virtualenv with `pip`, and runs `smoke_entry_points.py` under that virtualenv to launch every console entry point. A failure names the step that broke and states that nothing was tagged, pushed, or published.

The check lives in `_tag_and_publish`. That is the only code that runs `git tag` and `git push origin <tag>`, so no path reaches a tag without crossing it. Within that function it sits after the repository-state guards and before `confirm`. After, so a colliding tag fails in a second rather than after a build. Before, so the prompt tells the operator that the wheels install and launch.

It builds into a temporary directory, never `dist/`. A stale local build cannot be mistaken for this one, and the check leaves nothing behind.

There is no bypass flag. The check resolves third-party dependencies from PyPI rather than `uv.lock`, which is the divergence it exists to expose. It therefore needs the index: while PyPI is unreachable, a cut cannot be made.

The sequence matches CI exactly — `python -m venv` and `pip`, not `uv venv` and `uv pip`. A different resolver can pass locally and fail in the workflow. The install output is not quieted, because the resolved versions are what the operator is there to read.

`EXPECTED_WHEEL_STEMS` was the last hand-written copy of the published-member list, with nothing tying it to the workspace. It now derives from `PUBLISHED_DISTRIBUTIONS` through a new `wheel_stem` helper, placed beside its inverse `wheel_distribution`, so one file owns the filename-to-distribution mapping.

## Docs

`RELEASE.md` step 3 frames the artifact check as the other half of the pre-boundary review: the diff is half of what ships, and the other half cannot be read off any diff. Step 4 carries the mechanics and the network requirement. The `scripts/release.py` and `scripts/smoke_entry_points.py` automation-reference rows are updated. Two standing claims that this change makes false are corrected: the parenthetical saying only `verify` installs from PyPI, and `ci.yml`'s "the two gates that resolve dependencies from PyPI".

## Testing

| 1 | `smoke_build_and_install()` against the network | five wheels built, fresh install, both entry points launch | resolved `mcp-2.0.0` and `starlette-1.3.1` from PyPI; `pipefy --help` ok, `pipefy-mcp-server --help` ok | pass |
| 2 | Working tree after that run | nothing left behind | `git status` clean, temp tree removed | pass |
| 3 | Check ordering | runs before `git tag` and before `git push` | as expected | pass |
| 4 | Check raises | no tag, no push | no `git tag` or `git push` recorded | pass |
| 5 | Colliding tag | fails without building | the check never runs | pass |
| 6 | Confirmation prompt | states the built wheels install and launch | as expected | pass |
| 7 | Empty build output | reported as an incomplete wheel set | `has no wheel for:` | pass |
| 8 | Step failure | names the step, states nothing shipped | as expected | pass |
| 9 | `wheel_stem` round-trip over all five published members | the escaping survives both directions | as expected | pass |

Suite: 4185 passed, 5 skipped, 34 deselected — the 4172 baseline plus 13 new tests. `ruff check`, `ruff format --check`, `lint-imports`, `bump_version.py verify`, and the skill-ref linter all pass. No unit test reaches the network.

## Notes

Stacked on #578, one layer later at the same boundary. #578 moves the release workflow's own gate ahead of its publish steps. This one puts a gate before the tag exists at all. CI does not fire on a non-default base branch: retarget this to `dev` once #578 merges. The base has since gained a release-workflow drift guard, so the suite total here reaches 4187 after this rebases onto it.


RELEASE.md
```diff
@@ -88,14 +88,20 @@ It reads the current version and `## [Unreleased]` from `origin/dev` (not your c
...
+   `publish` asks for one confirmation before the irreversible tag push (pass `--yes` to skip it in automation). It refuses unless the checked-out branch matches the version's track and is in sync with its remote, so a release PR that has not merged yet is reported as such rather than surfacing as a rejected branch push.
+
+   Then, still before the prompt, it runs the **artifact check**: `uv build --all-packages --wheel` into a temporary directory (never `dist/`, so a stale local build cannot be mistaken for this one), one wheel per published member asserted before anything is installed, `pip install` of those wheels into a clean virtualenv, and `smoke_entry_points.py` under that virtualenv's interpreter to launch every console entry point. It is the same sequence `release.yml` runs from the same script, moved to the side of the boundary where a failure costs nothing — so the operator reads the real error locally, with no tag, no GitHub Release, and no manual withdrawal. The confirmation prompt then states that the built wheels install and launch, so what was checked is in front of you when you answer it.
```
adriannoes: **Optional.** Step 4 mentions the confirmation prompt, then says the artifact check still runs before it. Code order is smoke then confirm (`smoke_build_and_install` then `confirm` in `_tag_and_publish`). I would put the artifact-check paragraph first so the chronology matches the code. Your call whether to take this.

scripts/release.py
```diff
@@ -601,11 +612,95 @@ def watch_release_workflow(tag: str, *, exclude: frozenset[str] = frozenset()) -
...
+            f"(exit {exc.returncode}); the output above is the real error. "
+            "Nothing was tagged, pushed, or published."
+        ) from None
+
+
+def _venv_python(venv: Path) -> str:
+    return str(venv / ("Scripts" if sys.platform == "win32" else "bin") / "python")
+
+
+def smoke_build_and_install() -> None:
```
adriannoes, line 636: **Optional.** This PR takes `scripts/release.py` from 945 to 1045 lines. The new smoke helpers (`_smoke_run`, `_venv_python`, `smoke_build_and_install`, ~79 lines) fit next to `smoke_entry_points`. Keep only the call in `_tag_and_publish` so the no-bypass invariant stays. Your call whether to extract in this PR or follow up.

adriannoes (approved):
Reviewed at `571e1168`: full local ci.yml-equivalent gates (4185 unit + packaging clean-venv), release/smoke unit suite (99), and deep hunts on the pre-tag artifact check. No live MCP/CLI surface on this PR.

**Verdict: merge with notes** into `rc-dev/fix/release-validate-before-publish` (retarget to `dev` after #578).

### Decisions I would make

- Land #578 first, then retarget this PR to `dev` before merging. This head is CLEAN only onto the stacked base.
- Accept `release.py` briefly over 1000 lines for this #534 gate, or extract `smoke_build_and_install` beside `smoke_entry_points` in a follow-up. I would not block the gate on the extract.

### Optional

Your call on all of these.

- 2 smaller pointers are on their lines in the diff (RELEASE step-4 chronology; 1k extract).

### What worked well

- Artifact check sits in the only tag/push path, after cheap repo guards and before confirm, with no bypass.
- Membership before pip, temp build dir, shared `smoke_entry_points`, and `wheel_stem` SSOT close the #534 failure mode.
- Unit tests lock order and abort without network; Phase 3 packaging gate green.

<details>

Tier `deep`, trust `trusted`, reviewed at `977c79e2...571e1168`. CI class `none` on feature base; local full gate set pass. H16 n/a.

</details>

adriannoes (dismissed):
Re-reviewed at `c2505f59` after the docs follow-up.

What changed since `571e1168`: one commit, `docs(release): order step 4 the way publish runs`. Step 4 now narrates repo guards, then the artifact check, then the confirmation prompt, matching `_tag_and_publish`.

Re-verified on this tip: the RELEASE chronology matches smoke-then-confirm; `scripts/release.py` is unchanged at 1045 lines; DCO and packaging smoke are green; still stacked on #578 (retarget to `dev` after that lands). The optional extract of the smoke helpers is still open and still not merge-blocking.

**Verdict: merge with notes** into `rc-dev/fix/release-validate-before-publish`. Approve again on this tip.

adriannoes (approved):
Re-reviewed at `76b2d2d6` after the retarget onto `dev`.

What changed since `c2505f59`: merge of `origin/dev` into this branch (including landed #578). The unique three-dot delta versus current `dev` is still the six-file pre-tag artifact check; no conflict rewrite of that gate.

Re-verified on this tip: `smoke_build_and_install` still runs before confirm and tag; RELEASE step 4 still narrates guards, artifact check, then confirm; full CI on `dev` is green (lint, test, DCO, packaging smoke). The optional `release.py` extract past 1000 lines remains open and still not merge-blocking.

**Verdict: merge-ready** into `dev`. Approve again on this tip.

Issue #536, opened 2026-07-29
uninstall: complete cleanup across all channels and clients


mocha06:
## Problem

The toolkit ships through three independent channels — the Claude Code plugin, a local install (`install.sh` / `uvx` / `uv tool install`), and the hosted HTTP server — and there is no supported way to reverse any of them. The repo has no `uninstall`, `doctor`, `init`, or `config` command. `pipefy auth logout` is the only reversal primitive that exists, and it has a known bug (see below). The docs contain five scattered removal one-liners (`README.md:89`, `README.md:275`, `skills/onboarding/pipefy-toolkit-setup/SKILL.md:89`, `docs/cli/auth.md:223`, `packages/mcp/README.md:35`) and no uninstall section.

The result is that switching channels — the common case, e.g. local to hosted — leaves state from the old channel that interferes with the new one. Users who attempt this by hand tend to remove the cheap artifacts (binaries, tool envs) and leave exactly the state that causes conflicts (registrations, marketplace entries, stored credentials).

The sharpest example needs **no file deletion at all**: `pipefy` registered in two scopes with different endpoints — user scope pointing at the hosted HTTP server, project scope pointing at a local `uvx pipefy-mcp-server` from a git-tracked `.mcp.json`. Project scope wins, fails to start because no local server is installed, and silently shadows a perfectly working hosted registration. Claude Code emits a diagnostic for this, but nothing in the toolkit does, and no amount of `rm` fixes it. Detection is at least as important as removal.

`README.md:68` already states the invariant — "Register exactly one MCP server named `pipefy`" — and `skills/onboarding/pipefy-toolkit-setup/SKILL.md:76` asks an agent to "Confirm exactly one `pipefy` MCP registration". Today that invariant exists only as prose instructions to a language model. It should be a deterministic check.

## Goal

One curl-able command that reports the complete state across every channel and every client, and — after showing the user exactly what it will do and getting approval — leaves a genuinely clean machine.

```
curl -LsSf https://raw.githubusercontent.com/pipefy/ai-toolkit/main/uninstall.sh | sh -s -- --scan
curl -LsSf https://raw.githubusercontent.com/pipefy/ai-toolkit/main/uninstall.sh | sh
```

## Artifact inventory

Everything below has to be either removed, or named in the report as something the user must handle.

### Created by `install.sh`

| Artifact | Evidence | Note |
| --- | --- | --- |
[...]

PR #580, opened 2026-08-04
fix(auth): delete an unreadable keychain entry on logout

### LINKED ISSUE #538: auth: logout must delete an unreadable keychain entry
Supports #536. Independently shippable, and the teardown phase depends on it being correct.

`pipefy auth logout` early-returns when `load_session` cannot **read** the stored entry (`packages/cli/src/pipefy_cli/commands/auth.py:460-463`). `load_session` swallows both `KeyringError` and `ValidationError` and returns `None` (`packages/auth/src/pipefy_auth/storage.py:166-175`), so a stale, corrupt, or schema-drifted entry reports `Not signed in. Nothing to do.` while surviving on disk.

The caveat is already documented at `docs/cli/auth.md:223`. Any cleanup that trusts `logout` to have removed the credential is therefore wrong, which is why this is split out ahead of teardown.

## Scope

- Branch on entry **presence**, not parse success. Add a presence check to `packages/auth/src/pipefy_auth/storage.py` rather than inferring it from `load_session`
- Keep the existing revoke-then-delete ordering and its warning behaviour (`packages/cli/src/pipefy_cli/commands/auth.py:465-505`) — revocation still requires a readable refresh token, so an unreadable entry is delete-only and must say so
- Remove the caveat from `docs/cli/auth.md:223` once fixed

## Acceptance criteria

- [ ] `pipefy auth logout` deletes a stored entry that exists but cannot be parsed
- [ ] Output distinguishes three outcomes: signed out and revoked; an unreadable entry removed without revocation, stated plainly; and genuinely nothing to do
- [ ] A revocation failure still proceeds to delete, with the existing warning preserved
- [ ] `SessionDeleteError` still exits non-zero with the manual-removal hint
- [ ] Doc caveat removed
- [ ] Tests cover unreadable-entry, absent-entry, and revoke-failure paths using the in-memory keyring in `packages/cli/tests/conftest.py`. No real keychain access in tests

mocha06:
## Motivation

`pipefy auth logout` reported `Not signed in. Nothing to do.` while leaving the credential on disk. It branched on whether `load_session` returned a session, and `load_session` collapses three different states into one `None`: no entry, an entry that fails validation, and a keychain backend that refused the read. A stale, corrupt, or schema-drifted entry therefore looked like an already-clean machine.

Two things depended on that being wrong. The macOS `errSecInvalidOwnerEdit` remediation in `docs/cli/auth.md` told the user to fall back to `security delete-generic-password` precisely because logout could not be trusted, and the teardown work in #536 needs logout to actually delete.

## Outcome

Logout branches on entry presence. `session_entry_presence` in `pipefy_auth.storage` returns `present`, `absent`, or `unknown`, reading the entry without parsing it, so the decision rests on positive evidence gathered before anything is destroyed rather than on a delete failure after the fact.

Four outcomes are now distinguishable:

| State | stdout | Exit |
|---|---|---|
| Signed out and revoked | `Signed out of Pipefy (<issuer>).` | 0 |
| Entry present but unreadable | `Removed an unreadable Pipefy session entry (<issuer>).` plus a stderr line saying the refresh token was not revoked and stays valid at the IdP until natural expiry | 0 |
| No entry | `Not signed in. Nothing to do.` | 0 |
| Keychain read failed and nothing was deleted | stderr only: presence is unknown, so the command claims neither removal nor a clean machine | 1 |

Revocation needs a readable refresh token, so an unreadable entry is delete-only and says so instead of implying the credential is dead. The revoke-then-delete order, the revoke-failure warnings, and the exit-1 manual-removal hint on a rejected delete are unchanged — the revoke block moved into `_revoke_or_warning` verbatim so its existing tests cover it untouched.

An empty-string entry counts as present. `load_session` treats a falsy blob as `None`, so keying presence on readability would have let an empty entry survive a logout that reported success.

The stale caveat lived in three places, not the one the issue named: `docs/cli/auth.md`, `packages/mcp/README.md`, and the runtime macOS hint in `_auth_keychain_hints.py`. All three now describe presence-based deletion.

## Notes

Costs one extra `get_password` on the signed-in path, since presence is probed before the session is loaded. Inferring absence from `PasswordDeleteError` would avoid it, but that derives "there was nothing there" from a failure signal — the same class of bug this fixes.

Verified with the in-memory keyring in `packages/cli/tests/conftest.py`; no test touches a real keychain. Not verified against a real macOS keychain in the `errSecInvalidOwnerEdit` state, which is the condition the doc caveat described.

Closes #538

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **CLI (`pipefy auth logout`)**: a stored keychain entry that exists but cannot be parsed — stale, corrupt, or written by an older schema — is now deleted instead of being reported as `Not signed in. Nothing to do.` while surviving on disk. The decision comes from entry **presence** (`session_entry_presence` in `pipefy_auth.storage`: `present` / `absent` / `unknown`) rather than from `load_session`, which collapses all three into one `None`. Revocation needs a readable refresh token, so an unreadable entry is delete-only and stderr says the token stays valid at the IdP until natural expiry. When the presence probe itself fails and nothing is deleted, the command claims neither removal nor a clean machine and exits 1. Revoke-failure warnings, the revoke-then-delete order, and the exit-1 manual-removal hint on a rejected delete are unchanged. (#538)
 - **AI agents (active lifecycle)**: create can start inactive (`create_ai_agent(active=false)` / `pipefy agent create --inactive`) by sending `disabledAt` on create and the chained update so the second step does not revive; routine update preserves `disabledAt` (optional `disabled_at` / `--disabled-at` pass-through skips the re-read; use `toggle_ai_agent_status` / `pipefy agent toggle` to activate or deactivate); create/update success payloads expose `disabled_at` and `active`; partial-failure envelopes surface shell `disabled_at` and a toggle recovery hint. Skills document the contract. (#520)
 
 ### Added
--- a/docs/cli/auth.md
+++ b/docs/cli/auth.md
@@ -161,20 +161,25 @@ The shape is stable across all sources (fields you don't have are `null` rather 
 
 POSTs the stored refresh token to the IdP's `end_session_endpoint` (advertised in the OIDC discovery document) and removes the keychain entry. Without server-side revocation the refresh token would remain valid at the IdP until natural expiry — anyone who recovered it from a backup could still mint new access tokens.
 
-The command always clears the local keychain entry once it runs, even when the IdP round-trip fails. The two non-happy paths surface a stderr warning so the user knows whether server-side revocation succeeded:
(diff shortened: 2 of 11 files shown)


adriannoes (approved):
Reviewed at `0f415a3`: local ruff + targeted auth/logout pytest (40 passed), security static CLEAR, and hunts H1-H5/H7/H8/H10-H13/H15/H17 on the frozen `dev...head` range. Live CLI logout was not run (destructive keychain/IdP side effects); contract coverage uses the in-memory keyring.

### What worked well

- Presence vs readability split in `pipefy_auth.storage`, with logout claiming removal only after `deleted` and refusing a clean machine when presence is unknown.
- Unreadable path is honest about non-revocation; tests poison `revoke_session` so that contract cannot silently regress.
- Docs, MCP README, and the macOS login hint all leave the stale `load_session` early-return caveat.

<details>

Tier `standard`, trust `trusted`, reviewed at `05ed348e...0f415a30`.

| Gate | Result |
| :--- | :--- |
| CI lint/test/DCO | pass |
| Local ruff + pytest (auth logout/presence) | pass |
| H15 security | CLEAR |
| H16 live logout | blocked (destructive) |

Execution matrix: 11/11 local rows pass. H2/H3 n/a. Bugbot agent substituted by contracts/thermo/footguns/HVT.

Not covered: real macOS `errSecInvalidOwnerEdit` keychain (disclosed in PR Notes).

</details>

Issue #541, opened 2026-07-29
ci: shellcheck, install/uninstall parity, and a round-trip job


mocha06:
Phase 5 of #536. Neither `install.sh` nor `uninstall.sh` has any CI coverage today — `install.sh` is not referenced anywhere under `.github/`, and there is no shell linting in `.pre-commit-config.yaml`.

This phase also carries the user-facing documentation, which does not exist: the repo has five scattered removal one-liners and no uninstall section.

## Acceptance criteria

- [ ] `shellcheck` and `sh -n` over both scripts, wired into `.pre-commit-config.yaml` and CI
- [ ] Every deletion in `uninstall.sh` routes through a single guard that refuses an empty argument, `/`, or `$HOME` exactly — the cheapest defence against an unset-variable `rm -rf`
- [ ] `install.sh --dry-run` exercised for each `--client` value
- [ ] Parity test asserting every path `install.sh` writes appears in `uninstall.sh`'s removal table, so the two cannot drift. The scripts are colocated at the repo root specifically so this stays reviewable
- [ ] Linux round-trip job: snapshot, `install.sh --yes --client cursor`, `uninstall.sh --yes`, then assert the filesystem diff is empty except for an allowlisted set. This is the only check that proves completeness rather than correctness of the steps someone thought to write
- [ ] `docs/uninstall.md` covering full teardown and switching between hosted, local, and plugin
- [ ] README section, alongside the existing install paths
- [ ] Replace the prose invariant at `skills/onboarding/pipefy-toolkit-setup/SKILL.md:76` and the failure-mode row at `:89` with the scan command, turning an instruction to a language model into a deterministic check

mocha06:
Implementation note for `docs/uninstall.md` and the README switching section — two things the "switch to hosted" instructions must get right:

1. The hosted add command must include `--client-id pipefy-mcp` (matching `README.md:86`): `claude mcp add --transport http --scope user --client-id pipefy-mcp pipefy https://mcp.pipefy.com/mcp`. Without `--client-id`, Claude Code falls back to OAuth Dynamic Client Registration, which the auth server rejects, so the connection fails with an auth/registration error instead of prompting the browser login. Any doc snippet that omits the flag sends users straight into that failure.
[...]

PR #581, opened 2026-08-05
feat(uninstall): report toolkit state across every channel and client

### LINKED ISSUE #537: uninstall: non-destructive --scan across all channels and clients
Phase 1 of #536, which holds the shared artifact inventory, the eight recorded failure modes, and the design decisions. Read it first.

Ships `uninstall.sh` at the repo root with **only** the non-destructive `--scan` mode. No removal code in this phase, which is the cleanest guarantee that a scan cannot damage anything.

This phase carries its own weight even if nothing else lands: the sharpest observed failure is a `pipefy` server registered in two scopes with different endpoints, where the broken local registration wins and silently shadows a working hosted one. That needs detection, not deletion.

## Scope

- `uninstall.sh` at the repo root, POSIX sh, mirroring `install.sh`'s helper conventions (`say` / `warn` / `err` / `confirm` / `refuse_root` / `detect_platform`)
- `--scan`, `--client`, `--dry-run`, `--help`. Detection sweeps every client regardless of `--client`, since the point is finding state the user forgot about
- Reachable as `curl -LsSf .../uninstall.sh | sh -s -- --scan`

## Acceptance criteria

- [ ] Runs with nothing installed, with no `uv`, and with no `claude` on PATH
- [ ] Matches a registration by its **definition**, not by its config key name — a stdio command resolving to `pipefy-mcp-server` (including a runner such as `uvx` or `npx` with `pipefy-mcp-server` as an argument), or an HTTP/SSE URL whose host is `mcp.pipefy.com`. A real environment carried `pipefy-dev`, so name-keyed detection misses the entries that cause masking. Report the key name as data. Only the documented public endpoint may be hardcoded; any other Pipefy-shaped HTTP registration is reported generically for the user to judge
- [ ] Reports every matching MCP registration across every client and scope, with its endpoint and source file: `~/.claude.json` (user scope and `projects.<path>.mcpServers`), `~/.cursor/mcp.json`, Claude Desktop's `claude_desktop_config.json`, `~/.codex/config.toml`, and project `.mcp.json` in the repo root plus every path from `git worktree list`
- [ ] Flags multi-scope conflicts, naming both endpoints and which one wins
- [ ] Flags orphan state: a marketplace registered in `~/.claude/plugins/known_marketplaces.json` or `extraKnownMarketplaces` while `installed_plugins.json` has no plugin, and a marketplace clone on disk with no registration
- [ ] Flags broken registrations whose command does not resolve
- [ ] Reports every `pipefy` / `pipefy-mcp-server` on PATH with the resolved origin of each match, so a repo `.venv/bin` shadow is visible
- [ ] Enumerates keychain entries for service `pipefy` rather than assuming the production `(issuer, client_id)` tuple. macOS via `security`, Linux via `secret-tool`, plus `~/.config/pipefy/keyring.cfg` when the file backend is in use. Enumeration must not prompt for secret values
- [ ] Reports `PIPEFY_*` in the process env, in shell rc files, and in client config `env` blocks — **key name and location only, never the value**
- [ ] Reports `~/.config/pipefy` contents, shell completion files, and installed `pipefy-*` skills
- [ ] Detection is structural throughout: a JSON key path, a TOML section header, a keychain service attribute, an exact env-var name. No substring search over paths or file contents. Reuse the env-var allowlist from `_session_masking_env_vars()` (`packages/cli/src/pipefy_cli/commands/auth.py:86-102`)
- [ ] Exit codes: `0` clean, `1` findings present, `2` the scan itself failed

## Tests

Fixture `HOME` trees per scenario: clean, local-only, hosted-only, plugin-only, hosted plus a git-tracked project registration, orphan marketplace without plugin, file keyring backend, PATH shadowed by a repo venv, secret in a client config `env` block, and a **noise** home populated with unrelated `pipefy`-containing directory names, hostnames, and email addresses that must produce zero findings. Two specimens come from a real environment: an unrelated MCP server whose URL host contains `pipefy` but is not the documented endpoint, and a development repo's `.venv/bin/pipefy` on `PATH`, which is there by design — the latter must still be listed for shadowing purposes but classified as a project virtualenv binary rather than an installed artifact. Reproduce hostnames with synthetic names under an RFC 2606 reserved TLD; no internal hostname may enter this repo.

Note that `install.sh` has no CI coverage today, so shell testing here is greenfield. `packages/cli/tests/test_completion.py` already shells out to the real CLI with a temporary `HOME` and is a reasonable pattern to follow.

mocha06:
## Motivation

The toolkit ships through three channels — the Claude Code plugin, a local install, and the hosted HTTP server — and nothing reverses any of them. Switching channels is the common case, and it leaves state from the old channel that interferes with the new one.

The sharpest failure needs no file deletion at all. A `pipefy` server registered at user scope pointing at the hosted endpoint, shadowed by a project-scope entry from a git-tracked `.mcp.json` pointing at a local `uvx` server that is not installed. Project scope wins, fails to start, and silently hides a working hosted registration. No amount of `rm` fixes it. Detection is the missing piece.

`README.md` already states the invariant — register exactly one Pipefy MCP server — and the onboarding skill asks an agent to confirm it. That invariant existed only as prose addressed to a language model.

## Outcome

`uninstall.sh` at the repository root, POSIX sh, with only `--scan`. No removal code at all in this change, which is the cheapest guarantee that a scan cannot damage anything: a test bans `rm`, `mv`, `uv tool uninstall`, `claude mcp remove`, `security delete`, `secret-tool clear`, `os.replace` and `shutil.rmtree` from the file, with the tempfile trap as the sole exception.

**A registration is matched on what it runs, not on what it is called.** The config key is free text — a real environment had the server registered as `pipefy-dev` — so keying on the name misses exactly the entries that shadow a working one. An entry matches when its command is `pipefy-mcp-server`, or a known runner with `pipefy-mcp-server` as an exact argument, or its URL host is `mcp.pipefy.com`. The key is reported as data. This is what makes the one-server invariant checkable.

Detection is structural throughout — JSON key paths, a TOML section header, a keychain service attribute, exact environment-variable names. Never a substring search: `pipefy` turns up in unrelated hostnames, mail addresses and directory names, and `env | grep -i pipefy` matches `PWD` for anyone working in a directory named after it. A noise fixture carrying all three produces zero findings, and two mutations that relax an exact comparison to a substring both fail it.

Reported: every registration across every client and scope with its endpoint and source file, multi-scope conflicts naming both endpoints and which wins, orphan marketplace state in both directions, registrations whose command does not resolve, every match on `PATH` with its resolved origin so a repo `.venv` shadow is visible, enumerated keychain entries, `PIPEFY_*` by name and location, `~/.config/pipefy`, shell completions, and installed skills.

Credentials are reported by **name and location only, never by value**. `~/.claude.json` holds live third-party tokens in `env` and `headers` blocks, so this is a hard rule and three mutations that leak a value each fail the suite.

Exit `0` clean, `1` findings, `2` the scan itself failed. Every JSON source is read by one `python3` program; without `python3` those sources report *not inspected* and force the `2`, so a partial sweep can never read as a clean machine. Everything else — `PATH`, keychain, config directory, uv cache, completions, skills, the Codex TOML — needs no `python3`.

## Notes

Run against a real machine it reports the hosted user-scope registration shadowed by project-scope entries across 25 worktrees, and a marketplace registered with its clone on disk while no plugin is installed — the two failure modes this exists to catch.

Verified with fixture `HOME` trees and a stubbed `PATH`; no test touches a real keychain or client config. The suite also runs under `dash`, since `/bin/sh` on macOS is bash in POSIX mode and forgives what a Debian-family `/bin/sh` rejects.

Part of #536. Closes #537.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,0 +1,1465 @@
+#!/bin/sh
+# uninstall.sh — report Pipefy toolkit state across every install channel and MCP client.
+#
+# This version scans only. It reads; it never removes, moves, or edits anything,
+# and the single path it writes is a tempfile holding its own record stream.
+#
+# Channels: the Claude Code plugin, a local install (install.sh / uv tool /
+# uvx), and the hosted HTTP server. Clients: Claude Code, Cursor, Claude
+# Desktop, Codex.
+#
+# A registration is matched on what it *runs*, not on what it is called. The
+# registration key is free text — a real environment had the server registered
+# as `pipefy-dev` — so keying on the name misses exactly the entries that
+# shadow a working one. Matching stays structural: an exact command name, an
+# exact argument, an exact URL host, a JSON key path, a TOML section header, a
+# keychain service attribute, an exact environment variable name. Never a
+# substring search: "pipefy" turns up in unrelated hostnames, mail addresses
+# and directory names, and `env | grep -i pipefy` matches PWD for anyone
+# working inside a directory named after it.
+#
+# Exit codes: 0 nothing found, 1 findings present, 2 the scan itself failed.
+
+set -eu
+
+REPO="pipefy/ai-toolkit"
+CANONICAL_NAME="pipefy"        # the name install.sh writes; the documented invariant
+KEYCHAIN_SERVICE="pipefy"      # keyring service attribute
+SERVER_BINARY="pipefy-mcp-server"
+HOSTED_HOST="mcp.pipefy.com"   # the documented hosted endpoint, and the only one
+RUNNERS="uvx uv npx python python3 pipx"
+PLUGIN_ID="pipefy@pipefy"
+MARKETPLACE_ID="pipefy"
+
+CLIENT=""
+ALLOW_ROOT=0
+DRY_RUN=0
+
+OS=""
+CONFIG_DIR=""
+PYTHON3=""
+RECORDS=""
+FINDINGS=0
+SCAN_ERRORS=0
+TAB=$(printf '\t')
+
+# Secrets. `_session_masking_env_vars()` (packages/cli/.../commands/auth.py)
+# inspects PIPEFY_TOKEN plus both sides of the legacy alias map in
+# packages/auth/.../settings.py; the iPaaS client secret is added because this
+# list answers "what is a secret", which is a wider question than that
+# function's "what outranks a stored session". Pinned by
+# tests/test_uninstall_scan.py.
+CRED_ENV_NAMES="PIPEFY_TOKEN
+PIPEFY_SERVICE_ACCOUNT_CLIENT_ID
+PIPEFY_SERVICE_ACCOUNT_CLIENT_SECRET
+PIPEFY_OAUTH_CLIENT
+PIPEFY_OAUTH_SECRET
+PIPEFY_IPAAS_OAUTH_CLIENT_SECRET"
+
+# Non-credential configuration variables: every name in the docs/config.md
+# tables plus every name the pydantic settings models read. Informational —
+# they change behaviour rather than granting access. Pinned by the same test.
+CONFIG_ENV_NAMES="PIPEFY_ALLOW_INSECURE_URLS
+PIPEFY_AUTH_CLIENT_ID
+PIPEFY_AUTH_URL
+PIPEFY_BASE_URL
+PIPEFY_CONFIG_FILE
+PIPEFY_DEFAULT_WEBHOOK_NAME
+PIPEFY_DISABLE_STORED_SESSION
+PIPEFY_GQL_REUSE_FETCHED_GRAPHQL_SCHEMA
+PIPEFY_IPAAS_OAUTH_CLIENT_ID
+PIPEFY_IPAAS_OAUTH_REDIRECT_URI
+PIPEFY_IPAAS_URL
+PIPEFY_JWT_AUDIENCE
+PIPEFY_JWT_ISSUER_URL
+PIPEFY_JWT_JWKS_URI
+PIPEFY_JWT_VERIFY_AUDIENCE
+PIPEFY_KEYCHAIN_BACKEND
+PIPEFY_MCP_ALLOWED_HOSTS
+PIPEFY_MCP_ALLOWED_ORIGINS
+PIPEFY_MCP_ALLOW_INSECURE_HTTP_BIND
+PIPEFY_MCP_HOST
+PIPEFY_MCP_LOG_LEVEL
+PIPEFY_MCP_PORT
+PIPEFY_MCP_PROFILE
+PIPEFY_MCP_RS_REQUIRED_SCOPES
+PIPEFY_MCP_RS_RESOURCE_SERVER_URL
+PIPEFY_MCP_TOOLSETS
+PIPEFY_MCP_TRANSPORT
+PIPEFY_MCP_UNIFIED_ENVELOPE
+PIPEFY_ORG_ID
+PIPEFY_PERMISSION_DENIED_ENRICHMENT_TIMEOUT_SECONDS
+PIPEFY_PORTAL_ORG_UUID"
+
+say() { printf '%s\n' "$*"; }
+warn() { printf 'warning: %s\n' "$*" >&2; }
+err() { printf 'error: %s\n' "$*" >&2; exit 2; }
+
+section() { printf '\n== %s ==\n' "$*"; }
+# A finding is state on this machine that a teardown would have to deal with.
+finding() { printf '  * %s\n' "$*"; FINDINGS=$((FINDINGS + 1)); }
+note() { printf '  - %s\n' "$*"; }
+detail() { printf '      %s\n' "$*"; }
+# A source this run could not inspect, so a partial scan never reads as clean.
+uninspected() { printf '  ! %s\n' "$*"; SCAN_ERRORS=$((SCAN_ERRORS + 1)); }
+
+print_help() {
+    cat <<EOF
+Usage: uninstall.sh [OPTIONS]
+
+Report Pipefy toolkit state across every install channel and MCP client.
+This version detects only; it removes nothing.
+
+MCP registrations are matched on what they run — the $SERVER_BINARY
+command, a known runner invoking it, or the hosted endpoint's host — so an
+entry registered under any name is found.
+
+Options:
+  --scan              Report state and exit. The only mode this version has,
+                      and what a bare invocation does.
+  --client <id>       Accepted and validated for symmetry with install.sh.
+                      One of: claude-code, claude-desktop, cursor, codex, none.
+                      Detection always sweeps every client regardless, since
+                      the point is finding state you forgot about.
+  --dry-run           Accepted for symmetry with install.sh. A scan has no
+                      side effects, so this changes nothing.
+  --allow-root        Allow running as root (refuses by default).
+  -h, --help          Show this help.
+
+Exit codes:
+  0  nothing found
+  1  findings present
+  2  the scan itself failed (a source could not be inspected)
+
+Examples:
+  curl -LsSf https://raw.githubusercontent.com/$REPO/main/uninstall.sh \\
+    | sh -s -- --scan
+
+  ./uninstall.sh --scan
+EOF
+}
+
+parse_args() {
+    while [ $# -gt 0 ]; do
+        case "$1" in
+            --scan) shift ;;
+            --client) [ $# -ge 2 ] || err "--client requires a value"; CLIENT="$2"; shift 2 ;;
+            --client=*) CLIENT="${1#--client=}"; shift ;;
+            --dry-run) DRY_RUN=1; shift ;;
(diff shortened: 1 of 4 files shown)


uninstall.sh
```diff
@@ -0,0 +1,1465 @@
...
+        /^class: "genp"/ { genp = 1 }
+        /^[ \t]*"svce"<blob>=/ { line = $0; sub(/^[ \t]+/, "", line); if (line == svc) match_svc = 1 }
+        /^[ \t]*"acct"<blob>="/ { acct = $0; sub(/^[ \t]*"acct"<blob>="/, "", acct); sub(/"$/, "", acct) }
+        END { flush() }
+    '
+}
+
+scan_keychain_linux() {
+    if ! command -v secret-tool >/dev/null 2>&1; then
+        note "keychain: 'secret-tool' not on PATH; no Secret Service to query"
```
adriannoes, line 1161: **Optional.** Missing `secret-tool` on Linux is a `note` and can still exit 0; missing `security` on Darwin is `uninspected` and forces exit 2. That matches optional Secret Service tooling on headless Linux, but it is easy to misread exit 0 as "keychain fully enumerated." Your call whether to document the asymmetry in `--help` / Summary or harden Linux to `uninspected`.

adriannoes (changes requested):
Reviewed at `d59a061c`: detached worktree gates (ruff + 46 fixture tests under dash), CI green on `dev`, security/credential non-echo checks, and targeted refutations on conflict/Summary false-clean candidates.

### Required before merge

- Document the non-destructive `uninstall.sh --scan` curl on the README Installation / one-server callout. Hosted-only users have no clone, and in-script `--help` is not discoverable from the front door.
- Point the onboarding skill Verify / "two MCP entries" failure path at that scan instead of prose-only confirmation. Same enablement claim as the README gap.

### Decisions I would make

- Accept the 1465-line single-file `uninstall.sh` for curl|sh distribution for Phase 1, with an explicit size/growth budget before #582 lands more teardown surface on the same file.
- Keep bare invocation = scan in #581. When #582 adds teardown, require an explicit flag (or flip bare only with help/README updates in the same PR). Do not leave "bare is always safe" after teardown ships.

### Optional

Your call on all of these.

- Linux missing `secret-tool` stays note+exit 0 while Darwin missing `security` forces exit 2; document or align (pointer on the Linux keychain branch in `uninstall.sh`).
- Changelog Unreleased line for the new scan entry; Claude Desktop / `extraKnownMarketplaces` orphan fixtures; tighten the no-python3 skills assert; soften PR "mutations" wording to named behavioral tests.

### What worked well

- Match-by-what-it-runs (not config key) with structural detection, credential name+location only, and exit 2 when JSON sources are uninspected.
- Scan-only guarantee enforced by a static removal-verb ban plus tempfile trap exception.
- Strong fixture HOME suite (dash, noise home, sentinel leak tests) without touching a real keychain.

<details>

Tier `deep`, trust `trusted`, frozen `b180c5ab…d59a061c`.

| Gate | Result |
| :--- | :--- |
| GitHub CI (DCO/lint/test) | green |
| Local pytest `tests/test_uninstall_scan.py` | 46 passed (dash) |
| Secret/keychain/noise slice | 7 passed |
| H15 security | clean |
| H16 live MCP/CLI | n/a |
| Refute | disabledsrv refuted; Summary false-clean refuted; Linux keychain → SUGGESTION; H10 enablement survives |

Not covered: real-machine scan against reviewer HOME/keychain (by design).

</details>

uninstall.sh
mocha06: Fixed in `13aa5db` (lands at the tip of the stack, on #584 — see the note below on why).

You were right that this is the same false-clean class as the python3 rule, so I promoted it from optional. A missing `secret-tool` is now `uninspected` and forces exit 2, matching Darwin's missing `security`. An uninspectable credential store must not read as clean.

The scan harness now installs an empty-store stub by default, with an explicit `no_secret_tool=True` for the one test that is about its absence.

**Why the fix is not in this PR:** the four fix commits sit on `rc-540/ci/shell-script-coverage` (#584). Landing them here would mean force-pushing #582, #583 and #584, which destroys the incremental diff you are reviewing against. Your own merge-order guidance (581 → 582 → 583 → 584, no partial merge to `dev`) means no intermediate state carrying the bug ever reaches the default branch.

mocha06:
Thanks — thorough pass. All four fix commits land on `rc-540/ci/shell-script-coverage` (#584) rather than here, because putting them on this branch means force-pushing the three PRs stacked above it, which destroys the incremental diff. Your own merge-order guidance (581 → 582 → 583 → 584, no partial merge to `dev`) means no intermediate state carrying a fixed bug reaches the default branch.

**Optional, accepted:** the Linux `secret-tool` asymmetry — promoted to a fix in `13aa5db`, since it is the same false-clean class as the python3 rule.

**The two required items I am not actioning here, because they already ship in #584.** The README `--scan` curl and the onboarding-skill scan pointer both land in that PR: `README.md` gains an `Uninstalling, and switching between paths` section with the curl form, and the skill's Verify step and multi-registration failure row are replaced by the scan. Documentation for both scripts is #541's declared scope in #536, and #537 is deliberately scan-only — which is what lets the "a scan cannot damage anything" guarantee be enforced by a static removal-verb ban in this PR. Reviewed standalone this reads as a gap; across the stack it is not one.

Your #584 pass then caught a real defect in that very work — the skill said `./uninstall.sh --scan`, which assumes a clone that hosted-only users do not have. That one is fixed in `f9bc5b1`.

**Size budget:** noted, and it grew as you predicted. `uninstall.sh` is now ~3.2k lines. I have not split it, on the grounds that a single file is the whole point of a `curl | sh` entry point, but I agree the conversation is due before anything else lands on it.

adriannoes:
Follow-up after your note on stacking the docs into #584.

**Verdict stays needs changes** on this tip (`d59a061c`). Head is unchanged: still only `uninstall.sh` + `tests/test_uninstall_scan.py`. The two Required items from the prior review (README `--scan` curl near the one-server / install front door, and the onboarding skill Verify / multi-registration path pointing at that scan) are not on this PR.

I checked #584: the curl and skill pointer are there, and the Linux `secret-tool` → exit 2 fix is there too. That is good stack work. It does not clear Required on #581. Approving here would authorize merge to `dev` without a published entry point; if the train pauses after this PR, `dev` ships the script without the discovery path the Motivation asks for.

Force-push cost on #582–#584 is real. Prefer a small tip commit on this branch (README callout + skill Verify/failure rows only; no teardown docs) and let the later PRs rebase, or cherry-pick that pair onto this tip. Landing the full #584 doc surface here is not required.

`CHANGES_REQUESTED` remains until those two surfaces appear on `rc-dev/feat/uninstall-scan`. Happy to re-review as soon as they do.

adriannoes (changes requested):
Re-checked at `d59a061c` after the author reply: no new commits on this tip.

**Verdict: needs changes** into `dev` (unchanged).

### Required before merge

- README still has no `uninstall.sh --scan` curl / callout on this branch.
- Onboarding skill still asks for a prose “exactly one” check; no scan pointer on this branch.

Fixes confirmed on #584 do not satisfy Required for merging #581 alone. See the follow-up comment on the PR for the preferred small tip commit vs stack rebase tradeoff.

mocha06:
Fixed in `3222b67`, on this branch. You are right and I was wrong to argue it.

Your reasoning is the same risk I raised in my own note — if the train pauses after this PR, `dev` ships the script with no discovery path — and I then asked you to absorb it. That is not a reviewer's job. A Required item is about what this PR merges, not about what the stack eventually contains.

Landed the minimum you asked for and nothing more:

- The README one-server callout now carries the `curl … | sh -s -- --scan` block, what the scan matches on, that it changes nothing, and the three exit codes.
- The onboarding skill's Verify step, success criteria, and multi-registration failure row point at the same scan. Curl-led, since a hosted-only user has no checkout — that is the refinement your #584 pass caught, applied here from the start.

No teardown documentation, no `docs/uninstall.md` link, no forward reference to anything that does not exist at this commit. I checked for dangling links; the diff mentions teardown zero times.

`#246` is dropped from the skill here too, so it does not have to travel through the stack.

On propagation: I am merging this branch upward through #582 → #583 → #584 with merge commits rather than rebasing, so none of the three needs a force-push and your incremental diffs survive. #584 carries a richer version of both surfaces (it links `docs/uninstall.md`, which exists by then); I will resolve in its favour at that merge.

adriannoes (approved):
Re-checked at `3222b67` after `docs: publish the scan as the one-server check (#537)`.

**What changed:** README one-server callout now has the non-destructive `curl …/uninstall.sh | sh -s -- --scan` block (match-by-run, removes nothing, exits `0/1/2`). Onboarding Verify, success criteria, and multi-registration failure row point at the same curl-led scan (checkout path as alternate).

**Re-verified:** three-dot delta from `d59a061c` is only those two doc surfaces; CI green (DCO, lint, test, skills-lint); no teardown docs or dangling links on this tip. Prior Required items are satisfied.

Issue #540, opened 2026-07-29
install: write an install receipt so teardown is exact


mocha06:
Phase 4 of #536, which holds the shared artifact inventory and design decisions. Read it first.

`install.sh` currently discards knowledge that only it has, which forces teardown to guess:

- `--prefix` sets `UV_TOOL_DIR` process-scoped and forgets it (`install.sh:472-475`), so those tool environments become unfindable
- `UV_INSTALLED_THIS_RUN` is set, used for one banner, and dropped (`install.sh:159`), so nothing downstream can know whether `uv` is ours
- The JSON merge deliberately preserves a pre-existing `pipefy` entry and forgets it did so (`install.sh:351-353`), so the installer cannot distinguish entries it created from entries it found
- Where `npx skills add` landed is not recorded (`install.sh:303`)

A receipt fixes this for every future install. It is deliberately **data, not code**: a dropped script goes stale, while a receipt stays accurate about what happened at the time it was written.

## Acceptance criteria

- [ ] `install.sh` writes a receipt recording `uv_tool_dir` (including any `--prefix`), `uv_installed_by_us`, per-client `entry_created`, the skills landing directory, the resolved release tag, and a schema version
- [ ] Plain `key=value`, parseable in POSIX sh without `python3` — teardown must not inherit the hard `python3` dependency at `install.sh:311-315`
- [ ] Append-only, one record per run, so re-running with a different `--client` does not lose the earlier record
- [ ] Written incrementally as each step succeeds, so a crashed install leaves a truncated but usable record. Nothing written under `--dry-run`
- [ ] The merge at `install.sh:317-369` reports whether it created the entry or found one, and that result is recorded
- [ ] `uninstall.sh` prefers the receipt and falls back to heuristic mode
- [ ] Heuristic mode is permanent, not transitional — every existing install predates the receipt. It removes less: client entries only when the value matches a shape the installer writes, and `uv` is never treated as ours. It says so up front
- [ ] The receipt is itself in the teardown inventory and is removed

PR #582, opened 2026-08-05
feat(uninstall): teardown with tiered approval and client adapters

### LINKED ISSUE #539: uninstall: teardown with tiered approval and client adapters
Phase 3 of #536, which holds the shared artifact inventory, the eight recorded failure modes, and the design decisions. Read it first. Depends on the `--scan` phase and on the `auth logout` fix.

Adds the destructive path to `uninstall.sh`.

## Client adapters

Only Claude Code ships a removal CLI. `cursor-agent mcp` exposes `login` / `list` / `list-tools` / `disable`, where `disable` affects the approved list rather than the registration; Codex and Claude Desktop ship no CLI at all. **Direct file editing is the baseline mechanism; CLI delegation is an override for the one client that supports it.** Adding a client later should be a table row, not new logic.

| Client | Config | Format | Removal |
| --- | --- | --- | --- |
| claude-code | `~/.claude.json`, `~/.claude/settings.json` | JSON | Delegate: `claude mcp remove pipefy` (scopes are `local\|user\|project`; with no `-s` it removes from whichever scope holds it), `claude mcp logout pipefy` for the hosted OAuth token, `claude plugin uninstall pipefy@pipefy` (defaults to `-s user`, and removes `~/.claude/plugins/data/{id}/` unless `--keep-data`), `claude plugin marketplace remove pipefy`. Probe for each verb rather than checking a client version. Direct edit for keys with no verb |
| cursor | `~/.cursor/mcp.json` | JSON | Direct edit of `mcpServers.pipefy` |
| claude-desktop | `~/Library/Application Support/Claude/claude_desktop_config.json` | JSON | Direct edit of `mcpServers.pipefy` |
| codex | `~/.codex/config.toml` | TOML | Excise `[mcp_servers.pipefy]` through the next top-level section |

Reuse the atomic tempfile + `os.replace` JSON pattern at `install.sh:317-369`. The TOML excision needs section awareness, not a line delete, because `install.sh:382-388` appends the block as raw text.

## Tiered approval

Print the full plan up front, then confirm in tiers. Per-action prompting pushes everyone to `--yes` and loses the safety entirely.

```
[1] Ours, reversible      bulk confirm
[2] Credentials           separate confirm, "cannot be undone"
[3] Your files            separate confirm, timestamped backup first
```

## Sequence

```
scan → present plan → tiered approval → live-server guard → credentials
     → client configs → tools → skills → runtime state → re-scan and report
```

Credentials precede tools because only `pipefy auth logout` can revoke server-side, and that capability disappears with the tool environment. Client configs precede tools because the reverse strands a registration pointing at a missing binary.

## Acceptance criteria

- [ ] `--dry-run`, `--yes`, `--client`, `--keep-credentials`, `--keep-config`
- [ ] Tiered approval as above; `--yes` approves every tier
- [ ] Client adapter table covering claude-code, claude-desktop, cursor, codex
- [ ] Backs up any file it did not itself create before editing, with a timestamped suffix, and writes atomically
- [ ] Live-server guard before anything touches tool environments; detects a running client and asks the user to quit, since a running client rewrites its config and can resurrect deleted entries
- [ ] **Never invokes `uv cache clean`** — uv hardlinks tool environments into the cache, so cleaning it breaks a running server. Suggest `uv cache prune` only when no server is running, and never execute it
- [ ] Enumerated keychain entries are removed, not just the production tuple. Where revocation was not possible, say so explicitly rather than implying the credential is dead
- [ ] Removes known children of `~/.config/pipefy` and removes the directory only if empty afterwards. States that a later CLI invocation recreates it
- [ ] Never machine-edits a git-tracked `.mcp.json`. The repo-root file has an explicit carve-out at `.gitignore:44-48` and is the plugin's own MCP config. Report it, and offer `disabledMcpjsonServers` in `~/.claude.json` as the git-clean remediation
- [ ] Never removes `uv`, and never edits shell rc PATH lines it added. Report them
- [ ] `~/.config/pipefy/config.toml` and `.env` files are removed only under explicit consent
- [ ] Clears the hosted OAuth token with `claude mcp logout pipefy`, which is per-server and leaves other authenticated servers alone. Probe for the verb; fall back to the `/mcp` → `pipefy` → Clear authentication instruction only when `claude` is absent or does not offer it
- [ ] States the limits it cannot cross: a child process cannot unset a variable in the parent shell, so a `PIPEFY_*` variable exported in the live session needs `unset` or a restart and the report must say which it found and where
- [ ] Ends with a re-scan, reporting removed / left alone and why / do this yourself, each with a copy-pasteable command
- [ ] Closing block prints the exact install command for switching to another channel

## Tests

Extend the phase-1 fixtures with a codex config appended by the installer and one hand-edited afterwards, which must degrade to report-only. Put a stub directory first on `PATH` containing fake `uv`, `claude`, `security`, `secret-tool`, and `pipefy` that append their argv to a log, then assert the **command sequence** — ordering is the part that matters most and filesystem assertions cannot see it. Never touch a real keychain.

mocha06:
## Motivation

The scan reports; nothing removes. Teardown is the point of the milestone, and the ordering it needs is not obvious enough to leave to whoever writes it next.

## Outcome

The destructive path, on the client table the scan already uses.

**Clients are data.** One row carries id, platform, format, config path, servers key, scope label, state directory, capabilities and cli. Adding a client is a row and no new code — `test_a_new_client_costs_one_row_and_no_new_logic` appends a synthetic row to a copy of the shipped script and asserts detection, planned removal, the file edit, the backup and `--client` narrowing all work with nothing else changed. `--help` and the accepted values are both derived from the table so they cannot drift.

**Capabilities gate the client-specific work, not the client id.** `scopes` drives the project walk and precedence, `plugin-system` the marketplace and skills directory, `removal-cli` the delegation. This change also retrofits the scan, whose Cursor and Claude Desktop handling was inline: both surfaces now enumerate clients exactly once.

**Ordering is load-bearing.** Revoke first, because only `pipefy auth logout` reaches the identity provider and that ability leaves with the tool environment. Then credentials, then client configs — before tools, or a registration is left pointing at a binary that no longer exists. Runtime state last, because `pipefy auth logout` and `auth status` both recreate `~/.config/pipefy`, so clearing it earlier clears nothing. A re-scan closes the run: `uv tool uninstall` can succeed while a binary of the same name is still earlier on `PATH`. The stub log asserts the sequence, which no filesystem check can see.

**Approval is tiered.** What the toolkit installed, bulk; credentials, separately and worded as permanent; your own files, separately and copied to `<file>.bak.<timestamp>` first. Around twenty individual prompts pushes everyone to `--yes` and loses the safety entirely. `--yes`, `--dry-run`, `--keep-credentials`, `--keep-config` and `--client` are all covered.

**One deletion primitive.** Every `rm` and `rmdir` routes through `remove_path`, which refuses an empty argument, a relative path, `/` and `$HOME`. The script's own tempfiles go through it too. Eight refusals are tested under `--dry-run`, so a broken guard echoes rather than deletes.

Registrations are removed under the name they were found with. The hosted OAuth token goes through `claude mcp logout`, gated on a probe for the verb rather than a client version — a stub whose help omits the verb falls back to the `/mcp` instruction while other delegation still works.

Never done, and reported instead: `uv cache clean` in any form, removing uv, editing the shell rc lines uv added, machine-editing a git-tracked `.mcp.json`, and chasing the editable-install entries in the uv cache, which belong to other repositories' virtualenvs. Where a marketplace or a tracked file is reported, the report says removal is not durable and what brings it back.

## Notes

Exercised end to end against a fixture: a registration named `pipefy-dev` removed while an unrelated server in the same file survived, the Codex section excised, timestamped backups beside both files, and the config directory cleared. A Codex section edited by hand degrades to report-only, since `install.sh` appends exactly a header and one command line.

The hosted-token behavior corrects #536, which claimed the token was not separately removable. `claude mcp logout <name>` clears one server and leaves the rest, verified against Claude Code 2.1.221; the issue has been updated.

Based on #581. Part of #536. Closes #539.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,12 +1,15 @@
 #!/bin/sh
-# uninstall.sh — report Pipefy toolkit state across every install channel and MCP client.
+# uninstall.sh — report and remove Pipefy toolkit state across every install
+# channel and MCP client.
 #
-# This version scans only. It reads; it never removes, moves, or edits anything,
-# and the single path it writes is a tempfile holding its own record stream.
+# `--scan` reads and reports; it never removes, moves, or edits anything, and
+# the only paths it writes are tempfiles holding its own record streams. Any
+# other invocation scans, prints the plan it derived from that scan, asks for
+# approval in tiers, removes what was approved, and scans again.
 #
 # Channels: the Claude Code plugin, a local install (install.sh / uv tool /
-# uvx), and the hosted HTTP server. Clients: Claude Code, Cursor, Claude
-# Desktop, Codex.
+# uvx), and the hosted HTTP server. Clients come from CLIENT_TABLE below;
+# adding one is a row there and no new code.
 #
 # A registration is matched on what it *runs*, not on what it is called. The
 # registration key is free text — a real environment had the server registered
@@ -18,7 +21,11 @@
 # and directory names, and `env | grep -i pipefy` matches PWD for anyone
 # working inside a directory named after it.
 #
-# Exit codes: 0 nothing found, 1 findings present, 2 the scan itself failed.
+# Removal uses what the scan found, including the name a registration was
+# actually registered under, and every deletion goes through remove_path().
+#
+# Exit codes: 0 nothing found / nothing left, 1 findings remain, 2 the run
+# itself failed.
 
 set -eu
 
@@ -30,17 +37,65 @@ HOSTED_HOST="mcp.pipefy.com"   # the documented hosted endpoint, and the only on
 RUNNERS="uvx uv npx python python3 pipx"
 PLUGIN_ID="pipefy@pipefy"
 MARKETPLACE_ID="pipefy"
+UV_TOOLS="pipefy-cli $SERVER_BINARY"
 
+# The client table. One row per MCP client, pipe-separated because a config
+# path can contain spaces:
+#
+#   id|platform|format|config|servers|scope|statedir|capabilities|cli
+#
+#   id            the --client value; the allowlist is derived from this column
+#   platform      `uname -s` this row applies to, or * for every platform
+#   format        json or toml — the only thing that selects a reader or an editor
+#   config        the file holding server registrations; ~/ is expanded
+#   servers       the key path registrations live under
+#   scope         how a registration from this file is labelled in reports
+#   statedir      client state directory (settings, plugin registry), or -
+#   capabilities  comma-separated, see below
+#   cli           the client's own command-line tool, or -
+#
+# Capabilities gate the client-specific work, so nothing branches on an id:
+#
+#   scopes         resolves one registration name across local/project/user/
+#                  plugin sources, with the whole entry taken from the winner
+#   plugin-system  ships a plugin and marketplace registry under statedir
+#   removal-cli    ships a CLI that can remove registrations, tokens and
+#                  plugins; each verb is still probed for before it is used
+#
+# Direct file editing is the baseline for every row. Delegation is the
+# exception one row happens to enable.
+CLIENT_TABLE="claude-code|*|json|~/.claude.json|mcpServers|user|~/.claude|scopes,plugin-system,removal-cli|claude
+claude-desktop|Darwin|json|~/Library/Application Support/Claude/claude_desktop_config.json|mcpServers|client:claude-desktop|-|-|-
+codex|*|toml|~/.codex/config.toml|mcp_servers|client:codex|-|-|-
+cursor|*|json|~/.cursor/mcp.json|mcpServers|client:cursor|-|-|-"
+
+MODE=teardown
 CLIENT=""
 ALLOW_ROOT=0
 DRY_RUN=0
+YES=0
+KEEP_CREDENTIALS=0
+KEEP_CONFIG=0
 
 OS=""
 CONFIG_DIR=""
 PYTHON3=""
 RECORDS=""
+PLAN=""
+NOTES=""
+COLLECTING=0
 FINDINGS=0
 SCAN_ERRORS=0
+BACKUP_STAMP=""
+BACKUPS=0
+DONE_COUNT=0
+FAIL_COUNT=0
+SKIP_COUNT=0
+REVOKED=0
+CRED_DELETED=0
+TIER1_OK=0
+TIER2_OK=0
+TIER3_OK=0
 TAB=$(printf '\t')
 
 # Secrets. `_session_masking_env_vars()` (packages/cli/.../commands/auth.py)
@@ -103,57 +158,186 @@ detail() { printf '      %s\n' "$*"; }
 # A source this run could not inspect, so a partial scan never reads as clean.
 uninspected() { printf '  ! %s\n' "$*"; SCAN_ERRORS=$((SCAN_ERRORS + 1)); }
 
+trace() { printf '+ %s\n' "$*" >&2; }
+
+run() {
+    trace "$*"
+    if [ "$DRY_RUN" -eq 1 ]; then
+        return 0
+    fi
+    "$@"
+}
+
+confirm() {
+    _cf_msg="$1"
+    if [ "$YES" -eq 1 ]; then
+        return 0
+    fi
+    if [ -t 0 ]; then
+        printf '%s [y/N] ' "$_cf_msg"
+        read -r _cf_reply || _cf_reply=""
+    elif [ -r /dev/tty ]; then
+        printf '%s [y/N] ' "$_cf_msg" >&2
+        read -r _cf_reply < /dev/tty || _cf_reply=""
+    else
+        err "No TTY available for prompt: \"$_cf_msg\". Re-run with --yes to proceed non-interactively."
+    fi
+    case "$_cf_reply" in
+        [yY]|[yY][eE][sS]) return 0 ;;
+        *) return 1 ;;
+    esac
+}
+
+# ------------------------------------------------------------- client table
+
+# The table writes paths with a literal ~/ prefix; this is where one becomes a
+# real path.
+# shellcheck disable=SC2088
+expand_home() {
+    case "$1" in
+        '~/'*) printf '%s\n' "$HOME/${1#\~/}" ;;
+        *) printf '%s\n' "$1" ;;
+    esac
+}
+
+# Every row, paths expanded, platform ignored. The allowlist and the help text
+# both come from here, so they cannot drift apart.
+all_client_rows() {
+    printf '%s\n' "$CLIENT_TABLE" \
+        | while IFS='|' read -r _id _plat _fmt _cfg _srv _sc _st _caps _cli; do
+        [ -n "$_id" ] || continue
+        printf '%s|%s|%s|%s|%s|%s|%s|%s|%s\n' \
+            "$_id" "$_plat" "$_fmt" "$(expand_home "$_cfg")" "$_srv" "$_sc" \
+            "$(expand_home "$_st")" "$_caps" "$_cli"
+    done
+}
+
+# The rows that apply to this machine.
+client_rows() {
+    all_client_rows | while IFS='|' read -r _id _plat _fmt _cfg _srv _sc _st _caps _cli; do
+        case "$_plat" in
+            '*'|"$OS") ;;
+            *) continue ;;
+        esac
+        printf '%s|%s|%s|%s|%s|%s|%s|%s|%s\n' \
+            "$_id" "$_plat" "$_fmt" "$_cfg" "$_srv" "$_sc" "$_st" "$_caps" "$_cli"
+    done
+}
+
+client_ids() {
+    all_client_rows | cut -d'|' -f1
+}
+
+client_row() {
+    all_client_rows | awk -F'|' -v id="$1" '$1 == id { print; exit }'
+}
+
+# Field n of a row: 1 id, 3 format, 4 config, 5 servers, 6 scope, 7 statedir,
+# 8 capabilities, 9 cli.
+client_field() {
+    printf '%s\n' "$1" | cut -d'|' -f"$2"
+}
+
+client_has_cap() {
+    printf '%s\n' "$1" | cut -d'|' -f8 | tr ',' '\n' \
+        | awk -v c="$2" '$0 == c { f = 1 } END { exit !f }'
+}
+
+# The first row claiming a capability. The plugin registry, the marketplace,
+# the skills directory and CLI delegation all hang off this rather than off a
+# client id.
+client_row_with_cap() {
+    client_rows | while IFS= read -r _cr; do
+        if client_has_cap "$_cr" "$1"; then
+            printf '%s\n' "$_cr"
+            break
+        fi
+    done
+}
+
+# Which client owns a config path, for --client narrowing and for reports.
+client_id_for_file() {
+    client_rows | awk -F'|' -v f="$1" '$4 == f { print $1; exit }'
+}
+
+# --client narrows *registration edits* only; detection always sweeps every
+# client, and tools, credentials, skills and the plugin are unaffected.
+client_selected() {
+    [ -z "$CLIENT" ] && return 0
+    [ "$1" = "$CLIENT" ]
+}
+
+client_choice_list() {
+    _ccl=$(client_ids | tr '\n' '|')
+    printf '%s\n' "${_ccl%|}"
+}
+
 print_help() {
+    _choices="$(client_choice_list)|none"
     cat <<EOF
 Usage: uninstall.sh [OPTIONS]
 
-Report Pipefy toolkit state across every install channel and MCP client.
-This version detects only; it removes nothing.
+Report Pipefy toolkit state across every install channel and MCP client, then
+remove what you approve. A bare invocation scans, shows the plan, and asks.
 
 MCP registrations are matched on what they run — the $SERVER_BINARY
 command, a known runner invoking it, or the hosted endpoint's host — so an
-entry registered under any name is found.
+entry registered under any name is found, and removed under that name.
 
 Options:
-  --scan              Report state and exit. The only mode this version has,
-                      and what a bare invocation does.
-  --client <id>       Accepted and validated for symmetry with install.sh.
-                      One of: claude-code, claude-desktop, cursor, codex, none.
+  --scan              Report state and exit. Changes nothing.
+  --client <id>       Limit MCP registration edits to this client's config.
+                      One of: $_choices.
                       Detection always sweeps every client regardless, since
-                      the point is finding state you forgot about.
-  --dry-run           Accepted for symmetry with install.sh. A scan has no
-                      side effects, so this changes nothing.
+                      the point is finding state you forgot about, and tools,
+                      credentials, skills and the plugin are not narrowed.
+  --dry-run           Print the plan and exit without removing anything.
+  --yes, -y           Approve every tier. Required with no TTY.
+  --keep-credentials  Leave stored credentials alone (skips tier 2).
+  --keep-config       Leave user configuration alone: config.toml, .env, and
+                      non-credential PIPEFY_* settings in your own files.
   --allow-root        Allow running as root (refuses by default).
   -h, --help          Show this help.
 
+Approval is tiered, not per action:
+  [1] ours and reversible   one confirmation for the whole group
+  [2] credentials           its own confirmation; cannot be undone
+  [3] your own files        its own confirmation; backed up first
+
 Exit codes:
-  0  nothing found
-  1  findings present
-  2  the scan itself failed (a source could not be inspected)
+  0  nothing found, or nothing left after removal
+  1  findings remain
+  2  the run itself failed (a source could not be inspected)
 
 Examples:
   curl -LsSf https://raw.githubusercontent.com/$REPO/main/uninstall.sh \\
     | sh -s -- --scan
 
-  ./uninstall.sh --scan
+  ./uninstall.sh --dry-run
+  ./uninstall.sh --yes --keep-config
 EOF
 }
 
 parse_args() {
     while [ $# -gt 0 ]; do
         case "$1" in
-            --scan) shift ;;
+            --scan) MODE=scan; shift ;;
             --client) [ $# -ge 2 ] || err "--client requires a value"; CLIENT="$2"; shift 2 ;;
             --client=*) CLIENT="${1#--client=}"; shift ;;
             --dry-run) DRY_RUN=1; shift ;;
+            --yes|-y) YES=1; shift ;;
+            --keep-credentials) KEEP_CREDENTIALS=1; shift ;;
+            --keep-config) KEEP_CONFIG=1; shift ;;
             --allow-root) ALLOW_ROOT=1; shift ;;
             -h|--help) print_help; exit 0 ;;
             *) err "Unknown flag: $1 (try --help)" ;;
         esac
     done
     case "$CLIENT" in
-        ""|claude-code|claude-desktop|cursor|codex|none) ;;
-        *) err "Invalid --client: $CLIENT (use claude-code|claude-desktop|cursor|codex|none)" ;;
+        ""|none) ;;
+        *)
+            [ -n "$(client_row "$CLIENT")" ] \
+                || err "Invalid --client: $CLIENT (use $(client_choice_list)|none)" ;;
     esac
 }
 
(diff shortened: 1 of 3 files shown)


uninstall.sh
```diff
@@ -913,6 +1159,30 @@ codex_sections() {
...
+    AW_KEY="$2" AW_NAME="$3" AW_BIN="$SERVER_BINARY" awk '
+        function trim(s) { sub(/^[ \t]+/, "", s); sub(/[ \t]+$/, "", s); return s }
+        BEGIN {
+            sec = "[" ENVIRON["AW_KEY"] "." ENVIRON["AW_NAME"] "]"
+            want = "command = \"" ENVIRON["AW_BIN"] "\""
+        }
+        {
+            t = trim($0)
+            if (inside) {
+                if (substr(t, 1, 1) == "[") { inside = 0 }
```
adriannoes, line 1175: **Blocking.** `toml_section_is_pristine` stops at the next `[` header, so a sibling `[mcp_servers.<name>.env]` never marks the parent hand-edited. Teardown then runs `act_toml`, which deletes only until that next `[` and leaves the env table (and its secrets) in `~/.codex/config.toml`. The closing re-scan does not match orphan env tables, so the run can report Codex clean while credential material remains. The comment above the pristine helper already says a sub-table should be report-only; this path silently excises the parent instead.

I would treat any dotted child table under the section as non-pristine (or remove child tables with the parent), and make re-scan fail closed on leftover `[mcp_servers.<name>.env]`.

Done when:

- pristine parent + sibling `.env` with fake secrets is either report-only or fully cleared
- re-scan does not claim Codex clean while the env table remains
- regression test covers this shape (existing hand-edit fixture only covers parent+args+env together)

uninstall.sh
```diff
@@ -1450,10 +2611,98 @@ EOF
...
+
+    if [ -z "$PYTHON3" ]; then
+        err "removal needs python3: without it the JSON sources above were never read, and a teardown planned from a partial scan would leave exactly the state that causes conflicts. Install python3, or use --scan."
+    fi
+
+    present_plan
+    if [ "$(count_lines "$PLAN")" -eq 0 ]; then
+        say ""
+        say "Nothing to remove."
+        exit 0
```
adriannoes, line 2678: **Blocking.** After planning, if every remediation is `note_left` only (for example a hand-edited Codex section), `PLAN` is empty and the script prints `Nothing to remove.` then `exit 0` here. The same tree under `--scan` exits `1` because `FINDINGS>0`. Automation that treats exit 0 as clean is wrong, and this early exit also skips the teardown report so the Left alone notes never appear.

I would exit nonzero when `FINDINGS>0` even if `PLAN` is empty (align with `--scan`), and still print the queued Left alone / manual notes before exit.

Done when:

- hand-edited-Codex-only HOME: teardown (non-scan) exit ≠ 0
- Left alone / manual notes are visible on that path
- regression locks the exit contract

uninstall.sh
```diff
@@ -103,57 +158,186 @@ detail() { printf '      %s\n' "$*"; }
...
+}
+
+# Which client owns a config path, for --client narrowing and for reports.
+client_id_for_file() {
+    client_rows | awk -F'|' -v f="$1" '$4 == f { print $1; exit }'
+}
+
+# --client narrows *registration edits* only; detection always sweeps every
+# client, and tools, credentials, skills and the plugin are unaffected.
+client_selected() {
```
adriannoes, line 265: **Blocking.** Help appends `none` to the `--client` allowlist. `client_selected` only matches real client ids, so `none` matches nothing: every MCP registration edit is skipped while tools (and other non-narrowed work) still uninstall. That recreates registrations pointing at a missing binary, which this PR's ordering exists to prevent. `install.sh --client none` means "install without registering"; uninstall reuses the token with the opposite footgun and does not document a tools-only mode.

I would drop `none` from uninstall, or redefine it explicitly (refuse tool removal when every registration edit is excluded, or treat it as all-clients). Add a dry-run regression: `none` must not plan tool uninstall without registration removals unless the new contract says so out loud.

Done when:

- `--client none` cannot strand detected registrations after tool removal (or is rejected)
- help text matches the behavior
- dry-run fixture covers `none` vs bare

uninstall.sh
```diff
@@ -1358,12 +1742,731 @@ $(records err)
...
+    _mk_cli=$(removal_cli) || _mk_cli="-"
+    note_left "the '$1' marketplace can come back: a later session re-adds it from a settings file that still lists it or from a plugin that still needs it, so confirm with a fresh session"
+    if cli_verb_exists "$_mk_cli" remove "plugin marketplace"; then
+        run "$_mk_cli" plugin marketplace remove "$1"
+        return $?
+    fi
+    backup_file "$2" || return 1
+    json_remove_key "$2" "$1" || return 1
+    if [ "$3" != "-" ] && [ -d "$3" ]; then
+        remove_path "$3"
```
adriannoes, line 2407: **Blocking.** When the marketplace CLI verb is absent, `act_market` calls `remove_path` on whatever absolute directory `installLocation` names. Under a fake HOME, a poisoned `known_marketplaces.json` deleted a canary directory while the plan only said "unregister marketplace." Tier-1 consent does not disclose arbitrary path deletion. (Pairs with the `remove_path` canonicalize gap on the same chain.)

I would delete only if the path equals or is strictly under the canonical `…/plugins/marketplaces/$MARKETPLACE_ID` clone; otherwise skip and warn.

Done when:

- poisoned `installLocation` outside the canonical tree is not removed
- regression covers CLI-absent fallback
- see also the `remove_path` canonicalize thread

uninstall.sh
```diff
@@ -1358,12 +1742,731 @@ $(records err)
...
+    say "$*"
+    exit 1
+}
+
+# ------------------------------------------------------- removal primitives
+
+# The one path this script deletes through. Everything routes here so the
+# refusals cannot be skipped by a caller that forgot them. Tempfiles this
+# script creates come back through here too.
+remove_path() {
```
adriannoes, line 2043: **Blocking.** `remove_path` strips trailing slashes and string-compares to `/` and `$HOME` only. Dry-run under a fake HOME accepts `$HOME/.`, `$HOME/..`, and `/tmp/..` and traces `rm -rf`. Reachable today via marketplace `installLocation` (see sibling thread); the function is documented as the sole deletion chokepoint, so the guard's claim is false even if other feeders stay clean.

I would resolve/`realpath` (or reject any `.` / `..` segment) before the `/` / `$HOME` checks, and add refusal tests for those three shapes.

Done when:

- `$HOME/.`, `$HOME/..`, `/tmp/..` refuse (dry-run or live under fake HOME)
- exact `/` and `$HOME` still refuse
- see also the marketplace confinement thread

uninstall.sh
```diff
@@ -1358,12 +1742,731 @@ $(records err)
...
+    node = node[key]
+if not isinstance(node, dict) or keys[-1] not in node:
+    sys.exit(3)
+del node[keys[-1]]
+fd, tmp = tempfile.mkstemp(
+    prefix=os.path.basename(path) + ".", dir=os.path.dirname(path) or "."
+)
+try:
+    with os.fdopen(fd, "w", encoding="utf-8") as fh:
+        json.dump(data, fh, indent=2)
```
adriannoes, line 2141: **Blocking.** `json_remove_key` / `json_add_disabled` rewrite the whole file with `json.dump(..., indent=2)` and default `ensure_ascii=True`. Removing one registration turns unrelated UTF-8 sibling values into `\uXXXX` escapes. Live `~/.claude.json` already contains literal UTF-8; this is silent config mutation on every teardown JSON edit (semantic `json.loads` round-trip still works, but the on-disk file changes without consent).

I would pass `ensure_ascii=False` on both dumps (and keep install parity in mind) plus a regression with non-ASCII sibling keys.

Done when:

- UTF-8 siblings stay literal after a registration removal
- regression fails if `ensure_ascii` defaults return

uninstall.sh
```diff
@@ -1358,12 +1742,731 @@ $(records err)
...
+            [ -n "${_knd:-}" ] || continue
+            tier_approved "$_cls" || continue
+            _rc=0
+            do_action "$_cls" "$_knd" \
+                "$_p1" "$_p2" "$_p3" "$_p4" "$_p5" "$_p6" "$_p7" || _rc=$?
+            # 3 means the action decided there was nothing to do and said why
+            # in a note of its own; anything else non-zero is a failure.
+            if [ "$_rc" -eq 0 ]; then
+                DONE_COUNT=$((DONE_COUNT + 1))
+                note_done "$_dsc"
```
adriannoes, line 2445: **Blocking.** `execute_plan` calls `note_done` whenever the action returns 0. For hosted `mcplogout`, a no-op `claude mcp logout` that exits 0 yields `Removed: clear the stored OAuth token …` and process exit 0. The OAuth store is opaque to re-scan, so nothing contradicts the lie. (Filesystem CLI no-ops still get exit 1 from re-scan; that arm is messaging-only and not a separate blocker.)

I would record hosted logout as invoked-not-verified (or manual) unless there is an observable proof, and avoid `CRED_DELETED` / verified Removed language on exit 0 alone.

Done when:

- logout no-op stub does not print verified Removed / does not claim credential clearance
- process exit does not read as full success for that opaque clear
- optional: demote filesystem `done` notes still present after re-scan

adriannoes (changes requested):
Reviewed at `469f433`: detached worktree, Phase 3 full local gates (4256 unit + 84 uninstall), and fake-HOME runtime repros for the teardown path. No live MCP smoke (ops script only).

**Verdict: needs changes** into `rc-dev/feat/uninstall-scan`.

### Required before merge

- Codex pristine excision can leave `[mcp_servers.*.env]` secrets while re-scan reports clean. See the thread on `uninstall.sh:1175`.
- Empty PLAN with FINDINGS still exits 0 (`Nothing to remove.`), unlike `--scan`. See the thread on `uninstall.sh:2678`.
- `--client none` uninstalls tools and skips every registration edit. See the thread on `uninstall.sh:265`.
- Marketplace fallback deletes arbitrary `installLocation` paths. See the thread on `uninstall.sh:2407`.
- `remove_path` does not canonicalize, so `$HOME/..` / `/tmp/..` bypass the `/`/`$HOME` refusals. See the thread on `uninstall.sh:2043`.
- JSON rewrites use default `ensure_ascii=True` and mutate UTF-8 siblings. See the thread on `uninstall.sh:2141`.
- Hosted OAuth logout treats CLI exit 0 as verified `Removed:` with process exit 0. See the thread on `uninstall.sh:2445`.

### Decisions I would make

- Accept the ~2.7k-line `uninstall.sh` hotspot for this milestone unless you already planned a split; the Blocking items above are the merge gate, not a forced extract.
- Keep stack order #581 before #582; #583 can follow. `Closes #539` may need a manual close from a feature-branch base.

### Optional

Your call on all of these.

- Dry-run could print Left alone notes before exit; preview incompleteness alone would not block merge.
- Closing "Switching to another channel" copy could gate on a clean re-scan; mid-report already carries residuals.
- Teach the `_UV` stub to drop tools after uninstall so a clean re-scan exit can be asserted.

### What worked well

- Load-bearing ordering (revoke → creds → configs → tools → state → re-scan) with stub-sequence tests.
- Tiered approval and a single deletion choke intent.
- Never-do list (uv cache clean, tracked `.mcp.json`, shell-rc uv PATH) with behavioral coverage.
- Client table + capability gating as the extension point.

<details>

Tier `deep`, trust `trusted`, reviewed at `d59a061c…`…`469f4331…`. CI class `none`; local Phase 3 PASS. H16 n/a. Refutations executed under fake HOME for the seven Blocking threads (blocking uncapped past the discretionary 0–5 optional target).

</details>

uninstall.sh
mocha06: Fixed in `13aa5db`. Confirmed the mechanism before fixing: the section regex is `^\[key\.[^.]+\]$`, so a dotted child table is invisible to both the excision and the re-scan — secrets outlive a run that reports Codex clean.

`toml_section_is_pristine` now treats any `[key.name.*]` child as hand-edited wherever it sits, so parent and child are reported together and neither is excised. Test: a pristine parent plus `[mcp_servers.pipefy.env]` holding a token — file unchanged, "sub-table" named in the report, exit non-zero, and the token never printed.

uninstall.sh
mocha06: Fixed in `13aa5db`. `Nothing to remove.` now prints the queued *Left alone* / *Do this yourself* notes and takes `--scan`'s exit codes: 2 on scan errors, 1 on findings, 0 otherwise. A test asserts teardown's exit equals the scan's on a hand-edited-Codex-only home, with a clean-machine test pinning 0.

One consequence beyond the literal ask, flagged rather than folded in: `--dry-run` now also exits 1 when the plan is empty but findings remain, while still exiting 0 when it printed a plan. "Nothing to remove" is a claim about the machine; a printed plan is not.

uninstall.sh
mocha06: Fixed in `dead767`. Confirmed the stranding first — `--dry-run --yes --client none` on a fixture with one Cursor registration planned two `uv tool uninstall` actions and zero registration edits, exit 0.

Took your first option: `none` is refused outright. `print_help` now generates the allowlist from the client table alone, and `parse_args` errors with the reason — `none` belongs to `install.sh`, where it means "install without registering"; teardown has no equivalent, because "edit no client config but remove the binaries every config points at" has no coherent meaning. Inventing a tools-only mode would still have needed the stranding refusal to be useful.

`install_ids - {none} <= uninstall_ids` still holds. I also strengthened `test_the_advertised_client_values_are_the_accepted_ones`, which previously just returned early on `none`: it now asserts exit 2 with that message, so the cross-script contract is refused rather than merely unlisted.

uninstall.sh
mocha06: Fixed in `fbe6753`. Agreed on the principle — this was untrusted data authorising a delete, which is exactly the rule #540 established for the receipt and which the marketplace path predated.

The clone path is now derived from the client table and gated by a confinement check, canonicalized on both sides. An `installLocation` pointing anywhere else is reported and left, and the plan discloses the clone when it will be removed. Test: a poisoned registry pointing at a canary directory with the CLI absent — canary survives, registry key still removed.

uninstall.sh
mocha06: Fixed in `fbe6753`. Verified the bypass: no `realpath`, no `cd -P`, just string comparison after stripping trailing slashes, so `$HOME/..` and `/tmp/..` passed the refusals.

`canonical_path` does lexical `.`/`..` cancellation first — `realpath -e` is wrong here because the path often does not exist — then `cd -P`/`pwd -P` on the surviving parent. The final segment is deliberately left unresolved so a symlink is judged and deleted as a link rather than followed to its target. It now refuses `/`, `$HOME`, and any ancestor of `$HOME`; `$HOME/x` and `$HOME/link` still pass. 19 guard tests, including the traversal cases.

uninstall.sh
mocha06: Fixed in `f9bc5b1`. Real defect, and it was in three places, not two — `install.sh`'s merge has the same default. All three dumps now pass `ensure_ascii=False`, with tests on both sides asserting `café` / `naïve` stay literal and `\u` never appears in a rewritten file.

uninstall.sh
mocha06: Fixed in `dead767`, and this one got sharper while the branch was in flight.

A live teardown against a real machine showed `claude mcp logout` binding to a **rejected project-scope `.mcp.json` entry** instead of the user-scope hosted server, exiting 1. The verb resolves by name across scopes and has no `--scope` flag, and credentials ran before client configs, so the shadowing entry was always still present. `fbe6753` reorders the phases (6 → 8): unshadow project registrations at 3, hosted token at 4, remaining client configs at 5, with the hosted row keeping its own registration until after its logout. That makes a 0 exit *more* likely, which makes your point more urgent, not less.

So `act_mcplogout` now returns a distinct status and `execute_plan` has a fourth outcome: *Asked for, result not observable from here*, with no `DONE_COUNT` and no `CRED_DELETED`. The counter reads `N unverifiable` and the process exits 1 even on a clean re-scan. Test asserts a no-op stub yields no OAuth line under *Removed:*, `1 unverifiable`, and exit 1.

I scoped this to the opaque store as you marked it. Filesystem `done` notes surviving a re-scan are left alone — that state is observable and already forces exit 1, so it is messaging-only.

mocha06:
All seven blocking items are fixed, in four commits on `rc-540/ci/shell-script-coverage` (#584) rather than on this branch — putting them here would force-push #583 and #584 and destroy the incremental diff. Per-thread detail is in the replies above; each names its commit.

I verified every claim before fixing rather than taking them on trust, and every one held. Two were sharper than they first looked:

- The `installLocation` and `remove_path` findings are the same class — data the script did not write authorising a delete. #540 established that rule for the receipt ("the receipt can only ever authorise less") and both of these predate it.
- The hosted-logout finding got worse while the branch was in flight. A live teardown against a real machine showed `claude mcp logout` binding to a rejected project-scope `.mcp.json` entry rather than the user-scope hosted server, because the verb resolves by name across scopes with no `--scope` flag and credentials ran before client configs. Fixing the ordering makes a 0 exit more likely, so your point about exit 0 not being proof became more urgent, not less.

**One item I am declining:** requiring an explicit flag for teardown once bare invocation is no longer scan-only. #536 specifies `curl -LsSf … | sh` as the entry point, and bare already scans, prints the full plan, and asks for approval in three tiers — it is not unattended. `--yes` is the unattended path and it is opt-in. Adding a flag would contradict both the issue and the shipped `docs/uninstall.md`. Say the word if you want to overrule that and I will change it.

That live teardown also found a bug no reviewer could have seen from the diff: tier 1 deleted every `~/.claude/skills/pipefy-*` directory under the label "reinstalling restores it", and destroyed a user-authored skill this repository never shipped. Provenance now comes from the `skills add` lock file rather than the name prefix. It is fixed in `fbe6753`, and it is the strongest argument for your position that name-shaped matching is not provenance.

adriannoes (dismissed):
Re-checked after your thread replies and the summary on this PR. Head is still `469f433`; the seven Blocking items from the earlier review are fixed in four commits on #584 (`rc-540/ci/shell-script-coverage`, tip `dead767`) rather than on this branch, so the incremental stack (#583 / #584) stays intact.

I spot-checked that tip against each Done when: Codex sub-table treated as hand-edited with non-clean exit; empty plan with findings exits like `--scan` and prints Left alone; `--client none` refused; marketplace clone confined; `canonical_path` before `remove_path` refusals; `ensure_ascii=False` on install and uninstall JSON dumps; hosted logout recorded as unverifiable without `CRED_DELETED`. Accepting the decline of an extra bare-teardown flag (not part of those Blocking asks). The live finding that name-prefix skill deletion removed a user-authored skill, fixed via `skills add` lock provenance in `fbe6753`, is a good catch.

**Verdict: merge-ready** into `rc-dev/feat/uninstall-scan` under the stacked delivery model. Do not treat `469f433` alone as the shippable uninstall tip; land through the stack that includes #584 before anyone relies on teardown.

adriannoes (approved):
Reviewed at `d23bffe`: local ci.yml-equivalent gates green, plus isolated HOME probes on the prior Blocking modes. Those modes still reproduce on this SHA in isolation. They are fixed on the stack tip [#584](https://github.com/pipefy/ai-toolkit/pull/584) at `38ea762` (already APPROVED), after [#583](https://github.com/pipefy/ai-toolkit/pull/583) (also APPROVED).

**Verdict: merge-ready** into `dev` under stacked delivery.

### Decisions I would make

- Merge order is `#582` then `#583` then `#584`. Do not treat this head alone as the production-safe teardown tip.
- Do not re-open the same Blocking threads here; the Done-whens landed on `#584` (`dead767`, `13aa5db`, `fbe6753`, `f9bc5b1`, `1cb4cd8`, and siblings).
- Keep the declined bare-teardown flag closed.
- `Closes #539` may need a manual close when the stack reaches the default branch.

### Optional

Your call on all of these.

- Residual note already on `#584`: the Linux install/teardown round-trip job has never executed.
- Unrelated open work (`#588`, `#589`, dependabot, stale drafts) does not gate this stack.

### What worked well

- Tiered approval, client table as data, never-do list, and stub-sequence ordering tests on this PR.
- Fast follow-through on `#583`/`#584` so the ship tip is already signed off.

<details>

Tier `deep`, trust `trusted`, frozen `d52922a3...d23bffe`. Local full gates PASS. H16 n/a (no MCP/CLI package surface). Open-PR survey: only `#583`/`#584` are stack-critical.

</details>

PR #583, opened 2026-08-05
feat(install): write an install receipt so teardown is exact

### LINKED ISSUE #540: install: write an install receipt so teardown is exact

mocha06:
## Motivation

`install.sh` knows four things and then throws them away, so teardown has to guess at all four: the `UV_TOOL_DIR` a `--prefix` set, whether it installed uv itself, whether a client registration was created or found already present, and where the skills step landed.

The middle one matters most. The JSON merge deliberately preserves a pre-existing entry and forgets it did so, so the installer cannot tell an entry it created from one it found — which means teardown cannot either.

## Outcome

A receipt: plain `key=value` at `${XDG_STATE_HOME:-~/.local/state}/pipefy/install-receipt`, read by one awk program so teardown never inherits the installer's `python3` dependency.

**It is installer state, not user configuration**, which is why it is not in `~/.config/pipefy`. A file there would be reported as one this toolkit does not write, and would block the `rmdir` of a directory that must end empty.

**Append-only, one record per run, written incrementally**, so re-running with a different `--client` keeps the earlier record and a crashed install still leaves usable lines. Nothing is written under `--dry-run`.

**Records merge monotonically.** Every field answers "did this machine ever get X from an install run", which only becomes true. Directories and tool names union; booleans OR. Last-wins would strand the tools an earlier `--prefix` created, exactly as having no receipt does.

**The receipt can only ever authorise less.** No value in it names something to delete. It selects which directory to interrogate and whether a registration the structural scan already found may be removed. An unknown schema is skipped whole rather than half-read, a malformed line is counted and dropped, a recorded directory that no longer exists is reported rather than used. Eight corrupt shapes are covered.

Values are escaped for backslash, tab, CR and newline, and the reader validates the alphabet before expanding, so a value carrying an escape the writer never emits is dropped rather than interpreted. A `--prefix` containing a space round-trips through both scripts.

**Heuristic mode is permanent, not transitional** — every install predates the receipt, and only a version of `install.sh` that writes one produces one. It removes less: in a config the installer writes, it deletes a registration only where the value is exactly the single command the installer writes, reports the rest, and never treats uv as ours. It says so before anything else.

## Notes

Fixes a real bug in the shipped installer, found by the crashed-install test. `run_quiet` read `$?` after an `if` whose condition failed, and an `if` with no `else` exits 0, so a failing `uv tool install` reported success. Under `set -eu` the installer then carried on to register a server pointing at a binary it had not installed. Reproduced standalone: a command exiting 7 came back as 0. This is worth reading on its own — it affects the `install.sh` currently served from `main`.

The receipt records a closed key vocabulary, so an unrecognised key cannot be introduced silently, and a mutation test plants a credential into the writer to prove the leak check is not vacuous.

No test reaches the real `uv tool install`, `npx skills add`, or the network.

Based on #582. Part of #536. Closes #540.


uninstall.sh
```diff
@@ -374,6 +394,184 @@ config_toml_path() {
...
+            if (!ok) { if (newer) future++; else unreadable++ }
+            started = 0
+        }
+        BEGIN { max = ENVIRON["AW_MAX"] + 0 }
+        {
+            p = index($0, "=")
+            if (p == 0) { junk++; next }
+            k = substr($0, 1, p - 1)
+            v = substr($0, p + 1)
+            if (k !~ /^[a-z][a-z0-9_.]*$/) { junk++; next }
```
adriannoes: **Blocking.** The receipt writer records `entry_created.claude-desktop=…`, but this awk key check rejects hyphens (`^[a-z][a-z0-9_.]*$`). Those lines are dropped as malformed, so `entry_created=false` never reaches `registration_is_ours`.

On Darwin, an installer-shaped Claude Desktop `pipefy` entry that install recorded as already present is planned for removal, and with a normal PATH (including `cp`) the apply step empties the config. The same flow with `entry_created.cursor=false` correctly leaves Cursor alone, so the safety bit works only when the client id matches the charset.

I would either allow `-` in the reader key alphabet (and any mirrored writer validation), or write a hyphen-free client token in the receipt, then assert writer keys are a subset of the reader charset. Add a Darwin or forced-platform contract: `entry_created.claude-desktop=false` must not plan-remove an installer-shaped Claude Desktop registration.

Done when:

- `entry_created.claude-desktop=false` is accepted by the reader and leaves the entry alone
- a regression test fails if the hyphenated key is junked again
- closed-vocabulary / charset checks keep writer and reader in lockstep

```suggestion
            if (k !~ /^[a-z][a-z0-9_.-]*$/) { junk++; next }
```

adriannoes (changes requested):
Reviewed at `c9a58ec`: detached worktree, full local ci.yml gate set (CI class none on this feature-branch base), hermetic `tests/test_install_receipt.py` (43 passed), plus adversarial install/uninstall repros for the receipt reader.

**Verdict: needs changes** into `rc-537/feat/uninstall-teardown`.

### Required before merge

- Writer and reader disagree on whether `-` is legal in receipt keys, so `entry_created.claude-desktop=false` is dropped as junk and uninstall deletes a pre-existing Claude Desktop registration. See the thread on `uninstall.sh:476`.

### Decisions I would make

- Keep the stack merge order #581 → #582 → #583 → #584. Do not retarget this PR onto `dev` alone.
- `Closes #540` will not auto-close when this PR merges into a non-default base; close the issue when the stack reaches the default branch, or carry an explicit closer there.
- Deleting a future-schema-only receipt on uninstall matches #540 (receipt is inventory and is removed). I would not treat that as a bug.

### Optional

Your call on all of these.

- `uninstall.sh` continues the single-file growth path from #582 (2714 → 3114). I would keep the #582 waive/split conversation there rather than block #540 for density.
- JSON vs TOML hand-edit vs receipt provenance rules are asymmetric; worth documenting in one place if you touch that code again.

### What worked well

- Append-only incremental receipt, monotonic merge, authorize-less corrupt suite, permanent heuristic messaging, and the `run_quiet` exit-status fix with a crashed-install regression are solid.
- Hermetic tests avoid real `uv`/`npx`/network and the closed-vocabulary mutation check has teeth.

<details>

Tier `deep`, trust `trusted`, reviewed at `469f4331…c9a58ec14…`.

| Gate | Result |
| :--- | :--- |
| Local full ci.yml equivalent | pass |
| test_install_receipt | 43 passed |
| Full non-integration suite | 4299 passed |
| H16 live MCP/CLI | n/a |
| Refutation | hyphen SURVIVES; future-schema REFUTED |

Not covered by the 43-test suite: Darwin `claude-desktop` receipt round-trip (gap that hid the blocking finding).

</details>

uninstall.sh
mocha06: Fixed in `f9bc5b1`. Reproduced before fixing, and the side-by-side is stark: one receipt, one run, `entry_created.cursor=false` parsed and the Cursor entry preserved, `entry_created.claude-desktop=false` dropped as junk and the Claude Desktop entry deleted. The reader alphabet was `^[a-z][a-z0-9_.]*$` while the writer emits `entry_created.$CLIENT`, so every hyphenated client id — `claude-desktop` and `claude-code` — lost its provenance and fell through to the installer-shape heuristic.

Reader alphabet now accepts `-`. Added the Darwin `claude-desktop` round-trip the 43-test suite was missing, plus a writer-subset-of-reader charset contract test that expands `entry_created.$1` over the advertised `--client` ids, so the two ends cannot drift apart again.

adriannoes: Thanks for the fix write-up. I re-checked the tip of this PR: it is still `c9a58ec`, and `uninstall.sh` still has `^[a-z][a-z0-9_.]*$` (no hyphen).

`f9bc5b1` / `f9bc5b17` is on #584 (stacked on this tip), where the reader alphabet, Darwin `claude-desktop` round-trip, and writer-subset-of-reader contract look right. Until that change is on *this* PR's head (cherry-pick / restack), or we are reviewing the stacked tip that actually merges, I have to keep CHANGES_REQUESTED here. Happy to re-verify and approve as soon as the fix lands on the commit this PR would merge.

mocha06:
Fixed on this PR's head in `d2c8f6e`. You were right to hold, and the same point had just landed on #581 — a Required item is about the commit this PR merges, not what the stack eventually contains.

I did not cherry-pick `f9bc5b1` wholesale: it bundles the receipt fix with parity and teardown tests that belong to #584 and reference files that do not exist at this tip. What is here is only what #540 owns:

- The reader alphabet takes `-`, so `entry_created.claude-desktop` and `entry_created.claude-code` survive instead of being counted as malformed.
- `install.sh`'s JSON merge stops rewriting a sibling server's UTF-8 values into escapes.
- The Darwin `claude-desktop` round-trip the 43-test suite was missing, plus the UTF-8 sibling test.

Both tests fail against `3eac63d` and pass on `d2c8f6e`, so they have teeth rather than just documenting the new behaviour. `a9b2c20` is a formatting-only follow-up.

This head also now carries the README and onboarding-skill scan surfaces, merged up from #581 where the same argument applied.

adriannoes (dismissed):
Re-verified at `a9b2c20`.

The blocking hyphen gap is fixed in `d2c8f6e`: reader alphabet accepts `-`, Darwin `claude-desktop` found-not-created round-trip keeps the entry, and `tests/test_install_receipt.py` is green (45 passed locally).

**Verdict: merge-ready** into `rc-537/feat/uninstall-teardown` (stack order still #581 → #582 → this PR → #584).

adriannoes (approved):
Re-verified at `2446fce` after the retarget to `dev` and the merge of `origin/dev` (prior APPROVE on `a9b2c20` was dismissed).

Receipt surface unchanged vs the last approval: hyphen charset, Darwin `claude-desktop` found-not-created coverage, and CI green (lint/test/DCO/skills).

PR #584, opened 2026-08-05
ci: lint both shell scripts, prove parity, document teardown

### LINKED ISSUE #541: ci: shellcheck, install/uninstall parity, and a round-trip job

mocha06:
## Motivation

Neither `install.sh` nor `uninstall.sh` had any CI coverage: `install.sh` was referenced nowhere under `.github/`, and there was no shell linting in `.pre-commit-config.yaml`. The repository also had five scattered removal one-liners and no uninstall documentation.

## Outcome

**Both scripts are linted rather than trusted.** `shellcheck --shell=sh` and `sh -n` run from pre-commit and from a CI job that adds `dash -n`, since `/bin/sh` on macOS is bash in POSIX mode and forgives what a Debian-family `/bin/sh` rejects. The hook comes from `shellcheck-py` rather than the Docker-based alternative, so a contributor without Docker still gets it, and both sides pin the same release.

**The deletion guard is enforced, not assumed.** A check asserts every `rm` and `rmdir` in `uninstall.sh` sits inside `remove_path`, and that its four refusals — empty argument, relative path, `/`, `$HOME` — are still there. That is the cheapest defence against an unset-variable `rm -rf`, and the kind of absence a reviewer cannot see. Two mutations prove the check bites.

**The parity suite observes both scripts instead of reading them.** It snapshots a fixture `HOME` around a real install and a real teardown for each `--client` value, so a path a future installer step writes has to be one teardown accounts for, without anyone predicting it. Two leftovers are allowlisted with reasons — the client's own config file, which by teardown time may hold other servers, and the timestamped backup — and the config-file case additionally asserts the leftover no longer registers the server, so it is not a free pass. A mutation that makes the installer write a rogue path is caught.

**A Linux round-trip job** installs and tears down against real network and real uv, which is the only check that proves completeness rather than the correctness of steps someone thought to write. A middle snapshot means a teardown that removed nothing cannot pass by comparing equal. Its allowlist is explicit and each entry carries its reason.

**`docs/uninstall.md`** documents the scan and its exit codes, the `python3` degradation, the matching rule, the three approval tiers, every flag, the six-phase ordering and why each phase sits where it does, heuristic mode, the `PIPEFY_KEYCHAIN_BACKEND` store-move trap, what is never removed by design, the two things that come back, and remove-then-add recipes per channel. The five scattered one-liners point at it, and the onboarding skill's prose invariant becomes `./uninstall.sh --scan` — a deterministic check in place of an instruction to a language model.

## Notes

The second commit fixes a matching bug this phase's own new test found, in code from the earlier phases.

The rule is written twice — python for the JSON configs, shell for the Codex TOML — and the copies had drifted in opposite directions: python read `command` and never looked at `url`, shell read `url` and never looked at `command`. Underneath both was one client's reading rule applied to a function serving three. Cursor and Codex configure a remote server as a bare `url` with no `type`; Claude Code treats that as a configuration error and skips the entry. `describe()` encoded the Claude Code reading, so a hosted registration in a Cursor config was invisible to the scan and survived teardown.

An entry now matches when either half fires, the halves are independent, and a declared `type` is a precondition for neither. The caveat keeps its value without deciding the match: it hangs off a `typed-remote` capability on the client table rather than a client id in the matcher, because the source label cannot identify a reader — a project `.mcp.json` is a cross-client file with no client in its label.

`tests/test_matcher_parity.py` drives one case table into a JSON config and an equivalent Codex section and observes a single scan. It found all three divergences.

Two things not verified. The round-trip job has never executed; its risk points are `npx skills add` landing in the expected directory on a runner with no detected agent, and the runner's `HOME` staying quiet between snapshots. Both fail loudly with the offending paths named. The `--dry-run` client matrix runs in pytest rather than as a CI matrix step, because `--dry-run` does not gate release resolution and a five-way matrix would make ten unauthenticated GitHub API calls per push against a shared-IP limit; the stub harness also covers `claude-desktop` under a Darwin `uname`, which a Linux-only matrix cannot.

Based on #583. Part of #536. Closes #541.


skills/onboarding/pipefy-toolkit-setup/SKILL.md
```diff
@@ -73,11 +73,11 @@ Setup is outside the Pipefy MCP tool surface. After auth succeeds, verify with:
 
    - Shell (local / plugin / CLI): `pipefy --version`
    - MCP: call `list_organizations` (needs no id — the natural first read; confirms the credential works and surfaces the org ids other tools need) or another read-only tool the user allows
-   - Confirm exactly one `pipefy` MCP registration for the path you chose (hosted vs plugin)
+   - Run `./uninstall.sh --scan` (repository root; reports only, removes nothing) and confirm it reports exactly one registration, for the path you chose. It matches on what an entry **runs** — the `pipefy-mcp-server` command, a known runner invoking it, or the host `mcp.pipefy.com` — so a server registered under any other name is still found. Exit `0` nothing found, `1` findings remain, `2` a source could not be inspected.
```
adriannoes: **Blocking.** Verify and Success criteria now require `./uninstall.sh --scan` at the repository root. The same skill's When-to-use and path table lead with Hosted MCP and the Claude Code plugin, which do not give the agent a checkout that contains that script.

README and `docs/uninstall.md` already lead with the curl pipe for `--scan`. Agents that follow only Steps + Success criteria hit a dead command on the paths this skill is written for.

I would use the curl-led form (or "curl if no clone; `./uninstall.sh` from a checkout") in Verify, Success criteria, and the multi-registration recovery row.

Done when:

- Skill verify works without a git checkout of ai-toolkit
- Matching-rule / exit-code prose stays

adriannoes: **Optional.** Exit `1` ("findings remain") is the healthy outcome after a successful one-registration setup, and your success criteria already judge stdout. A one-line note that verify should expect exit `1` (and ignore the scan footer's uninstall recipes) would stop exit-code-first agents from "fixing" a good setup. Your call.

adriannoes (changes requested):
Reviewed at `60131b11`: detached worktree, full local ci.yml-equivalent gates (4363 pytest, 148 shell tests, shellcheck, deletion guard), security clean, matcher/parity contracts green. Round-trip Actions job not executed here or on GitHub (feature-branch base).

**Verdict: needs changes** into `rc-539/feat/install-receipt` (stack toward `dev`).

### Required before merge

- Onboarding skill verify must not require a repository checkout of `uninstall.sh`. See the thread on `skills/onboarding/pipefy-toolkit-setup/SKILL.md:76`.

### Decisions I would make

- Accept that `shell` / `roundtrip` CI did not run on this PR base; treat first green on a `main`/`dev` (or retargeted) run as the completeness proof the Notes already flag. I would not block this PR solely for that once the skill gate is fixed.
- Empty-plan `uninstall --yes` exit 0 with leftovers is pre-existing heuristic leave-alone behavior. I would document "re-run `--scan`; do not trust uninstall exit as clean" rather than redesign the exit in this CI phase, unless you want the stronger postcondition now.
- Merge order remains 581 → 582 → 583 → 584.

### Optional

Your call on all of these.

- Exit-`1`-is-healthy + ignore teardown footer on verify (pointer on the same skill line).
- Drop `#246` from `skills/onboarding/pipefy-toolkit-setup/SKILL.md` (product skills should not carry ticket refs; that line is outside the diff hunk so it could not be anchored).
- Local pytest teeth for `check_roundtrip.py` (parity already covers content clearer than the CI path allowlist).

### What worked well

- Either-half matcher fix with `test_matcher_parity.py` is the right dual-implementation contract.
- Observational install↔uninstall parity + deletion-guard teeth are high-value.
- `docs/uninstall.md` matches script behavior on exits, flags, phases, and switching recipes.

<details>

Tier `deep`, trust `trusted`, reviewed `c9a58ec147d5868bd56a6587343aa1d57086e7d9`…`60131b11bb4fa951c43acfe90b36e65cd372ce84`.

| Gate | Result |
| :--- | :--- |
| Local full suite | 4363 passed |
| Shell lint/guard | pass |
| Security H15 | clean |
| Live MCP/CLI | n/a |
| Refutation | checkout-only finding survives; exit-clarity adjusted to optional |

Not covered: live GitHub Actions `roundtrip` job (network/HOME); intentional.

</details>

skills/onboarding/pipefy-toolkit-setup/SKILL.md
mocha06: Fixed in `f9bc5b1`. Correct catch — hosted-only users are precisely the ones with no clone, and they are the population most likely to need the scan.

Verify, the success criteria, and the multi-registration failure row now lead with the curl form, naming the checkout form beside it rather than instead of it. A test asserts the verify block works without a checkout and that no line offers `./uninstall.sh` without the curl form alongside.

mocha06: Two parts here.

`#246` is dropped — agreed that a product skill should not carry a ticket reference, and thanks for flagging it even though it sat outside the diff hunk.

The exit-1 note I am declining, as your optional marking allows: the step already spells out all three exit codes immediately after the command, so "expect 1" would restate the line below it. Happy to add it if you would rather have it explicit.

uninstall.sh
```diff
@@ -2055,28 +2237,90 @@ $_dirs
...
+# What the entry in the skills directory points at, and the lock entry that
+# describes it. A skill directory is often a link into a shared store, so
+# removing only the link leaves the content and the record of it behind, and a
+# run that says the skill is gone would be wrong.
+plan_skill_store() {
+    if [ -L "$1" ]; then
+        _pss_target=$(canonical_path "$(resolve_link "$1")")
+        if [ -d "$_pss_target" ] && [ "$_pss_target" != "$(canonical_path "$1")" ]; then
+            detail "content at $_pss_target"
+            plan_add 7 ours rmpath "$_pss_target" - - - - - - \
```
adriannoes: **Blocking.** `plan_skill_store` resolves an "ours" skill symlink and plans `rmpath` on that target with no store confine. Marketplace gained `marketplace_clone_confines` in this same delta; skills did not.

A lock entry that names this toolkit plus a link under `~/.claude/skills/pipefy-*` pointing outside the skill store is enough for `--yes` to delete the resolved directory. That is the same class of "path from disk is not authority to delete" bug the marketplace fix closed.

I would confine store targets the way marketplace does (only under the expected skills store / recorded path), otherwise report and leave, and add a canary test that a symlink to a directory outside the store is never deleted.

Done when:

- An out-of-store symlink target is reported and left, not planned for `rmpath`
- A regression test fails if that confine is removed

adriannoes (changes requested):
Follow-up at `dead767d` (four commits after `60131b11`). Prior blocking skill verify is fixed (curl-led `--scan` plus a no-checkout regression test). `#246` is gone. Empty-plan teardown now exits `1` with the Left alone notes, matching `--scan`. The optional exit-`1` note stays declined, which is fine.

**Verdict: needs changes** into `rc-539/feat/install-receipt`.

### Required before merge

- Skill-store teardown still follows an "ours" skill symlink and plans `rm -rf` on the resolved target with no confine. Marketplace got that guard in this same delta; skills did not. See the thread on `uninstall.sh:2316`.

### What I re-verified

218 tests green on this tip (teardown, scan, parity, matcher, receipt). Hosted OAuth unverifiable clear, `--client none` refuse, and `remove_path` `$HOME/..` look sound.

Approve once skill-store targets are confined (or reported and left) with a canary test parallel to the marketplace outside-tree case.

mocha06:
Fixed in `1cb4cd8`, and you were right that this one was worse than the marketplace case: following a link makes the reach unbounded.

`plan_skill_store` resolved a skill entry's symlink and planned `rm -rf` on whatever it landed on, with no check, while `marketplace_clone_confines` sat twenty lines above doing exactly that for the clone. I added that guard in the same delta and did not extend it to skills.

The permitted root is derived, not hardcoded: `skills add` keeps its lock file beside the store it writes into, so the store is `dirname(lock)/skills` — which yields `~/.agents/skills` for a global install and `<project>/.agents/skills` for a project one from the same rule. `skill_store_confines` canonicalizes both sides as the marketplace check does, and requires the target **strictly inside** the root, which is stricter than the marketplace check's allowance of equality: the root holds every agent's skills, so no single skill is ever the root. No lock file means no derived store and therefore no permitted target.

Outside that root the link is still ours and still goes, the directory it pointed at is untouched, and both the plan detail and a *Left alone* note name the surviving path, so nothing reads as "the skill is gone" while its content is on disk.

Tests mirror your marketplace outside-tree case rather than inventing a shape: `test_a_skill_linked_outside_the_store_keeps_its_content` is the canary parallel, and `test_a_skill_link_with_no_lock_file_to_derive_a_store_from_keeps_its_content` covers receipt-only provenance where the skill is ours but there is no store to derive. Both fail against `dead767`.

I also checked the lock-entry reach you did not ask about. It needs no guard: `skill_lock_file` returns a path the probe constructs itself as `join(base, ".agents", ".skill-lock.json")` where `base` is `$HOME` or a project directory the scan already sweeps — never a string read out of a file's content, which is precisely what made `installLocation` a bug. The action is also a `jsonkey` edit, which backs the file up first, not a deletion. A confinement check there would restate what construction already guarantees, so I recorded the property as a comment instead. Say the word if you would rather have the guard.

Head is now `38ea762`: the stack merged upward from #581 → #582 → #583, so the README and skill surfaces and the receipt fix are all on this head too. Merge commits, not rebases, so none of the PRs below needed a force-push.

adriannoes (dismissed):
Follow-up at `38ea7628`. The skill-store symlink confine landed in `1cb4cd8d` (`skill_store_confines`, store root from the lock file, out-of-store targets reported and left). Canary `test_a_skill_linked_outside_the_store_keeps_its_content` passes beside the marketplace outside-tree twin.

Earlier items stay closed: curl-led onboarding `--scan`, empty-plan exit aligned with `--scan`, `#246` dropped.

**Verdict: merge-ready** into `rc-539/feat/install-receipt`.

adriannoes (approved):
Re-approve after the base retarget to `dev`. Tip is still `f093e8aa`: no new head commits since the prior approve at `38ea7628`, only the parent-branch merge that was already on this tip. Skill-store confine, curl-led onboarding `--scan`, and the empty-plan exit contract remain in place.

PR #585, opened 2026-08-05
fix(mcp): field-condition honesty (#516 + #523)

### LINKED ISSUE #516: Field conditions: create reports success but rule does not persist (or lands on wrong phase)
## Problem
`create_field_condition` can return success while the show/hide rule does not persist, or is saved on the wrong phase (often the start form instead of the requested phase). Agents then report conditional visibility as done when it is not.

## Evidence
From internal Builder product testing (report 23/07/2026 — 23 builds, 7+ testers):
- Appeared in **5 builds**.
- Testers observed “created successfully” responses where the rule was missing on reload, or always landed on the start form.
- Report detail: `create_field_condition` appears to ignore `phaseId` / perform a “ghost write”.

## Impact
False-green delivery: the Builder trusts the tool response and marks visibility rules complete. Reviewers and customers discover missing conditionals only later — full rework cycles.

## Status on `dev`
**PARTIAL / UNVERIFIED against live** — MCP exposes `create_field_condition` with a `phase_id` argument mapped to API `phaseId`. No post-create read-back / persistence check in the tool shell. Whether the API still drops or remaps `phaseId` needs confirmation when picking this up.

- After create, re-read via `get_field_conditions(phase_id)` and fail (or warn) if the rule is absent or on another phase
- Surface clearer API errors when `phaseId` is ignored upstream
- Document the quirk at the tool docstring / field-conditions skill until the API is fixed
- Route persistent API ignore of `phaseId` to the platform team; keep toolkit work focused on honest success and guidance

### LINKED ISSUE #523: Field conditions: required fields that are hidden still block card moves silently
## Problem
A field marked `required=true` remains required even when a field condition hides it. Automations or moves that should advance the card then fail with little or no visible explanation — the hidden required field blocks progress.

## Evidence
- Appeared in **1 build**, across **3 consecutive cycles**.
- Report detail: hidden required field blocks `move_single_card`; users do not understand why the card will not advance.

## Impact
Silent operational breakage after the Builder “successfully” configured show/hide rules. Looks like automation or MCP failure; root cause is a conflicting required + hide setup.

## Status on `dev`
**MISSING** as a lint/validation surface — field condition and phase field tools do not warn about required+hidden combinations. Product semantics (whether required should auto-relax when hidden) may be platform policy.

- Lint/validate tool or skill checklist listing required fields that also have hide conditions
- Stronger move/automation error messages naming the blocking required field
- Document the platform rule in field-conditions / automations skills
- Cross-link ghost field-condition writes (same report) — bad conditionals worsen this

adriannoes:
## Summary

- After `create_field_condition`, re-read the condition and only report success when it exists on the requested phase (`verified: true`). Ghost / wrong-phase → `success: false`; both verify reads down → success + warning.
- Block `hide` (and legacy `hidden` / case-whitespace variants) on `required=true` fields before the mutation (`INVALID_ARGUMENTS`).
- Enrich `move_card_to_phase` failures that name a required field into a `success: false` envelope (optional hide hint when cross-check finds required+hidden).
- Skill, MCP docs, and `docs/parity.md` updated; CLI honesty checks deferred as known MCP-ahead behavior.

Closes #516
Closes #523

## Decisions

| # | Topic | Choice | Rationale |
| --- | --- | --- | --- |
| 1 | Ghost write / wrong phase after create | Hard fail (`success: false`) with requested `phase_id` + `condition_id` (+ actual phase when known) | Agents key off `success`; a warning would false-green |
| 2 | Both verify reads failed | Success + warning “could not verify” | Do not mask a successful create as an error; outage ≠ ghost write |
| 3 | Lint required + `hide` | Blocking error (`success: false`) listing field ids; mutation does not run | Soft warnings are easy to ignore; toolkit is stricter than the API on purpose |
| 4 | Move path (`move_card_to_phase`) | Enrich when GraphQL error names the required field | Live repro: `Field "…" is required!…`; return tool envelope instead of crash |
| 5 | Packaging | One PR closing #516 and #523 | Same module; slightly larger review OK |

## Move enrich evidence (task 3.1)

Live Cursor MCP / branch MCP: unfilled required field blocks `move_card_to_phase` to an allowed destination. Raw error: `Field "Required note" is required! Please fill it and you'll be ready to go!` — **ENRICH** (identifiable client-side).

## Smoke (branch MCP + live API)

| Scenario | Result |
| --- | --- |
| Create hide on required | `success: false`, mutation not called |
| Create with `actionId: "hidden"` | Same block |
| Create on Inbox (API remaps to Start form) | Honest `success: false` wrong-phase |
| Create on Start form | `success: true`, `verified: true` |
| Move with empty required | `success: false` naming the field |

## Test plan

- [x] `uv run pytest -m "not integration"` (4216 passed)
- [x] Live smoke against branch MCP (table above)
- [x] CI green on this PR

## Review

Local T2 review: Approved with caveats (post-C.7). Deferred: large test-file split.


adriannoes:
## Live smoke (post-review fixes) — OVERALL PASS

Ran against working-tree MCP (`build_pipefy_mcp_server` + keychain session), disposable pipe in test org; cleanup OK. Head: `65ce3b2d`.

| # | Scenario | Result |
|---|----------|--------|
| A | `create_field_condition` hide on required | PASS — `success: false`, `INVALID_ARGUMENTS`, field id listed |
| A2 | alias `actionId: "hidden"` | PASS — blocked same way |
| B | create on Inbox (non-start) + read-back | PASS — remapped to Start form → `success: false` wrong_phase |
| F3 | recovery guidance on verify fail | PASS — message includes `delete_field_condition` / no blind-retry |
| D | create on Start form | PASS — `success: true`, `verified: true` |
| E | `update_field_condition` hide on required | PASS — `success: false` pre-mutation |
| C | `move_card_to_phase` empty required | PASS — envelope names `Field "Required note" is required!` |

Unit: `73 passed` (`field_condition` focused). Skills-lint + disclosure grep clean.

mocha06 (changes requested):
## Review

I checked each claim against the code rather than reading the diff in isolation — the core design holds up (`phase { id }` really is in `GET_FIELD_CONDITION_QUERY`, `phase_fields_from_payload` matches the real `get_phase_fields` shape, and the move-error regex won't misfire on GraphQL schema-validation messages). Three things worth fixing, three minor, and two I'd explicitly leave alone.

### Fix these

**1. `field_condition_tools.py:124` — the "verification unavailable" gate is the wrong condition.** This is the one real bug.

`needs_list` is true in two cases: the get-by-id raised, *or* it succeeded and returned a condition whose `phase` is null/absent (`_field_condition_has_usable_phase_id` false). But the `except` branch only degrades to success+warning when `fetch_raised`. So in the second case — get-by-id *found the condition under that id*, and the list call then fails — `listed_ids` stays `None`, `evaluate_condition_persistence` falls through to an empty `listed` set, and the tool hard-fails with "did not persist on requested phase" plus an instruction to delete it. There is no evidence it didn't persist: the read found it. That's precisely the inconclusive case this PR says should be success + warning.

Gate on the absence of conclusive evidence (`listed_ids is None` after `needs_list`), not on `fetch_raised`. Note that `test_field_condition_planner.py:40` covers `fetched={"phase": None}` only *with* a populated `listed_ids`, so this branch is currently untested.

**2. `field_condition_planner.py:96` — the required+hide lint never fires for a `uuid` `phaseFieldId`.**

`required_tokens` is built from `internal_id` and `id` only, but a uuid is a first-class `phaseFieldId` in this toolkit: `field_condition_phase_field_id_looks_like_slug` explicitly exempts `UUID_RE` (`pipe_config_tool_helpers.py:270`), there's a dedicated accepts-uuid test, `GET_PHASE_FIELDS_QUERY` already selects `uuid`, and `resolve_phase_field_identifiers` in the same module treats internal_id/uuid/slug as the identifier triple. So an agent that hides a required field by uuid sails past the guard this PR exists to add. Adding `"uuid"` to the key tuple closes it with no other change.

**3. `field_condition_tools.py:44` — adding `"actions"` to `_UPDATE_FIELD_CONDITION_EXTRA_RESERVED` silently drops a value that previously reached the API.**

The set was `{"id"}` before, so `extra_input={"actions": [...]}` was forwarded and normalized by the SDK. Now it's filtered out of `update_attrs`; if `actions` was supplied *only* via `extra_input`, the mutation runs without it and the envelope still reports `action: "updated"`. Unlike create — where `actions` is a required positional and therefore always a genuine duplicate — on update it's optional, so `extra_input` can be the only place it appears. Given this PR's own honesty premise, that should be an `INVALID_ARGUMENTS` error naming the top-level `actions` argument, not a silent drop.

### Minor

- **`field_condition_tools.py:548`** — dead re-validation. Because `actions` is now stripped from `extra_input`, `update_attrs["actions"]` can only be the top-level list, which lines 515-520 already checked. Either drop the second `isinstance` + `field_condition_actions_error_message` pair, or keep it as the explicit error from item 3.
- **`field_condition_tools.py:51`** — `_CREATE_VERIFY_FAIL_RECOVERY` says "Delete this condition with `delete_field_condition`", but that tool is `destructiveHint=True` and returns a confirmation preview unless `confirm=true`. An agent following the instruction literally gets a preview right after being told the create failed. One clause makes the remediation executable.
- **`field_condition_tools.py:141,147`** — both verify-failure payloads pass `message` only: no `code`, and the ids live in prose. Every other failure in the module carries a code, and `SKILL.md:233` tells the agent to delete "the returned `condition_id`" — which isn't a key anywhere in the envelope. Put the ids in `error.details` with a stable code so clients can branch on the verify outcome.

### Deliberately not asking for these

- `str(exc)` in `pipe_tools.py:945` bypasses `extract_error_strings`. It's safe as written (`PipefyGraphQLError.__str__` is the joined API messages, so the regex finds its match); it only matters on a multi-error response, where the enriched message would carry unrelated errors. Fine to leave.
- A retry around the read-after-write. It would dodge a rare replica-lag false negative at the cost of latency on every failure, and it works against the point of a check whose value is being honest about what it actually saw.

`_action_id_is_hide` being broader than the SDK's exact-`"hidden"` mapping is harmless — `"HIDE"` would be rejected by the API anyway.

adriannoes:
Thanks @mocha06 for the thorough review — all three required fixes and the three minors are in.

| Commit | Scope |
|--------|--------|
| `0c621cb5` | required+hide lint matches field `uuid` |
| `2d194b9a` | verify gate on `listed_ids is None`; reject `extra_input.actions`; verify-fail codes/details + `confirm=true` recovery |
| `6277c020` | skill / MCP docs / parity |

### Smoke (working-tree MCP + keychain)

| # | Scenario | Result |
|---|----------|--------|
| A | `create_field_condition` hide on required | PASS — `INVALID_ARGUMENTS` |
| A2 | alias `actionId: "hidden"` | PASS |
| A3 | hide by field `uuid` | PASS |
| B | create on Inbox → wrong phase | PASS — `FIELD_CONDITION_WRONG_PHASE` + `error.details` |
| F3 | recovery guidance | PASS — `delete_field_condition` + `confirm=true` + stable code/details |
| D | create on Start form | PASS — `verified: true` |
| C | `move_card_to_phase` empty required | PASS |
| E | `update_field_condition` hide on required | PASS |
| E2 | `actions` via `extra_input` | PASS — `INVALID_ARGUMENTS` |

Unit: `75 passed` (`field_condition` focused). Cleanup OK.

mocha06 (approved):

Issue #522, opened 2026-07-29
Writes: tool/GraphQL can report failure while side effects already applied (retry duplicates)


adriannoes:
## Problem
Some write paths return failure (sometimes with an empty error message) even though the mutation already applied in Pipefy. Agents that retry on “error” duplicate cards or other side effects.

## Evidence
From internal Builder product testing (report 23/07/2026):
- Appeared in **1 build**, with high severity.
- `execute_graphql` returned `success: false` with an empty message while `cards_count` increased.
- **18 cards** were created despite the reported error — real risk of duplicating customer data on retry.

## Impact
Worst failure mode next to silent automations: the agent believes nothing happened and retries. Customer data duplication; trust loss in the toolkit.

## Status on `dev`
**UNVERIFIED** which specific mutations still exhibit empty-error-after-success. No general `idempotency_key` on create-card / batch paths. Related to batch scale pressure (#518 family) where retries are more common.

## Possible directions (non-binding)
- Propagate the real upstream error text; never empty failure after partial success
- Accept optional `idempotency_key` (or equivalent) on create paths if API supports it
- Skill rule: on ambiguous failure, re-read (`get_cards` / counts) before retrying creates
- Post-write verification playbook (see the Guidance issue from the same report)

Related: batch operations issue from the same report; Guidance post-write verification.

Solution and acceptance criteria are decided when this issue is picked up.

adriannoes:
Pairs with post-write verification in the Guidance issue from the same product-testing report, and with batch scale pressure (#518).

adriannoes:
## Follow-up (deferred shells)

Remaining raw `str(exc)` / empty-message shells outside the evidence paths of this issue are tracked in **#587** (same milestone). They were deferred on purpose so this fix stays focused; #587 is the place we will finish the sweep after this ships.

adriannoes:
Shipped on `dev` via [#588](https://github.com/pipefy/ai-toolkit/pull/588) (`70d3489a`, tip includes review follow-up `cbc56484`). Closing manually because `Closes` on a `dev` PR does not auto-close until `main`.

Remaining raw `str(exc)` shells stay open in **#587** (deferred sweep).

PR #588, opened 2026-08-06
fix(mcp): write failure without blank message or blind retry (#522)

adriannoes:
## Summary

- Shared `ensure_non_empty_error_message` so wired evidence-path write/error shells never return a blank failure to agents (remaining raw `str(exc)` shells → #587).
- Wired on #522 evidence paths: `execute_graphql` / introspection excepts, portal `PortalPermissionError`, `create_card` mutation except.
- Skills: re-read counts/ids before retrying ambiguous writes; hard stop when re-read is inconclusive; no fake client-side idempotency (`CreateCardInput` has no `idempotency_key`).

Closes #522

**Deferred** remaining raw `str(exc)` shells → **#587** (same milestone; intentional split for reviewability).

## Audit (write-error paths)

| Path | Verdict |
|------|---------|
| `execute_graphql` except | **fixed** |
| introspect_* / `search_schema` except | **fixed** |
| portal `PortalPermissionError` | **fixed** |
| portal generic fallback | **already OK** |
| `create_card` mutation except | **fixed** (does not use `handle_tool_graphql_error`) |
| `handle_tool_graphql_error` + most card writes | **already OK** |
| automation empty-message (#288) | **already OK** — untouched |
| Blind retry after ambiguous write | **skill-only** (+ envelope re-read steer on write fallbacks) |
| Other raw `str(exc)` (iPaaS, KB/LLM, …) | **deferred → #587** |

Grep used: `rg -n 'str\(exc\)' packages/mcp/src/pipefy_mcp/tools/` and `rg -n 'build_error_payload\(|tool_error\(' packages/mcp/src/pipefy_mcp/tools/`.

## Review follow-ups (post-#588 review)

Addressed on tip after PR-driven review (merge-with-notes), before merge:

| Item | Change |
|------|--------|
| Blank `PortalPermissionError` fallback | Uses `_PORTAL_PERMISSION_GUIDANCE` (`create_portal` / `manage_portals`), not “Try again” |
| Envelope re-read steer | `create_card` / `execute_graphql` empty fallbacks + tool docstrings point at re-read / no blind-retry |
| Soft GraphQL blank/whitespace | Empty/`"   "` `result.error` and whitespace-only `errors[].message` → `success: false` with stable fallback |
| DRY introspection excepts | Shared `_exception_error_payload` / `_soft_result_error_payload` |
| Session tests | Whitespace `str(exc)` + preserve-non-blank matrix for introspect tools / `create_card` / `execute_graphql` |

[...]

PR #589, opened 2026-08-07
refactor(mcp): drop 785 no-op parametrize decorators and the duplicate transition_hints module

Danielmoraisg:
## Summary

- 785 tests across 18 tool test modules carried `@pytest.mark.parametrize("<name>_session", [None], indirect=True)`, whose only effect was to set `request.param` to `None`. The session fixtures read it as `getattr(request, "param", None)`, so the decorator produced the same `elicitation_callback=None` as omitting it. Removed. The 14 decorators that pass a real accept/decline callback are untouched.
- Two of those modules (`test_pipe_tools.py`, `test_field_condition_tools.py`) read bare `request.param` with no fallback, so their 100 decorators were load-bearing. Their fixtures now default the param like the other sixteen, which is why that lands as its own commit first.
- `packages/mcp/src/pipefy_mcp/tools/transition_hints.py` was byte-identical to the SDK copy and had zero importers. Deleted.

Three atomic commits, reviewable in order. The first two have a hard ordering dependency: until the fixtures default the param, removing those 100 decorators breaks the tests. Each commit passes on its own. The third is independent and can be split out if you would rather keep test cleanup separate from the source deletion.

### Rebased onto dev

Rebased on `dev` at 70d3489. The count moved from 777 to 785 because three commits landed on `dev` in the meantime and two of them added tests carrying the same no-op decorator: 2 in `test_pipe_tools.py` and 6 in `test_introspection_tools.py`, all from the write-failure work in #522. Those 8 are removed here on the same grounds as the original 777, so the invariant a reviewer would grep for still holds: `[None], indirect=True` returns zero hits under `packages/mcp/tests/`.

The other two test files `dev` touched (`test_graphql_error_helpers.py`, `test_portal_tool_helpers.py`) use no indirect parametrization at all, so nothing was missed there.

### On the duplicate module

`packages/sdk/src/pipefy_sdk/transition_hints.py` is the live copy, with three importers, one of which is MCP's own `tools/phase_transition_helpers.py:9`. It already reaches across to `pipefy_sdk.transition_hints` rather than to the file that sat beside it. The SDK version was added later by the AI preflight work that needed the same constant; the MCP file has not been touched since the package moved under `packages/mcp`.

Verified structurally rather than by grep alone: no module lists `pipefy_mcp.tools.transition_hints` as a dependency in the import graph, and it is not re-exported from any `__init__.py`. Re-checked after the rebase, since `dev` added roughly 200 lines across four MCP source files.

### Behavior

None. Test IDs lose their `[None]` suffix, which is the only externally visible effect. Nothing in CI selects on them.

## Test plan

All re-measured after the rebase.

- [x] `uv run pytest -m "not integration"`: 4345 passed, 5 skipped, 34 deselected
- [x] `uv run pytest packages/mcp/tests -m "not integration"`: 1872 passed, 5 skipped, 4 deselected, identical to the `dev` baseline measured by checking out `dev` and running the same command
- [x] Collected test count unchanged at 1876 for `packages/mcp/tests`, same as `dev`
- [x] Each commit passes `uv run pytest packages/mcp/tests -m "not integration"` on its own: 1872 passed at commit 1, 1872 at commit 2, 1872 at commit 3
- [x] `uv run ruff format --check .`: 415 files already formatted
- [x] `uv run lint-imports` (from `packages/mcp`): 2 contracts kept, 0 broken, which independently confirms the deleted module had no importers
- [x] `[None], indirect=True` returns zero hits under `packages/mcp/tests/`, and the 14 decorators passing a real callback are still present

`uv run ruff check .` reports two `I001`s, in `tests/test_parity.py` and `tests/test_uninstall_scan.py`. Both are pre-existing on `dev` (the second arrived with the uninstall-scan work) and neither file is touched here, so they are left for a separate change.

## Docs / skills

No MCP or CLI surface changed, so `docs/parity.md` is unaffected. No tool names, arguments, or behavior changed, so no skill references anything altered here.

## Legal / contributions

- [x] Regulated-domain skills include `COMPLIANCE.md` when applicable (not applicable)

## Note on size

This is 19 files and 824 changed lines, over the 10 files / 300 lines guidance in `AGENTS.md`. Nearly all of it is one mechanical deletion repeated 785 times, and the commit split is what keeps it reviewable: commit 2 is 785 deletions and zero added lines, so a reviewer can confirm it by reading the diff for anything that is not a removed decorator. Happy to split commit 3 into its own PR if you prefer a stricter reading of the cap.


adriannoes (dismissed):
Reviewed at `cbaea031`: deep checkout review with local MCP gates (ruff, lint-imports, 1846 MCP unit tests), targeted elicitation runs (14 load-bearing callbacks + broader elicitation suite), and structural confirmation that the deleted MCP `transition_hints` twin was byte-identical and unused.

### Decisions I would make

- Prefer merging this PR before #588 (or rebasing #588 afterward) so new tests there do not reintroduce the no-op `[None]` session parametrize this cleanup removes.
- Rebase onto latest `dev` before merge if you want a linear tip; the three-dot diff reviewed here does not include #581 and stays correct either way.

### What worked well

- Atomic commit order (fixture defaults, then 777 decorator deletions, then dead-module delete) keeps each step independently reviewable.
- Real accept / decline / raises elicitation coverage was left intact; only the no-op `[None]` noise went away.
- Production blast radius is a single unused duplicate; SDK remains the sole owner.

<details>

Tier `deep`, trust `trusted`, reviewed at `d52922a3...cbaea031`.

| Gate | Result |
| :--- | :--- |
| GitHub CI (lint, test, DCO) | green |
| Local ruff / format / lint-imports | pass |
| pytest MCP not integration | 1846 passed |
| 14 elicitation callback tests | pass |
| Byte-identical + zero importers for deleted MCP transition_hints | confirmed |

Live smoke: n/a (no agent-visible runtime surface). Not covered: full root suite beyond MCP (CI already ran test job green); security specialist (no auth surface).

</details>

adriannoes (approved):
Follow-up at `a3168a87` after the rebase onto `dev`. Prior approve on `cbaea031` was dismissed by the force-push.

Re-checked: 785 no-op `[None]` session parametrize removals (+8 from #522 on `dev`), 14 real elicitation callbacks kept, zero residual no-ops, unused MCP `transition_hints` twin still gone. Local MCP suite 1872 passed; CI green.

One leftover note: watch #588 so new tests there do not bring the no-op `[None]` pattern back.

Issue #92, opened 2026-04-23
Tech debt: PipefyClientError hierarchy and narrow tool-layer except Exception


adriannoes:
## Problem
- `packages/mcp/src` carries 198 `except Exception` handlers, 193 of them silenced with `# noqa: BLE001`, and every one is inside `tools/`. The count this issue was filed with was "100+", so the number moved the wrong way.
- A blanket handler renders a permission denial, a malformed argument and a programming bug as the same tool envelope. The caller cannot tell a retryable API failure from a defect in this repo.
- The typed errors the issue asked for already exist in the SDK, and the tool layer does not catch a single one. Step 1 is done in a different place than the body proposed; steps 2 and 3 did not happen.

## Evidence
Verified read-only against `main` @ 3442f0c8 on 2026-09-16. All counts are grep over the tree.

- Broad handlers by package: `mcp` 198 (193 with `# noqa: BLE001`), `sdk` 22 (14 with the noqa), `auth` 1, `cli` 0, `infra` 0. In `packages/mcp/src` there are zero outside `tools/`.
- The same directory already has typed handlers, so the pattern is mixed rather than absent: 32 `except ValueError`, 15 `except ValidationError`, 7 `except MalformedFieldDefinitionError`, 3 `except UserCancelledError`, 3 `except AttachmentUploadError`, 2 `except ToolError`, plus single uses of `NoBackChannelError`, `KnowledgeBaseDocumentUploadError` and `AutomationPreflightError`.
- `packages/mcp/src/pipefy_mcp/tools/attachment_tools.py` is the reference: 349 lines, zero `except Exception`, and its three API failure paths catch `AttachmentUploadError` at `:198`, `:288` and `:347`.
- The hierarchy exists in the SDK. `packages/sdk/src/pipefy_sdk/exceptions.py` defines `PipefyError` and `PipefyAPIError`. `packages/sdk/src/pipefy_sdk/graphql_executor.py:46` adds `PipefyGraphQLError(PipefyAPIError)` carrying the raw per-error dicts on `.errors`, and `:208` maps `TransportQueryError` at the executor boundary so callers see one SDK type instead of a gql exception.
- `packages/sdk/src/pipefy_sdk/graphql_problem.py` is the closed-set classifier: `GraphQLProblemKind` with permission denied, not found, invalid arguments, feature not enabled and runtime, reached through `classify_exception` and `classify_graphql_error_dicts`.
[...]

PR #590, opened 2026-08-07
fix(sdk): root PipefyGraphQLError in the error hierarchy (groundwork for #92)

Danielmoraisg:
## Summary

- `PipefyGraphQLError` subclassed `Exception` directly, so the hierarchy the SDK exports described something it never raised. `PipefyError` and `PipefyAPIError` are both in `pipefy_sdk.__all__`, documented as the base and the API-payload error, and neither is raised anywhere in the package. Every GraphQL failure arrives as a `PipefyGraphQLError` that `except PipefyError` did not catch, so a consumer following the documented root caught nothing it would ever see.
- Parented to `PipefyAPIError`, whose docstring ("Raised when the Pipefy GraphQL API returns an error payload") already describes exactly this type. That roots it under `PipefyError` too, and gives a previously dead exported class a referent rather than leaving it declared and unused.
- The two CLI runners in `_common.py` are reordered to catch the specific type before the root, in a separate preparatory commit.

Groundwork for #92. That issue has been deferred since 2026-04-23 as "an architectural change, not a one-file fix", and this is the part that was actually blocking it: with the hierarchy connected, the ~190 broad `except Exception` handlers in the tool layer can be migrated to typed catches one domain at a time. **No handler is migrated here.**

## What this does not change

Nothing at runtime, for the CLI or the MCP server. The tool layer already catches `Exception` everywhere, so a widened `PipefyError` reaches nothing new there. The value is that the hierarchy stops lying, not that anything behaves differently today. Worth saying plainly so the diff is not read as a fix for a live failure.

## Commit order is deliberate

The reparent alone breaks `test_portal_sub_portal_detach_graphql_error_exits_1`: once `PipefyGraphQLError` is a `PipefyError`, the generic branch catches it first, the GraphQL formatter stops running, and stderr collapses from `Couldn't find PortalInterface with (RECORD_NOT_FOUND)` to the bare message. I confirmed that failure before writing the fix.

So the reorder lands first, where it is a provable no-op (the two types are disjoint at that commit, and the full CLI suite passes either way), and the reparent lands second. Neither commit is red. A fourth commit carries the tests and docs corrections from review.

## `PortalPermissionError` is deliberately untouched

It stays a `ValueError`, outside the hierarchy. The `ValueError` parentage is the mechanism holding the exit 2. Rooting it under `PipefyError` would silently move it to exit 1, which is a product decision about CLI exit-code semantics, not a refactor.

No SDK-side pin was added for it either. `test_portal_sub_portal_detach_permission_error_still_exits_2` already asserts that coupling end to end, in the layer that owns the exit-code mapping: reparenting under `PipefyError` sends it to the `except PipefyError` branch above `except ValueError` in `run_cli_command`, and that test goes red. An `issubclass` assertion in the SDK suite would restate the same thing one layer away from where it is decided. `40e53eb` introduced the type for MCP isinstance mapping and touched no CLI file, so whether exit 2 was intended is a question this PR does not answer either way.

## Breaking-ish, for library consumers

`except PipefyError` now catches GraphQL failures. That is the intent, but code that catches the root for a narrow purpose and expected GraphQL failures to pass through will newly intercept them. In the CHANGELOG, and `docs/sdk/README.md` gains an "Errors" section with the hierarchy and the catch-specific-before-root rule.

I checked the blast radius inside the repo: there are exactly two `except PipefyError` sites (both reordered here) and zero `except PipefyAPIError` sites anywhere.

## Test plan

- [x] `uv run pytest -m "not integration"`: 4560 passed, 5 skipped, 34 deselected. Measured both ends: 4555 on this branch with the tests added here removed, so +5
- [x] `uv run ruff check packages/sdk packages/cli && uv run ruff format --check packages/sdk packages/cli`: clean, 218 files already formatted
- [x] Each commit passes independently on sdk+cli: 1713 at the reorder, 1718 at the reparent, 1713 at the removal, 1718 at the tests commit
- [x] Reverting the reparent at `ce616a8` left all 1713 sdk+cli tests green. That is what showed the tip had nothing pinning this change: the portal CLI twins catch `PipefyGraphQLError` by name, so they pass whichever base it has
- [x] `packages/sdk/tests/test_exception_hierarchy.py` pins the parentage. With the reparent reverted at the new tip its 3 guards fail and `test_api_error_is_rooted`, a pre-existing invariant, passes. Verified by reverting the source change and re-running, so they are guards rather than tautologies
- [x] `test_run_pipefy_client_coroutine_graphql_error_keeps_the_code_suffix` pins the sibling runner's catch order: swapping its two catches drops the `(CODE)` suffix and the test goes red on the message alone
- [x] Reorder-only (reparent reverted) passes all CLI tests, which is what makes commit 1 a safe no-op

`uv run ruff check .` reports one pre-existing `I001` in `tests/test_parity.py`, untouched here and already present on `dev`.

## Docs / skills

- [x] `docs/sdk/README.md` gains an "Errors" section. `PipefyError` is described as the root of the API error types, with the carve-outs named: `PortalPermissionError` on `ValueError`, `AttachmentUploadError` and `KnowledgeBaseDocumentUploadError` on `Exception`
- [x] `docs/parity.md` not applicable: no MCP or CLI surface changed
- [x] No skill references any of these types

## Legal / contributions


mocha06:
tip: update with dev branch to fix failing job

packages/cli/src/pipefy_cli/commands/_common.py
```diff
@@ -216,6 +216,9 @@ async def _run() -> _T:
 
     try:
         return asyncio.run(_run())
+    except PipefyGraphQLError as exc:
```
adriannoes, line 219: **Optional.** The tip also dropped the short WHY comments that said GraphQL must be caught before `PipefyError` or the `(CODE)` suffix is lost. `run_cli_command` is still pinned by the portal detach test; `run_pipefy_client_coroutine` is not. Your call whether to restore a one-line comment here (and at the sibling runner) or add a coroutine-runner CODE format test.

docs/sdk/README.md
```diff
@@ -10,6 +10,40 @@ This tree summarizes how to work with **`pipefy`**: the vendor GraphQL client, s
 
 For a short in-repo overview and dev commands, see **[`../../packages/sdk/README.md`](../../packages/sdk/README.md)**.
 
+## Errors
+
+`PipefyError` is the root. Catch it to handle any failure the SDK raises:
```
adriannoes: **Optional.** This line says catch `PipefyError` for any SDK failure, and the bullet below says it is the root of every error the SDK raises. The same section (and other exported types such as upload errors) sit outside that tree. Softening to the typed GraphQL/API subtree would match the carve-outs. Your call.

adriannoes (changes requested):
Reviewed at `ce616a8a`: detached worktree gates (ruff + 1713 sdk/cli tests), targeted portal GraphQL/permission CLI pins, parentage smoke, and a mutation litmus on the deleted hierarchy suite. Live MCP/CLI smoke was not required (local contracts cover the changed surfaces).

### Required before merge

- Tip commit `ce616a8a` removes `packages/sdk/tests/test_exception_hierarchy.py`, the five parentage pins that `9fb4b219` added for this change. The remaining portal CLI twins assert exit codes and `message (CODE)` formatting. They catch `PipefyGraphQLError` by name, so they stay green if that class is put back under bare `Exception` while this PR's library contract (`except PipefyError` catches GraphQL failures) quietly breaks. Please restore that file (or an equivalent `issubclass` / raise-and-catch suite covering GraphQL→API→root and `PortalPermissionError` outside `PipefyError`), and update the PR Test plan so it no longer claims the +5 hierarchy tests while the tip has none.

### Decisions I would make

- I would not treat merge as closing #92. This remains groundwork only (tool-layer BLE001 sites untouched).
- I would keep `PortalPermissionError` as `ValueError` for exit 2, as the PR already documents.

### Optional

Your call on all of these.

- Restore short catch-order WHY comments (or a coroutine-runner CODE pin). See the pointer on `_common.py`.
- Soften Errors "any/every" wording. See the pointer on `docs/sdk/README.md`.

### What worked well

- Deliberate CLI reorder before reparent so stderr keeps `message (CODE)`.
- Honest groundwork framing for #92; docs/CHANGELOG disclose the consumer catch widen.
- Portal detach twins still pin exit 1 + CODE and exit 2 for permission.

<details>

Tier `standard`, trust `trusted`, reviewed at `667fa984...ce616a8a`.

| Gate | Result |
| :--- | :--- |
| CI on tip | green (6/6) |
| Local sdk+cli pytest | 1713 passed |
| Parentage litmus | recovered 5 pass; GraphQL→Exception: hierarchy fails, CLI twins pass |
| H16 live | not required |

Not covered: live Pipefy org smoke (not required for this matrix).

</details>

Danielmoraisg:
Reviewed and acted on at `cf8bd50`.

**The required item, confirmed.** I reverted the reparent at `ce616a8` and ran the sdk+cli suite: 1713 passed, fully green. Nothing at the tip held the contract, exactly as you said. The portal CLI twins catch `PipefyGraphQLError` by name, so they pass whichever base it has.

`packages/sdk/tests/test_exception_hierarchy.py` is restored. With the reparent reverted at the new tip:

```
FAILED test_graphql_error_is_an_api_error
FAILED test_graphql_error_is_catchable_as_the_root
FAILED test_raised_graphql_error_is_caught_by_the_root
3 failed, 1 passed
```

3 guards and 1 pre-existing invariant (`test_api_error_is_rooted`), not the 3+2 the old body claimed.

**Declining one named sub-item: the `PortalPermissionError` pin.** Not because the coupling doesn't matter, but because `test_portal_sub_portal_detach_permission_error_still_exits_2` already asserts it end to end, in the layer that owns the mapping. Reparenting under `PipefyError` sends it to the `except PipefyError` branch above `except ValueError` in `run_cli_command`, and that test goes red. The SDK-side version asserted the same coupling one layer away from where it is decided, and its docstring called the parentage deliberate. `40e53eb` introduced the type "for MCP isinstance mapping" and touched no CLI file, so I don't think this PR can claim intent behind exit 2 either way. If you want it pinned by name regardless, two `issubclass` lines appended to that CLI test would satisfy it without moving the pin out of the mapping layer. Say the word.

Separately, and not blocking: you wrote `40e53eb`. Was exit 2 for portal permission denials intended, or did it fall out of the `ValueError` base while the other GraphQL failures moved from 2 to 1? If it is incidental it deserves an issue rather than an unexamined invariant, and I would rather ask than keep guessing at it in a docstring.

**Both optional items taken**, one as a test rather than a comment. `run_pipefy_client_coroutine` now has `test_run_pipefy_client_coroutine_graphql_error_keeps_the_code_suffix`; swapping its two catches turns it red on the lost `(CODE)` suffix. Details in the two threads.

**Test plan rewritten** with re-measured numbers. The old ones were stale from before `dev` moved: 4560 passed / 5 skipped / 34 deselected on the full suite, against 4555 with these tests removed, and per-commit 1713 / 1718 / 1713 / 1718 on sdk+cli rather than the 1710 / 1715 the body claimed.

packages/cli/src/pipefy_cli/commands/_common.py
Danielmoraisg: Went with the test rather than the comment, on the reasoning that a comment cannot fail.

`test_run_pipefy_client_coroutine_graphql_error_keeps_the_code_suffix` in `test_common_helpers.py` asserts exit 1 and `Couldn't find PortalInterface with (RECORD_NOT_FOUND)` on stderr. Swapping the two catches in this runner makes it fail on the message alone, with the suffix gone, which is the exact regression the comment described.

Both runners are pinned now: `run_cli_command` by the portal detach test, this one by the new test.

docs/sdk/README.md
Danielmoraisg: Narrowed. The line now reads "root of the API error types" and the bullet "root of the API error types below", with the carve-outs named rather than left implicit: `PortalPermissionError` on `ValueError`, `AttachmentUploadError` and `KnowledgeBaseDocumentUploadError` on `Exception`.

Two more in the same section that your point applies to:

- The code example still had `# anything else the SDK owns` under `except PipefyError`, which was the same overclaim three lines down from the one you flagged.
- I dropped "deliberately" from the `PortalPermissionError` paragraph. `40e53eb` introduced the type for MCP isinstance mapping and touched no CLI file, so the docs now state the mechanism (the `ValueError` parentage is what holds exit 2) without asserting intent behind it.

adriannoes (approved):
Thanks for the careful follow-up, Daniel. The way you addressed the review was thoughtful and stronger than the minimum ask.

Since `ce616a8a`, `cf8bd50e` restores `test_exception_hierarchy.py` so the GraphQL→API→root contract is pinned again, adds `test_run_pipefy_client_coroutine_graphql_error_keeps_the_code_suffix` so both CLI runners guard the `(CODE)` suffix (a test that can fail beats a comment), and narrows the Errors docs to the API subtree with the carve-outs named. Leaving the Portal `issubclass` pin out in favor of the existing exit-2 CLI twin is a clear, defensible call, and the updated Test plan matches the tip.

Re-checked on this tip: CI green, parentage suite present, coroutine CODE pin in place, docs wording fixed. Nothing blocking remains.

**Verdict: merge-ready** into `dev`. Approved. Appreciate the solid groundwork for #92.

PR #591, opened 2026-08-07
docs(mcp): warn connector field replace-all (#525)

### LINKED ISSUE #525: Card connections: updating a connector field replaces the entire list (no add/remove)
## Problem
Updating a connector/connection field replaces the entire related list. Agents cannot append or remove a single item; they must resend the full list every time.

## Evidence
- Appeared in **1 build**.
- Report detail: connector update is replace-all; no incremental append operation.

## Impact
Fragile under concurrency — if another automation changes the field between read and write, connections are lost. Forces error-prone full-list rewrites for simple “add one link” tasks.

## Status on `dev`
**PARTIAL** — card relation create/delete tools exist for pipe relations; connector **field** updates via `update_card_field` still behave as replace-all for multi-value connection fields. Confirm exact field-type semantics when picking up.

- Incremental `add_card_connection` / `remove_card_connection` (or field-value append/remove) if API supports
- Document replace-all semantics loudly on `update_card_field` for connector types
- Skill: always read current connections before write; never assume append
- Cross-link with card relation tools so agents pick the right surface

adriannoes:
## Summary
- Document **replace-all** semantics for connector/connection multi-value fields on `update_card_field` (read-merge-write with `get_card(include_fields=true)`, concurrency caveat).
- Cross-link skills: pipes-and-cards failure mode + relations Key concepts distinguishing connector **fields** vs pipe-relation card links.
- Point agents at existing incremental path: `update_card` + `field_updates` with `operation` ADD/REMOVE (`updateFieldsValues`).

Closes #525

## Research note — incremental API

**YES** (existing surface; no new tools in this PR).

- `updateCardField` / MCP `update_card_field`: replace-all only (no `operation`).
- `updateFieldsValues` / MCP `update_card` + `field_updates[].operation`: `ADD` | `REMOVE` | `REPLACE`.
- Live introspection: `UpdateFieldValuesOperators` and `NodeFieldValueInput.operation` explicitly list **Connections** as compatible with ADD/REMOVE.
- Manual Cursor MCP smoke (ephemeral fixtures): replace-all, ADD, and REMOVE confirmed; `include_fields=true` required for merge reads.
- **Follow-up:** no dedicated add/remove tools here; reopen only if agents still miss the existing `update_card` ADD/REMOVE path after this guidance.

## Test plan
- [x] `uv run pytest packages/mcp/tests -k "update_card_field" -q`
- [x] `uv run pytest -m "not integration"`
- [x] Live MCP smoke: replace-all / ADD / REMOVE / include_fields
- [ ] CI green on this PR


packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -975,6 +975,16 @@ async def update_card_field(
             Use this tool for simple, single-field updates. The entire field value
             will be replaced with the new value provided.
 
+            **Connector / connection (multi-value):** never assume append. Prefer
```
mocha06: The warning is scoped to connector/connection fields, but the same replace-all semantics apply to the whole multi-value family: the live `UpdateFieldValuesOperators` enum names "Attachments, Assignees, Labels, Connections and Checklists" as the ADD/REMOVE-compatible set, and `CardField.array_value` covers the same list.

This matters because the repo already ships a path that does exactly the destructive thing being warned about here. `create_attachment_presigned_url` step 3 (`packages/mcp/src/pipefy_mcp/tools/attachment_tools.py:322`) tells agents to "set the attachment field to `[storage_path]` via `update_card_field`", and `packages/sdk/src/pipefy_sdk/services/attachment_service.py:346` calls `update_card_field(card_id, field_id, [storage_path])`. A card with attachment A loses it when an agent uploads B through either path.

Either widen the wording to "connector / attachment / checklist / label / assignee (any list-valued field)", or say the connector scoping is deliberate.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
```diff
@@ -233,6 +233,7 @@ Do not hide a required field — MCP rejects `hide`/`hidden` on `required=true`
 - **`create_field_condition` fails with missing/wrong phase:** use `error.details.condition_id` (or the id in the message) with `delete_field_condition` and `confirm=true` before recreating; do not blind-retry create on the same requested phase.
 - **`create_card` fails with missing required fields:** call `get_start_form_fields` first to discover required `field_id` values.
 - **`create_card` / write reports failure (empty or unclear message):** do not blind-retry. Re-read `get_cards` / `get_phase_cards_count` (or pipe `cards_count`) before any retry — see [Ambiguous write failure](../../api-troubleshoot/pipefy-api-fallback/SKILL.md#ambiguous-write-failure-re-read-before-retry).
+- **Connections missing after a connector field update:** `update_card_field` is replace-all — writing one related card id drops the rest. Prefer `update_card` / CLI `card update --field-updates` with `operation` ADD/REMOVE and related **card ids**. Do not rebuild a full list from `get_card` `value` (display titles only, not ids). For links via a pipe relation (not a connector field), use `create_card_relation` / `delete_card_relation` — see `skills/relations/`.
```
mocha06: This closes the only read path it names (`get_card` `value` = titles, "do not rebuild from it") without naming a working one — and there is one.

Verified live: `get_card_relations` returns connector-field links with ids. For a sample card, `child_relations[0]` came back as `{name: "Use case", cards: [{id: "1204385469", …}]}` for the connector field `golden_set_use_cases`. `CardField.array_value` (`["1204385469"]`) is the other source, reachable via `execute_graphql`.

An agent asked to *remove* one related card needs that id for `operation: "REMOVE"`. With no documented source it either falls back to the titles it was just told not to use, or does the full-list rewrite this bullet warns against. One clause pointing at `get_card_relations` closes the loop.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
```diff
@@ -156,8 +156,8 @@ This returns valid `type` enum values and their descriptions.
...
 | `fill_card_phase_fields` | `pipefy card fill <id> --phase <id>` | No | Fill phase fields non-interactively; filters to editable IDs. Uses `--fields` JSON object; for ad-hoc updates use `card update --field-updates` (JSON array). |
-| `update_card` | `pipefy card update <id>` | No | Update title, assignee, due date, fields. |
-| `update_card_field` | `pipefy card update <id> --field-updates` | No | Single-field update (`updateCardField`); for several fields prefer `update_card` + `field_updates`. |
+| `update_card` | `pipefy card update <id>` (`--field-updates` = `updateFieldsValues`) | No | Update title, assignee, due date, fields. For connector/multi-value, prefer `--field-updates` / `field_updates` with `operation` ADD/REMOVE and related **card ids**. |
+| `update_card_field` | *(MCP only; no direct CLI twin — use `card update --field-updates`)* | No | Single-field `updateCardField`; connector/multi-value is **replace-all** of related **card ids** (not display titles). Do not rebuild the list from `get_card` `value` (titles only). Prefer `update_card` + ADD/REMOVE instead. |
```
mocha06: This cell is accurate — no `packages/cli/src` caller exists — but `docs/parity.md:189` still reads `| update_card_field | pipefy card update | shipped | Use --field-updates JSON array (). |`.

`parity.md` is the CI-checked canonical MCP↔CLI mapping (`tests/test_parity.py`), so the repo now answers the same question two opposite ways depending on which file the reader opens. Worth updating that row in the same change; the trailing empty `()` in its note is stale too.

skills/relations/pipefy-relations/SKILL.md
```diff
@@ -18,6 +18,7 @@ Link processes and cards across workflows. **8 MCP tools.**
 - **Pipe relations** define parent/child structure between pipes (who connects to whom, constraints, auto-fill). Use `get_pipe_relations` on a pipe to list relation IDs and metadata.
 - **Card relations** connect individual cards through an existing pipe relation: pass `source_id` = the pipe relation's ID (from `get_pipe_relations`). Default `sourceType` is `PipeRelation`; use `extra_input` (e.g., `sourceType: Field`) when the API requires a field-based link — see `introspect_type` on `CreateCardRelationInput`.
 - **Table relations** in GraphQL are loaded by table-relation ID, not by database table ID: `get_table_relations` takes a non-empty list of those IDs.
+- **Connector fields** are card fields (connection type), not pipe relations. Prefer `update_card` + `field_updates` with `operation` ADD/REMOVE (related **card ids**). `update_card_field` is **replace-all** of the full id list; `get_card` `value` is display titles only — do not rebuild from it. This skill's `create_card_relation` / `delete_card_relation` link cards through a pipe relation and are incremental — pick the surface that matches the link type.
```
mocha06: "Connector fields are card fields (connection type), **not pipe relations**" is right for writes but over-drawn as a blanket statement.

The write split checks out — I confirmed `pipe.childrenRelations` is empty for a pipe whose cards carry a connector field, so there is no pipe-relation id to feed `create_card_relation`. But `card.child_relations` — i.e. this skill's own `get_card_relations` — does surface connector-field links, with ids.

A reader takes "not pipe relations" as "the relations tools won't help here" and skips the one tool that would have given them those ids. Suggest: "…not pipe relations for *writes*; `get_card_relations` still reads them."

mocha06 (commented):
The premise holds up. I checked the load-bearing claims against the live schema and a real card rather than the PR body:

- Connector `CardField.value` is a JSON array of display titles (live sample: `"[\"Data extraction for .pdf file\"]"`, while the connected card id is `1204385469`), and `GET_CARD_QUERY` (`packages/sdk/src/pipefy_sdk/queries/card_queries.py:60`) only selects `name`/`value` — so `get_card(include_fields=true)` genuinely cannot return ids. The "do not rebuild the list from `value`" warning is correct.
- Live `UpdateFieldValuesOperators` = `ADD | REMOVE | REPLACE`, and it explicitly lists Connections as compatible. `UpdateCardFieldInput.new_value` has no operation arg, so replace-all is right.
- The `operation` key does pass through CLI → SDK (`convert_values_to_camel_case`, `formatters.py:168`), and `update_card`'s docstring already documents it — the redirect target is real.
- `update_card_field` really has no CLI caller under `packages/cli/src`, so the corrected table cell is accurate.
- Gates pass locally: `test_parity.py` (189 passed), CLI help-golden (16 passed), `lint_skill_refs.py` clean.

Four comments below, none blocking.

One more, out-of-diff and pre-existing, so noting it here rather than inline: this PR makes `update_card` the preferred connector path, but `update_card` (`pipe_tools.py:1067-1075`) has no `try/except handle_tool_graphql_error` and no `debug` parameter, unlike `update_card_field` (lines 1009-1015, with `resource_kind="phase_field"` enrichment). An `operation: "ADD"` against a wrong field slug now surfaces as a raw exception instead of the structured `tool_error` envelope with the field-not-found hint. Not introduced here, but this PR is what routes traffic there — worth a follow-up issue.

mocha06 (dismissed):

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
adriannoes: Done in 8c6aa9e0: widened the warning to list-valued fields (connections, attachments, checklists, labels, assignees) and kept ADD/REMOVE as the preferred path.

The destructive attachment upload path (`create_attachment_presigned_url` / SDK `update_card_field([storage_path])`) is out of scope here — tracked in #595.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
adriannoes: Done in 8c6aa9e0: failure mode now points at `get_card_relations` (and GraphQL `array_value`) for current related-card ids before REMOVE or a full rewrite.

skills/relations/pipefy-relations/SKILL.md
adriannoes: Done in 8c6aa9e0: wording is now "not pipe relations for *writes*" and explicitly notes that `get_card_relations` still reads connector-field links with ids.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
adriannoes: Done in 8c6aa9e0: `docs/parity.md` now says MCP-only `updateCardField` with the `card update --field-updates` workaround, and drops the empty `()`.

adriannoes:
Addressed the four inline notes in 8c6aa9e0.

Follow-ups opened for the out-of-diff items:
- #595 attachment upload path still does replace-all via `update_card_field`
- #596 `update_card` structured `tool_error` / `debug` parity with `update_card_field`

@mocha06 ready for another look when you have a minute.

docs/parity.md
```diff
@@ -186,7 +186,7 @@ For **database records**, `find_records` result nodes may use **`fields`** while
 | `update_ai_knowledge_base_plain_text` | `pipefy kb plain-text update` | shipped | Knowledge bases; partial update (`--id`, `--pipe-uuid`, optional `--name`/`--content`/`--description`, at least one). CLI gates on the read-access probe; limits fail fast client-side. |
 | `update_automation` | `pipefy automation update` | shipped | (`--condition` and/or `--extra` JSON; at least one required). First-class typed `condition` replaces the rule's condition. |
 | `update_card` | `pipefy card update` | shipped | (`--field-updates` JSON, optional title/labels/assignees/due-date). |
-| `update_card_field` | `pipefy card update` | shipped | Use `--field-updates` JSON array (). |
+| `update_card_field` | `pipefy card update` | shipped | MCP-only `updateCardField` (no CLI twin). Workaround: `card update --field-updates` (`updateFieldsValues`; optional `operation` ADD/REMOVE/REPLACE). |
```
mocha06: The note is right now, and I confirmed the `operation` passthrough it describes is real (`formatters.py:135` uppercases it, defaults to `REPLACE`) and that no `packages/cli/src` caller exists. But the row kept `| pipefy card update | shipped |`, so it now says "no CLI twin" in a row whose CLI column names a command and whose status means, per the legend at lines 19-24, "CLI command exists in `packages/cli` today."

The cross-file disagreement became an intra-row one. The legend already has the exact entry for this case — **N/A**: "MCP- or IDE-oriented surface with no first-class CLI twin planned" — and all four existing no-twin rows (`call_ipaas_tool`, `create_ipaas_connection`, `get_ipaas_connection_auth_url`, `get_ipaas_tools`) use `| — | deferred |`.

Suggest matching that convention:

```
| `update_card_field` | — | N/A | MCP-only `updateCardField`. Workaround: `card update --field-updates` (`updateFieldsValues`; optional `operation` ADD/REMOVE/REPLACE). |
```

I tried that locally and `test_parity.py` still passes 189 — the test skips CLI resolution for `deferred`/`n/a` statuses, so nothing breaks.

mocha06 (commented):
Rechecked at `8c6aa9e`. Three of the four are cleanly closed; one is half-done (inline below).

**Closed:**

- **Widened scope** — `pipe_tools.py` now reads "List-valued fields (connections, attachments, checklists, labels, assignees)". I re-introspected the schema to check the widened list rather than take it on faith: the `ADD`/`REMOVE` enum descriptions say "compatible with field types that support lists like Attachments, Assignees, Labels, Connections and Checklists" — the list matches verbatim, and it is mirrored consistently into both SKILL.md files and `docs/mcp/tools/pipes-and-cards.md`.
- **Read path named** — `get_card_relations` (and GraphQL `array_value`) is now cited in all four places, and the pointer resolves: `GET_CARD_RELATIONS_QUERY` selects `cards { id title }` (`packages/sdk/src/pipefy_sdk/queries/card_queries.py:123-126`).
- **Relations wording** — "not pipe relations for *writes*" plus "this skill's `get_card_relations` still **reads** connector-field links (with card ids)" is exactly the distinction that was missing.

**Gates at this head:** `test_parity.py` 189 passed, CLI help/golden 16 passed, `skills-lint` 17 skills validated.

**Still open, both fine as follow-ups rather than changes here:**

1. `update_card` (`pipe_tools.py:1067-1075`) still has no `handle_tool_graphql_error` wrapper or `debug` parameter, so the path this PR now prefers gives worse errors than the one it steers away from.
2. The attachment upload path is unchanged — `attachment_tools.py:322` tells agents to "set the attachment field to `[storage_path]` via `update_card_field`", and `attachment_service.py:346` does exactly that. The widened docstring now tells agents never to assume append for attachments, while the repo's own attachment upload assumes precisely that. Docs and code disagree there; worth an issue.

docs/parity.md
adriannoes: Done in the tip: row is now `| — | N/A |` with the workaround kept in Notes, matching the legend and the other no-twin rows. `test_parity.py` still 189 green.

adriannoes:
Follow-up on the parity row: tip is now `de3a34a5` (`update_card_field` → `| — | N/A |`). Ready for another look when you have a minute.

docs/parity.md
mocha06: Resolved by `de3a34a`. Row now reads `| update_card_field | — | N/A | MCP-only updateCardField. Workaround: card update --field-updates … |`.

`N/A` is the right pick over `deferred`: the four existing no-twin rows are all iPaaS and all say a twin is "considered once agent usage settles", whereas here `updateFieldsValues` supersedes the tool and no twin is planned — which is exactly what the legend reserves `N/A` for. `test_parity.py` passes 189 on the new head, and `skills/pipes-and-cards/.../SKILL.md:160` already said "MCP only; no direct CLI twin", so the two files agree now.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
mocha06: Addressed in `8c6aa9e`. The docstring now reads "List-valued fields (connections, attachments, checklists, labels, assignees)", and I checked that list against the schema rather than the diff: `UpdateFieldValuesOperators` `ADD`/`REMOVE` describe themselves as "compatible with field types that support lists like Attachments, Assignees, Labels, Connections and Checklists" — verbatim match. Mirrored consistently into both SKILL.md files and `docs/mcp/tools/pipes-and-cards.md`.

The attachment path itself is unchanged (`attachment_tools.py:322`, `attachment_service.py:346` still do the single-element replace-all the widened warning now cautions against). That is behavior, not docs, so it belongs in a follow-up issue rather than this PR.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
mocha06: Addressed in `8c6aa9e` — `get_card_relations` (and GraphQL `array_value`) is now cited here and in the three other places that repeat the warning. The pointer resolves: `GET_CARD_RELATIONS_QUERY` selects `cards { id title }` (`packages/sdk/src/pipefy_sdk/queries/card_queries.py:123-126`), so an agent doing `operation: "REMOVE"` has a documented source for the ids.

skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md
mocha06: Addressed across `8c6aa9e` and `de3a34a`. This cell was already the accurate side; `docs/parity.md:189` now reads `| — | N/A |`, so the two files answer the question the same way. `test_parity.py` passes 189 on the current head.

skills/relations/pipefy-relations/SKILL.md
mocha06: Addressed in `8c6aa9e` — now "not pipe relations for *writes*", plus "this skill's `get_card_relations` still **reads** connector-field links (with card ids)". That is exactly the distinction that was missing: the write split stands (no pipe-relation id exists for a connector field), while the read path stays discoverable.

mocha06 (approved):

Issue #518, opened 2026-07-29
Scale: no batch update_card_field / create_cards — one call per field or card


adriannoes:
## Problem
- Bulk card creation needs one `create_card` call per card, and a multi-card field correction needs one `update_card` call per card. No tool accepts a list of cards and returns a per-item success or error.
- Reported from product testing: one correction pass over 517 field values had to be hand-assembled into 13 command blocks, another issued about 70 individual calls, and most of the observed errors came from assembling those call lists rather than from latency.
- `fill_card_phase_fields` sends every editable field of a phase in one `updateFieldsValues` call with no chunking, so a phase with more than 30 editable fields fails as a whole, mid-task.

## Evidence
- Multiple fields on one card already works and is not part of this gap. `update_card(card_id, field_updates=[...])` takes a list and routes to `updateFieldsValues` with a per-item `operation` of `ADD`, `REMOVE` or `REPLACE` (`packages/mcp/src/pipefy_mcp/tools/pipe_tools.py:1039`; dispatch at `packages/sdk/src/pipefy_sdk/services/card_service.py:217`). Live validation: 30 field updates in one call succeed, 31 fail with `You exceeded the maximum number of fields` and `field: null`, and the message does not state the maximum.
- The remaining gap is one call per card. Verified by introspection on 2026-09-15: `UpdateFieldsValuesInput.nodeId` is a single `ID!` ("The card or table record ID that will be updated"), and `UpdateFieldsValuesInput.values` is documented as "Limit of 30 fields per call".
- No chunking exists on the path. `fill_card_phase_fields` (`packages/mcp/src/pipefy_mcp/tools/pipe_tools.py:1179`) filters the caller's keys to the phase's editable fields, builds one `field_updates` list and issues one `client.update_card` call, with no caveat in its docstring.
- The API does have multi-item writes and this repo uses none of them. `batchUpsert`, `cardsImporter`, `recordsImporter` and `batchExportData` appear nowhere in `packages/`, `skills/` or `docs/`.
- `batchUpsert` is "Submit an asynchronous batch upsert of records or cards (Service Account only)" and returns `BatchUpsertPayload { batchOperation }`. `BatchUpsertInput`:

| Input field | What it provides |
| --- | --- |
| `repoId` (`ID!`) | the pipe or table whose cards or records are upserted |
[...]

Issue #288, opened 2026-06-03
AI automation error helper can return empty message after stripping diagnostic markers


gbrlcustodio:
## Summary

When `create_ai_automation` / `update_ai_automation` started routing failures through `handle_automation_tool_graphql_error`, the old stable fallback for failures whose text is only `[code=…]` / `[correlation_id=…]` markers was lost.

The handler falls back to `"Automation request failed."` when there are no extracted strings, but after `strip_internal_api_diagnostic_markers` the joined message can still end up empty.

## Suggested fix

Add a guard in `automation_tool_helpers.py`:

```python
if not base.strip():
    base = "Automation request failed."
```

This matches what `enrich_behavior_error` already does for the same shape.

## Verification

A unit test on the helper using `ValueError(" [code=X] [correlation_id=Y]")` confirms the behavior.

## Context

Deferred from PR review: https://github.com/pipefy/ai-toolkit/pull/287#discussion_r3351348469

PR #592, opened 2026-08-07
fix(mcp): apply ensure_non_empty_error_message to remaining high-priority error shells (#587)

### LINKED ISSUE #587: MCP: apply ensure_non_empty_error_message to remaining raw str(exc) error shells
## Problem

Some MCP tool error paths still return `tool_error(str(exc))` / `build_error_payload(str(exc))` (or equivalent) without a non-empty fallback. If upstream yields an empty or whitespace-only message, the agent sees a blank failure and may blind-retry writes.

## Relationship to #522

**Deferred from [#522](https://github.com/pipefy/ai-toolkit/issues/522)** (PR05 / M18). That issue ships:

- Shared helper `ensure_non_empty_error_message`
- Wiring on the evidence paths: `execute_graphql` / introspection excepts, portal `PortalPermissionError`, `create_card` mutation except
- Skill rule: re-read counts/ids before retrying ambiguous writes; no fake client-side idempotency

The remaining shells were audited in that PR and **intentionally deferred** so #522 stays reviewable and focused on the reported failure mode. **We will resolve them in this issue after #522 merges.**

## Scope (suggested)

Apply the existing helper (prefer shared wrappers over copy-paste) to remaining raw `str(exc)` error construction, with regression tests (empty/whitespace → stable fallback; real text preserved).

Priority slices (from the #522 audit):

1. **Shared wrappers** — iPaaS shared except, knowledge-base / LLM `classify_exception is None` fallthrough, AI agent `error_payload_from_exception`-style paths
2. **Call sites** — other `build_ai_tool_error(str(exc))` / portal-org read excepts / pre-mutation `ValueError` → `tool_error(str(exc))` where still unguarded
3. **`message=str(exc)` builders** — member / report / webhook / pipe-config style payloads (lower urgency if messages are usually non-empty)

Out of scope: fake idempotency keys; automation empty-message (#288 already fixed); batch APIs (#518).

## Acceptance criteria

- [ ] Remaining high-priority raw `str(exc)` write/error shells use `ensure_non_empty_error_message` (or an equivalent non-empty guard)
- [ ] Unit tests for empty/whitespace → fallback and non-blank preservation on touched paths
- [ ] No reopen of #288 / no drive-by automation changes unless required
- [ ] Disclosure-clean diff

## Notes

Helper home: `packages/mcp/src/pipefy_mcp/tools/graphql_error_helpers.py` (`ensure_non_empty_error_message`).

adriannoes:
## Summary

- Apply shared `ensure_non_empty_error_message` to remaining high-priority MCP error shells deferred from #522 / #588: iPaaS `_run_ipaas_tool`, KB/LLM classify fallthrough + ValueError paths, and `build_ai_tool_error` (covers AI agent/automation ValidationError envelopes).
- Guard portal/org read excepts the same way.
- Follow-up on the same branch: portal reads go through `map_portal_error_to_message`; org/portal read fallbacks stay plain (no anti-retry); write portals keep anti-retry copy.
- Review follow-ups (@mocha06): `extract_error_strings` skips whitespace; SDK `graphql_problem` normalizes blank/whitespace messages; drop duplicate KB/LLM helpers; revert unreachable `MalformedFieldDefinitionError` `ensure_*` wrappers and reject blank messages at the exception constructor.

Closes #587

**Out of scope (intentional):** Priority-3 `message=str(exc)` builders (member/report/webhook/pipe-config/observability/attachment); automation GraphQL empty-message (#288); fake idempotency; batch APIs (#518); central `tool_error` non-blank backstop → **#593**.

## Relationship to #522

#522 shipped the helper and evidence-path wiring. This PR is the deferred sweep for the remaining high-priority funnels so #522 stayed reviewable. Envelope-level invariant follow-up is tracked in #593.

## Test plan

- [x] `uv run pytest -m "not integration"` — **4381** passed, 5 skipped, 34 deselected (post review follow-ups A–C)
- [x] Focused empty-message / preserve matrix and touched MCP tool suites
- [x] `uv run ruff check` + `ruff format --check` on packages + touched tests
- [x] `uv run python scripts/bump_version.py verify`
- [x] `cd packages/mcp && uv run lint-imports` — 2 contracts kept
- [x] Disclosure sweep on diff — clean
- [x] No drive-by automation (#288) changes

### Cursor MCP smoke (server on prior tip)

Against a disposable test org / pipe (ids omitted):

| Scenario | Result |
| --- | --- |
| `list_organizations` / `get_organization` | success |
| `list_portals` / `get_portal` | success |
| `get_llm_providers` / `get_ai_agents` / `get_ai_knowledge_bases` | success |
| `execute_graphql` `{ __typename }` | success |
| `get_ipaas_tools` + sample `call_ipaas_tool` | success (org with iPaaS enabled) |
| `create_card` → `fill_card_phase_fields` → `delete_card` | success (card cleaned up) |
| Invalid org/portal/KB/LLM/iPaaS ids | `success: false` with **non-empty** messages |
| KB blank `name` ValueError | `name must be a non-empty string` |
| `create_ai_agent` empty `behaviors` | non-empty validation envelope via `build_ai_tool_error` |
| iPaaS on org without the feature | non-empty disabled/unavailable message |

### Local proof (empty/whitespace cannot be forced via live API)

Injected `RuntimeError("")` / `"   "` through the same wrappers MCP loads: blank → domain fallback; `"real boom"` preserved. Same for LLM ValueError shells (missing config file / `set_default` neither id) without live writes.

## Docs / skills

No MCP tool names, arguments, or success behavior changed — only blank-failure envelopes, shared extractors, and read/write fallback copy alignment.

## Legal / contributions

- [x] Regulated-domain skills include `COMPLIANCE.md` when applicable (not applicable)


packages/mcp/src/pipefy_mcp/tools/portal_tool_helpers.py
```diff
@@ -71,7 +80,7 @@ def map_portal_error_to_message(exc: BaseException) -> str:
         if messages:
             return "; ".join(messages)
 
-    return ensure_non_empty_error_message(text, _PORTAL_OPERATION_FAILED)
+    return ensure_non_empty_error_message(text, empty_fallback)
```
mocha06, line 83: This line is unreachable for a `PipefyGraphQLError` — the most common portal failure — because the branch just above returns `"; ".join(messages)` first, and `extract_error_strings` collects whitespace-only messages (it filters on `and msg`, not `and msg.strip()`).

Reproduced on the branch tip:

```python
map_portal_error_to_message(
    PipefyGraphQLError([{"message": "   "}]),
    empty_fallback="Portal request failed.",
)
# -> '   '
```

So `empty_fallback` never fires for the case it was added for, and a blank message still reaches the agent. The new `test_map_portal_error_whitespace_permission_denied_uses_guidance` covers the whitespace input *with* `extensions.code = PERMISSION_DENIED`, which short-circuits at the permission branch above, so the hole stays untested.

Fixing `extract_error_strings` (see the review summary) closes this without adding anything to this function.

packages/mcp/src/pipefy_mcp/tools/knowledge_base_tools.py
```diff
@@ -20,13 +20,24 @@
...
 
+_KB_REQUEST_FAILED = (
+    "Knowledge base request failed. Re-read knowledge base state "
+    "before retrying; do not blind-retry."
+)
+
+
+def _classified_problem_message(message: str) -> str:
+    """Normalize classifier text so whitespace matches the empty-message path."""
+    return message.strip() or "Unknown error"
```
mocha06: `message.strip() or "Unknown error"` never returns a blank string, so the `ensure_non_empty_error_message(..., _KB_REQUEST_FAILED)` that wraps this at lines 77 and 105 can never fire. The code expresses two intents and only one of them runs — and it's the less useful one: a whitespace-only classified error now surfaces `"Unknown error"` (the new test asserts exactly that) instead of `_KB_REQUEST_FAILED` with its retry guidance.

Worth picking one:

- return `""` for blank input here, so `ensure_non_empty_error_message` supplies the domain fallback and the guard actually means something; or
- keep `"Unknown error"` and drop the guard, since it's dead either way.

Separately, this helper is duplicated verbatim at `llm_provider_tools.py:45`, and it duplicates normalization the SDK classifier already owns at `packages/sdk/src/pipefy_sdk/graphql_problem.py:127`. Normalizing there removes both copies and also covers the probe paths that still surface `problem["message"]` verbatim (`knowledge_base_service.py:714`, `llm_provider_service.py:341`), which this PR doesn't touch.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
```diff
@@ -191,7 +196,11 @@ async def create_card(
                         phase_id, required_fields_only
                     )
                 except MalformedFieldDefinitionError as exc:
-                    return tool_error(str(exc))
+                    return tool_error(
+                        ensure_non_empty_error_message(
+                            str(exc), _MALFORMED_FIELD_DEFINITION_FAILED
+                        )
+                    )
```
mocha06: `_MALFORMED_FIELD_DEFINITION_FAILED` is unreachable. `MalformedFieldDefinitionError` has exactly one raise site in the repo — `packages/sdk/src/pipefy_sdk/models/field_definition.py:71` — and it is always constructed from a hardcoded f-string:

```python
raise MalformedFieldDefinitionError(
    f"Cannot {action}: {malformed} field "
    "definition(s) from Pipefy are missing required 'id' or 'type'. "
    "The pipe configuration may be incomplete or unsupported."
)
```

`str(exc)` can't be blank, so the fallback can't be reached at any of the seven guarded sites (198, 211, 239, 266, 290, 1218, 1252), and the accompanying test constructs `MalformedFieldDefinitionError("")` — a shape production never produces.

Suggest reverting these hunks. If the concern is a future raise site passing a blank message, asserting non-blankness in `field_definition.py` is one check instead of seven wrappers, and it holds for every caller rather than the ones remembered here.

mocha06 (commented):
Reviewed the four error-shell surfaces plus the helpers they call. Three notes are inline; the fourth has no anchor because the file it concerns isn't in this diff, so it's here.

### `extract_error_strings` is the root cause, and it also leaves the main funnel exposed

`packages/mcp/src/pipefy_mcp/tools/graphql_error_helpers.py:95` filters on truthiness, not blankness:

```python
if isinstance(msg, str) and msg:
    structured.append(msg)
```

So `"   "` is collected and joined, and every consumer that returns the join before reaching a fallback keeps emitting whitespace. Two consequences beyond the portal case in the inline comment:

- `handle_tool_graphql_error` (`graphql_error_helpers.py:481`) — `msgs` is non-empty but whitespace, so `base` is whitespace and `fallback_msg` is never used. This is the funnel most tools go through, and it isn't covered by this PR.
- `build_ai_tool_error` via `ai_agent_tools.py:98-100` — two whitespace-only GraphQL errors join to `"  ;  "`, which passes the `.strip()` truthiness check downstream, so the agent receives `{"error": {"message": "  ;  "}}`.

Changing that condition to `and msg.strip()` closes the whitespace class of #587 for roughly eight consumers at once, including the two above, and makes the portal guard in this PR redundant rather than bypassed.

Worth considering more broadly: every builder touched here funnels through `tool_error` (`core/tool_error_envelope.py:74`), as do the sites deferred to a follow-up. A non-blank backstop there would hold the invariant by construction instead of per call site, with domain constants kept only where they add actionable guidance.

adriannoes:
## Review follow-up plan (@mocha06)

Addressing on this branch (uncommitted until re-test):

| Item | Action |
| --- | --- |
| `extract_error_strings` whitespace | **A** — filter with `strip()` so shared consumers (portal map, `handle_tool_graphql_error`, …) skip blank messages |
| KB/LLM `_classified_problem_message` | **B** — normalize blank/whitespace in SDK `graphql_problem`; drop duplicate MCP helpers |
| `MalformedFieldDefinitionError` guards | **C** — revert unreachable wrappers; reject blank messages at the exception constructor |

Deferred (new issue, linked to this initiative):

- **D** — non-blank backstop in `tool_error` envelope → **#593**

Will push after local re-test; no commit yet on this comment.

packages/mcp/src/pipefy_mcp/tools/portal_tool_helpers.py
adriannoes: Done. Root fix is in `extract_error_strings` (`msg.strip()` / item strip). Whitespace-only GraphQL messages no longer join, so `empty_fallback` runs for `PipefyGraphQLError([{"message": "   "}])` → `Portal request failed.`. Added `test_map_portal_error_whitespace_graphql_message_uses_empty_fallback` (no `PERMISSION_DENIED` short-circuit) plus `handle_tool_graphql_error` whitespace coverage.

packages/mcp/src/pipefy_mcp/tools/knowledge_base_tools.py
adriannoes: Picked the SDK path: `graphql_problem` now does `strip() or "Unknown error"`. Removed the duplicate `_classified_problem_message` helpers from KB and LLM tools. Classified blanks stay `Unknown error` (shared with CLI/probes); `ensure_non_empty_error_message` remains on unclassified `str(exc)` paths where the domain fallback still applies.

packages/mcp/src/pipefy_mcp/tools/pipe_tools.py
adriannoes: Reverted the seven MCP `ensure_*` wrappers and the artificial blank-message test. `MalformedFieldDefinitionError` now rejects blank/whitespace messages in its constructor so a future raise site cannot invent an empty string.

adriannoes:
## Review summary follow-up (@mocha06)

- **`extract_error_strings`:** fixed with `strip()` — closes whitespace for portal map, `handle_tool_graphql_error`, and other join consumers.
- **KB/LLM classified path:** normalized in SDK `graphql_problem`; MCP duplicates removed.
- **`MalformedFieldDefinitionError`:** wrappers reverted; blank rejected at the constructor.
- **`tool_error` envelope backstop:** deferred → **#593** (linked from this PR / #587).

Commits on tip: `95dca003`, `d312ebf5`, `72786f88`. Local CI + Cursor MCP smoke green before push.

mocha06:
Confirmed green here, and both CI commits are right — I had a competing attempt in #594 keyed on the session bus and it does not work: `ubuntu-latest` has a bus address and no Secret Service client, so the runner takes the bus-present branch and still exits 2. Your read of the environment was correct and mine was not. I have stripped that change out; #594 is now only the unrelated stderr fix.

**One finding on `4d40153`, worth a look before merge.**

Pruning `work` removes the only artifact tree this job actually creates from the completeness comparison. Skills do not install to `$HOME/.claude` — in both the failing run and your passing one they land in:

```
/home/runner/work/_temp/.claude/skills/pipefy-*
```

`RUNNER_TEMP` is under `$HOME` on hosted runners, so `work` being pruned means the skills tree is no longer compared. Teardown still removes them — 17 `rm -rf` lines in your passing log — but if it ever stopped, the filesystem check would not notice. That is the part of the script I would least like to lose coverage on: the provenance logic is the newest code in it, and a teardown that deleted a skill it did not own is what prompted it.

The root cause looks like the step above it rather than the prune. `Give the skills installer a target` runs `mkdir -p "$HOME/.claude"`, but `npx skills add … -y` picks its scope from the working directory, and `Install` runs with `working-directory: ${{ runner.temp }}`. So the target that step creates is not the one used.

If `Install` and `Uninstall` ran from a directory outside `$HOME`, skills would land where the comparison still covers them and pruning `work` would be harmless rather than load-bearing. I have not tested that on a runner, so treat it as a suggestion rather than a patch — you have more context on the image than I do.

`actions-runner` pruning looks unambiguously right either way.

mocha06 (approved):

mocha06:
Correcting myself on the previous comment — I tested the suggestion and it is wrong.

Running `npx skills add pipefy/ai-toolkit -y` from a **non-project** directory with `$HOME/.claude` already created still installs into `$CWD/.claude/skills`, not `$HOME`. So "run install from a cwd outside `$HOME`" would push the skills further out of the compared tree, not into it. Scratch that.

The diagnosis was also narrower than I made it sound. Teardown is not relying on the prune being absent: `scan_skills` takes its directories from the plugin client state dir **and from the skills directories the receipt recorded**, and `install.sh` records the one it actually observed. That is why your log shows the 17 removals under `/home/runner/work/_temp/.claude/skills` even though `$HOME/.claude/skills` was the nominal target. The script handles it correctly.

What remains is only the filesystem comparison: with `work` pruned, the tree those skills live in is no longer diffed, so a future regression in skills teardown would not fail the job. Teardown still does the right thing today, and your log proves it.

Since I own that job, I will carry the coverage fix in #594 rather than adding scope to yours — the owner asked to keep it there. Nothing blocking here; both your commits are right and I would merge as is.

PR #594, opened 2026-08-10
fix(uninstall): drain the streaming consumers, and read both skills layouts

mocha06:
## Motivation

Three defects, each found by the one before it.

A scan wrote `uninstall.sh: 258: printf: printf: I/O error` to stderr on every Linux run, in a job that passed. Chasing that exposed a round-trip job that could not have caught it: the skills the job installs were landing in a directory the filesystem comparison prunes, so the one artifact set produced by running the real installer sat outside the check that exists to prove teardown is complete. Restoring that coverage then failed on its first run, with 17 `SKILL.md` files and a lock file left behind by a teardown that reported success.

## Outcome

**Streaming consumers read to the end.** `client_rows`, `all_client_rows` and `records` stream through a loop; five consumers left early — four `awk … exit` and one `break` — closing the pipe mid-write, which dash reports as EPIPE. The head job now logs zero occurrences, down from two.

A behavioural test cannot hold this: whether the producer is still writing when the consumer leaves depends on the pipe buffer, so it reproduces on Linux and not on macOS. That is how the `break` survived a pass that fixed the other four. `check_pipe_drain.py` is structural instead, in the shape of the deletion guard, and rejects `break` / awk `exit` / `head` / `grep -q` in any pipeline whose producer is a function in the script. An `exit` inside an awk `END` block is allowed, since `END` runs only after input is consumed. It fails against the previous commit naming all four sites, and runs in pre-commit and CI.

**Both skills layouts are read.** `npx skills add` uses two:

| Install | Lock | Store |
|---|---|---|
| Global | `~/.agents/.skill-lock.json` | `~/.agents/skills` |
| Project | `<base>/skills-lock.json` | `<base>/.agents/skills` |

The probe knew only the first, so a project install had no lock, hence no derived store, so the confinement guard refused its target and left the content — the guard working as written on an incomplete premise. The store is now derived from the base for both, and both lock names are read.

Lookups also keyed on the skill name alone, so with both layouts present the first record won for every skill of that name and a project skill could bind to the global store. Store *and* provenance now key on `(lock, name)`, resolved from the directory being swept.

**The lock file is retired, not deleted.** It belongs to `skills add` and can list other sources — the one on the machine this was verified against holds 33 entries from three other repositories. Only this toolkit's entries are dropped, and the file goes only if nothing remains, the same contract `rmdir` already uses.

**The round-trip job runs both scripts from `$HOME/roundtrip`**, which the snapshot covers, and asserts outright that the install produced skills and that none survive. The `work` prune stays; nothing this job cares about is written there now.

## Notes

The keychain change this branch opened with is gone. It was aimed at the exit `2` on `dev`, on the premise that a machine without `secret-tool` has no session bus and therefore no Secret Service. `ubuntu-latest` disproves it — bus present, no client — and the job still exited `2` with the change applied. #592 fixed that correctly in CI.

The project *lock* path is corroborated by CI only. Two project checkouts on the verification machine have the store and links but no lock of either name, having been installed by an older CLI. The project *store* path is corroborated three ways.


uninstall.sh
```diff
@@ -2307,19 +2332,19 @@ scan_skills_dir() {
...
-# each get their own, and neither path is one this script may hardcode.
-skill_store_root() {
-    [ -n "${1:-}" ] || return 1
-    printf '%s\n' "$(dirname "$1")/skills"
+# The store `skills add` wrote a skill's content into, as the probe derived it
+# from the base its lock file was found at. The lock's own name differs between
+# the global and project layouts and its directory differs with it, so the
+# store is derived from the base, which does not.
+skill_store_dir() {
+    records skilllock | awk -F"$TAB" -v n="$1" '$3 == n { print $6; exit }'
```
adriannoes: **Blocking.** `skill_store_dir` (and `skill_lock_file` / `skill_provenance`) key only on skill name and `exit` on the first skilllock row. The probe now emits both global and project locks, in `~`-first order, so a project teardown can bind HOME's store and lock when the same name exists in both layouts.

On that path the project link fails confinement against the HOME store, only the link is removed, `jsonkey`/`rmlock` touch the HOME lock, and the project `SKILL.md` plus `skills-lock.json` stay while the run exits 0. Re-scan only walks skills directories, so the orphan store is invisible. Independent reproduction confirmed EXIT 0 with leftovers.

I would carry `(lock, store)` from the skillsdir / lock under teardown (not first name match), and add a regression for global+project same name (ours/ours and other/ours).

Done when:

- dual same-name project store and project lock are cleared (or honestly left with a non-zero exit / finding)
- a unit test fails if first-match-by-name returns

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 ### Fixed
 
+- **`uninstall.sh` (skills, project layout)**: `skills add` has two layouts and only one was read. A global install keeps its lock at `~/.agents/.skill-lock.json`; an install run from inside a project keeps it at `<project>/skills-lock.json`, and the probe looked for the first name only. No lock meant no derived store, so the confinement guard correctly refused a target it had no store for and the content stayed — a real round trip left seventeen `SKILL.md` files and the lock file behind while reporting success. Both names are read now, per base. The store is derived from the base rather than from the lock's own directory, since it is `<base>/.agents/skills` in both layouts while the lock's place is what differs. The skills directory beside a project lock is swept too, so a project install is visible without a receipt as well as with one. The lock file is `skills add`'s own and can record skills from several sources, so this run drops its own entries and removes the file only when those were the only ones in it — the contract `rmdir` already uses for a directory. The confinement guard is unchanged: a link out of the store still costs the link and nothing more. (#541)
```
adriannoes: **Optional.** This Unreleased bullet correctly says the store is derived from the base. The older skill-store Fixed bullet still says the permitted root comes from the lock file beside its store. Readers of Unreleased get both formulas. Your call whether to amend the older bullet here.

tests/test_uninstall_scan.py
```diff
@@ -813,6 +813,22 @@ def test_a_missing_secret_tool_is_uninspected_rather_than_clean(tmp_path):
...
+    dash reported the resulting EPIPE as `printf: I/O error`. Whether the
+    producer is still writing when the consumer goes depends on the pipe
+    buffer, so this reproduces on Linux and not on macOS: it guards the
+    regression where CI runs, and passes either way where it does not.
+    """
+    home = _home(tmp_path)
+
+    result = _run(home, _stub_path(tmp_path))
+
+    assert "I/O error" not in result.stderr, result.stderr
```
adriannoes, line 828: **Optional.** This guard is honest about Linux-only reproduction, and on Darwin it stays green if the three awk `!seen` fixes are reverted. It also does not cover `client_row_with_cap`, which is the residual EPIPE path on CI. Your call whether to harden it alongside the `client_row_with_cap` drain.

adriannoes (changes requested):
Reviewed at `9c1968ee`: detached worktree gates (ruff + 113 uninstall tests), Phase 4 hunts, refute on the two merge gates, and head CI round-trip log.

### Required before merge

- The titled stderr fix is incomplete: `client_row_with_cap` still early-closes `client_rows`, and head Linux Uninstall still prints `printf: I/O error`. The titled fix drains three `awk … exit` consumers with `!seen`, but `client_row_with_cap` still `break`s after the first capability match while `client_rows` is writing. Head job `install.sh then uninstall.sh` still logs `uninstall.sh: 258: printf: printf: I/O error` on Uninstall, so the CHANGELOG line that a scan no longer prints that error is not true on Linux CI. I would drain the pipe after the first match (same read-to-EOF / `!seen` shape as the awk sites), and keep a regression that fails if that early close returns. (No RIGHT-side hunk on the unchanged `break`, so this stays in the body.)
- Skilllock lookups keyed only by skill name mis-bind when global and project layouts both list the same name, leaving project `SKILL.md` / lock with exit 0. See the thread on `uninstall.sh` `skill_store_dir`.

### Decisions I would make

- Retitle / expand the PR Outcome to match CHANGELOG (stderr + dual layouts/`rmlock` + `$HOME/roundtrip`). Reviewers who stop at the title miss most of the tip. I would not block merge on the title alone once the two Required items land.

### Optional

Your call on all of these.

- Align `docs/uninstall.md` Never-removed “lock file names” with store-from-base. The new skills paragraph correctly teaches store-from-base; the Never-removed bullet still points at the old `dirname(lock)/skills` model, which for the project layout is `<project>/skills`, not `<project>/.agents/skills`.
- One CHANGELOG Unreleased formula for store derivation (see the pointer on the new Fixed bullet).
- Harden the stderr unit so it fails when `client_row_with_cap` breaks early on Linux (see the pointer on the new test).

### What worked well

- Root-cause fix for project-layout leftovers (store-from-base + both lock leaves + phase-8 `rmlock`) with strong teardown tests and a revert litmus that fails closed.
- Moving the round-trip cwd to `$HOME/roundtrip` with an install pin so skills cannot hide under pruned `work`.
- Dropping the wrong keychain/session-bus product change and leaving #592's CI stub as the honest runner fix.

<details>

Tier `standard`, trust `trusted`, reviewed at `667fa984...9c1968ee`.

| Gate | Result |
| :--- | :--- |
| uv sync / ruff / 113 uninstall tests | pass |
| H15 security | clean |
| H16 MCP | n/a |
| CI roundtrip @ head | pass teardown; residual one I/O error |
| Refute EPIPE / skilllock | survives / adjusted IMPORTANT |

Live smoke: GHA `install.sh then uninstall.sh` job 93568210763. Not covered: mutating Pipefy MCP (n/a).

</details>

mocha06:
Both Required items fixed in `36e57ad`. I checked each against the code and the head log before touching anything, and you were right on both — no push-back.

**The stderr fix was incomplete and my CHANGELOG claim was false.** `client_row_with_cap` still `break`s out of a `while read` fed by `client_rows`, and the head job still logged one `printf: I/O error`. I had that log and did not check it against the claim I was making. Head now logs **zero**, down from two.

Your suggestion to harden the unit turned into a structural check instead, because a behavioural one provably cannot hold here: whether the producer is still writing when the consumer leaves depends on the pipe buffer, so it reproduces on Linux and not on macOS — which is exactly how the `break` survived the pass that fixed the other four. `check_pipe_drain.py` follows the deletion-guard shape and rejects `break` / awk `exit` / `head` / `grep -q` in any pipeline whose producer is a function in this script, allowing `exit` inside an awk `END` block since `END` runs only after input is consumed.

It earned its place immediately. Run against the previous commit it names **four** sites, not one:

```
uninstall.sh:286   break leaves the loop while client_rows is still writing
uninstall.sh:2281  awk leaves on its first match, and records is still writing
uninstall.sh:2289  awk leaves on its first match, and records is still writing
uninstall.sh:2340  awk leaves on its first match, and records is still writing
```

Exit 1 against that file, 0 against this one, and it runs in pre-commit and CI.

**The skilllock mis-bind is real and I extended the fix one step further than the thread.** Keying on name alone mis-binds the store, and it mis-binds **provenance** the same way — a project skill taking its `source` from the global record — so both now key on `(lock, name)`, resolved from the directory being swept rather than looked up by name. `skill_lock_file` is gone entirely; the caller owns the lock, which removes the lookup that could not be disambiguated at all. `test_both_layouts_at_once_each_answer_from_their_own_lock` covers it and fails against `9c1968e` with the project store surviving at exit 0.

Both optional docs items taken: the never-removed bullet and all three CHANGELOG entries now use one formula, `<base>/.agents/skills` for the base its lock was found at, instead of the superseded `dirname(lock)/skills`.

Retitled and rewrote the Outcome — you were right that the old title undersold the tip.

All six checks green at `36e57ad`, and I verified the round-trip log directly rather than inferring it from a green tick this time.

adriannoes (dismissed):
Re-checked at `36e57ad` after the follow-up commit.

The two merge gates from the earlier review are addressed: `client_row_with_cap` drains the pipe (head round-trip Uninstall has zero `printf: I/O error`), and skilllock lookups are keyed on lock plus name with a dual-layout regression. Docs/CHANGELOG/title and the static `check_pipe_drain.py` guard line up with that.

CI is green on this tip. **Verdict: merge-ready** into `dev`.

mocha06:
Rebuilt on current `dev` in `bbcd9cd`. Content is unchanged from the tip you approved at `36e57ad`; this is the merge and one conflict resolution.

The conflict was `CHANGELOG.md` under `## [Unreleased]`, and it was not the usual both-sides-appended case. `#590` and `#591` carried the `--client none` and skill-store entries to `dev` in their original wording, while this branch had rewritten those same two entries for the one-formula change you asked for. A plain union kept both versions — four entries where there should be two. Resolved by entry subject instead: this branch's wording for the two that exist on both sides, and nothing from `dev` dropped (there was nothing unique to it in that hunk).

Verified after the merge: two entries, not four; `beside its store` gone everywhere; `<base>/.agents/skills` for the base its lock was found at, once.

All seven checks green on `bbcd9cd`, and the round-trip log still shows zero `printf: I/O error`. Re-requesting review only because the merge commit reset the approval.

adriannoes (approved):
Re-checked at `bbcd9cd` after the merge of `origin/dev`.

Delta since the prior approve is that merge only (#590 / #591 already on `dev`). Uninstall fixes for the pipe drain and lock-keyed skilllock lookups are unchanged. CI is green; round-trip Uninstall still shows zero `printf: I/O error` and a clean filesystem check.

Issue #601, opened 2026-08-11
refactor: apply the four architecture decisions (ADR rollout)


gbrlcustodio:
Umbrella tracker for applying the architecture decisions recorded in `docs/contributing/adr/`. This is a first task list to refine and split later.

## Typed-output rollout (ADR-0002)

- [ ] Card first, the highest-traffic and most-duplicated shape
- [ ] Remaining resources, one at a time behind the stable facade
- [ ] A typed fake of the SDK surface so client tests stop mocking flat wire methods

## MCP outcome tools (ADR-0003)

- [ ] Audit the tool surface for endpoint-shaped tools
- [ ] Consolidate into outcome-shaped tools, keeping primitive escape hatches
- [ ] Move remaining write gates onto protocol-native elicitation

## Vertical-slice refactor (ADR-0004)

- [ ] `PipefyClient` to `Pipefy` composition-root rename
- [ ] Per-slice import-linter contract: facade to use case to client to model, `models.py` free of transport and framework
- [ ] Split `webhook_service`: System Integration (webhook CRUD), Communication (email), Work Execution (`get_card_inbox_emails`)
- [ ] Split `observability_service`: Billing (usage/credits), Performance and Oversight (execution metrics, automation logs), Governance and Audit (AI-agent logs)
- [ ] Split `report_service`: Performance and Oversight (reports), Governance and Audit (`export_pipe_audit_logs`)
- [ ] Split `relation_service`: Work Execution (card relations), Process Modeling (pipe-relation definitions), Business Records (table relations)
- [ ] Split `pipe_service`: Process Modeling (definition reads), Work Execution (phase-card reads), Identity and Access Management (`get_pipe_members`)
- [ ] Rename and split `pipe_tools.py` (MCP) across Work Execution, Process Modeling, Request Intake, and Identity and Access Management

PR #598, opened 2026-08-11
docs: add documentation authoring guideline

gbrlcustodio:
Stacked on #597. Base is `docs/architecture-principles`, so this diff shows only the Phase 2 change. Review and merge #597 first.

## What

Add `docs/contributing/authoring.md`, a short contributor guideline for how the `docs/` tree is organized and where a new doc goes. It joins its siblings `architecture.md` and `conventions.md`, and is linked from `docs/README.md` and `AGENTS.md`.

## Why

Phase 1 derived a set of rules for the docs tree but never wrote them down. This records them, cut to the irreducible set (mantra: less brings clarity, more brings confusion):

- Where a doc goes: audience first (contributor, consumer-by-surface, durable cross-cutting at root), then Diataxis kind, one kind per file where practical.
- Generate fast-changing reference from code (docstrings, `Field(description=...)`, the registry, Typer help); hand-author only where no code source exists; never mix a generated table with durable prose.
- Keep it small: keep recognizable names, do not invent buckets for a few files, and send concrete migration steps to issues rather than a checklist in the doc.

## The target, not the current state

The guideline describes the target the tree converges to. The current tree is mid-migration. It states the rules, not a status list, because status rots. The known divergences (generate `parity.md` and the `config.md` table; relocate `ipaas.md` under `docs/mcp/`; split the four-quadrant `cli/auth.md` and portal's contributor testing content; separate the contributor tails in `MIGRATION.md`/`parity.md`) are staged as separate, reviewed steps tracked in issues. Each is load-bearing in code or tests, so none happen here.

## Testing

Docs-only: one new file (23 lines) plus two index rows. No code, test, or fixture path changed, so the parity test, the auth `_docs.py` assertions, and the CLI golden snapshot are untouched. Links resolve; no em or en dashes or curly quotes.

PR #602, opened 2026-08-11
docs: add architecture decision records and graduate their rules

gbrlcustodio:
| Stack | PR | Base | Content |
|---|---|---|---|
| 1 of 2 | #597 | `dev` | The living docs: architecture and conventions. |
| 2 of 2 | **#602 (this one)** | `docs/architecture-principles` | The decision records under `docs/contributing/adr/`. |

Review #597 first. This diff shows only the second step.

## What

Record the architecture decisions from the refactor spike as four consolidated decision records, add the guideline that says where a doc belongs, and connect both to the living docs.

- `docs/contributing/authoring.md` (new): where a doc goes. Audience first (contributor, consumer by surface, durable cross-cutting at root), then Diataxis kind, one kind per file where practical. Generate fast-changing reference from code. Keep recognizable names and do not invent buckets for a few files. It also holds the rule for where a decision record lives.
- `docs/contributing/adr/` (new): four records. Layered responsibility (0001), typed single-form contract (0002), MCP tools express outcomes (0003), vertical-slice structure and naming (0004). An index `README.md` maps each record to the living doc that carries its current rule, and lists the rollout work tracked in #601.
- `docs/mcp/README.md`: gains the tool-design section that `architecture.md` points to.
- `docs/contributing/architecture.md`: four lines change, each one a back-link to a record this PR creates. That is the whole diff on this file.

## All four records land as proposed

None is adopted. A fact-check against the code found that the rules are not yet true of the code, so the set stays one editable proposal rather than a set of immutable records.

This matters for how you read them. An adopted record is immutable, and changing the decision means adding a record that supersedes it. A proposed record is still editable in place, so a correction here is cheap. Say so now if a decision reads wrong.

The rollout that would earn adoption is tracked in #601.

## Why the records are separate from the rules

The rule a contributor follows day to day lives in a living doc (`architecture.md`, `conventions.md`, `mcp/README.md`). The record keeps the reasoning that produced it. A living doc can be edited freely as the code moves. A record, once adopted, cannot. Splitting them means the durable why does not rot alongside the current how.

## Scope notes

- The ten exploratory records from the spike are consolidated into these four. The full exploration stays in the branch history of `refactor/mcp-named-concerns`.
- No `domain-surface-map.md` ships. It derived from an internal model, so its durable slice list moved into ADR-0004 and its migration work-list into #601.
- Existing record prose is kept intact beyond the targeted status and scope changes, so the recorded reasoning is not altered after the fact.

## Testing

Docs-only. No test asserts a `docs/` path, and no workflow checks links, so verification is manual:

- Every relative link in the changed files resolves against the tree this PR produces, stacked on #597 (66 links checked). The four new back-links in `architecture.md` point at records this same PR adds, so no merge point has a dangling link.
- No code, test, or fixture path changed.


gbrlcustodio:
Closing in favor of #627, which carries the same branch (`docs/decisions`) with the correct base.

The content is unchanged and the head commit is the same. The only reason for a new number is that GitHub refuses a base change on a member of a native stack, and this PR still pointed at `docs/authoring`. After #598 closed, that base was no longer an ancestor of this branch, so the merge base fell back to `dev` and the diff inflated to 15 files, showing #597's changes as well as this PR's. #627 has the correct base and shows the intended 10 files.

This PR also now includes the authoring guideline from #598 and the record statuses from #625, so the four-PR stack is two PRs: #597 then #627.

PR #577, opened 2026-08-04
chore: back-merge main into dev (9 commits)

mocha06:
## Motivation

`origin/main` carries 9 commits that `origin/dev` does not, and that blocks every release right now.

The commits are the Apache-2.0 LICENSE and NOTICE, the DCO workflow with the issue and pull request templates, the conduct, security and compliance docs, and the README banner. None of them touched the version, so `release-pr`'s version-string guard could not see them. The commit-count guard added in #572 does see them, and it refuses to cut a release until this lands.

Two earlier attempts, #564 and #565, were closed without merging. The drift has stood since 1 August.

## Outcome

`origin/dev` contains every commit on `origin/main`. `release-pr` runs again.

**This merge changes no files.** The same content had already been applied to `dev` independently through separate pull requests — `dev` already carried `LICENSE`, `NOTICE`, `.github/workflows/dco.yml`, `CODE_OF_CONDUCT.md`, and `SECURITY.md` before this. The merged tree is byte-identical to `origin/dev`. What this reconciles is history, so that `git rev-list origin/dev..origin/main` reaches zero.

Verify that claim directly:

```bash
git fetch origin dev rc-dev/chore/back-merge-main-4691d13
git diff origin/dev origin/rc-dev/chore/back-merge-main-4691d13   # prints nothing
```

### Two resolutions

Both are worth a look, because neither was mechanical.

**`docs/images/pipefy-developers-banner.png`** conflicted. Both branches replaced the same broken asset independently: `dev` carried a JPEG misnamed `.png` since May, `main` fixed it on 1 August, and `dev` fixed it differently on 3 August in #568. I kept `dev`'s version. It is 1200x563 and complete; `main`'s is 1024x285 and crops the top of the "pipefy Developers" heading while dropping the descriptive line under the divider entirely. Say so if you want `main`'s instead and I will swap it.

**`skills/AGENTS.md`** auto-merged into a bug. Git saw `main` add two paragraphs about regulated-domain review and the skill template, did not know `dev` already carried the same two paragraphs from a separate commit, and appended them a second time. The duplicate is removed in the merge commit. A clean auto-merge is not the same as a correct one, which is the argument for this landing as a reviewed pull request rather than a direct push.

## Notes

The branch name matches what `.github/workflows/backmerge.yml` generates for this `main` head (#576). Once that workflow is on `main`, it recognises this branch and its pull request and leaves both alone rather than opening a second one.

[...]

PR #576, opened 2026-08-04
ci: open the main to dev back-merge pull request automatically

mocha06:
## Motivation

A commit that reaches `main` outside the release flow leaves `dev` behind, and nothing on the GitHub side notices.

`release-pr` already refuses to cut while `origin/main` carries commits absent from `origin/dev`. That guard turns a silent hazard into a hard failure, but it fires only at the next release. The drift persists until someone tries to cut one, and it is diagnosed by whoever happens to be cutting rather than by whoever caused it.

The case is live now. `origin/main` carries 9 commits that `origin/dev` does not: the Apache-2.0 LICENSE, the DCO workflow, the conduct and security docs, and the README banner. None of them touched the version, so the version-string guard could not have seen them. Two manual back-merge attempts, #564 and #565, were closed without merging, so this has stood since 1 August and blocks the next release today.

## Outcome

`.github/workflows/backmerge.yml` counts `origin/dev..origin/main` on every push to `main`, daily, and on demand. When the count is zero it closes any open tracking issue and stops. When the count is non-zero it cuts `rc-dev/chore/back-merge-main-<short-sha>` from `origin/dev`, merges `origin/main` in, and opens a pull request into `dev`. The reconciliation waits for review instead of waiting to be remembered.

The decision logic lives in `.github/workflows/scripts/backmerge.py`. Read that file first. The workflow supplies a checkout and credentials, nothing else.

### Behaviour boundaries

- **A conflict is a normal outcome.** The merge is aborted, nothing is pushed, and an issue labelled `back-merge` names the conflicting paths and the commands to resolve them. The run exits zero, because a red run on every release trains people to ignore the signal.
- **Re-runs are idempotent.** The branch name derives from the `main` head, so the same drift always resolves to the same branch. An existing branch or pull request is recognised and left alone. A closed back-merge pull request counts too: closing one is a human decision.
- **Nothing is resolved automatically.** `-X ours` and `-X theirs` both silently discard real work.
- **The tracking issue is one thread.** It is found by label rather than by title, its title carries the live count, it closes on its own once the branches are back in sync, and a re-run that would repeat itself stays silent.
- **The workflow cannot push to `dev` or merge its own pull request.** The ruleset forbids both, and that human approval is the point.

### CI on the back-merge pull request

[...]

Issue #561, opened 2026-08-01
ci: open the main -> dev back-merge pull request automatically


mocha06:
## Summary

Anything merged directly into `main` — a hotfix, a docs fix, a release bump — leaves `dev` behind until a human remembers to back-merge. Nothing on the GitHub side notices. The drift is silent from the moment it happens until the next release cut trips over it, which can be days later and is diagnosed by whoever happens to be cutting rather than by whoever caused it.

Add a workflow that watches for the divergence and opens the back-merge pull request itself, so the reconciliation is waiting for review instead of waiting to be remembered.

## Why the existing issues do not cover this

Two open issues already address the release path, and this one deliberately does not overlap with either:

- **#528** makes `release-pr` fail when `origin/main` has commits absent from `origin/dev`. That is a gate at the *next release*, so the drift still persists for however long passes before someone cuts one.
- **#532** makes `publish` close the loop for the drift *the release itself creates* (the version bump and the stamped `CHANGELOG` section).

Neither sees a commit that reaches `main` outside the release flow. That is the case here, and it is the more common one: #455 (`rc-main/fix/marketplace-version-lockstep`) and #457 (`docs/beginner-onboarding`) both merged straight into `main` without touching the version, so #528's guard would not have fired and #532's post-`publish` hook would never have run. #530 was the manual cleanup after a release cut, and it is a good picture of the pull request this workflow should be producing on its own.

## Constraints to design against

- **A pull request is mandatory on both branches.** The repository ruleset (`Push Ruleset`, active on `~DEFAULT_BRANCH` and `refs/heads/dev`) requires a pull request with one approving review, and blocks deletion and non-fast-forward pushes. The workflow cannot push to `dev` directly, and it cannot merge its own pull request. That is fine and intended — a human approval is the point. The workflow's job is to make sure the pull request *exists*.
- **A pull request opened with `GITHUB_TOKEN` does not trigger `on: pull_request` workflows.** Left as-is, the back-merge pull request would show zero checks, and CI (`.github/workflows/ci.yml`) is exactly what should be validating the merge result. Resolve this with a GitHub App token or a PAT, or state in the pull request body that checks must be re-run manually. Whichever is chosen, say so explicitly in the workflow comments — a green-looking pull request with no checks is worse than an obviously unchecked one.
[...]

PR #605, opened 2026-08-12
ci: put the back-merge workflow on main so its triggers can fire

mocha06:
## Motivation

The back-merge automation from #576 has never run. It landed on `dev` alone, and every trigger that would run it for real resolves the workflow file somewhere it is not:

- `push: branches: [main]` resolves the workflow from the pushed commit **on `main`**. The file is not there.
- `schedule` and `workflow_dispatch` are read from the **default branch**, which is `main`. The file is not there either. `.github/workflows/backmerge.yml:25-27` already says this about those two.

The run history is the proof — two runs, both `pull_request` dry runs on the branch that introduced the workflow:

```
$ gh run list --workflow=backmerge.yml
pull_request  rc-dev/ci/backmerge-main-to-dev  success  2026-08-05
pull_request  rc-dev/ci/backmerge-main-to-dev  success  2026-08-04
```

Zero scheduled runs in the five days since it merged, against a `cron: "37 6 * * *"`. The `back-merge` label does not exist in the repository either, which agrees: `ensure_label()` creates it on the first real run, and there has not been one.

The fix for `main`/`dev` drift is stranded on `dev` by that drift.

## Outcome

`.github/workflows/backmerge.yml`, `.github/workflows/scripts/backmerge.py`, and `tests/test_backmerge.py` now exist on `main`, so `push`, `schedule`, and `workflow_dispatch` all resolve to a real file.

All three are copied **verbatim** from `dev` (`git checkout origin/dev -- <paths>`), so the two copies are byte-identical and the next back-merge is a no-op on these paths rather than a conflict.

`RELEASE.md` and `CHANGELOG.md` were also touched by #576. Those stay on `dev` and reach `main` through the usual release pull request — carrying them here would put a stamped-changelog edit on `main` outside a release, which is the shape of drift this whole mechanism exists to prevent.

## Scope

This restores the trigger. It does **not** by itself clear the current 9-commit divergence, because that divergence is an ancestry problem, not a content one: #577 and #570 were squash-merged, so `main`'s commits are on `dev` in content but not as ancestors, and `git rev-list --count origin/dev..origin/main` stays at 9 no matter how many times the workflow runs. A follow-up pull request repairs the ancestry with a real two-parent merge.

## Verification

| # | Check | Result |
|---|---|---|
| 1 | The three paths are absent from `origin/main` before this change | Confirmed via `git ls-tree origin/main .github/workflows/` |
| 2 | Files are byte-identical to `origin/dev` | Staged with `git checkout origin/dev -- <paths>`; no edits after |
| 3 | `tests/test_backmerge.py` passes against `main`'s tree | 35 passed |
| 4 | The test needs nothing that is `dev`-only | It loads the script by path and stubs every subprocess call; only `pytest` is required |
| 5 | Nothing else is staged | `git status --short` shows exactly three added paths |

## After this merges

`BACKMERGE_TOKEN` is not set, so back-merge pull requests will open with no checks on them. That is the documented behaviour from #576, not a regression: GitHub does not trigger `on: pull_request` for a pull request opened with `GITHUB_TOKEN`, and the generated body says so. Setting the secret turns checks on with no change to the workflow.

Refs #561, #576.

### DIFF
--- a/.github/workflows/backmerge.yml
+++ b/.github/workflows/backmerge.yml
@@ -1,0 +1,82 @@
+name: Back-merge
+
+# Anything merged straight into `main` leaves `dev` behind until a human
+# remembers to reconcile it. `release.py release-pr` refuses to cut while that
+# is true, so forgotten drift does not corrupt a release -- it blocks one, at
+# the least convenient moment, for whoever happens to be cutting. This opens the
+# back-merge pull request itself so the fix is waiting for review instead of
+# waiting to be remembered.
+#
+# All the logic is in `scripts/backmerge.py` rather than inline shell: the
+# idempotency rules (an existing branch, an existing pull request, an unchanged
+# tracking issue) are decisions worth unit-testing, and `tests/test_backmerge.py`
+# does exactly that. This file only supplies credentials and a checkout.
+
+on:
+  push:
+    branches: [main]
+  schedule:
+    # Daily, off the hour -- GitHub drops scheduled runs queued at popular
+    # times. This is the backstop for drift the push event missed (a failed
+    # run, a merge during an Actions outage) and for re-checking a divergence
+    # that is still unresolved.
+    - cron: "37 6 * * *"
+  workflow_dispatch:
+  # This workflow's own changes are exercised by the pull request that makes
+  # them. `schedule` and `workflow_dispatch` stay dormant until the file
+  # reaches `main`, since GitHub reads both from the default branch alone.
+  pull_request:
+    paths:
+      - .github/workflows/backmerge.yml
+      - .github/workflows/scripts/backmerge.py
+
+permissions:
+  contents: write
+  pull-requests: write
+  issues: write
+
+concurrency:
+  # Every real trigger shares one group: two at once would race each other
+  # between "does the branch exist" and "push the branch". Pull request dry
+  # runs write nothing, so they get their own per-ref group rather than queuing
+  # behind a real run.
+  group: backmerge-${{ github.event_name == 'pull_request' && github.ref || 'default' }}
+  cancel-in-progress: false
+
+jobs:
+  backmerge:
+    name: Open the main to dev back-merge
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # The merge needs real history, not the default depth-1 cone.
+          fetch-depth: 0
+          # A pull request opened with GITHUB_TOKEN does not trigger
+          # `on: pull_request` workflows, so the back-merge pull request shows
+          # no checks and its body says so plainly. Set a BACKMERGE_TOKEN
+          # secret -- a GitHub App token or a fine-grained PAT scoped to this
+          # repository with contents:write, pull-requests:write and
+          # issues:write -- and checks start running with no change here.
+          token: ${{ secrets.BACKMERGE_TOKEN || secrets.GITHUB_TOKEN }}
+
+      - name: Configure the committer
+        # Identity for the merge commit, and the DCO sign-off derives from it.
+        run: |
+          set -euo pipefail
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Open the pull request, or track why it could not be opened
+        env:
+          GH_TOKEN: ${{ secrets.BACKMERGE_TOKEN || secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          BACKMERGE_HAS_CI: ${{ secrets.BACKMERGE_TOKEN != '' }}
+          # On the pull request that edits this workflow, run every real step --
+          # fetch, count, merge, conflict detection -- but announce the writes
+          # instead of performing them. Otherwise testing a change here would
+          # open a genuine back-merge pull request off the branch under review.
+          BACKMERGE_DRY_RUN: ${{ github.event_name == 'pull_request' }}
+        run: python .github/workflows/scripts/backmerge.py
--- a/.github/workflows/scripts/backmerge.py
+++ b/.github/workflows/scripts/backmerge.py
@@ -1,0 +1,636 @@
+#!/usr/bin/env python3
+"""Open the ``main`` -> ``dev`` back-merge pull request, or record why one could not be.
+
+Anything merged directly into ``main`` -- a hotfix, a docs fix, a LICENSE --
+leaves ``dev`` behind until a human remembers to back-merge. Nothing on the
+GitHub side notices, so the drift is silent from the moment it happens until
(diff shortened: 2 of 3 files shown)


.github/workflows/scripts/backmerge.py
```diff
@@ -0,0 +1,636 @@
...
+    return int(
+        capture(
+            ["git", "rev-list", "--count", f"origin/{DEV_BRANCH}..origin/{MAIN_BRANCH}"]
+        )
+    )
+
+
+def main_head() -> str:
+    """The short SHA of ``origin/main``."""
+    return capture(["git", "rev-parse", "--short", f"origin/{MAIN_BRANCH}"])
```
adriannoes, line 115: **Optional.** `main_head()` keys the branch name, `existing_pr` lookup, remote-branch check, and state marker off `git rev-parse --short` with no fixed length. Git can lengthen that abbreviation as the object DB grows, so a later scheduled run can miss the open PR and treat the same tip as new.

I would pin a fixed abbrev (for example `--short=12`) or a stable prefix, on `dev` and `main` together so the copies stay identical. Your call whether to take this before or after this trigger restore lands.

.github/workflows/scripts/backmerge.py
```diff
@@ -0,0 +1,636 @@
...
+    sha = main_head()
+    branch = branch_for(sha)
+    commits = carried_commits()
+    print(f"origin/{MAIN_BRANCH} is {behind} commit(s) ahead at {sha}.")
+
+    # Idempotency, in the order the states can occur.
+    pr, pr_state = existing_pr(branch)
+    if pr and pr_state == "OPEN":
+        print(f"Pull request #{pr} already covers {sha}; leaving it alone.")
+        return 0
```
adriannoes, line 530: **Optional.** When an OPEN PR already covers the tip, this early return never calls `supersede_tracking_issue`. Create paths do. A prior conflict tracking issue can stay open beside the live PR.

I would call `supersede_tracking_issue(branch)` before returning here, with a regression test. Your call whether to fold that into a follow-up on both branches.

.github/workflows/backmerge.yml
```diff
@@ -0,0 +1,82 @@
+name: Back-merge
+
+# Anything merged straight into `main` leaves `dev` behind until a human
+# remembers to reconcile it. `release.py release-pr` refuses to cut while that
+# is true, so forgotten drift does not corrupt a release -- it blocks one, at
+# the least convenient moment, for whoever happens to be cutting. This opens the
+# back-merge pull request itself so the fix is waiting for review instead of
+# waiting to be remembered.
+#
+# All the logic is in `scripts/backmerge.py` rather than inline shell: the
```
adriannoes, line 10: **Optional.** The header says the logic lives in `scripts/backmerge.py`, but the job runs `.github/workflows/scripts/backmerge.py` and the short path is not in the tree. Your call on fixing the comment when you next touch the file.

adriannoes (approved):
Reviewed at `94478809`: detached worktree gates (`uv sync`, ruff, 35 backmerge tests), CI green including the intentional `pull_request` dry-run, security/ops/contracts/thermo hunts, and byte-identity vs `origin/dev` confirmed.

**Verdict: merge with notes** into `main`.

### Decisions I would make

- Accept that `RELEASE.md` / `CHANGELOG.md` stay on `dev` for now. Generated PR bodies that say "see RELEASE.md" for `BACKMERGE_TOKEN` will dangle on `main` until a normal release promotion. That is the drift shape this mechanism exists to avoid inventing on `main` outside a release.
- Treat #606 as the ancestry companion (either-first). This PR restores triggers; it does not clear the 9-commit squash ancestry gap.

### Optional

Your call on all of these.

- Dry-run PR checks still use the job name **Open the main to dev back-merge** while `mutate` only announces writes. Renaming or annotating the dry-run check would stop a green conclusion from reading like a completed back-merge. Live fail-closed behavior is already solid.
- Three smaller pointers are on their lines in the diff (short-SHA idempotency key, supersede on OPEN early return, workflow header script path).

### What worked well

- Precise Scope: restore triggers without claiming to fix squash ancestry, and keep RELEASE/CHANGELOG off this `main` landing.
- Verbatim copy from `origin/dev` tip is actually byte-identical; future back-merge is a no-op on these paths.
- Live fail-closed compensating-signal tests and the dry-run mutation gate are deliberate and well covered.

<details>

Tier `standard`, trust `trusted`, reviewed at `4691d132…94478809`.

| Gate | Result |
| :--- | :--- |
| Phase 3 baseline | pass (35 tests) |
| CI | green |
| H15 security | pass |
| H14 ops | pass |
| H13 claims | pass (verbatim) |
| H16 live MCP/CLI | n/a |
| Phase 7 refute | n/a (nothing blocking) |

Live smoke: n/a. Not covered: first real post-merge `push`/`schedule` write (expected after land; token unset documented).

</details>

.github/workflows/scripts/backmerge.py
mocha06: Real, and it predates this change. `main_head()` has used a bare `--short` since the file landed on `dev` in #576. This pull request copies it byte for byte and adds no line.

Not fixed here, because these three paths must stay byte-identical to `origin/dev`. A change on `main` alone puts the next back-merge conflict inside `backmerge.py`, which is the script that performs back-merges.

Current exposure is bounded: the abbreviation sits at git's floor of 7 characters against roughly 16k objects, so a step to 8 needs substantial growth, and the worst case is a duplicate pull request rather than a bad merge.

Filed as #610 against `dev`, pinning an explicit length with a test. It reaches `main` through the back-merge this pull request restores. Does that work?

.github/workflows/scripts/backmerge.py
mocha06: Real. `supersede_tracking_issue` runs at `:590` and `:631`, both create paths, and the OPEN early return at `:530` reaches neither.

The sequence is reachable: a conflicting run files the issue, someone opens the pull request by hand, and the next run returns at `:530` with the issue still open. That is the state the docstring says the function prevents.

Not fixed here, for the byte-identity reason on the sibling thread: any edit to these paths on `main` alone makes the next back-merge conflict inside `backmerge.py`.

Filed as #610 against `dev`, with the regression test beside the existing idempotency cases. Does that work?

.github/workflows/backmerge.yml
mocha06: Fair. Line 10 reads `scripts/backmerge.py`, which resolves against the workflow's own directory, while line 12 reads `tests/test_backmerge.py` from the repository root. One comment block, two bases.

Not fixed here, for the same byte-identity reason as the two inline threads on the script.

Folded into #610 against `dev`. Does that work?

mocha06:
Taking the dry-run check name as a follow-up, not here.

`jobs.backmerge.name` is `Open the main to dev back-merge` for every event, and `pull_request` runs it with `BACKMERGE_DRY_RUN=true`. A green check does read like a completed back-merge.

Fixing it means editing `backmerge.yml`, and these three paths must stay byte-identical to `origin/dev` or the next back-merge conflicts inside the script that performs it. Same reason as the three inline threads.

All four are filed as #610 against `dev`. They reach `main` through the back-merge this pull request restores, which is the intended path for them. Does that work?

PR #606, opened 2026-08-12
chore: restore main's ancestry on dev after the squashed back-merges

mocha06:
> **Merge this with "Create a merge commit".** Squashing it reintroduces exactly the problem it repairs, and the next person gets to diagnose it again.

## Motivation

`origin/main` has counted 9 commits ahead of `origin/dev` since 3 August, and re-running the back-merge never clears it. The content is not missing — the **ancestry** is.

#577 and #570 were squash-merged. Their merge commits carry one parent each:

```
$ git log -1 --format='%h parents: %p' 8c43a82
8c43a82 parents: 05ed348          # PR #577, one parent
$ git log -1 --format='%h parents: %p' 33c7593
33c7593 parents: 3f0c0e5          # PR #570, one parent
```

A squash copies the content and records no ancestry, so `main`'s commits are on `dev` as text while `main`'s head stays unreachable from `dev`. `git rev-list --count origin/dev..origin/main` is stuck at 9 permanently. #553 and #530 were proper two-parent merges, so this is a recent change in habit rather than a long-standing one.

Two things break as long as that holds:

- `release.py release-pr` refuses to cut at all — the guard from #572 counts commits, and the count never reaches zero.
- The back-merge workflow, once #605 puts it on `main` where it can actually run, would re-propose an identical pull request on every scheduled run forever.

## Outcome

A two-parent merge commit that reconnects the histories and changes no file.

```
d69f90d parents=6f4b43b 4691d13
```

The resulting tree is **byte-identical to `origin/dev`**, verified against `git rev-parse origin/dev^{tree}`. After this merges, `git rev-list --count origin/dev..origin/main` is `0`.

## Why `-s ours` and not a plain merge

Because `dev`'s tree is already correct and complete, and a plain 3-way merge actively corrupts one file.

A plain `git merge origin/main` changes exactly one path against `origin/dev` — it appends a **second copy** of two paragraphs `dev` already carries in `skills/AGENTS.md`:

```
$ grep -c "Regulated domains" skills/AGENTS.md
2                     # after a plain 3-way merge
1                     # on origin/dev, and 1 on origin/main
```

That is the squash artifact. With no shared ancestry, git reads one paragraph on each side as two independent additions and keeps both. There is nothing to gain by merging the text and a duplicated section to lose.

The rest of the tree was checked directly rather than assumed:

| Check | Finding |
|---|---|
| Files on `main` but absent from `dev` | `fastmcp_tool_lifecycle.py`, `transition_hints.py` — both deleted by #589 on `dev`. A 3-way merge keeps them deleted; so does this. |
| Content delta of a plain 3-way merge vs `origin/dev` | One file, `skills/AGENTS.md`, `+4` lines, all of them duplicates |
| Resulting tree vs `origin/dev` | Identical |

## The one real conflict

`docs/images/pipefy-developers-banner.png` is the only path that conflicts, and it is binary, so there is no textual resolution. `dev`'s version is kept:

| | `main` | `dev` |
|---|---|---|
| Landed | 1 Aug (`1866b51`) | 3 Aug (#568, `2f8b61e`) |
| Dimensions | 1024x285 RGB | 1200x563 RGBA |
| Content | Same lockup, cropped | Same lockup plus the descriptive tagline |

Both are valid PNGs and the README markup around them is identical on both branches, so this is purely about which artwork ships. `dev`'s is newer, higher resolution, and a superset. Overriding it is a one-commit change on this branch if you disagree.

## Ordering against #605

Independent — either can merge first.

- **#605 first** (puts the workflow on `main`): `main` gains a commit, this branch needs one more `git merge origin/main` before it merges. Additive, no force-push.
- **This one first**: `dev` is clean, then #605 merges and leaves `main` one commit ahead. The workflow is live on `main` by then, so it opens that back-merge itself — the mechanism proving itself on its own first run.

## Follow-up this does not cover

Nothing currently stops the next back-merge from being squashed again. The repository allows all three merge methods, and the workflow has no way to detect after the fact that its pull request was flattened. Worth deciding separately: drop squash from the allowed methods, or have the generated pull request body lead with the required merge method.

Refs #561, #576, #605. Supersedes the drift left by #577 and #570.


adriannoes (approved):
Reviewed at `d69f90d`: verified the two-parent `-s ours` merge against git objects (parents `6f4b43be` + `4691d132`, tree `2267e1f9` identical to `origin/dev`, empty three-dot file delta, `rev-list` from head to `origin/main` is 0). CI install/uninstall is red on a GitHub releases API 403 with the same tree that already passed on `dev`.

### Decisions I would make

- Merge with **Create a merge commit** only. Squash or rebase drops the second parent and leaves `origin/dev..origin/main` stuck at 9 again (same failure mode as #577/#570). After merge: `git fetch && git rev-list --count origin/dev..origin/main` should print `0`.
- Treat the install/uninstall failure as environmental (unauthenticated releases list returned 403). Re-run the job or waive the required check; do not ask for a content change on this empty tip.
- Ordering vs #605: either first is fine. If #605 lands on `main` first, refresh this branch with one more `git merge origin/main` before merging.

### Optional

Your call on all of these.

- Repo still allows squash on future back-merges. A follow-up can disable squash and/or harden the generated back-merge PR body (this PR correctly leaves that out of scope).
- Tiny PR-body typo: #570's single parent is `ff826581`, not `3f0c0e5`. Conclusion (squash / one parent) is still right.

### What worked well

- The PR leads with the merge-method constraint and explains why `-s ours` avoids the `skills/AGENTS.md` duplicate and keeps the newer banner.
- Claims are checkable with a handful of git commands and match the same `rev-list` metric `release.py` and the back-merge tooling use.

<details>

Tier `lite`, trust `trusted`, reviewed at `6f4b43be...d69f90d`.

| Gate | Result |
| :--- | :--- |
| Tree / parents / counts | pass |
| Product surfaces H1–H3/H16 | n/a |
| CI install/uninstall | red, environmental 403 |
| CI test/lint/DCO/shell/skills | green |

Live smoke: n/a. No worktree.

</details>

PR #625, opened 2026-08-12
docs: falsify architecture claims and mark ADRs as proposed

gbrlcustodio:
Stacked on top of #`docs/decisions`. Review after that base merges.

## Why

The base branch reads as if the ADR set had already been adopted. A fact-check against the code found claims that overreach or state a not-yet-shipped rule as present. No ADR is adopted yet, so the whole set is one editable proposal, and this pass falsifies the weak parts before adoption.

## What changed

`architecture.md`:
- Drop the universal-ports claim. Ports exist only in the SDK today (`GraphQLExecutor`, `S3Uploader`, `UrlDownloader`). `auth`, `infra`, and `cli` have none.
- Add a shared-support-libraries section placing `auth`, `infra`, and the CLI in the model, with ports governed by payoff.
- Make the import-linter claim precise (layer order plus the tools-read-no-settings rule). Mark the framework-free core a target, since its contract is written but disabled.
- Add the composition-root per-app nuance (MCP `core/runtime.py`; CLI at its entry point) and back-links to the ADRs.
- Keep the coherent proposed rules (numeric-first SDK, CLI name-resolution flag, MCP intent-first, elicitation).

ADRs:
- Mark all four `proposed`, not `adopted`. Correct the README status table and the intro wording that presumed adoption.
- Add a shared-library scope note to ADR-0001.

Prose follows an ASD-STE100 pass. Docs only, no code changes.

## Note

Existing ADR body prose is kept intact beyond the targeted changes, to avoid altering the recorded reasoning.


gbrlcustodio:
Closing. The content ships in the two PRs below, moved earlier in the stack so no reviewer reads a claim that this PR then retracts.

This PR was a correction to its own stack. It rewrote 31 of the 51 lines of `docs/contributing/architecture.md` that #597 introduces, and it flipped all four record statuses one commit after #602 wrote them. None of the 27 lines it deleted had reached `dev`, so the review cost was real and the shipped result was identical.

Where each part went, split by whether it depends on the record files:

- #597 takes the architecture claims: the universal-ports removal, the shared-support-libraries section, the precise import-linter contract, the framework-free-core caveat, the per-app composition-root detail, and the prose pass. None of these needs a record file to exist.
- #602 takes the record parts: all four statuses written as `proposed` from the start, the corrected index table and intro, the ADR-0001 scope note, and the four back-links in `architecture.md`.

The split is load-bearing rather than cosmetic. Folding this PR wholesale into #597 would have left four links to `adr/*.md` on `dev` with no target, because the records do not arrive until #602.

The tip tree is unchanged: tree `2be365da`, identical to this branch. The branch stays until #597 and #602 merge.

Issue #128, opened 2026-05-19
fix(mcp): replace silent degradation with friendly errors (post-#30 hardening)


adriannoes:
## Problem

- MCP tools report `success: true` on responses that carry no evidence the write happened: an empty automation payload, a `createCard` response with no `card.id`, a phase-field update where every provided key was dropped.
- Eleven destructive tools report a deletion the API declined. The mutation answers `success: false`, the tool ignores the flag and tells the agent the resource is gone.
- The follow-up read that would catch either case is exactly what an agent skips after a reported success.

## Evidence

State of each item verified against `main` on 2026-09-16.

### Item 1, validate the raw field list before filtering. Done

- `parse_field_definitions` runs on the raw list from Pipefy before any filtering: `packages/sdk/src/pipefy_sdk/services/pipe_service.py:311` (phase fields) and `:78` (start form).
- It raises `MalformedFieldDefinitionError` when any entry is not a dict or is missing `id` / `type`, including GraphQL `null`: `packages/sdk/src/pipefy_sdk/models/field_definition.py:43-86`.
- Seven sites in `packages/mcp/src/pipefy_mcp/tools/pipe_tools.py` catch it and return `tool_error`: `:193`, `:202`, `:226`, `:249`, `:269`, `:1226`, `:1256`.
- A list such as `[null, "x"]` can no longer reach an empty elicitation form. No work left; listed so nobody reopens it.

### Item 2, `fill_card_phase_fields` cannot tell "nothing to do" from "input dropped". Half done

- `pipe_tools.py:1266-1287` splits the message: "No field values were collected, so nothing was updated." when the phase has editable fields, "No fields to update." when it has none.
- Both branches still return `success: True` at `:1283`, so a caller that reads the envelope and not the prose still concludes the card is filled.
- Neither branch names the dropped ids. `skipped_field_ids` exists at `packages/sdk/src/pipefy_sdk/field_filters.py:38` and the CLI already surfaces it (`packages/cli/src/pipefy_cli/commands/card.py:427`), but it appears nowhere under `packages/mcp/src`.

### Item 3, `create_card` response without `card.id`. Open

- `pipe_tools.py:303-318`: `card_id` is read defensively, and when it is falsy the function falls through to `return result` at `:318` with no `card_link` and no error signal. The caller gets the raw GraphQL payload shaped like a success.

### Item 4, comment mutations parsed in the tool layer. Open

[...]

PR #629, opened 2026-08-14
feat(mcp): require signed preview token before destructive confirm

adriannoes:
## Summary

- Dedicated destructive MCP tools no longer execute on one `confirm=true`. Call 1 returns a preview with an HMAC `confirmation_token` and mutates nothing; call 2 runs only when `confirm=true` echoes a token valid for the same tool, resource identity, caller key, token version and 300s TTL.
- 29 tools gain optional `confirmation_token` (28 `delete_*` + `remove_member_from_pipe`). Preview wording puts "get their explicit approval" before the mechanics.
- Token failures are told apart: missing / invalid-or-expired / identity mismatch. A changed `pipe_uuid` between preview and confirm no longer reads as "expired".
- Version signed inside the MAC; non-canonical base64url rejected; identity canonicalised on both mint and verify (a stringified value used to mint a token that could never verify, so the caller previewed forever).
- 829 deletions are consolidation: `build_delete_pipe_preview_payload`, `build_delete_table_preview_payload` and `format_json_preview` gone, every tool through one guard.

**Part 1 of 3.** Merge #629 → #630 → #631. Retarget #630 to `dev` once this lands.

## Accepted limits (ADR-001, decisions not oversights)

| Limit | Why it stands |
| --- | --- |
| Hosted caller can mint its own token (key = `sha256(bearer)`) | Token orders the protocol; authorization stays the API permission |
| Replay inside the 300s TTL | Stateless by design, no ticket store |
| Auto-approving client can preview + confirm back to back | Guarantee is that the preview reached the transcript |
| `unpublish_sub_portal` ungated | Reversible |
| CLI keeps `--yes`, no tokens | TTL 300s and the six needles are frozen |

## Test plan

- [x] `uv run pytest -m "not integration"` (4618 passed)
- [x] `uv run lint-imports` from `packages/mcp`
- [x] Live smoke: preview without token does not delete; confirm with token does. Disposable resources only.
- [x] CI green here (DCO, shell scripts, install/uninstall, lint, test). Only PR in the stack based on `dev`, so it is the stack's CI anchor.

## Docs / skills

- [x] No docs/skills in this PR. `docs/parity.md`, skills and the CHANGELOG entry land in #631.

## Legal / contributions

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py
@@ -612,17 +612,20 @@ class AiAgentTools:
             meta=REMOTE,
         )
         async def delete_ai_agent(
-            ctx: Context, uuid: str, confirm: bool = False
+            ctx: Context,
+            uuid: str,
+            confirm: bool = False,
+            confirmation_token: str | None = None,
         ) -> dict:
             """Delete an AI Agent permanently. This action is irreversible.
 
-            Two-step operation: preview with ``confirm=False`` (default), then execute with
-            ``confirm=True`` after explicit human approval. Elicitation does not authorize
-            deletion (only ``confirm=True`` does).
+            Two-step operation: preview with ``confirm=False`` (default), then echo
+            ``confirmation_token`` from the preview on step 2.
 
             Args:
                 uuid: Agent UUID.
-                confirm: Must be ``True`` to run the delete mutation.
+                confirm: Set to True with the preview token to execute the deletion (step 2).
+                confirmation_token: Token from the preview response.
             """
             client = get_pipefy_client(ctx)
             agent_uuid = uuid.strip()
@@ -634,6 +637,9 @@ class AiAgentTools:
                 ctx,
                 confirm=confirm,
                 resource_descriptor=f"AI agent (UUID: {agent_uuid})",
+                resource_identity={"uuid": agent_uuid},
+                tool_name="delete_ai_agent",
+                confirmation_token=confirmation_token,
             )
             if guard is not None:
                 return guard
--- a/packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
@@ -245,17 +245,19 @@ class AiAutomationTools:
             ctx: Context,
             automation_id: PipefyId,
             confirm: bool = False,
+            confirmation_token: str | None = None,
             debug: bool = False,
         ) -> dict:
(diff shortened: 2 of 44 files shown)


adriannoes:
## Reviewed and deliberately deferred: the `success` flag on delete mutations

While reviewing this PR we checked whether the destructive tools honor the `success` boolean the delete mutations return. Thirteen of the 29 do not: they discard the flag and report success unconditionally, so an API-declined delete is reported to the agent as done.

It is **not** part of this PR and should not be fixed here:

- It is pre-existing. This branch touches none of those return lines (`git diff 2655d123...ba78dba1` contains no `success` line in the affected modules).
- Folding it in would mix a pre-existing correctness fix into a contract change and inflate a diff that is already 41 files.

Documented with the full tool list and evidence in #128, whose stated goals ("no silent partial success", "no false `success: true`") already cover it: https://github.com/pipefy/ai-toolkit/issues/128#issuecomment-5304192710

Recording it here so the omission is visibly a decision rather than an oversight.

mocha06 (approved):

Nothing survived verification against this PR. The guard is sound and its identity binding is complete. Three findings landed on #630 and one on #631, none blocking.

### The limits you flagged

- **Hosted caller mints its own token** — holds, and it is the right shape for the deployment. The key is `sha256(bearer)`, so any replica verifies the same token with no ticket store, and one caller's token does not verify for another.
- **Replay inside the 300s TTL** — holds. `exp` is exclusive, so a token dies at `exp` rather than after it.
- **`unpublish_sub_portal` ungated** — holds, and it is stronger than stated. The tool exposes no `confirm` argument at all, not merely no token.
- **CLI keeps `--yes`, no tokens** — holds.
- **The deferred `success` flag (#128)** — agreed, and out of scope here. This branch touches none of those return lines.

### Identity coverage

I enumerated every gated tool's input schema and compared it to the arguments each mutation actually sends. The only arguments outside `resource_identity` are `pipe_id` and `phase_id` on `delete_phase`, `delete_label`, and `delete_phase_field`. Those feed the preview dependents and the cascade diagnosis, never the mutation. `delete_phase_field` binds the optional `pipe_uuid`, the one optional argument that does reach its mutation. So nothing selecting can change between step 1 and step 2 while a token still verifies.

### Probes

- Mint and verify: `1` and `"1"` bind one token; `exp` is exclusive; `user_ids` order does not matter but a superset fails; the three failure classes classify correctly.
- Registry built from this branch: 187 tools, 30 with `destructiveHint`, 29 with `confirmation_token`, matching the 28 `delete_*` plus `remove_member_from_pipe` in the description.

### Findings

0 on this PR. 3 on #630, 1 on #631, none blocking.

### Read at

PR #630, opened 2026-08-14
feat: gate GraphQL and iPaaS destructive calls behind confirmation token

adriannoes:
## Summary

Closes the two one-call destruction paths left open by #629.

- **GraphQL:** mutations take the same two-step ticket, queries stay one-shot. Identity is `sha256(document)` + canonical `variables` hash, so changing either invalidates the token. No `destructiveHint` on `execute_graphql`.
- **iPaaS:** judged destructive in fixed order: annotation `true` → `arguments.operation` needle-equality → annotation `false` stops → catalog-name substring. Mixed `operation=DELETE` two-step; `ADD` / `UPDATE` one-shot. Token binds the whole `arguments` map, so a leftover `DELETE` token cannot run an `ADD`. One OAuth handshake per invocation.
- **Too-nested documents refused on both surfaces.** Unclassifiable means it could be an unconfirmed mutation, so nothing is sent. `pipefy graphql exec` exits 2 even with `--yes` (previously an uncaught `RecursionError`). Ceiling is stack headroom, not a depth; recorded in the SDK docstring.
- Preview text names the operation it would run, so two pending mutations stop rendering identical approval text.

Spans `packages/sdk` (document inspection), `packages/mcp` (gates) and `packages/cli` (matching refusal), hence no scope in the title.

**Part 2 of 3.** Stacked on #629; retarget to `dev` when #629 merges.

## Verification

Base is a feature branch, so GitHub runs DCO only. Full `ci.yml` set run locally on this head:

| Gate | Result |
| --- | --- |
| `pytest -m "not integration"` | 4684 passed |
| `ruff check` / `ruff format --check` | clean |
| banned imports (sdk/auth/infra), `lint-imports` (mcp) | pass |
| version lockstep, wheel build, clean-venv install | pass |

## Test plan

- [x] `uv run pytest packages/sdk/tests/test_graphql_document.py packages/mcp/tests/tools/test_introspection_tools.py packages/mcp/tests/tools/test_ipaas_tools.py packages/mcp/tests/core/test_ipaas_gateway.py`
- [x] `uv run pytest packages/cli/tests -k graphql`
- [x] `uv run pytest -m "not integration"` (4684 passed)
- [x] Live smoke: query one-shot; `mutation { __typename }` previews then executes with the returned token; iPaaS destructive call previewed, not confirmed

## Docs / skills

- [x] No docs/skills in this PR. Parity rows, skills and CHANGELOG for both gates land in #631.

## Legal / contributions

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -210,6 +210,10 @@ class IntrospectionTools:
                     ctx,
                     confirm=confirm,
                     resource_descriptor="GraphQL mutation",
+                    irreversible_sentence=(
+                        "⚠️ This GraphQL mutation's effects are permanent "
+                        "and cannot be undone."
+                    ),
                     resource_identity={
                         "document": hashlib.sha256(query.encode("utf-8")).hexdigest(),
                         "variables": hashlib.sha256(
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -190,7 +190,7 @@ class IntrospectionTools:
             Mutations need a preview token: call once to receive ``confirmation_token``,
             then echo that token with ``confirm=True`` on step 2. A token is replayable
             within its TTL, so a non-idempotent mutation can run twice if the caller
-            resends it. Prefer dedicated tools.
+            resends it.
 
             On ambiguous write failure (``success: false`` with empty or unclear message),
             re-read counts/ids before retrying; do not blind-retry creates.
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -2,11 +2,15 @@
 
 from __future__ import annotations
 
+import hashlib
+import json
 from typing import Any
 
 from mcp.server.mcpserver import Context, MCPServer
 from mcp.types import ToolAnnotations
+from pipefy_sdk.graphql_document import document_contains_mutation
 
+from pipefy_mcp.tools.destructive_tool_guard import check_destructive_confirmation
 from pipefy_mcp.tools.graphql_error_helpers import ensure_non_empty_error_message
 from pipefy_mcp.tools.introspection_tool_helpers import (
     build_error_payload,
@@ -175,11 +179,19 @@ class IntrospectionTools:
             ctx: Context,
             variables: dict[str, Any] | None = None,
             include_parsed: bool = False,
+            confirm: bool = False,
+            confirmation_token: str | None = None,
         ) -> dict:
             """Run arbitrary GraphQL against Pipefy (queries or mutations).
 
             Prefer dedicated tools when available. Use this as a fallback when no specific
             tool exists. Always introspect the mutation's input shape before executing.
+
+            Mutations need a preview token: call once to receive ``confirmation_token``,
+            then echo that token with ``confirm=True`` on step 2. A token is replayable
+            within its TTL, so a non-idempotent mutation can run twice if the caller
+            resends it. Prefer dedicated tools.
+
             On ambiguous write failure (``success: false`` with empty or unclear message),
             re-read counts/ids before retrying; do not blind-retry creates.
             Returns ``result`` (pretty-printed JSON string).  Set ``include_parsed=True``
@@ -189,8 +201,30 @@ class IntrospectionTools:
                 query: Full GraphQL document (query or mutation).
                 variables: Optional variable map for the operation.
                 include_parsed: When True, include ``data`` dict alongside ``result``.
+                confirm: Set to True with the preview token to execute a mutation (step 2).
+                confirmation_token: Token from the preview response; echo it on step 2.
             """
             client = get_pipefy_client(ctx)
+            if document_contains_mutation(query):
+                guard = await check_destructive_confirmation(
+                    ctx,
+                    confirm=confirm,
+                    resource_descriptor="GraphQL mutation",
+                    resource_identity={
+                        "document": hashlib.sha256(query.encode("utf-8")).hexdigest(),
+                        "variables": hashlib.sha256(
+                            json.dumps(
+                                variables or {},
+                                sort_keys=True,
+                                separators=(",", ":"),
+                            ).encode("utf-8")
+                        ).hexdigest(),
+                    },
+                    tool_name="execute_graphql",
+                    confirmation_token=confirmation_token,
+                )
+                if guard is not None:
+                    return guard
             try:
                 result = await client.execute_graphql(query, variables)
             except Exception as exc:  # noqa: BLE001
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -210,7 +210,9 @@ class IntrospectionTools:
             inspection = inspect_graphql_document(query)
             if inspection.too_nested:
                 return build_error_payload(
-                    "GraphQL document is too deeply nested to parse."
+                    "This document is too deeply nested to parse, so it cannot "
+                    "be classified as a query or a mutation; nothing was sent. "
+                    "Reduce the nesting and retry."
                 )
             if inspection.contains_mutation:
                 guard = await check_destructive_confirmation(
@@ -218,8 +220,8 @@ class IntrospectionTools:
                     confirm=confirm,
                     resource_descriptor=inspection.mutation_descriptor,
                     irreversible_sentence=(
-                        "⚠️ This GraphQL mutation's effects are permanent "
-                        "and cannot be undone."
+                        f"⚠️ Executing {inspection.mutation_descriptor} is "
+                        "permanent and cannot be undone."
                     ),
                     resource_identity={
                         "document": hashlib.sha256(query.encode("utf-8")).hexdigest(),
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -220,8 +220,9 @@ class IntrospectionTools:
                     confirm=confirm,
                     resource_descriptor=inspection.mutation_descriptor,
                     irreversible_sentence=(
-                        f"⚠️ Executing {inspection.mutation_descriptor} is "
-                        "permanent and cannot be undone."
+                        f"Executing {inspection.mutation_descriptor} needs "
+                        "approval because this server cannot tell what the "
+                        "mutation changes."
                     ),
                     resource_identity={
                         "document": hashlib.sha256(query.encode("utf-8")).hexdigest(),
--- a/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
@@ -8,7 +8,7 @@ from typing import Any
 
 from mcp.server.mcpserver import Context, MCPServer
 from mcp.types import ToolAnnotations
-from pipefy_sdk.graphql_document import document_contains_mutation
+from pipefy_sdk.graphql_document import inspect_graphql_document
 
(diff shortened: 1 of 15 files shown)


packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
```diff
@@ -175,22 +179,65 @@ async def execute_graphql(
...
+                    "be classified as a query or a mutation; nothing was sent. "
+                    "Reduce the nesting and retry."
+                )
+            if inspection.contains_mutation:
+                guard = await check_destructive_confirmation(
+                    ctx,
+                    confirm=confirm,
+                    resource_descriptor=inspection.mutation_descriptor,
+                    irreversible_sentence=(
+                        f"⚠️ Executing {inspection.mutation_descriptor} is "
```
mocha06, line 223: Every mutation preview tells the user the write is permanent, including a create.

How:
1. `inspect_graphql_document` reports only whether a document holds a mutation.
2. It cannot say whether that mutation creates, updates, or deletes.
3. This line writes "is permanent and cannot be undone" for all three.
4. A `createLabel` preview reads `⚠️ Executing GraphQL mutation CreateLabel is permanent and cannot be undone.`
5. Deleting that label undoes it. The sentence is false, so callers learn to skip the warning.

Fix: state what the code knows — the server cannot tell what this mutation changes, so it needs approval. `call_ipaas_tool` already does this. It drops the `⚠️` when the classifier clears the call.

Evidence:
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/introspection_tools.py:222-225 @ f7c2e01` (the sentence, applied to every mutation)
- `ai-toolkit:packages/sdk/src/pipefy_sdk/graphql_document.py:16-19 @ f7c2e01` (`GraphqlDocumentInspection` carries `contains_mutation`, never a kind)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:307-314 @ f7c2e01` (the precedent: no glyph when the classifier clears the call)
- `ai-toolkit:packages/mcp/tests/tools/test_introspection_tools.py:352 @ f7c2e01` (the one assertion the fix also touches)

Probe, in-process server built from this branch:

```
=== CREATE ===
message: ⚠️ Executing GraphQL mutation CreateLabel is permanent and cannot be undone. Show this
preview to the user and get their explicit approval before continuing. ...
resource: GraphQL mutation CreateLabel
```

A replica of the guard with the corrected sentence keeps the token line, the approval line, the proceed path, and the identity-mismatch refusal.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
```diff
@@ -43,6 +46,102 @@
...
+    return name if isinstance(name, str) else ""
+
+
+def _operation_equals_destructive_needle(arguments: dict[str, Any] | None) -> bool:
+    if arguments is None or "operation" not in arguments:
+        return False
+    value = arguments["operation"]
+    if not isinstance(value, str):
+        return False
+    return value.casefold() in IPAAS_DESTRUCTIVE_NEEDLES
```
mocha06, line 100: A delete can run on one call when `operation` carries a trailing space.

How:
1. This line compares `operation` to the needles after `casefold()` and nothing else.
2. It does not strip whitespace, so `"DELETE "` matches no needle.
3. The catalog name `manage_records` carries no needle either, so the name fallback misses.
4. With no `destructiveHint` on the entry, `ipaas_call_is_destructive` returns `False`.
5. The call runs with no preview and no token.

Fix: `value.strip().casefold()`. Stripping normalizes the input, so the frozen needle set stays at six.

Whether the iPaaS host accepts `"DELETE "` decides whether step 5 destroys a record or only errors. I could not check that from here.

Evidence:
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:94-100 @ f7c2e01` (equality after `casefold()`, no `strip()`)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:59-78 @ f7c2e01` (order: needle-equality runs before the annotation-`false` stop, so a miss here falls through to the name)

Probe, `ipaas_call_is_destructive` from this branch:

```
'no hint, operation DELETE'    -> destructive=True
"no hint, operation 'DELETE '" -> destructive=False
'no hint, name has needle'     -> destructive=True
```

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
```diff
@@ -173,10 +280,52 @@ async def call_ipaas_tool(
...
+                confirmation_token: Token from the preview response; echo it
+                    on step 2.
             """
 
             async def work(gateway: IpaasGateway, token: str) -> dict:
-                result = await gateway.call_tool(token, tool_name, arguments)
+                async with gateway.mcp_session(token) as ipaas_session:
+                    catalog = await ipaas_session.list_tools()
+                    entry = _catalog_entry_named(catalog, tool_name)
+                    judged = {"name": tool_name} if entry is None else entry
```
mocha06, line 293: A catalog tool marked destructive can run on one call when the gate cannot find it.

How:
1. `IpaasMcpSession.list_tools` sends `tools/list` with no cursor, so it reads one page.
2. `_catalog_entry_named` returns `None` for a tool outside that page.
3. This line then judges the call from `{"name": tool_name}` alone.
4. The entry's `annotations.destructiveHint` never reaches `ipaas_call_is_destructive`.
5. A destructive tool whose name carries no needle runs with no preview.

Fix: treat a catalog miss as unclassifiable and gate it, the way `execute_graphql` refuses a document it cannot classify.

Whether the host paginates `tools/list` I could not check. `get_ipaas_tools` calls the catalog large.

Evidence:
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:292-294 @ f7c2e01` (a miss becomes `{"name": tool_name}`)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/core/ipaas_gateway.py:110-114 @ f7c2e01` (`tools/list` sent with `{}`, no cursor, one page read)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:137-143 @ f7c2e01` (`_catalog_entry_named` returns `None` on no exact name match)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:195-197 @ f7c2e01` (`get_ipaas_tools` docstring: "a large tool catalog")

Probe, `ipaas_call_is_destructive` from this branch:

```
'catalog MISS (entry not found)' -> destructive=False
'hint true, no operation'        -> destructive=True
```

mocha06 (approved):

Nothing here is broken today. All three findings sit in the new iPaaS and GraphQL gates: one preview sentence that is false for a create, and two classifier paths that let a destructive call through on one round trip. None blocks.

### The decisions you flagged

- **`operation` needle-equality, not substring** — holds. `operation=DELETE` gates, `ADD` does not, and a supplied token routes any call through the guard, so a leftover `DELETE` token cannot run an `ADD`. Two gaps in the *input* to that comparison are findings below.
- **Annotation `false` stops before the name needles** — holds, and that ordering is what makes `manage_records` with `destructiveHint: false` plus `operation=DELETE` two-step.
- **Too-nested refused on both surfaces** — holds. `pipefy graphql exec` exits 2 with and without `--yes`, where `dev` let `RecursionError` escape: `_document_contains_mutation` caught `GraphQLSyntaxError` only.
- **Identity binds document plus variables** — holds, and it is complete. `include_parsed` is the only other argument and it changes no write.
- **One OAuth handshake per invocation** — holds, and the invocation budget survives it. `IpaasMcpSession.call_tool` passes `CALL_TIMEOUT_SECONDS` positionally into `_mcp_request`, which forwards it to `http.post(timeout=...)`, so 120s still overrides the session client's 30s.

### Findings

3, none blocking:

- `packages/mcp/src/pipefy_mcp/tools/introspection_tools.py:223`
- `packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:100`
- `packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:293`

The last two are PLAUSIBLE. Each mechanism is in the code; what I could not read is the iPaaS host's own behaviour. Each comment says which part.

### Read at

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
adriannoes: Fixed in 3bf540d3. A catalog miss is unclassifiable: the call site gates it instead of judging from the name alone, so a hinted tool missing from the first `tools/list` page cannot run one-shot.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
adriannoes: Fixed in 3bf540d3. `operation` is stripped then casefolded before needle equality, and the identity digest uses the same normalisation so `DELETE ` confirms `DELETE`.

packages/mcp/src/pipefy_mcp/tools/introspection_tools.py
adriannoes: Fixed in 3bf540d3. The preview now states what the server knows — it cannot tell what the mutation changes, so it needs approval — and no longer claims the write is permanent.

mocha06: Dropped. Verified on `3bf540d`, not taken on trust.

The preview now reads `Executing GraphQL mutation CreateLabel needs approval because this server cannot tell what the mutation changes.` The `⚠️` is gone and so is the irreversibility claim, which matches the `call_ipaas_tool` precedent.

Probe, in-process server built from this branch:

```
=== execute_graphql CREATE preview ===
has warning glyph: False
says permanent: False
client called: 0
```

`packages/mcp/tests/tools/test_introspection_tools.py` and `test_ipaas_tools.py` pass on this head (151 tests).

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
mocha06: Dropped, and the normalisation is scoped the way it should be.

`_normalized_operation` strips and casefolds `operation` only, in the classifier, in `_arguments_digest`, and in the identity. Every other argument still binds exactly, so the token did not get broader than the call.

Probe at the call site on `3bf540d`:

```
=== 'DELETE ' first call ===
requires_confirmation: True  host called: []
=== confirmed with 'DELETE' (no space), same token ===
success: True  host called: [('manage_records', {'operation': 'DELETE'})]
=== same token, extra argument ===
requires_confirmation: True  host called: []
```

The third case is the one that matters: adding `record_id` to the same token re-previews instead of running.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
mocha06: Dropped. The miss moved to the call site and fails closed.

My original probe called `ipaas_call_is_destructive` directly, which is now the wrong instrument — the classifier no longer sees a miss at all. Probed end to end instead, with a catalog holding only `something_else`:

```
=== catalog miss, first call ===
requires_confirmation: True
host called: []
message: Running iPaaS catalog tool 'manage_records' on pipe 1 ({"operation":"ADD"}) could not be
classified because it was not in the catalog page this server read, so it needs approval.
=== catalog miss, confirmed ===
success: True  host called: [('manage_records', {'operation': 'ADD'})]
```

`operation=ADD` on an unfound tool is the case that used to run one-shot. It previews now, and the preview says why.

mocha06 (dismissed):
Re-approving on 3bf540d. All three findings verified fixed at the call site on this head, not taken on trust: the GraphQL preview no longer claims a mutation is irreversible, `operation` is stripped before needle equality with the normalisation scoped to that one key, and a catalog miss fails closed into the two-step. Details in each thread. The 151 tests in the three files you touched pass locally, which matters here because this PR's base is a feature branch so GitHub only runs DCO.

adriannoes:
CI is green on `dev` after retarget (lint, test, DCO, shell, install/uninstall). Merge is blocked because GitHub now reports `REVIEW_REQUIRED`: the HEAD re-approval on `3bf540d` shows as dismissed (the earlier approval on `f7c2e01` is still listed but does not count for this head).

@mocha06 could you re-approve this head so we can merge-commit into `dev`? No new commits since `3bf540d`.

mocha06 (approved):

PR #631, opened 2026-08-14
docs: document the two-step destructive confirmation and guard it in the registry

adriannoes:
## Summary

- **Drift guard** (`test_destructive_confirm_ticket_drift.py`): walks the live registry, not a hardcoded list, so a destructive tool shipped without the two-step fails the build.
- **Canonical docs**: `docs/mcp/tools/cross-cutting.md` holds the protocol once; restating pages link there. States what it does not buy: no server-enforced human approval, auto-approving clients can chain both calls, tokens replay within the TTL.
- **Late behavior from #630 covered**: too-nested document refused and nothing sent; iPaaS token binds the whole `arguments` map so step 2 resends it unchanged; a supplied token routes even a non-destructive call through the guard; a rejected confirm returns a *fresh* token (a client retrying the old one cannot progress).
- `docs/parity.md` destructive rows + `execute_graphql` / `call_ipaas_tool`; `packages/mcp/AGENTS.md` no longer claims `execute_graphql` has no confirm gate; CHANGELOG records the new preview key and the new refusal.

**Part 3 of 3.** Retarget to `dev` when #630 merges.

## What the drift guard catches

| Break it detects | Fails by name |
| --- | --- |
| Destructive tool missing `confirmation_token` in schema | yes |
| `resource_identity` narrower than the required selectors | yes |
| Tool that writes before consulting the guard | yes |
| Destructively named tool without `destructiveHint` | yes |
| `unpublish_sub_portal` gaining a token | yes |

Each row was checked by mutation, not assumed.

## Verification

Base is a feature branch, so GitHub runs DCO + skills frontmatter only. Full `ci.yml` set run locally on this head:

| Gate | Result |
| --- | --- |
| `pytest -m "not integration"` | 4691 passed |
| `ruff check` / `ruff format --check` | clean |
| banned imports, `lint-imports`, `lint_skill_refs.py` | pass |
| version lockstep, wheel build, clean-venv install | pass |

## Test plan

- [x] `uv run pytest packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py`
- [x] `uv run pytest -m "not integration"` (4691 passed)
- [x] Live smoke on a disposable pipe: one-shot delete previews; two-step delete succeeds; query ungated / mutation gated; iPaaS destructive call previewed not confirmed; `unpublish_sub_portal` exposes no token. Ran against an in-process server built from this branch (editor server was on an older build). Fixtures removed afterwards.

## Docs / skills

- [x] `docs/parity.md` updated
- [x] Affected `skills/` updated in this PR
- [x] Canonical protocol in `docs/mcp/tools/cross-cutting.md`; restating pages link there

## Legal / contributions


docs/parity.md
```diff
@@ -58,36 +60,36 @@ For **database records**, `find_records` result nodes may use **`fields`** while
...
+| `delete_ai_knowledge_base_plain_text` | `pipefy kb plain-text delete` | shipped | Knowledge bases; pipe-scoped (`--id`, `--pipe-uuid`). MCP two-step with `confirmation_token`; CLI `--yes` (or interactive prompt). |
+| `delete_automation` | `pipefy automation delete` | shipped | MCP two-step with `confirmation_token`; CLI `--yes` or interactive prompt. |
+| `delete_card` | `pipefy card delete` | shipped | MCP two-step with `confirmation_token`; CLI `--yes` or interactive prompt. |
+| `delete_card_relation` | `pipefy relation card delete` | shipped | requires OAuth (internal API); MCP two-step with `confirmation_token`; CLI `--yes`. |
+| `delete_comment` | `pipefy card comment delete` | shipped | MCP two-step with `confirmation_token`; CLI `--yes` or interactive prompt. |
+| `delete_field_condition` | `pipefy field-condition delete` | shipped | MCP two-step with `confirmation_token`; CLI `--yes` or interactive prompt. |
+| `delete_label` | `pipefy label delete` | shipped | MCP two-step with `confirmation_token`; CLI `--yes`. |
```
mocha06, line 73: The table says 18 CLI deletes take `--yes` and nothing else. Every one of them also prompts.

How:
1. Each CLI delete calls `confirm_destructive(yes=..., description=...)`.
2. That helper returns early on `yes`, else it calls `typer.confirm` and aborts on no.
3. So `--yes` skips a prompt that is always there.
4. 18 rows read "CLI `--yes`"; 8 rows read "CLI `--yes` or interactive prompt".
5. A reader compares two rows and concludes `pipefy label delete` deletes without asking.

Fix: use one phrasing on all 26 rows, because one helper implements all 26.

Evidence:
- `ai-toolkit:docs/parity.md:73 @ 3453eb2` (`delete_label`: "CLI `--yes`") against `docs/parity.md:69 @ 3453eb2` (`delete_card`: "CLI `--yes` or interactive prompt")
- `ai-toolkit:packages/cli/src/pipefy_cli/commands/_common.py:342-347 @ 3453eb2` (the one helper: prompts unless `yes`)
- `ai-toolkit:packages/cli/src/pipefy_cli/commands/member.py:123`, `webhook.py:136`, `phase.py:262`, `field.py:108` @ `3453eb2` (all route through it)

Probes:

```
$ pipefy label delete 1 < /dev/null
Permanently delete label 1? [y/N]:
Aborted.

$ pipefy member remove --pipe 1 --user-ids 2 < /dev/null
Permanently remove 1 member(s) from pipe 1? [y/N]:
Aborted.
```

mocha06 (approved):

One finding, non-blocking, in `docs/parity.md`. The drift guard is the strongest part of the stack and I could not break it by inspection.

### The rows you flagged

Each row holds, with one narrowing worth recording:

- **Destructive tool missing `confirmation_token`** — caught. The walk keys on `destructiveHint` over the live registry, so a new tool cannot escape by not being listed.
- **`resource_identity` narrower than the required selectors** — caught, and "required" is exact. `_required_identity_keys` reads the schema's `required` list, so an *optional* argument that reaches a mutation is not covered. Only `delete_phase_field.pipe_uuid` sits in that position today, and it is bound. A comment on `PROTOCOL_KEYS` would stop the next optional selector being assumed covered.
- **Tool that writes before consulting the guard** — caught, through `PRE_GUARD_CLIENT_READS`.
- **Destructively named tool without `destructiveHint`** — caught, in any position, not only a prefix.
- **`unpublish_sub_portal` gaining a token** — caught while it stays unhinted. A tool given both a token and `destructiveHint` lands in `skip` instead, which is the correct outcome for a real destructive tool and the reason the row reads as it does.

### One documented rule is stricter than the code

`cross-cutting.md` says step 2 must resend iPaaS `arguments` "byte-for-byte". `_arguments_digest` normalizes key order and casefolds `operation`, so the real rule is looser. It errs safe, so I am not filing it, but the sentence claims more than the code checks.

### Findings

1, non-blocking: `docs/parity.md:73`.

### Read at

packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py
```diff
@@ -0,0 +1,391 @@
...
+                    required_keys,
+                )
+                assert_preview_honors_guard(tool.name, result)
+
+    @pytest.mark.anyio
+    async def test_no_destructive_tool_writes_before_consulting_the_guard(self):
+        """A preview must describe the deletion, never perform part of it.
+
+        The identity walk proves the guard was consulted and its verdict
+        returned. It cannot see a tool that deletes first and previews after,
```
mocha06, line 367: A hosted delete becomes unconfirmable on one pod while working on another, and CI stays green.

How:
1. `signing_key_for` falls back to a per-process random key when the request carries no bearer.
2. That never fires today: every gated tool calls `get_pipefy_client` first, which raises on a missing bearer.
3. Nothing in this guard pins that ordering.
4. Move one guard call above its `get_pipefy_client` and the hosted profile mints process-keyed tokens.
5. Across replicas a preview from pod A never verifies on pod B, reported as `invalid_or_expired`.

Fix: add a sibling check to this one. Walk each gated tool's AST and fail when `check_destructive_confirmation` appears before `get_pipefy_client`.

Evidence:
- `ai-toolkit:packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py:363-368 @ 3453eb2` (this test's docstring names the same class of blind spot: what the identity walk cannot see)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/tools/destructive_tool_guard.py:50-56 @ 3453eb2` (the fallback: `except Exception` returns `_PROCESS_SIGNING_KEY`)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/auth/session_identity.py:98-99 @ 3453eb2` (`RequestScopedIdentity.resolve` raises rather than falling back; this is the only reason the fallback is unreachable)
- `ai-toolkit:packages/mcp/src/pipefy_mcp/core/runtime.py:172 @ 3453eb2` (`session_for_request` resolves eagerly, so the raise lands at `get_pipefy_client`)
- `ai-toolkit:packages/mcp/tests/tools/test_destructive_tool_guard.py:357` and `test_introspection_tools.py:505 @ 3453eb2` (the only two tests touching `signing_key_for` patch it to inject a key; neither asserts which branch the remote profile takes)

Probe, AST walk over the 49 gated call sites on this branch:

```
49 gated call sites; 3 need a closer look
  - ipaas_tools.py:call_ipaas_tool -> NO get_pipefy_client in this function
```

Those three are `call_ipaas_tool` and its two enclosing scopes, and they are fine: `_run_ipaas_tool` resolves the client at `ipaas_tools.py:167` and mints the pipe token at `:169`, both before `work()` reaches the guard. Every other gated tool is client-first in its own body. So the property holds today at all 30 tools, on ordering alone.

mocha06 (commented):
One more finding, non-blocking. It came out of a follow-up question about whether the two-step holds up on the hosted profile. It does — the signing key is `sha256(bearer)`, so any replica verifies the same token with no ticket store. The finding is that the property rests on call ordering that this guard does not pin. The approval stands, but a test may be worth having if not already present (I couldn't find any).

docs/parity.md
adriannoes: Fixed in ee81c1dc. All 26 CLI delete rows (plus `remove_member_from_pipe`) now read `CLI --yes or interactive prompt`, matching the one helper.

packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py
adriannoes: Fixed in ee81c1dc. The drift guard now walks each gated call site and fails if `check_destructive_confirmation` appears before `get_pipefy_client`. `call_ipaas_tool` is allowed only because `_run_ipaas_tool` still resolves the client before `work()`, and that preamble order is pinned too.

docs/parity.md
mocha06: Dropped. Counted on `ee81c1d`: 0 rows read `CLI --yes` alone, and 29 read `CLI --yes or interactive prompt` — the 28 `delete_*` rows at lines 63-90 plus `remove_member_from_pipe` at line 169.

Worth recording that you did **not** sweep `execute_graphql` into the new phrasing. That is correct: `pipefy graphql exec` is the one destructive-ish CLI command not routed through `confirm_destructive`, so it exits 2 and never prompts. Its row keeps `CLI mutations require --yes (exit 2 without)`.

packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py
mocha06: Dropped. I mutation-tested the new check rather than reading it, since the table in your description sets that bar.

Three mutations, three failures, each naming the offender:

```
1. removed `client = get_pipefy_client(ctx)` from delete_webhook
   -> ['webhook_tools.py:delete_webhook:497']
2. moved that resolve to after the guard
   -> ['webhook_tools.py:delete_webhook:497']
3. removed the resolve from _run_ipaas_tool
   -> AssertionError: _run_ipaas_tool never calls get_pipefy_client
```

Mutation 3 confirms the `_run_ipaas_tool` escape hatch is not a hole: the exemption is pinned by `test_ipaas_preamble_resolves_the_client_before_work`. All three reverted; the tree is clean and the 9 drift tests pass.

One limit worth a note, not a change: the check compares line numbers, so a client resolved inside an `if` would satisfy it while leaving the `else` path unresolved. No tool branches that way today, and my proposed check had the same limit.

Thanks for the `PROTOCOL_KEYS` comment on optional selectors — that closes the narrowing I raised in the summary.

mocha06 (dismissed):
Re-approving on ee81c1d. Both findings verified fixed. The parity table now uses one phrasing across all 29 destructive rows, and correctly leaves `execute_graphql` alone since `pipefy graphql exec` never prompts.

I mutation-tested the new ordering guard rather than reading it: three mutations, three failures, each naming the offender, including one that confirms the `_run_ipaas_tool` exemption is pinned rather than assumed. Reverted after; the tree is clean.

You also fixed the `byte-for-byte` overstatement in cross-cutting.md that I raised in the summary but did not file. The replacement is accurate: key order does not matter, `operation` compares after strip and casefold, any other value invalidates.

adriannoes:
Merged `dev` into this branch to clear the conflict after #630. One content conflict in `docs/mcp/tools/pipes-and-cards.md`: kept dest's exclusive-mode wording for `update_card` and this PR's `confirmation_token` contract for `delete_card`. Head is `902851ca`. CI should run on this synchronize; please re-approve so we can merge-commit into `dev`.

mocha06 (approved):

PR #632, opened 2026-08-17
feat: package the repo as a Cursor Marketplace plugin

adriannoes:
## Summary

- Package this repo as a **Cursor plugin** on the hosted MCP (`mcp.pipefy.com`) with in-client OAuth, skill-catalog lockstep, and a CI packaging lint.
- **Manifest declares `"commands": []`**, so Cursor's per-component folder discovery does not ship `/install` or `/pipefy-login`. Verified by loading a repository tarball through `cursor-agent`: neither command reaches the palette, the skills still do.
- **`mcp.json` is validated by allowlist.** ADR-004 pins the whole eight-line file, so the lint permits `mcpServers` → `pipefy` → `{url, auth}` → `{CLIENT_ID}` and rejects the rest, `headers` and `CLIENT_SECRET` and a top-level `inputs` prompt included. Messages name the offending key and never echo its value, since they reach public CI logs (`4ec5013e`).
- **Onboarding skill** routes by observed symptom: an absent catalog listing and a token-endpoint `unauthorized_client` get separate failure rows, each permitting the `install.sh --client cursor` fallback. The `unauthorized_client` row is **regression recovery**, not the current expected state. `uninstall.sh --scan` exit `1` no longer reads as "a competing Cursor entry, delete it".
- **Desktop OAuth gate is green** (2026-08-18). Keycloak made `pipefy-mcp` a public PKCE client (confidential client credentials removed). Dummy token POST as `pipefy-mcp` now returns `400 invalid_grant` (was `401 unauthorized_client`). Live `plugin-pipefy-pipefy` sign-in and `list_organizations` succeeded, and **Q8 passed**: after a full Cursor restart the server came up `ready` with no second prompt. ADR-005 Decision 1 is satisfied; this PR is no longer held by `blocked:platform`. Marketplace listing still submits from `main` after merge, so a catalog 404 until then is expected.

**History.** Rebased onto `dev` and regrouped from seventeen commits into four, one per phase, so the diff reads in order: package, CI, skill, docs. A fifth commit (`4ec5013e`) redacts received `mcp.json` values from public CI diagnostics. The `CHANGELOG.md` conflict with #631 is resolved by keeping this `### Added` block above the `### Changed` block that landed on `dev`. The tree is unchanged by the rebuild apart from that merge, and the add-then-delete of the old PNG logo is gone from the history.

## Accepted limits

| Topic | Choice | Rationale |
|---|---|---|
| Cursor path | Marketplace stays the recommendation | The fallback is a documented failure branch, not the default |
| `commands/` | Suppress in the manifest | Relocating breaks the Claude Code plugin, which needs repo-root `commands/` |
| `mcp.json` | No JSON Schema check in CI | The allowlist already pins every legitimate key |
| Logo | `assets/logo.svg`, vector | Cursor's manifest example uses `assets/logo.svg`, relative paths resolve to `raw.githubusercontent.com` which serves it as `image/svg+xml`, and a vector stays crisp at every catalog size |

## Test plan

- [x] `uv run python scripts/bump_version.py verify` (exit 0)
- [x] `uv run ruff check .` and `uv run ruff format --check .` (431 files)
- [x] `python3 .github/workflows/scripts/lint_cursor_plugin.py` (exit 0)
- [x] `uv run pytest tests/test_lint_cursor_plugin.py tests/test_bump_version.py -q` (99 passed at `4ec5013e`)
- [x] `uv run pytest -m "not integration" -q` (4735 passed, 5 skipped, 34 deselected)
- [x] Nine credential and endpoint mutations of `mcp.json` each exit 1, none echoes the planted value, the unmodified file still exits 0
- [x] `/install` and `/pipefy-login` absent from the palette when a repository tarball loads as a plugin, skills still present
- [x] After the OAuth client is public PKCE: browser sign-in, then `list_organizations` through the plugin (2026-08-18)
- [x] Manual smoke of hosted tools through Cursor MCP (`plugin-pipefy-pipefy` ready; 184 hosted tools plus Cursor-injected `mcp_auth`)
- [x] Q8: session survives a full Cursor restart without a second prompt (`ready` + `list_organizations`, no `mcp_auth`)

## Docs / skills

- [x] Affected `skills/` updated (`skills/onboarding/pipefy-toolkit-setup`)
- [x] README, CHANGELOG, `docs/uninstall.md`, CONTRIBUTING, RELEASE updated
- [x] `RELEASE.md` carries the post-publish step that submits the listing
- [x] `docs/parity.md` not needed (no MCP or CLI surface change)

## Legal / contributions

- [x] No regulated-domain skills in this PR


adriannoes:
Desktop OAuth gate is green.
- Keycloak dropped confidential client credentials on `pipefy-mcp`. 
- Control probe: dummy token POST as `pipefy-mcp` now returns `400 invalid_grant` / `Code not valid` (was `401 unauthorized_client` on 2026-08-17). `pipefy-cli` on the same shape is also `400 invalid_grant`.

Live Cursor plugin (`plugin-pipefy-pipefy`):
- `mcp_auth` completed; server `ready`
- `list_organizations` succeeded
- hosted unauth `POST /mcp` still `401` + RFC 9728
- catalog listed 185 descriptors including Cursor-injected `mcp_auth`; hosted floor remains 184

Removed `blocked:platform`. Marketplace submit still waits for `dev` → `main` (catalog 404 until then is expected).

This maintainer machine still has competing stdio/user-config Pipefy registrations (`uninstall.sh --scan` exit 1). That does not invalidate the plugin handshake.

adriannoes:
Q8 passed after a full Cursor restart (2026-08-18).

Did **not** invoke `mcp_auth`. `plugin-pipefy-pipefy` came up `ready`; `list_organizations` succeeded. Phase 1 desktop gate is complete (Q4/Q7/Q8). Ready for human review. Marketplace submit still waits for `dev` → `main`.

.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,34 @@
...
+    "./skills/ipaas/pipefy-ipaas",
+    "./skills/members-email-webhooks/pipefy-members-email-webhooks",
+    "./skills/observability/pipefy-observability",
+    "./skills/onboarding/pipefy-toolkit-setup",
+    "./skills/pipes-and-cards/pipefy-pipes-and-cards",
+    "./skills/portal-setup/pipefy-portal-setup",
+    "./skills/process-design/pipefy-process-design",
+    "./skills/process-intelligence/pipefy-process-intelligence",
+    "./skills/relations/pipefy-relations",
+    "./skills/reports/pipefy-reports"
```
mocha06, line 31: any way to use a wildcard so this catalog can increase dynamically without us having to remember to update this file when we add a new skill?

README.md
```diff
@@ -142,9 +144,29 @@ npx skills add pipefy/ai-toolkit --skill pipefy-pipes-and-cards
 
 Catalog and authoring guide: [`skills/README.md`](skills/README.md).
 
+### 6. Cursor Marketplace plugin
+
+**Pick this when:** you're in Cursor and want the hosted MCP server with browser sign-in and no local Python. The plugin ships the skill catalog and points Cursor at `https://mcp.pipefy.com/mcp`. Cursor runs the OAuth flow. The surface is the hosted deployment's remote-safe floor: **184** tools as of 2026-08-17 (Cloud Agents against production). Withheld are the tools whose input is a file on your machine; attachment uploads still work from a URL or a presigned upload target. Per-client toolset selection (`PIPEFY_MCP_TOOLSETS`) does not apply to this URL — use the [Quick-install script](#3-quick-install-script) when you need that, or the local-file tools.
```
mocha06: please remove tool counts like _**184** tools_, for example. This will grow stale quickly and one more place where we'd have to remember to update once we add more tools.

(this "counts in docs" rule should apply as a general rule)

README.md
```diff
@@ -186,7 +208,7 @@ Deprecation and semver (post-1.0): [`docs/DEPRECATION.md`](docs/DEPRECATION.md).
 
 ## MCP server
 
-The server registers **187 tools** across fourteen domains. Canonical names: `PIPEFY_TOOL_NAMES` in [`packages/mcp/src/pipefy_mcp/tools/registry.py`](packages/mcp/src/pipefy_mcp/tools/registry.py).
+The server registers **187 tools** across fourteen domains locally. Canonical names: `PIPEFY_TOOL_NAMES` in [`packages/mcp/src/pipefy_mcp/tools/registry.py`](packages/mcp/src/pipefy_mcp/tools/registry.py). The Cursor Marketplace plugin and Hosted MCP path expose the hosted remote-safe floor instead: **184** tools as of 2026-08-17 (the three-tool gap is the local-file tools that URL withholds).
```
mocha06: same as above. remove counts

README.md
```diff
@@ -219,7 +241,7 @@ Not every client wants every tool. Three independent controls decide what `tools
...
 | **Power discovery** | `--toolsets power` | Replaces the curated tools with four catalog meta-tools (`get_tool_categories`, `search_tools`, `describe_tool`, `execute_tool`) plus the raw-GraphQL tools, so the working set stays small no matter how large the catalog grows. |
 
-The toolset names are a **different grouping from the table above**: that table is organized by documentation area (the reference docs you read), while subject domains partition tools by the job they serve — card relations land in `workflow`, table relations in `database`. Passing an unrecognized name is a startup error that prints the full list of valid ones.
+The toolset names are a **different grouping from the table above**: that table is organized by documentation area (the reference docs you read), while subject domains partition tools by the job they serve — card relations land in `workflow`, table relations in `database`. Passing an unrecognized name is a startup error that prints the full list of valid ones. `--toolsets` / `PIPEFY_MCP_TOOLSETS` is a process-level switch: it applies to the local stdio server only, not to the Cursor Marketplace plugin or Hosted MCP URL.
```
mocha06: is it necessary to mention Cursor here specifically? Because this would also apply to claude desktop/codex/etc etc. (This question also applies to other places in this PR)

RELEASE.md
```diff
@@ -105,6 +105,8 @@ It reads the current version and `## [Unreleased]` from `origin/dev` (not your c
 
    Once the workflow finishes it asserts, and fails loudly on any gap, that the GitHub Release ships all five wheels (`pipefy-*.whl`, `pipefy_mcp_server-*.whl`, `pipefy_cli-*.whl`, `pipefy_auth-*.whl`, `pipefy_infra-*.whl`), that the published version installs from PyPI (`uvx --from "pipefy-cli==<PEP 440>" pipefy --version`), and that the `install.sh` dry-run resolves the just-cut tag (`Resolved tag: vX.Y.Z`). Re-run those checks any time with `uv run python scripts/release.py verify vX.Y.Z`.
 
+5. Submit or refresh the Cursor Marketplace listing, from `main`, once the tag is published. The listing is not cut by any workflow: `.cursor-plugin/plugin.json` and `mcp.json` ship with the repository, and a human submits them at [cursor.com/marketplace/publish](https://cursor.com/marketplace/publish). Until that submission lands, README section 6, `docs/uninstall.md`, and the onboarding skill all say the listing publishes from `main`, so this step is what makes those sentences true. Cursor reviews every plugin by hand and re-reviews each update, so budget for a review round rather than treating the submission as instant.
```
mocha06: `Until that submission lands, README section 6....`

are we going to remember to update this doc 😅 

this can easily go forgotten and stale

mocha06 (changes requested):

.cursor-plugin/plugin.json
adriannoes: No glob in the Cursor manifest — `skills` is a string or an array of paths, not a wildcard. Default discovery only looks at immediate children of `skills/` that contain `SKILL.md`. Our catalog is `skills/<domain>/<skill>/`, so that would miss every published skill, and `"skills": "./skills/"` would also risk picking up `skills/_template/`.

The explicit list is the same override pattern as `"commands": []`. We do not rely on remembering to edit it: `lint_cursor_plugin.py` already diffs the array against `git ls-files 'skills/**/SKILL.md'` (minus the template) and fails CI in both directions. CONTRIBUTING.md calls that out when adding a skill.

Leaving the array as-is.

README.md
adriannoes: Dropped the headline counts (and the as-of date) across the README, not only this line — overview, install table, §6, and the MCP section. Named the contract instead: local catalog vs hosted remote-safe floor. Also removed the per-domain Tools column so those numbers do not go stale the same way. Left `docs/parity.md` / `packages/mcp/README.md` for a follow-up; they were not introduced here.

Addressed in e07f7346.

README.md
adriannoes: Same pass as above — this paragraph no longer names a count, and the per-domain Tools column is gone.

Addressed in e07f7346.

RELEASE.md
adriannoes: Rewrote step 5 as a recurring operator action (submit/refresh from `main` after every tag that touches `.cursor-plugin/` or `mcp.json`) with no "until that lands" clause. Same class of sentence removed from README §6 and `docs/uninstall.md`.

Addressed in e07f7346.

README.md
adriannoes: Agreed — `PIPEFY_MCP_TOOLSETS` is a local stdio process switch, not a Cursor quirk. It does not apply to the hosted URL for any client. Reworded the three README sites that tied toolsets to the Marketplace plugin; Cursor stays named where the subject is the plugin path itself (Marketplace, `~/.cursor/mcp.json`).

Addressed in e07f7346.

.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,34 @@
...
+  "author": { "name": "Pipefy" },
+  "homepage": "https://github.com/pipefy/ai-toolkit",
+  "repository": "https://github.com/pipefy/ai-toolkit",
+  "license": "Apache-2.0",
+  "logo": "assets/logo.svg",
+  "keywords": ["pipefy", "mcp", "workflow", "kanban", "automation"],
+  "category": "productivity",
+  "tags": ["mcp", "workflow", "skills"],
+  "skills": [
+    "./skills/ai-agents/pipefy-ai-agents",
```
mocha06, line 15: A Cursor user asks to add a PDF knowledge base, and the tool the skill names is not on the server.

How:
1. This plugin points Cursor at the hosted server, which serves 184 of 187 tools.
2. It withholds three: `create_ai_knowledge_base_document`, `create_llm_provider`, `update_llm_provider`.
3. The `pipefy-ai-agents` skill on this line tells the agent to call all three.
4. That skill marks no tool local-only, and line 34 says its tools are "all remote-safe".
5. `lint_skill_refs.py` checks skill tool names against all 187, so CI stays green.

Every other channel that ships skills runs the local server, so this is the first pairing of the catalog with the hosted floor.

Fix: mark those three rows the way `pipefy-pipes-and-cards` marks `upload_attachment_to_card` — "local profile only … required on the hosted server".

Evidence:
- Probe on this branch: `apply_remote_profile(remote_mode=True)` → `Remote profile: exposed 184, withheld 3` — `create_ai_knowledge_base_document`, `create_llm_provider`, `update_llm_provider`.
- Live `mcp.pipefy.com` `tools/list` omits exactly those three; it serves `get_ai_knowledge_base_document`, `update_ai_knowledge_base_document`, `delete_llm_provider`, `get_llm_providers` (so the gap is per-tool, not per-domain).
- `skills/ai-agents/pipefy-ai-agents/SKILL.md:169` @ `e07f7346` — `create_llm_provider` then `update_llm_provider`, no profile note.
- `skills/ai-agents/pipefy-ai-agents/SKILL.md:225` and `:240` @ `e07f7346` — `create_ai_knowledge_base_document` with `file_path`, no profile note.
- `skills/ai-agents/pipefy-ai-agents/SKILL.md:34` @ `e07f7346` — "all remote-safe: available under the hosted (`profile=remote`) surface".
- `skills/pipes-and-cards/pipefy-pipes-and-cards/SKILL.md:167` and `skills/database-tables/pipefy-database-tables/SKILL.md:66` @ `e07f7346` — the note pattern that already exists.
- `packages/mcp/src/pipefy_mcp/tools/remote_profile.py` @ `e07f7346` — default-deny, `meta=REMOTE` is the only opt-in.
- `.github/workflows/scripts/lint_skill_refs.py:92` @ `e07f7346` — reads `PIPEFY_TOOL_NAMES` from `registry.py`, the full 187.

README.md
```diff
@@ -44,23 +44,25 @@ Feedback and issues: [GitHub Issues](https://github.com/pipefy/ai-toolkit/issues
...
 - **In Claude Code and want the fastest start with no local setup?** → **Hosted MCP**.
 - **In Claude Code and want the CLI, `/pipefy:*` slash commands, or the few local-only tools?** → **Claude Code plugin**.
-- **On Cursor, Claude Desktop, or Codex — or want one command for everything?** → **Quick-install script**.
+- **On Cursor, Claude Desktop, or Codex and need the local-file tools, the CLI, or one command for everything?** → **Quick-install script**.
 - **Terminal, scripting, or CI, with no agent?** → **CLI only**.
 - **Just want the workflow playbooks in any agent?** → **Skills only**.
 
 | Install path | MCP server runs on | Tools available | Auth | Also installs | Best for |
 |---|---|---|---|---|---|
+| **[Cursor Marketplace plugin](#6-cursor-marketplace-plugin)** | Pipefy cloud (HTTPS) | Remote-safe surface; local-file tools withheld | In-client OAuth | nothing else | Fastest start in Cursor; zero local Python |
```
mocha06: Someone picking a path from this table reads that the Cursor plugin installs nothing else. It installs the whole skill catalog.

How:
1. The manifest's `skills` array ships every published skill.
2. Section 6 says so too: "The plugin ships the skill catalog".
3. This cell says `nothing else`.
4. The same column writes `skills` for the Claude Code plugin row and `markdown playbooks` for Skills only, so the reader takes `nothing else` literally.

Fix: write `skills` in that cell. The docs pass in `e07f7346` rewrote this row's Tools cell and left this one.

Evidence:
- `.cursor-plugin/plugin.json:14`–`32` @ `e07f7346` — 17 skill directories.
- `README.md:149` @ `e07f7346` — "The plugin ships the skill catalog".
- `README.md:60` and `README.md:61` @ `e07f7346` — the same column naming `skills` on the two rows that install them.

.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,34 @@
...
+  "version": "0.5.0-alpha.1",
+  "author": { "name": "Pipefy" },
+  "homepage": "https://github.com/pipefy/ai-toolkit",
+  "repository": "https://github.com/pipefy/ai-toolkit",
+  "license": "Apache-2.0",
+  "logo": "assets/logo.svg",
+  "keywords": ["pipefy", "mcp", "workflow", "kanban", "automation"],
+  "category": "productivity",
+  "tags": ["mcp", "workflow", "skills"],
+  "skills": [
```
mocha06, line 14: If Cursor discovers skill folders the way it discovers `commands/`, `pipefy-skill-template` reaches a user's palette as a Pipefy skill.

Does the explicit `skills` array suppress folder discovery under `skills/`, the way `"commands": []` suppresses it under `commands/`?

I could not answer this from the code in pipefy-workspace. The lint deliberately excludes `skills/_template/` from the manifest, so the intent is clear, but the tarball and the README copy command both carry that directory. In Claude Code the template does surface as a loadable skill, which is why the question is not theoretical. Your `cursor-agent` tarball run already proves the listed skills load — the open half is whether the unlisted one loads too.

mocha06 (commented):

Read at `e07f7346`. The packaging, the version lockstep, and the `mcp.json` allowlist hold up. What does not is the pairing this PR creates for the first time: the skill catalog served by the hosted floor. Every earlier channel that ships skills runs the local server, so no skill ever had to say which of its tools the hosted URL withholds — and one skill names all three of them.

### Verdicts

- **Premise: CONFIRMED.** Before this PR the only Cursor path was `install.sh` (local Python); `#1 Hosted MCP` is Claude Code only. No `.cursor-plugin` exists on `dev`. Nothing here reimplements an existing mechanism.
- **Mechanism: DELIVERS.** The lint's source of truth is sound — `git ls-files -- 'skills/**/SKILL.md'` returns 18 of 18 tracked files, and `skills/_template/` is dropped downstream. The `mcp.json` allowlist rejects `headers`, `env`, `command`, `CLIENT_SECRET`, and a top-level `inputs`, and no message echoes a received value.

### Findings

2 findings, 1 blocking.

- Blocking — `.cursor-plugin/plugin.json:15`: the shipped `pipefy-ai-agents` skill names the three tools the hosted server withholds.
- Non-blocking — `README.md:58`: the install table says this path installs `nothing else`.

1 question, on `.cursor-plugin/plugin.json:14`.

Not in the diff, so no inline anchor: `skills/README.md:47` still describes the onboarding skill as "First-time install: hosted MCP, `install.sh`, or Claude Code plugin". That skill's own frontmatter gained the Cursor path here; its row in the index did not.

### Read at

- `ai-toolkit` @ `e07f7346`, base `origin/dev`. The head moved from `4ec5013e` mid-review; every citation below was re-checked against `e07f7346`, and both findings survive that commit.
- No sibling repository is called by this change, so none was read.
- Probes: the hosted floor computed in-process from this worktree, and `mcp.pipefy.com` `tools/list` read read-only. No test suite, linter, or build was run — the pipeline is green by assumption.

.cursor-plugin/plugin.json
adriannoes: Agreed — this pairing is new, and the skill was written for the local catalog. Marked the three withheld tools local-profile-only at the point of use: the agent-tools paragraph (those seven stay remote-safe; the later file tools do not), the BYOM `create_llm_provider` / `update_llm_provider` note, the `create_ai_knowledge_base_document` table row, and the PDF step (hosted callers create plain text, or use the CLI / Quick-install path).

Addressed in f1549dad.

README.md
adriannoes: The cell now says `skills`. Hosted MCP (Claude Code) still says `nothing else` — that path does not ship the catalog. The onboarding row in `skills/README.md` now lists the Cursor Marketplace plugin as well.

Addressed in a5cabb4e.

.cursor-plugin/plugin.json
adriannoes: Yes. Cursor's plugin docs: if the manifest sets `skills`, that field **replaces** folder discovery — the default `skills/` scan does not also run ([component discovery](https://cursor.com/docs/reference/plugins#cursor-plugin-component-discovery)). Same mechanism as `"commands": []` suppressing `commands/`.

So `pipefy-skill-template` should not reach the palette even though the tarball and the README `cp -R skills` copy still include `skills/_template/` on disk. The earlier `cursor-agent` run only proved listed skills load; it did not list the template. Happy to run a second palette check for `pipefy-skill-template` if you want that confirmed empirically.

Leaving the explicit array as-is.

.cursor-plugin/plugin.json
mocha06: Verified at `a5cabb4e`. Dropping this.

Those four sites are the only four in the skill tree that name the three tools — `grep -rn` over `skills/` returns `SKILL.md:34`, `:169`, `:225`, `:240` and nothing else — so the coverage is complete, not partial.

The hosted alternatives you added also check out: `get_llm_providers` and `create_ai_knowledge_base_plain_text` are both on the hosted floor, and `providerId` / `systemProviderId` ride on `create_ai_agent` / `update_ai_agent`, which are too. `systemProviderId` is real (`packages/mcp/src/pipefy_mcp/tools/ai_agent_tools.py:256` @ `a5cabb4e`), and "no hosted-safe source" for the PDF row is right — unlike the attachment tools, there is no `file_url` variant, and `update_ai_knowledge_base_document` is metadata-only.

README.md
mocha06: Verified at `a5cabb4e`. Dropping this.

The cell reads `skills`, and `skills/README.md:47` now lists the Cursor Marketplace plugin. Hosted MCP keeping `nothing else` is correct — that path registers a URL and ships no catalog.

.cursor-plugin/plugin.json
mocha06: Answered. I read the page you cited, and this sentence settles it:

> If a manifest field **is** specified (e.g., `"skills": "./my-skills/"`), it **replaces** folder discovery for that component.

The array is specified, so the `skills/` scan does not also run and `skills/_template/` stays out of the palette. No empirical check needed — thanks for the offer.

The same page lists `rules/` as a default discovery folder. This repository has no root `rules/` directory today, so nothing is picked up there; worth knowing if one is ever added for the Claude Code side, since `commands: []` only suppresses `commands/`.

.cursor-plugin/plugin.json
mocha06: Checked this against Cursor's docs — the wildcard part holds. The field is typed "string or array" of "Path(s) to skill directories", and no glob syntax is documented for any path field. There is nothing to wildcard with. The lint does fail both ways too, so nobody can forget. Keeping the explicit array.

One premise is thinner than it reads: the page never states discovery depth. "Each subdirectory containing a `SKILL.md` file" is all it says, so immediate-children-only is an inference, not documented. It does not change the conclusion — the `_template` objection rules out the directory form on its own.

Where I would still push, and it is not really about wildcards.

`_lint_skill_set` demands **exact set equality** in both directions, so the array carries no information the tree does not already have. It is a hand-maintained mirror of the filesystem. That is the thing worth removing, not the edit cost.

**The one I would actually do: move `skills/_template/` out of `skills/`.** That directory being inside the catalog is already special-cased in four separate CI code paths — `lint_cursor_plugin.py:14` (`SKILL_TEMPLATE_PREFIX`), `lint_skill_refs.py:108` (`"_template" not in p.parts`), and two `os.walk` prunes in `skills-lint.yml` (`:42-44`, `:91-92`) — plus a test and three docs that explain the exclusion. Moving it deletes all four rules, and that payoff does not depend on Cursor at all.

It also settles this thread either way. With `_template` gone, the only remaining objection to `"skills": "./skills/"` is the discovery depth, which your `cursor-agent` rig can answer in one run. Recursive, and the array disappears. Immediate-children-only, and the array is forced and this is closed for good.

Two things I floated earlier and am withdrawing, so they do not cost you time:

- **A `--fix` mode on the lint.** The manifest is hand-formatted (`"author": { "name": "Pipefy" }` on one line), so a faithful rewrite needs a surgical splice rather than `json.dump`. That is real code plus tests to automate pasting one line into a sorted array a few times a year, and it puts a second writer on a manifest `bump_version.py` already writes. Not worth it.
- **A pre-commit hook.** Weaker than I made it sound. It needs a `files:` regex or it runs on every commit, and pre-commit does not fire reliably on deletions — so it would catch adding a skill but not removing one, for an event CI already catches in minutes.

Evidence:
- [Cursor plugin component discovery](https://cursor.com/docs/reference/plugins#cursor-plugin-component-discovery) — field type "string or array"; no glob or wildcard syntax on any path field; no documented skipping of `_`- or `.`-prefixed directories, which is what would ship `skills/_template/`.
- `.github/workflows/scripts/lint_cursor_plugin.py` `_lint_skill_set` @ `a5cabb4e` — errors on both `tree - declared` and `declared - tree`, which is what makes the array a pure mirror.
- The four exclusion sites @ `a5cabb4e`: `lint_cursor_plugin.py:14`, `lint_skill_refs.py:108`, `skills-lint.yml:42-44`, `skills-lint.yml:91-92`.
- `git ls-files -- 'skills/**/SKILL.md'` returns 18 of 18 tracked files, so the lint's source of truth cannot under-match.

adriannoes: Doing 1 and 2 in this PR: `d53a7eef` adds `--fix` (rewrites only the `skills` array from `git ls-files -- 'skills/**/SKILL.md'` minus `_template`) and a local `cursor-plugin` pre-commit hook that runs that with `--fix`. CI still runs the lint with no flag. That is the answer to "I still hand-edit JSON" — the hook rewrites the array at commit, and the author re-stages like ruff.

Deferring 3. Cursor-agent discovery depth is still undocumented. `"skills": "./skills/"` is already ruled out while `skills/_template/` lives under `skills/` — a recursive walk would ship the template as a loadable skill. Moving `_template` is a catalog layout change that also hits Claude Code (it currently surfaces there as a loadable skill). Not needed to close the original "don't remember to update the file" ask once `--fix` + pre-commit land; we can revisit directory-form discovery later if we want it.

adriannoes: Thanks for the edit — we had already landed `--fix` plus the local `cursor-plugin` pre-commit hook from the earlier wording (`d53a7eef`). Reverted that in `74f68a2f` because those two are not worth a second writer on the same hand-formatted JSON, or a hook that is weaker than it sounded.

Did the remaining ask in `b4ba26f8`: moved `skills/_template/` to `.github/skill-template/` (copyable `SKILL.md` kept) and dropped the four `_template` exclusion sites (`lint_cursor_plugin.py`, `lint_skill_refs.py`, and both `os.walk` prunes in `skills-lint.yml`).

Left the explicit `skills` array in `.cursor-plugin/plugin.json`. Cursor discovery depth is still undocumented here, so we are not switching to `"./skills/"` without evidence that discovery is recursive.

mocha06: Verified `b4ba26f8` rather than taking it from the summary. The part that could have gone wrong did not: dropping the four prunes is safe because both linters are rooted at `skills/` only — `os.walk("skills")` at `skills-lint.yml:41` and `:86`, and `skills_root = REPO_ROOT / "skills"` at `lint_skill_refs.py:148`. So `.github/skill-template/` is out of scope by location now, not by exclusion, which is why deleting the rules is a simplification and not a hole.

Also checked: all seven relative links resolve (`skills/` → `../.github/…`, and `../../…` back out of the template), no stale `skills/_template` path survives, `lint_cursor_plugin.py` exits 0, and tracked skills are 17 against 17 manifest entries.

Your Claude Code note checks out from the other side. `pipefy-skill-template` was in my loaded skill list from this repo before the move, so it was a real leak rather than a theoretical one.

The `--fix` detour is on me — my first wording ranked it as the answer when it was the weakest of the three. Sorry for the build-and-revert.

One thing left, and it is one line. `.github/skill-template/README.md:5` bills itself as "Before opening a PR here" but names only `lint_skill_refs.py`. This PR adds a second hard requirement, the `.cursor-plugin/plugin.json` skills array, enforced in both directions. `CONTRIBUTING.md` step 4 covers it; this checklist does not. So someone authoring their first skill runs the one command named, passes, pushes, and CI fails on the other lint — the exact failure step 4 exists to prevent. Adding `python3 .github/workflows/scripts/lint_cursor_plugin.py` to that line, plus a clause that a new skill needs a manifest entry, closes it.

Deferring the discovery-depth question is fine by me. The array is honest and CI-enforced either way, so nothing rides on it now.

adriannoes: Added. `.github/skill-template/README.md` step 5 now names the manifest `skills` array and both lints (`lint_skill_refs.py` and `lint_cursor_plugin.py`), matching CONTRIBUTING step 4. `ea8b6018`.

mocha06: Verified at `ea8b6018`. Step 5 now names the manifest edit and both lints, and the invocation forms match CONTRIBUTING step 4 and CI (`uv run python` for `lint_skill_refs.py`, `python3` for `lint_cursor_plugin.py`). Both pass on this head, 17 skill files against 17 manifest entries.

That closes everything I raised here. Resolving.

mocha06 (approved):

adriannoes:
Thanks @mocha06 — the review caught the hosted-skill pairing, the install-table lie, and the template living inside the catalog. Appreciate the probes and the patience through the `--fix` detour. Merging to `dev`.

PR #626, opened 2026-08-12
chore: back-merge main into dev (3 commits)

mocha06:
> **Use "Create a merge commit".** A squash or a rebase discards the second parent, so `dev` gains the *content* of these commits but never their *ancestry*. `git rev-list --count origin/dev..origin/main` would still report 3, `release-pr` would still refuse to cut, and the next run of the workflow would open this same pull request again.

## What this is

`.github/workflows/backmerge.yml` built and pushed this branch on the push of `7052d6b`. It could not open the pull request, so this one was opened by hand off the branch the workflow produced. The merge commit is the workflow's, untouched:

```
33c7583 parents=2655d12 7052d6b
chore: back-merge main into dev (3 commits)
Signed-off-by: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
```

## Why the workflow could not open it

```
pull request create failed: GraphQL: GitHub Actions is not permitted to
create or approve pull requests (createPullRequest)
```

**Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests"** is off. Everything upstream of that call worked: the drift count, the branch, the clean merge, the sign-off, and the fallback to the tracking issue in #611.

The earlier run on `e428409` failed differently — `HTTP 502` from the GraphQL API, transient — and the script attributed it to this same setting without having checked. The guess was right the second time and wrong the first. Reporting `gh`'s actual stderr instead of a presumed cause belongs in #610.

## Contents

Three commits from `main`, of which exactly one changes a file against `dev`:

```
$ git diff --stat origin/dev origin/rc-dev/chore/back-merge-main-7052d6b
 AGENTS.md | 1 +
```

That line is the DCO guideline from #603. The other two commits are #605 and its merge, whose files were copied from `dev` verbatim and are byte-identical, so they carry no delta.

No duplicated content: `skills/AGENTS.md` holds one copy of each paragraph. This merge is clean because #606 restored the ancestry — before that, the same merge duplicated two paragraphs.

## Checks

None. The branch was pushed by `GITHUB_TOKEN`, and GitHub does not trigger `on: pull_request` for its events. Push any commit to the branch, or close and reopen, to run them. A `BACKMERGE_TOKEN` secret makes checks run on their own and also lifts the restriction above.

Refs #611, #610.

Issue #610, opened 2026-08-12
backmerge: pin the idempotency key, supersede the stale tracking issue, and fix two labels


mocha06:
## Summary

Four follow-ups raised in the review of #605, all in the back-merge automation from #576. None blocks anything today. Two are correctness, two are labels that read as more certain than they are.

They are grouped because they touch the same three files, and those three files must stay **byte-identical on `main` and `dev`**. A change to any of them on one branch alone makes the next back-merge conflict inside `backmerge.py` — the script that performs back-merges. So all four land here on `dev`, together, and reach `main` through a back-merge or a release promotion. Never directly on `main`.

## 1. `main_head()` keys everything off an unpinned short SHA

`.github/workflows/scripts/backmerge.py:113-115`

```python
def main_head() -> str:
    """The short SHA of ``origin/main``."""
    return capture(["git", "rev-parse", "--short", f"origin/{MAIN_BRANCH}"])
```

`--short` with no length uses `core.abbrev=auto`, which grows with the object count. That value keys four things: the branch name via `branch_for()` (`:118-120`), the `existing_pr` lookup (`:527`), the remote-branch check, and `state_marker()` (`:261-264`).

If the abbreviation lengthens between two runs, the same `main` head produces a different branch name. The `existing_pr` lookup then misses the open pull request and the run treats an already-handled tip as new.

**Current risk is low, not zero.** The abbreviation sits at git's floor of 7 characters against roughly 16k objects, so a step to 8 needs substantial growth. The worst case is a duplicate pull request, not a bad merge.

**Fix:** pin an explicit length, for example `--short=12`.

## 2. The OPEN early return skips `supersede_tracking_issue`

`.github/workflows/scripts/backmerge.py:527-530`

```python
pr, pr_state = existing_pr(branch)
if pr and pr_state == "OPEN":
    print(f"Pull request #{pr} already covers {sha}; leaving it alone.")
    return 0
```

`supersede_tracking_issue()` runs at `:590` and `:631`, both create paths. This early return reaches neither.

Reachable sequence:

1. A run hits a merge conflict and files the tracking issue. No pull request is opened.
2. Someone resolves it by hand, pushes the branch, and opens the pull request.
3. The next run finds an OPEN pull request for that head and returns at `:530`. The tracking issue from step 1 stays open.

That is the state the function's own docstring says it exists to prevent:

[...]

PR #603, opened 2026-08-11
docs: require commit sign-off (DCO) in commit guidelines

gbrlcustodio:
Adds a bullet to the AGENTS.md commit guidelines: sign off every commit with `git commit -s`, because CI enforces the Developer Certificate of Origin (DCO) and an unsigned commit fails the check.

PR #633, opened 2026-08-18
feat(backmerge): repair a squashed back-merge, and name expected DCO failures

mocha06:
## Motivation

Two things happen to a back-merge pull request after it leaves the workflow's hands, and neither was handled.

**A squash or a rebase strands it.** Both apply every file and discard the second parent, so `dev` gains the content and never the ancestry. `git rev-list --count origin/dev..origin/main` cannot reach zero from there, `release-pr` stays blocked, and the workflow re-proposed the same pull request on every run. This is not hypothetical — #577 and #570 were both squash-merged, and the count sat at 9 for eleven days with every file already in place until #606 repaired it by hand.

**A back-merge fails DCO on history nobody can rewrite.** #626 is red right now on `7052d6b`:

```
sign-off not found for commit author: Gabriel Custódio gabriel.custodio@pipefy.com
found: ['Gabriel <gabriel.custodio@pipefy.com>']
```

Same email, different name. That commit passed DCO on its own pull request (#603). The squash then rewrote the author to the merging account's GitHub profile identity while the `Signed-off-by` trailer, being message text, was carried verbatim. Once it is on the default branch it cannot be rewritten, so the check fails on every back-merge that carries it from then on.

### The workflow repairs a squashed back-merge

When `existing_pr` reports the pull request for this `main` head is `MERGED` while the divergence still stands, `restore_ancestry` cuts `rc-dev/chore/back-merge-main-<short-sha>-ancestry` from `dev`, merges `main` with `-s ours`, and opens it.

`-s ours` is correct here for one reason: the squash already delivered the files, so `dev`'s tree is current and ancestry is the sole thing missing. A plain merge is the wrong tool — with no shared ancestry, git reads one paragraph on each side as two independent additions and keeps both. That is observable, not theoretical: a 3-way merge during the #606 repair duplicated two paragraphs in `skills/AGENTS.md`.

Every idempotency rule that guards the ordinary back-merge guards the repair. The branch is keyed to the same `main` head, an open repair is left alone, and the branch is never pushed over.

**Where it stops.** Squashing the repair in turn leaves the count exactly where it was. Opening a third pull request for one `main` head would loop, so the automation stops there and records it with the commands to finish by hand. Nothing short of disabling squash removes the human step; this makes each iteration one click instead of a manual git session.

### The pull request body names expected DCO failures

`dco_mismatches()` lists carried commits whose author disagrees with their own sign-off, and `dco_section()` renders the explanation. Run against the live drift it finds exactly the commit the check actually failed on, and nothing else:

```
$ python3 -c '...; print(m.dco_mismatches())'
drift commits: 3
mismatches found: 1
  `7052d6b` author `Gabriel Custódio <...>`, signed off by `Gabriel <...>`
```

Merge commits are skipped, because `dco-check` runs with `--no-merges`.

## Deliberately not done: skipping DCO for back-merge branches

The obvious fix is a `dco.yml` condition on the head branch matching `rc-dev/chore/back-merge-main-*`. That is not a guard. `head_ref` on a fork pull request is the branch name in the **contributor's own repo**, so anyone could open a pull request from a branch with that name and skip the provenance check. This repository is public and invites outside contributions, and DCO is the one check that cannot have a name-matched bypass.

If the red ever needs to go away, the condition has to be something an outsider cannot satisfy — the pull request author being the bot identity, or a `back-merge` **label**, which needs triage access to apply. Never the branch name.

I also checked whether `dco-check` can simply be told to ignore commits already on `main`, which would sidestep the question. It cannot: `--exclude-emails` and `--exclude-pattern` both match author email only (`dco_check.py:1043,1049-1050`), and `--default-branch` affects only the `GitRetriever` path, while on GitHub CI the `GitHubRetriever` wins and uses the event's `base..head`.

## Verification

| # | Scenario | Result |
|---|---|---|
| 1 | `dco_mismatches()` against the live `origin/dev..origin/main` | Finds `7052d6b` alone — the commit #626's DCO actually failed on |
| 2 | Matching author and sign-off | Not reported |
| 3 | Missing sign-off | Reported separately from a mismatch |
| 4 | `--no-merges` in the argv | Pinned by test |
| 5 | Squashed back-merge, no repair yet | Builds, pushes, and opens the repair; never calls `attempt_merge` or `open_pr` |
| 6 | Repair already open | Nothing rebuilt, nothing tracked |
| 7 | Repair itself squashed | Tracks the loop, opens no third pull request |
| 8 | Ancestry merge fails | Pushes nothing, tracks the reason |
| 9 | `-s ours` in the merge argv | Pinned by test — dropping the strategy silently turns it into the merge it must not be |
| 10 | Repair branch name distinct from `branch_for` | Asserted, or the repair would collide with the merged pull request and read as handled |
| 11 | Recovery body demands a merge commit | Asserted |

Gates: `pytest -m "not integration"` (4707 passed, 5 skipped), `ruff check`, `ruff format --check`, `lint_skill_refs.py`. `tests/test_backmerge.py` goes from 35 to 44 tests.

## Docs

- `RELEASE.md` drops three claims the trigger restore made false: that back-merge pull requests carry no checks by default, and twice that the workflow is inert until it reaches `main`. Adds why a red DCO on a back-merge is expected and does not block.
- `CHANGELOG.md` entry under `## [Unreleased]`.

Note `.github/workflows/backmerge.yml` is unchanged, so `main`'s copy of it stays byte-identical. `backmerge.py` now differs between the branches until this reaches `main` through a back-merge or a release promotion.

Refs #610, #611, #626.


.github/workflows/scripts/backmerge.py
```diff
@@ -477,11 +587,153 @@ def open_pr(
...
+Pull request #{merged_pr} is merged, yet `origin/{MAIN_BRANCH}` is still {behind} commit{plural} ahead of `origin/{DEV_BRANCH}`. That is what a squash or a rebase does to a back-merge: it applies every file and discards the second parent, so `{DEV_BRANCH}` gained the content but never the ancestry.
+
+`git rev-list --count origin/{DEV_BRANCH}..origin/{MAIN_BRANCH}` cannot reach zero on its own from here, `release-pr` stays blocked, and this workflow would keep proposing the same back-merge. This pull request closes that loop.
+
+## What it changes: nothing
+
+The merge is `-s ours`. It takes no files from `{MAIN_BRANCH}` and records it as a parent, because #{merged_pr} already delivered every file. Confirm with one command — it prints nothing:
+
+```bash
+git diff origin/{DEV_BRANCH} {recovery_branch_for("<sha>")}
```
adriannoes: **Optional.** This `git diff` calls `recovery_branch_for("<sha>")`, so the generated command names `rc-dev/chore/back-merge-main-<sha>-ancestry` rather than the branch that was just pushed. A reviewer who pastes it gets a missing ref, not an empty diff. `open_recovery_pr` already has the real `branch`. I would thread that through (or tell them to `git diff origin/dev HEAD`) and pin the concrete `-ancestry` name in `test_the_recovery_pull_request_body_demands_a_merge_commit`. Your call whether to take this.

.github/workflows/scripts/backmerge.py
```diff
@@ -477,11 +587,153 @@ def open_pr(
...
+    plural = "" if behind == 1 else "s"
+    listing = "\n".join(f"- `{line}`" for line in commits)
+    checks = (
+        "This pull request was opened by a token that triggers workflows, so its checks run normally."
+        if has_ci
+        else (
+            "**This pull request carries no CI checks.** It was opened by `GITHUB_TOKEN`, and GitHub "
+            "does not trigger `on: pull_request` workflows for events raised by it. To run them, close "
+            "and reopen this pull request, or push any commit to its branch."
+        )
```
adriannoes: **Optional.** The `has_ci=False` arm here names `GITHUB_TOKEN` and close-and-reopen, and drops the `BACKMERGE_TOKEN` / `*looks* green` warning that `pr_body` still carries. A recovery PR opened without the secret has the same CI hole. The recovery-body test passes `has_ci=True`, so this arm never runs in CI. I would share one checks blurb. Your call whether to take this.

adriannoes (dismissed):
Reviewed at `67781ba5`: worktree pytest on `tests/test_backmerge.py` was 44 passed, ruff was clean, and H1 probes showed exit 0 only when a repair PR stands or tracking landed. CI check-runs on `dev` are green. No MCP or CLI surface, so live smoke is n/a.

### Decisions I would make

- Land #626 with a merge commit on its own clock. This PR does not skip DCO, does not rewrite #626, and does not turn that check green. The Push Ruleset wants one review and no required status checks, so red DCO is expected noise, not the merge gate. Waiting on #633 only delays the live `main` into `dev` reconciliation.

- After this lands on `dev`, merge-commit #626 (or otherwise make `main` an ancestor of `dev`) before the next real merge of `origin/main` into that tree. `git merge-tree` of this head against `origin/main` is an add/add conflict on `backmerge.py` while squash ancestry is still missing. #626 versus this PR does not conflict on that file. The new repair logic runs from `main`'s copy of the script, so it stays diagnose-only until a release promotion. A main-to-dev back-merge cannot carry this file onto `main`.

### Optional

Your call on all of these.

- Two pointers sit on `recovery_pr_body`: the advertised `git diff` uses a literal `<sha>` branch name, and the no-CI arm dropped the `BACKMERGE_TOKEN` warning that `pr_body` still has. Neither blocks the repair.

### What worked well

- `restore_ancestry` is a real extraction: `main()` shrank, `-s ours` is pinned as a full argv, and squashing the repair stops instead of looping.
- DCO is named in the generated body and is not skipped by branch name. That is the right call on a public repo.

<details>

Tier `standard`, trust `trusted`, reviewed at `9d426ae08ac0934449a1126ca74d329c613733ab...67781ba57bed0999ab870b0d9184493eeac25924`.

| Gate | Result |
| :--- | :--- |
| CI check-runs | green (one recovered checkout flake, not caused-by-pr) |
| Local pytest `tests/test_backmerge.py` | 44 passed |
| ruff (scoped) | pass |
| H1 matrix S1–S9 | pass |
| H15 | pass, no secret BLOCKER |
| H16 live MCP/CLI | n/a |
| Phase 7 refute | n/a (nothing blocking) |

Not covered: installing the new script on `main` (needs a release promotion, not this PR). #610 follow-ups remain open by design.

</details>

.github/workflows/scripts/backmerge.py
mocha06: Done in 31adad0. `recovery_pr_body` now takes the real `branch` from `open_recovery_pr`, and the command reads:

```bash
git fetch origin rc-dev/chore/back-merge-main-<short-sha>-ancestry
git diff origin/dev FETCH_HEAD
```

FETCH_HEAD rather than the branch name, so it resolves without a checkout. `test_the_recovery_body_advertises_a_command_that_resolves` pins the concrete `-ancestry` name and asserts `<sha>` is absent.

.github/workflows/scripts/backmerge.py
mocha06: Done in 31adad0. Both bodies now call `checks_blurb(has_ci)`, which carries the `pr_body` wording including the `BACKMERGE_TOKEN` remedy.

You are right that the arm never ran: `BACKMERGE_HAS_CI` comes from a secret that is set, so CI only renders `has_ci=True`. `test_both_bodies_share_one_checks_blurb` calls both bodies with `has_ci=False` and asserts they carry the same string, so the arm is pinned rather than reached.

adriannoes (approved):
Re-reviewed at `eceda37` after `origin/dev` was merged in. The earlier approve on `67781ba5` was dismissed when follow-up commits landed.

What changed since that head: `31adad0` threads the real recovery branch into the advertised `git fetch` / `FETCH_HEAD` command and shares `checks_blurb` across both PR bodies. `eceda37` then merges current `dev` (#626 as a merge commit, then #632). The CHANGELOG conflict kept the back-merge Unreleased bullet and the entries already on `dev`.

Re-verified on this tip: the three-dot vs `ec00104` is still the four files. Both review pointers are still present and pinned by tests. `origin/main` is an ancestor of `dev`, so the ancestry count is 0. CI on `eceda37` is green.

No new findings.

PR #634, opened 2026-08-19
test(mcp): cover the SSE attach vector of session ownership, and fix its record

### LINKED ISSUE #548: Hosted profile: session ownership degrades to client_id, so end users of one OAuth client are not isolated

mocha06:
## Motivation

#548 reported that hosted session ownership degraded to the OAuth client, so end users of one client were not isolated. The report is accurate, and the fix already shipped in #552 (`8775143`, then `3003e51`) — that PR never referenced the issue, so it stayed open. Verifying the report against the shipped code left two things worth closing.

The deliverable is coverage. `tests/auth/test_session_ownership.py` pinned `POST /mcp` only. `GET /mcp` opens the standalone SSE stream, which is a separate branch of the transport and the second way onto a session. #548 observed a cross-user attach on it directly, so it was an exploited vector with no regression test.

The CHANGELOG record of that fix also carries two errors, corrected alongside: the entry sits under a release tagged before the fix landed, and it states a version claim that does not hold.

## Outcome

The SSE attach is asserted. A second user presenting the owner's `mcp-session-id` on `GET /mcp` gets 404, and the owner's own attach still opens with 200 as the control. Reverting `subject` to `None` fails the cross-user case, so the assertion catches a real regression rather than passing incidentally.

Asserting the accepted case needs a different harness than the rest of the file, which is why `_attach_status` exists. `httpx.ASGITransport` runs the app to completion before it yields a response, and a successful attach is a stream that never ends, so the status never arrives through the client. The helper drives the ASGI app directly, reads the status off `http.response.start`, then cancels. Read that helper first.

The record is corrected on two points. The entry moves out of `[0.4.0-beta.2]`, which was tagged two days before the fix and does not contain it, into `[Unreleased]`. And the claim that `mcp` 1.x had no ownership check and no fields to populate now states the affected range instead: `mcp 1.27.2` added the check and `AccessToken.subject`/`claims` in the same release, verified against the published wheels, where `1.27.1` has neither and `1.27.2` has both.

That range is real but narrow. The published `pipefy-mcp-server==0.4.0b2` allows `mcp[cli]<2` and resolves to `1.29.0`, so it runs the ownership check against a `subject` nothing populates. Reaching it needs a self-hosted `remote` deployment on a pre-`0.5.0-alpha.1` release: the `local` profile validates no inbound bearer, and the hosted deployment pins `pipefy-mcp-server==0.5.0a1`, which contains both fix commits and resolves `mcp 2.0.0`. No action is planned on the published wheel; the entry just stops claiming the range is unaffected.

## Testing

Verified against the real `MCPServer` and session manager over ASGI, comparing the pre-fix mapping (`client_id` only) with the shipped one.

| 1 | `POST /mcp` `ping`, second user on the owner's session, pre-fix mapping | 200 | 200 | Reproduced |
| 2 | `GET /mcp` attach, second user on the owner's session, pre-fix mapping | 200 | 200 | Reproduced |
| 3 | `POST /mcp` `ping`, second user on the owner's session, shipped | 404 `Session not found` | 404 | Isolated |
| 4 | `GET /mcp` attach, second user on the owner's session, shipped | 404 | 404 | Isolated |
| 5 | `POST /mcp` `ping`, owner on their own session | 200 | 200 | No regression |
| 6 | `GET /mcp` attach, owner on their own stream | 200 | 200 | No regression |
| 7 | `subject` reverted to `None`, new attach test | fails | fails | Assertion is load-bearing |
| 8 | Published `0.4.0b2` with resolved `mcp 1.29.0`, cross-user `ping` | unknown | 200 | Affected range confirmed |

Gates run from the directories CI uses: full suite 4700 passed / 39 skipped, `ruff check`, `ruff format --check`, `lint-imports` (2 contracts kept), `scripts/bump_version.py verify`, and `lint_skill_refs.py` (17 files).

Not verified: whether server-initiated traffic reaches the wrong end user over a hijacked stream on an affected release. #548 raised it and left it open. Case 4 makes it moot for the shipped code, since the attach is refused before any stream is framed.

Closes #548.


adriannoes (approved):
Reviewed at `0ce5a8a9` after a rebase onto current `dev`. The last freeze was `6caba45`. Title and commit message changed (`correct` to `fix`); the two file trees did not (`git diff 6caba45 0ce5a8a9` on `CHANGELOG.md` and `test_session_ownership.py` is empty). Isolated pytest of that file on this tip: 12 passed. GitHub test, lint, and DCO are green on this SHA.

### Decisions I would make

- Leave #548 closed. The mapping is already on `dev` via #552. This PR is the GET test and the record correction. GitHub will not list a closer on merge because the issue was closed by hand at PR-open. Production on `0.5.0a1` is patched; reopening a `hosted-mcp` bug that is already fixed would mislead.

### What worked well

- Owner 200 control plus a shared `_attach_status` helper, so a GET that always 404s cannot green the hijack case.

<details>

Tier `standard`, trust `trusted`. First pass `9d426ae0...6caba45`; re-verified `ec001046...0ce5a8a9`. H16 n/a. No inline threads.

</details>

Issue #636, opened 2026-08-20
chore(release): add DCO sign-off (-s) to release.py git commits


adriannoes:
## Problem

`scripts/release.py` creates release commits with:

```python
run(["git", "commit", "-m", f"chore: release v{version}"])
```

Those commits do not include a DCO `Signed-off-by` trailer, so the **Developer Certificate of Origin** CI check fails on release PRs (e.g. #635 required a manual `git commit --amend -s` + force-push).

## Proposed fix

Pass `-s` / `--signoff` on every `git commit` invocation in `release.py` (at minimum the release bump commit in `_apply_prepare`).

Example:

```python
run(["git", "commit", "-s", "-m", f"chore: release v{version}"])
```

## Acceptance criteria

- [ ] Release PRs opened by `release-pr` / `alpha-pr` pass DCO without manual amend
- [ ] Trailer matches commit author (`user.name` / `user.email`), per `.cursor/rules/dco-signoff.mdc`

## Context

- Repo enforces DCO via `.github/workflows/dco.yml`
- Prior release commits (e.g. `chore: release v0.4.0-beta.2`) had the same gap; easy to forget because most day-to-day commits are signed off manually

PR #565, opened 2026-08-03
chore: Update dev branch with Main's changes

mocha06:
## Summary

updating dev branch with changes pushed directly to main

-

## Test plan

- [ ] `uv run pytest -m "not integration"`
- [ ] `uv run ruff check . && uv run ruff format --check .`
- [ ] Manual smoke (if applicable): callable via Cursor MCP

## Docs / skills

- [ ] `docs/parity.md` updated when MCP ↔ CLI coverage changed
- [ ] Affected `skills/` updated in this PR (or a paired PR)
- [ ] No docs/skills update needed

## Legal / contributions

- [x] Commits include DCO sign-off (`git commit -s`)
- [x] Regulated-domain skills include `COMPLIANCE.md` when applicable

PR #638, opened 2026-08-21
chore: release v0.5.0-beta.1

### LINKED ISSUE #528: fix(release): release-pr must fail when main has commits absent from dev

### LINKED ISSUE #531: ci: run the packaging smoke install on a schedule to catch upstream breakage

### LINKED ISSUE #534: fix(release): publish tags without ever building or installing the wheels

### LINKED ISSUE #535: fix(release): the release workflow creates the GitHub Release before validating the wheels

### LINKED ISSUE #540: install: write an install receipt so teardown is exact

### LINKED ISSUE #548: Hosted profile: session ownership degrades to client_id, so end users of one OAuth client are not isolated

### LINKED ISSUE #550: telemetry: stamp the deployment (local vs hosted) into the client headers

### LINKED ISSUE #561: ci: open the main -> dev back-merge pull request automatically

mocha06:
Cuts `v0.5.0-beta.1`, capturing everything merged into `dev` since `v0.5.0-alpha.1`. Tagging publishes a GitHub Release with wheels and a pre-release PyPI upload.

## What this PR contains

- Lockstep version bump `0.5.0-alpha.1` -> `0.5.0-beta.1` across every workspace package's `__init__.py`, the root `pyproject.toml`, `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`, `.cursor-plugin/plugin.json`, and `uv.lock`.
- `CHANGELOG.md`: `## [Unreleased]` finalized to `## [0.5.0-beta.1] - 2026-08-21`; the Release workflow uses this section as the GitHub Release body.

### Added

- **Cursor Marketplace plugin**: `.cursor-plugin/plugin.json` and root `mcp.json` package the skill catalog plus the hosted MCP server (`mcp.pipefy.com`) with in-client OAuth. The Cursor manifest declares `"commands": []` so the Claude Code files in `commands/` are not discovered. Version lockstep and CI packaging lint cover the new manifest. The listing is submitted from `main` after this packaging is released.

### Changed

- **Skill authoring template**: the copyable `SKILL.md` starter lives at `.github/skill-template/` rather than under `skills/`, so the catalog directory is only published skills.
- **Skills (`pipefy-ai-agents`)**: `create_ai_knowledge_base_document`, `create_llm_provider`, and `update_llm_provider` take a local file and are withheld on the hosted server. The skill now marks them local-profile-only so an agent on the Cursor Marketplace plugin (or any hosted URL) does not call tools the hosted floor omits.
- **Back-merge workflow (squashed back-merges, DCO):** a back-merge that was squashed or rebased is now repaired without being asked. The next run sees its pull request merged while the divergence still stands, cuts `rc-dev/chore/back-merge-main-<short-sha>-ancestry` from `dev`, merges `main` with `-s ours`, and opens that. The repair takes no files, because a squash applies every file and drops only the second parent — `dev`'s tree is already current and ancestry is the sole thing missing. A plain merge there would duplicate text `dev` already carries, since with no shared ancestry git reads one paragraph on each side as two independent additions. Squashing the repair in turn leaves the count where it was, and that is where the automation stops and records the loop rather than opening a third pull request for one `main` head. Separately, the generated pull request body now names any carried commit whose author disagrees with its own `Signed-off-by`: a squash rewrites the author to the merging account's profile identity while the trailer is carried verbatim, so a commit that passed DCO on its own pull request fails it on every back-merge that carries it afterwards, and it cannot be rewritten once it is on the default branch. The check does not block the merge, and naming the commits keeps a red DCO from reading as a new problem.
- **MCP (destructive confirmation):** every destructive preview now includes a required `confirmation_token` key, and 31 tool schemas gain an optional `confirmation_token` argument (29 dedicated destructive tools plus `execute_graphql` and `call_ipaas_tool`). A first call with `confirm=true` and no valid token returns that preview instead of mutating. Existing clients that ignore unknown keys still parse the preview, and clients that omit the new argument still get a preview rather than an execute, so neither change is a parse break; both change the published contract. GraphQL mutations through `execute_graphql` and destructive `call_ipaas_tool` calls follow the same two-step. The GraphQL preview names the mutation and does not claim it is irreversible (the server cannot tell create from delete). iPaaS `arguments.operation` is stripped then casefolded before needle equality, and a catalog miss is unclassifiable so it takes the two-step rather than running one-shot. The CLI takes no tokens and keeps confirming with `--yes` (or an interactive prompt).
- **GraphQL (documents too nested to parse):** `execute_graphql` and `pipefy graphql exec` refuse a document whose nesting exhausts the parser, and send nothing. Such a document cannot be classified as a query or a mutation, so running it could execute a mutation with no confirmation. `pipefy graphql exec` exits 2 for it even with `--yes`, where it previously failed with an uncaught `RecursionError`.

### Fixed

- **Release tooling (DCO):** `release.py` signs off the release commit it writes. The commit was created with a plain `git commit -m`, so the trailer arrived only from a `prepare-commit-msg` hook the author happened to have — a fresh clone has none, and the release pull request then failed DCO. A trailer lives inside the commit message, so no additive commit repairs it: the only way out was an amend and a force-push on an open release pull request, which is what #635 needed. `-s` takes the identity from `user.name`/`user.email`, the same one git stamps as author, so the trailer matches by construction. Both `release-pr` and `alpha-pr` reach the commit through `_apply_prepare`, so one call site covers every flow. (#636)
- **MCP (hosted `remote` profile): per-session credential binding now discriminates per user.** `mcp` 2.0 refuses a request whose authorization context differs from the one that created its Streamable HTTP session, comparing the `(client_id, issuer, subject)` triple the SDK derives from the validated bearer. `JwtTokenVerifier` supplied only `client_id`, which for the hosted server is the one public `pipefy-mcp` OAuth client every end user authorizes through, so the check could not tell two users of that client apart and either could present the other's `mcp-session-id`. The JWT `sub` is now mapped onto the SDK's own `subject` field and `iss` is carried on `claims`, the two components the comparison reads on top of `client_id`. The same triple also binds sealed `requestState`. `mcp` 1.27.2 added the ownership check and the `subject`/`claims` fields together, so a build resolving the toolkit's former `mcp[cli]>=1.25.0,<2` range to 1.27.2 or later ran the check against a `subject` nothing populated, and compared the client alone. Releases before `0.5.0-alpha.1` are affected on the hosted `remote` profile; the `local` profile validates no inbound bearer and is not. `PipefyAccessToken` is removed, since `subject` is a base `AccessToken` field in 2.0 and the subclass that carried `sub` under 1.x is redundant. Part of #543, and closes #548. The mapping supplies `subject` for every `sub` shape, not only a `str`. It had kept the `isinstance(sub, str)` guard from the `PipefyAccessToken.sub` field it replaced, where `sub` fed request logging only and degrading to `None` cost nothing; once the field became an ownership component that same line became a fail-open, collapsing `{absent, null, every non-string}` onto one principal. Two bearers of the one public `pipefy-mcp` client whose `sub` landed in that class produced an identical triple and the manager served one caller's request on the other's session (`ping` and `tools/list` both 200 for `sub` 101 vs 102, against the real manager). A non-string `sub` is now coerced with `str` rather than dropped, so an IdP emitting a numeric `sub` keeps distinguishing its users, the way the SDK's own `principal_components` coerces a non-string `iss`. An empty or whitespace-only `sub` is rejected outright (401), not mapped to `None`: `None` is the no-subject class, so folding a blank `sub` into it would make two blank-`sub` bearers share a context where they previously did not. A non-blank value is carried verbatim, never stripped. Not reachable through the deployed issuer, which emits a string `sub` on every access token it mints; the exposure is a self-hosted `PIPEFY_JWT_ISSUER_URL` pointed at another IdP. A bearer carrying no `sub` at all still maps to `subject=None`, and two such bearers of one client still share a context, waived deliberately: the identity precedence exists to accept a bearer whose only claim is `azp`/`client_id`, which names no end user, and it is the degradation the SDK documents for an unsupplied component. Deriving a `subject` from `jti` or a bearer hash was rejected, because it puts a credential id in the field every reader takes for an end user (the request log emits it as `sub`) and breaks a caller's own session on every token refresh. The waiver is pinned against the real session manager.
- **`uninstall.sh` (skills, project layout)**: `skills add` has two layouts and only one was read. A global install keeps its lock at `~/.agents/.skill-lock.json`; an install run from inside a project keeps it at `<project>/skills-lock.json`, and the probe looked for the first name only. No lock meant no derived store, so the confinement guard correctly refused a target it had no store for and the content stayed — a real round trip left seventeen `SKILL.md` files and the lock file behind while reporting success. Both names are read now, per base. The store is derived from the base rather than from the lock's own directory, since it is `<base>/.agents/skills` in both layouts while the lock's place is what differs. The skills directory beside a project lock is swept too, so a project install is visible without a receipt as well as with one. The lock file is `skills add`'s own and can record skills from several sources, so this run drops its own entries and removes the file only when those were the only ones in it — the contract `rmdir` already uses for a directory. The confinement guard is unchanged: a link out of the store still costs the link and nothing more. Both lookups are keyed on the lock as well as the skill name, since a machine carrying a global and a project install has the same name in both: keyed on the name alone the first record answered for the other one's store, so a project skill was judged against the global root, read as outside it, and left behind on a run that exited 0. (#541)
- **`uninstall.sh` (stderr noise)**: a scan no longer prints `printf: I/O error`. `client_rows`, `all_client_rows` and `records` emit a line at a time, and five consumers stopped reading early — four `awk … exit` and one `break` out of a `while read` — which closes the pipe while the producer is still writing; dash reports that EPIPE. They read to the end and keep the first match instead, which for a four-row table costs nothing. A static check enforces the shape, because running the code cannot: whether the producer is still writing when the consumer leaves depends on what fits in the pipe buffer, so the same early exit reproduces on Linux and not on macOS. `check_pipe_drain.py` runs beside the deletion guard in pre-commit and in CI and rejects an early exit in any pipeline fed by a function defined in the script; an `exit` inside an awk `END` block is not one, since END runs after the input is consumed. (#541)
- **CI (round-trip coverage)**: the install/uninstall round trip runs both scripts from `$HOME/roundtrip` rather than `RUNNER_TEMP`. `npx skills add` writes into the working directory, and `RUNNER_TEMP` sits under `$HOME` inside the pruned `work` tree, so the skills the job installed were never part of the filesystem comparison it exists to make. The job now also asserts outright that the install produced skills and that none survive the teardown, so the guarantee does not rest on which directories stay pruned. (#541)
- **`uninstall.sh` (skills)**: a skill is removed only where a record says this toolkit installed it. Teardown deleted every `~/.claude/skills/pipefy-*` directory under a tier-1 label promising a reinstall would restore it, which is false for anything this repository does not ship: `pipefy-` is a namespace anyone may write in, and a skill someone authored there went through `remove_path` with no backup. The prefix is now read as a namespace rather than a claim of ownership. Provenance comes from two records and nothing else — the install receipt, which lists the skills a run added by difference against what was already there, and the lock file `skills add` keeps for the base it wrote into, whose `source` is compared for equality against this repository. Anything else under the same prefix is reported and left in place, including in heuristic mode, which is where most machines are and stay. A skill directory is usually a link into a shared store, so the store content and the lock entry now go with the link instead of being stranded behind a run that reported the skill gone. (#541)
- **`uninstall.sh` (hosted token)**: the hosted OAuth logout runs where it can reach its target. `claude mcp logout <name>` resolves the name across scopes and takes no scope flag, so running it in the credentials phase — before any registration was touched — bound it to whichever entry outranked the hosted one and failed, leaving the token stored. Removal now has a phase for unshadowing local- and project-scope registrations, then the logout, then the hosted entry itself; the credentials-before-configs-before-tools-before-skills-before-state order it sits inside is unchanged. A logout that fails is reported as failed with the in-client fallback, never as a cleared credential. (#541)
- **`uninstall.sh` (deletion guard)**: `remove_path` resolves a path before judging it. It compared strings after stripping trailing slashes, so `$HOME/..` was neither `/` nor `$HOME` as text and named the directory holding every account. Paths are now normalised — `.` and `..` cancelled, the surviving directory resolved through the shell so a symlink cannot hide behind it, the final segment deliberately left alone because a symlink is deleted as a link — and a path containing `$HOME` is refused alongside `/` and `$HOME` themselves. (#541)
- **`uninstall.sh` (marketplace clone)**: an `installLocation` read out of `known_marketplaces.json` is a claim to check, not a deletion target. With no marketplace CLI verb to delegate to, teardown deleted whatever absolute directory that registry named while the plan said only _unregister marketplace_. The path is now required to be the plugin directory's own clone, derived from the client table; anything else is reported and left, and the plan discloses the clone it will delete. (#541)
- **`uninstall.sh` (Codex sub-tables)**: `[mcp_servers.<name>.env]` is a separate table header, invisible to a scan that stops at the next `[`, so excising the parent section left the sub-table and its secrets behind while the closing re-scan reported Codex clean. A section with any dotted child table is now treated as hand-edited: both are reported, both stay, and the run does not claim the file is clear. (#541)
- **`install.sh` / `uninstall.sh` (JSON rewrites)**: `ensure_ascii=False` on every `json.dump`. Adding or removing one registration rewrote unrelated servers' UTF-8 values as `\uXXXX` escapes — semantically identical, and a change to a file the user did not ask to have rewritten. (#541)
- **`uninstall.sh` (receipt keys)**: the receipt reader accepts the hyphen its writer emits. `entry_created.claude-desktop` and `entry_created.claude-code` failed an `^[a-z][a-z0-9_.]*$` key check and were dropped as malformed, so the bit recording that a registration was already there never reached the decision, and teardown fell back to the heuristic and deleted an installer-shaped Claude Desktop entry the installer had explicitly recorded leaving alone. A parity test now expands the writer's key vocabulary over the advertised `--client` values and asserts the reader's alphabet accepts every one. (#541)
- **`uninstall.sh` (exit contract)**: a teardown that removed nothing while state remains no longer exits `0`. `Nothing to remove.` returned success on exactly the tree `--scan` exits `1` on — a hand-edited Codex section, a registration a receipt records as pre-existing — and skipped the report, so the _Left alone_ notes explaining why nothing was planned never printed. Both now match `--scan`. (#541)
- **`uninstall.sh` (Linux keychain)**: a missing `secret-tool` reports the Secret Service as uninspected and exits `2`, as a missing `security` already did on macOS. A credential store nothing could read is not a store known to be empty, whatever the reason it could not be read. (#541)
- **`uninstall.sh` (`--client none`)**: the word is refused instead of silently disarming every registration edit. `--help` advertised `none` in the `--client` allowlist, inherited from `install.sh` where it means *install without registering*, but teardown narrows registration edits and nothing else — so `none` matched no client, skipped every config edit, and still removed the tools, leaving each registration the scan found pointing at a missing command. That is the stranding the phase order exists to prevent, and "tear down but touch no client config" has no coherent meaning, so the value is rejected with a message naming the difference rather than given a second one. The help text is generated from the client table alone. (#541)
- **`uninstall.sh` (skill store)**: a skill entry that links out of the store is not licence to delete what it points at. Resolving the link and deleting the target gave the teardown unbounded reach — a skill linked to a checkout of the user's own would have had that checkout removed — which is the marketplace `installLocation` problem with a longer arm. The permitted root is derived the same way the marketplace clone's is, from the lock file `skills add` keeps rather than from a hardcoded path — `<base>/.agents/skills` for the base that lock was found at — and both sides are canonicalized before the comparison. A target outside it leaves the link removed, the directory untouched, and a report line naming what stayed and where, so nothing reads as "the skill is gone" while its content is still on disk. (#541)
- **`uninstall.sh` (hosted token, reporting)**: an exit status is not proof for a store this run cannot read. `execute_plan` reported any action exiting `0` as `Removed:`, so a `claude mcp logout` that did nothing produced a verified removal claim, set the credential-deleted flag, and let the process exit `0` — with the client's OAuth store opaque to the closing re-scan, nothing ever contradicted it. The clear is now recorded as invoked-but-unobservable: its own report group, no removal claim, no credential-deleted accounting, and an exit of `1` rather than `0`, since a machine with an unverifiable clear is not one this run can call clean. This is the rule the local credential path already followed. (#541)
- **Skills (onboarding)**: the toolkit-setup skill's verify step uses the `curl` form of `uninstall.sh --scan`. It required `./uninstall.sh` at a repository root, which the hosted and plugin paths the skill leads with never produce. (#541)
- **`install.sh`**: a failed `uv tool install` is no longer swallowed. `run_quiet` read `$?` after the `if` that ran the command, which is the `if`'s own status (0) when its condition failed, so every failure reported success and the install carried on to register a server binary that was never installed. (#540)
- **`uninstall.sh` (matching rule)**: a registration is found whichever of its two fields identifies this toolkit, and a declared transport `type` is required for neither. An entry carrying a `url` and no `type` was treated as malformed and skipped, which is Claude Code's reading of that shape — it reads a type-less entry as stdio and refuses to start it — applied to every client. Cursor and Codex serve a bare `url` over HTTP, so a hosted registration in a Cursor config was invisible to the scan and survived a teardown, and the same shape was judged three different ways across the two client formats. An entry carrying both a `command` and a `url` diverged the same way in the other direction. The shape is now matched as written, described by whichever half is this toolkit's, and the per-client caveat comes from a `typed-remote` capability in the client table rather than from a client id in the matcher. The rule is implemented twice — python for the JSON configs, shell for the Codex TOML, which must work without python3 — and a suite drives one case table into both formats and one scan, so the copies cannot drift again. (#541)
- **SDK (error hierarchy)**: `PipefyGraphQLError` now subclasses `PipefyAPIError`, and so `PipefyError`. It previously subclassed `Exception` directly, which left the declared hierarchy disconnected from what the SDK actually raises: `PipefyError` and `PipefyAPIError` are both exported from `pipefy_sdk.__all__` and documented as the base and the API-payload error, but neither was raised anywhere in the package, while every GraphQL failure arrived as a `PipefyGraphQLError` that `except PipefyError` did not catch. A consumer following the documented root caught nothing it would ever see. **This widens what `except PipefyError` catches**, which is the point, but it is visible to a library consumer: code that catches `PipefyError` for a narrow purpose and expected GraphQL failures to pass through will now intercept them. Catch `PipefyGraphQLError` first where the distinction matters. The two handlers in `pipefy_cli.commands._common` are reordered for exactly that reason, so the GraphQL formatter still runs and CLI stderr keeps its `message (CODE)` shape rather than collapsing to `str(exc)`; `test_portal_sub_portal_detach_graphql_error_exits_1` guards `run_cli_command` and fails on the reparent alone; `test_run_pipefy_client_coroutine_graphql_error_keeps_the_code_suffix` guards the sibling runner. The parentage itself is pinned by `packages/sdk/tests/test_exception_hierarchy.py`. `PortalPermissionError` is untouched and still subclasses `ValueError`, which is what keeps a portal permission denial on exit code 2 while other GraphQL failures exit 1. No CLI or MCP behavior changes. The reparent is groundwork for #92 (replacing the ~190 broad `except Exception` handlers in the tool layer with typed catches); no handler is migrated here.
- **CLI (`pipefy auth logout`)**: a stored keychain entry that exists but cannot be parsed — stale, corrupt, or written by an older schema — is now deleted instead of being reported as `Not signed in. Nothing to do.` while surviving on disk. The decision comes from entry **presence** (`session_entry_presence` in `pipefy_auth.storage`: `present` / `absent` / `unknown`) rather than from `load_session`, which collapses all three into one `None`. Revocation needs a readable refresh token, so an unreadable entry is delete-only and stderr says the token stays valid at the IdP until natural expiry. When the presence probe itself fails and nothing is deleted, the command claims neither removal nor a clean machine and exits 1. Revoke-failure warnings, the revoke-then-delete order, and the exit-1 manual-removal hint on a rejected delete are unchanged. (#538)
- **AI agents (active lifecycle)**: create can start inactive (`create_ai_agent(active=false)` / `pipefy agent create --inactive`) by sending `disabledAt` on create and the chained update so the second step does not revive; routine update preserves `disabledAt` (optional `disabled_at` / `--disabled-at` pass-through skips the re-read; use `toggle_ai_agent_status` / `pipefy agent toggle` to activate or deactivate); create/update success payloads expose `disabled_at` and `active`; partial-failure envelopes surface shell `disabled_at` and a toggle recovery hint. Skills document the contract. (#520)
- **Release workflow (validate, then publish)**: `.github/workflows/release.yml` created the GitHub Release before anything installed the wheels, so a failing smoke install left a published Release advertising artifacts that could not start. `v0.4.0-beta.1` was exactly that shape — the Release was created, the smoke install failed with `ModuleNotFoundError`, the PyPI upload was skipped, and withdrawing the Release and the tag was manual work. The job now runs in two bands: build, bound the wheel set, stage the PyPI set, and install it into a clean virtualenv launching every console entry point — and only then create the GitHub Release and upload to PyPI. A validation failure now leaves nothing behind but the tag, which is unavoidable because the tag is what triggers the workflow. PyPI stays last within the publish band, since an upload cannot be taken back while a Release can be deleted — which leaves one narrower gap by choice, a failing upload leaving a Release with no matching PyPI artifacts — and a comment in the file states the boundary so later validation is not added between the two publish steps. The two hand-rolled wheel-set assertions the job carried — an inline Python check on `dist/`, and a five-glob bash check while staging — collapse onto `smoke_entry_points.py --check-wheels`, already used by the CI and scheduled-smoke workflows, so every workflow reads one definition of the published-member list rather than three that can drift apart. Closes #535.
- **Release CLI (build before the boundary)**: `scripts/release.py publish` pushed the tag — the one irreversible step — without ever having built or installed the artifacts it was about to release. Every guard before the confirmation prompt was about repository state (branch, working tree, tag collision); the first time a wheel existed was on the CI runner, past the point of no return, and `verify` inspected the _published_ result. `publish` now builds all five wheels with `uv build --all-packages --wheel` into a temporary directory, asserts one wheel per published member, installs them into a clean virtualenv, and launches every console entry point through `smoke_entry_points.py` — the same sequence the release workflow runs, moved to the side of the boundary where a failure costs nothing but a minute. The check sits in the shared irreversible half, after the repository-state guards so a wrong checkout still fails in a second rather than after a build, and before the confirmation prompt so the operator answers it knowing the artifacts start; nothing can reach a tag push without crossing it. A failure names which step broke and states that nothing was tagged, pushed, or published. `v0.4.0-beta.1` is the shape this catches: `mcp` 2.0.0 removed a module the MCP server imports at startup, so a fresh resolve produced a server that could not start — detectable locally in about a minute, and invisible in the release commit's diff, because nothing in the repository changed to cause it. Building into a temporary directory rather than `dist/` keeps a stale local build from being mistaken for a fresh one and leaves nothing behind. The confirmation prompt states that the built wheels install and launch, so what was checked is in front of the operator answering it. The check resolves third-party dependencies from PyPI rather than `uv.lock`, so it will occasionally pick a newer permitted version than the lock pins — that divergence is its subject, not a defect — and it needs the index, so a cut cannot be made while PyPI is unreachable. The wheel set `verify` requires on the GitHub Release is now derived from the published-member list in `smoke_entry_points.py` via a `wheel_stem` helper, so the escaped filename forms are no longer a hand-written third copy of that list. Closes #534.

### Added

- **CI / docs (shell script coverage and teardown documentation)**: `install.sh` and `uninstall.sh` are linted rather than trusted. `shellcheck --shell=sh` and `sh -n` run from `.pre-commit-config.yaml` and from a CI job that adds `dash -n`, since `/bin/sh` on macOS forgives what a Debian-family `/bin/sh` rejects; the hook comes from `shellcheck-py` so a contributor without Docker still gets it. A check asserts every `rm` and `rmdir` in `uninstall.sh` sits inside `remove_path`, and that the guard still refuses an empty argument, a relative path, `/`, and `$HOME`. A parity suite observes both scripts instead of reading them, snapshotting a fixture `HOME` around a real install and a real teardown per `--client`, so a file a new installer step writes has to be one teardown accounts for. A Linux round-trip job proves completeness against a real install with real network and real uv; its allowlist is explicit and each entry carries its reason. `docs/uninstall.md` documents the scan, the tiered teardown, heuristic mode, what is never removed by design, and the per-channel switching recipes; the removal one-liners scattered across the READMEs and the onboarding skill point at it, and the skill's prose invariant becomes a `--scan` run, in the `curl` form the hosted and plugin paths can actually reach. (#541)
- **`install.sh` / `uninstall.sh` (install receipt)**: `install.sh` appends one `key=value` record per run to `${XDG_STATE_HOME:-~/.local/state}/pipefy/install-receipt`, recording the `UV_TOOL_DIR` it used, whether it installed uv, whether it created each client registration or found one already there, where the skills landed, the release tag, and a schema version. It is installer state rather than user configuration, so it sits outside `~/.config/pipefy`, which holds the user's own `config.toml`. Each line is written as its step succeeds and the closing `record=end` marks a finished run, so a crashed install still leaves a usable record; records accumulate, so a second run with a different `--client` does not erase the first, and the merge takes the union of tool and skills directories because two runs with different `--prefix` both need tearing down. The format is parsed with POSIX text tools alone — teardown does not inherit the installer's python3 dependency — and a value carrying a space, an `=`, a tab or a newline round-trips through a defined escape. `uninstall.sh` prefers the receipt and falls back to heuristic mode when there is none: heuristic mode is permanent rather than transitional, since every install made before the receipt existed has none, and it removes less — in a config the installer writes it deletes a registration only where the value is exactly the single command the installer writes, and it never treats uv as this toolkit's. A record from an unknown schema is skipped whole rather than half-read, a malformed line is dropped, and no value from the receipt ever names something to delete: it only picks which tool directory to ask uv about, which skills directory to enumerate, and whether a registration the structural scan already found may be removed. The receipt is itself in the teardown inventory. Closes #540.
- **CI**: the `main` → `dev` back-merge pull request now opens itself (`.github/workflows/backmerge.yml`). Anything merged straight into `main` — a hotfix, a docs fix, a LICENSE — left `dev` behind until a human remembered to reconcile it, and nothing on the GitHub side noticed. `release-pr` refuses to cut while that is true, so forgotten drift did not corrupt a release; it blocked one, at the least convenient moment, for whoever happened to be cutting rather than whoever caused it. On every push to `main`, daily, and on demand, the workflow counts `origin/dev..origin/main` and, when it is non-zero, cuts `rc-dev/chore/back-merge-main-<short-sha>` from `origin/dev`, merges `origin/main` in, and opens a pull request into `dev`. The branch name is derived from the `main` head, which is what makes a re-run idempotent: the same drift resolves to the same branch, so an existing branch or pull request is recognized and left alone rather than pushed over — the ruleset blocks non-fast-forward pushes, and rewriting a branch under review destroys the reviewer's incremental diff. A _closed_ back-merge pull request is neither reopened nor re-pushed, since closing one is a human decision, but its unresolved drift is recorded on the tracking issue rather than treated as handled — #564 and #565 were both closed without merging and the divergence then sat unnoticed for days. A conflict is treated as a normal outcome rather than an error: release drift conflicts on `CHANGELOG.md` by construction and binary assets conflict whenever both branches touched one, neither is resolvable automatically (`-X ours` and `-X theirs` both silently discard real work), so the merge is aborted, nothing is pushed, and an issue labelled `back-merge` carries the conflicting paths and the resolution commands. That issue is one thread found by its label, closes on its own once the branches are back in sync or when a later pull request answers it, and stays silent on a re-run that would say exactly the same thing. The run exits zero when either the pull request or the tracking issue landed, because a red X on every release trains people to ignore it — and non-zero when neither did, since exiting zero is defensible only while the issue is the compensating signal, and a green run that surfaced nothing at all is the silent drift this removes. The pull request carries no CI checks by default and says so plainly — GitHub does not trigger `on: pull_request` for a pull request opened with `GITHUB_TOKEN` — and a `BACKMERGE_TOKEN` secret turns checks on with no change to the workflow. The pull request body requires a **merge commit**: a squash or rebase discards the second parent, so `dev` gains the content of `main`'s commits but never their ancestry, the count never reaches zero, and the workflow reproposes the same back-merge forever. When a merged pull request is found alongside surviving drift, the tracking issue names the squash as the cause and gives the commands to redo it. The decision logic is unit-tested in `tests/test_backmerge.py`, and the `pull_request` trigger that exercises the workflow's own changes runs as a dry run, so editing it cannot open a genuine back-merge off the branch under review. Closes #561.
- **CI**: the packaging smoke install now also runs on a **daily schedule** (`.github/workflows/packaging-smoke.yml`), not only on push and pull request. `uv.lock` pins every dependency, so the suite passes against known-good versions no matter what is published, and this is the only gate that resolves from PyPI — which left the window between merges with no observer at all. A breakage published upstream armed itself with zero commits here and detonated on whatever release came next, regardless of content; `mcp` 2.0.0 was caught by the release workflow's own smoke step, after the tag was pushed and the GitHub Release created. Two jobs run, because they see different failures: wheels **built from the checkout** catch an upstream release outside the declared bounds (the failure that would land on the next release), and the wheels **already on PyPI**, installed by name the way a user installs them, catch a broken or incomplete upload plus upstream breakage against the bounds of the release that already shipped. A failure opens an issue labelled `packaging-smoke-failure`, or comments on the open one, so a break persisting across days stays one thread rather than a daily pile. Two limits are documented rather than solved: the schedule fires from the default branch only, so a change on `dev` is inert until promoted, and the gate proves process startup, not behaviour — a dependency release that breaks behaviour without breaking imports passes it, and the locked suite would not see it either. Closes #531.
- **CI**: `scripts/smoke_entry_points.py` becomes the single definition of "every published console entry point", shared by the CI, release, and scheduled-smoke workflows, which each carried their own launch list. It reads _installed_ distribution metadata rather than the checkout, so it fails when a published member is missing from the install or when a wheel stops shipping its script — a wheel-filename glob could not see either. Its `--check-wheels` mode runs before the install and asserts one wheel per member, because an incomplete set is not an error for `pip`: the sibling `==` pins let it satisfy the absent member from the index, and the smoke then passes against a wheel the build never produced. Every discovered script is launched, so a newly added entry point is covered with no edit, while an explicit floor catches a lost one; the suite asserts both of its constants still match the workspace's `pyproject.toml` files, so adding a package or a script fails a test instead of quietly narrowing what the packaging gate checks.
- **Skills**: thin `pipefy-building` router (`skills/building/pipefy-building/`) maps build/configure intent to the correct domain skill; catalog and authoring domain list include `building`. Consulting-only `pipefy-process-design` links execution asks to the router instead of growing a delivery playbook.
- **Skills (Builder guardrails)**: `pipefy-automations` requires event×action `triggerEvents` checks before create, documents the known non-firing `field_updated` + `move_single_card` combo, and consolidates a did-not-fire / empty-logs debug checklist; `pipefy-ai-agents` (and AI automations) require an explicit AI ask plus post-write active/`disabledAt` re-read; `pipefy-ipaas` routes external-integration asks away from native automations/AI by default; `pipefy-pipes-and-cards` states phase connections are UI-only and requires verify-after-write before reporting success.
- **Release process**: alpha releases ship from `dev`, beta and stable releases from `main`. `scripts/release.py alpha-pr <bump>` branches off `origin/dev`, bumps the lockstep version, and opens a release PR into `dev`; once merged, `publish` from `dev` tags the commit and pushes the tag, whose wheels reach PyPI as a pre-release so the hosted MCP server's deployment wrapper can pin an exact version and exercise a release in staging before it ships. `release-pr beta` then promotes the alpha to the first beta of the same `X.Y.Z` (`0.5.0-alpha.3` -> `0.5.0-beta.1`) through the `dev` -> `main` release PR, and `publish` from `main` ships it. `alpha-pr` is `release-pr` based on `dev`, so both lines take the same reviewed path — a repository ruleset requires a pull request on `main` and `dev` alike, and no flow pushes a release branch: the bump always arrives via a merged PR and `publish` pushes only the tag, asserting the branch matches its remote so an unmerged release PR is reported as such. An alpha deliberately does not stamp `CHANGELOG.md`: `## [Unreleased]` keeps accumulating across `alpha.1..alpha.N` and is what the Release workflow uses as an alpha's release body, so the beta promotion still stamps the whole set into one section. Which branch a release ships from is derived from the version's own pre-release track rather than the operator's checkout, and enforced before the bump, again before the tag, and in the Release workflow itself, which asserts the tagged commit is an ancestor of `origin/dev` for an alpha or `origin/main` otherwise before any wheel is built.
- **SDK / MCP**: outbound Pipefy requests now carry the client **deployment** beside the surface, so hosted MCP traffic can be told apart from local installs. The two axes are orthogonal: the surface says which entry point ran (`mcp`, `cli`, `sdk`), the deployment says where it runs (`local`, `hosted`). With both, `User-Agent` reads `pipefy-sdk/<version> (<surface>; <deployment>)` and a parsed `X-Client-Deployment` joins the `X-Client-Name` / `X-Client-Version` pair, so a consumer never matches on the `User-Agent` string to group traffic. The MCP server derives the deployment from its resolved profile (`remote` → `hosted`, otherwise `local`) at the composition root: it is never read from env or TOML and no `PIPEFY_*_DEPLOYMENT` knob exists, keeping the property #336 established for the surface. Both sides are labelled explicitly, so a bare `(mcp)` unambiguously reads as a client older than this change rather than a local one — the one-time cost is that the existing `(mcp)` series is renamed rather than continued. The CLI keeps plain `(cli)` and sends no `X-Client-Deployment` (a CLI is local by definition), direct SDK use keeps plain `(sdk)`, and OAuth requests to the sign-in host are unchanged: the IdP path has neither a surface nor a deployment. Closes #550.

### Changed (breaking, internal API)

- **MCP**: ported to the `mcp` SDK 2.x (`mcp[cli]==2.0.*`), superseding both the `>=1.29.0,<2` floor bump and the `<2` cap entry listed below in this same cycle, whose closing note that the upper bound stays no longer describes the shipped pin. 2.0 removes the `mcp.server.fastmcp` module and renames `FastMCP` to `MCPServer`, so the pin could not move without the port. Nothing changes for a caller using the server over stdio or HTTP: the 187-tool surface, the `--toolsets` selections, the remote-safe floor, the tool-error envelope, and the structured log shapes are all unchanged, verified by the full suite plus a live smoke against the API. What changes is the code that embeds this package.
  - **The tool-call middleware seam moved onto the SDK's own.** The chain used to wrap the low-level server's private `request_handlers[CallToolRequest]` slot; 2.0 removes that dict and ships `MCPServer(middleware=[...])` instead. `install_tool_call_middleware` becomes `build_tool_call_middleware`, which returns one `ServerMiddleware` for the composition root to pass to the constructor. `extra_tool_middlewares` on `build_pipefy_mcp_server` is unchanged, so an embedder registering middleware needs no edit; an embedder that called `install_tool_call_middleware` directly does. Three shape changes for a middleware author. `ToolCallContext.req` is gone (read `tool_name` and `arguments`, now raw pre-validation values off the params). The seam's return type is the SDK's `HandlerResult` (`BaseModel | dict[str, Any] | None`) rather than a `types.ServerResult`, and a middleware that short-circuits returns the serialized wire dict `short_circuit_error` builds, not a model. And `ToolCallContext` carries a required `protocol_version`: `short_circuit_error` shapes its result against it, so any code constructing a context by hand (a middleware test fixture, a synthetic call) has to supply one.
  - **`serverInfo.version` now reports this package's version** (`pipefy_mcp.__version__`) instead of the SDK's. FastMCP neither accepted nor forwarded a version, so the advertised string tracked `mcp` itself; `MCPServer` takes one.
  - `host`, `transport_security`, and `json_response` leave the server constructor for the per-transport call, so `wire_hosted_observability` takes them. `port` is unaffected: it only ever reached the uvicorn bind. An embedder that calls `streamable_http_app()` itself must pass `json_response` there, or its POST replies become SSE-framed instead of plain JSON. `core/fastmcp_tool_lifecycle.py` is deleted in favour of the public `MCPServer.remove_tool`. `transport_security_for` pairs the resource-server URL parse with the allowlist build in one place.
  - **`context_from_request` is renamed `context_from_server_request`**, and its input changes from a `types.CallToolRequest` model to a `ServerRequestContext`. An embedder importing the old name from `pipefy_mcp.core.tool_middleware` gets an `ImportError`; the arguments and request id are read off the context (`ctx.params`, `ctx.request`) rather than off `req.params`.
  - **`McpRuntime.transport_security` is removed.** The SDK takes the DNS-rebinding allowlist per transport, on `streamable_http_app()`, so it no longer travels on the runtime; reading the attribute raises `AttributeError`. Resolve it with `transport_security_for(settings)` from `pipefy_mcp.core.transport_security` and pass the result to the serving call, as `run_server` does.
  - **What `call_next` returns inside a tool-call middleware is now polymorphic.** The SDK's own handler returns the serialized wire dict, keyed `isError` in camelCase, and so does `short_circuit_error`; a `CallToolResult` model, with a snake_case `is_error`, only arrives from an inner middleware that built its result by hand. Read the flag through `result_is_error(result)`, the supported accessor. A 1.x-style attribute read silently returns the default on the dict, so every real tool failure reads as a success. That exact read caused a regression during this port.
  - **A short-circuiting middleware owns its response envelope, and `short_circuit_error` now shapes one.** The SDK shapes a handler's result for the wire per negotiated revision _inside_ the middleware chain, so a middleware that returns without awaiting `call_next` is never shaped. Returned as a `CallToolResult`, its 2.0 `resultType` default (`"complete"`, a 2026-07-28 field) reached the wire on a connection that negotiated 2025-11-25, invisibly, because a client's surface validation ignores unknown fields. `short_circuit_error` now takes the `ToolCallContext` as its first argument and returns the wire dict, shaped by the SDK's own `serialize_server_result` against `ctx.protocol_version`. A middleware author updates the call (`short_circuit_error(ctx, "quota exceeded", code="RATE_LIMITED")`); nothing changes for a client. `ToolCallContext` gains `protocol_version`.
  - A `tools/call` that names no tool is logged under `tool: "<unnamed>"` rather than an empty string, so it does not share a dashboard bucket with the real tool names. The call still reaches the middleware chain, and `ctx.tool_name` is still `""`, the raw view of what the client sent.
  - **Mounting the HTTP app carries a lifespan obligation.** The app `wire_hosted_observability` returns owns the lifespan that enters `session_manager.run()`. Serving it directly (what `run_server` does, and what a uvicorn deployment does) needs nothing extra. An embedder that `Mount`s it under a host Starlette or FastAPI app must enter `app.session_manager.run()` from the host's own lifespan, because Starlette does not run a mounted app's lifespan; skipping it leaves the session manager unstarted and every request fails.
  - The 4 MiB Streamable HTTP body limit (a `413` above it) carries forward from the 1.29.0 floor bump documented below, unchanged: the wiring passes no `max_request_body_size`, so the SDK default stands. Attachments are unaffected (no tool takes bytes or base64); a long free-text argument is the reachable case. A hosted deployment that needs a larger body raises it on the `streamable_http_app()` call.
  - **The pin is the 2.0 patch line, not `<3`.** Two of the surfaces this package builds on sit outside what SemVer covers: the `MCPServer(middleware=[...])` list, which upstream marks provisional, and the `Tool.run` patch behind the argument-validation envelope. A minor release can move either with no signal, and no CI job resolves past `uv.lock`, so nothing would catch it before a release. Patch releases are accepted; widening to a new minor is a deliberate change that re-tests both surfaces.
  - Transitive dependency changes worth noting for image builds: the SDK now uses `httpx2` rather than `httpx` (this package still declares `httpx` itself, and nothing hands an httpx object into the SDK), and adds `mcp-types`, `opentelemetry-api`, and `truststore` while dropping `httpx-sse`. `truststore` validates TLS against the OS trust store, so a minimal container with no system CA store needs `SSL_CERT_FILE`. **Two transitive majors arrive together:** `starlette` 0.46.1 to 1.3.1 and `sse-starlette` 2.2.1 to 3.4.6. `starlette` is now a declared dependency of `pipefy-mcp-server` (`starlette>=1.3.1,<2`) rather than an undeclared transitive, because four modules import it directly (the `Starlette` app `wire_hosted_observability` returns, and the `Request` annotation on `McpRuntime.session_for_request` and the two identity resolvers), so an SDK release that dropped or vendored it would have broken those imports with no bound to catch it. `uvicorn` is now declared for the same reason (`uvicorn>=0.34.0,<1`): `run_server` imports it and builds the `Config`/`Server` pair on the `--transport http` path, while the SDK requires it only under a `sys_platform != 'emscripten'` marker, so the transitive guarantee was conditional.
  - The SDK deprecates the logging capability in the 2026-07-28 revision (SEP-2577), so every `ctx.debug` call in the tool modules emits `MCPDeprecationWarning` — around a thousand warning records in a suite run, against a previously clean baseline. Migrating those calls is a behavior change deserving its own review, so the suite carries one scoped `filterwarnings` ignore instead, matched on the message rather than the class so the SDK's other deprecation subjects still surface. The ignore is bounded by construction: when the SDK removes the capability the calls break outright and the suite fails regardless.
  - Part of #543. **The server negotiates `2025-11-25` on the `initialize` handshake, but a client that opens with the enveloped `server/discover` reaches `2026-07-28`**, which a default `mcp` 2.x `Client` (`mode="auto"`) does. Adopting that revision as the served default is still separate work; what changed here is that the tools now behave correctly on both.

### Changed (installer)

- **Installer**: `install.sh` resolves the newest GitHub Release whose tag is **not** an alpha, so a staging alpha published from `dev` never becomes what a default `curl … | sh` install hands out; the public install path stays on the beta line. Alphas are also flagged as GitHub pre-releases, keeping the "Latest" badge on the newest beta. Filtering is by tag shape rather than the API's `prerelease` flag, because the whole pre-1.0 line is betas that must stay installable. `install.sh --version vX.Y.Z-alpha.N` still installs an alpha on purpose.

### Fixed

- **Release process**: `release-pr` (and `alpha-pr`, which delegates to it) now refuses to cut when `origin/main` carries commits absent from `origin/dev`, naming the count and prescribing the back-merge. The existing staleness guard compared version strings, so it only saw divergence that moved the version; work merged straight into `main` without a bump — a doc fix, a CI workflow — passed it unnoticed. Since the cut branches off `dev` and the released notes come from `dev`'s `## [Unreleased]` alone, those entries were stamped out of the release section and left orphaned under `## [Unreleased]` on `main`, and the resulting tag covered a combination only ever exercised by the PR merge check. The new guard runs before the branch is created, so a stale `dev` costs nothing to walk away from. `prepare` is unaffected: it runs on `main` and stamps `main`'s own section, never consulting `dev`. Closes #528.

## After merge


adriannoes (approved):
LGTM

PR #639, opened 2026-08-21
chore: back-merge main into dev (2 commits)

mocha06:
## Motivation

`origin/main` carries 2 commits absent from `origin/dev` after the `v0.5.0-beta.1` release: `bd57ef3` (the #638 merge) and `70cd0c4` (the version bump and CHANGELOG stamp). `release-pr` refuses to cut while that count is non-zero, so this blocks the next release.

`.github/workflows/backmerge.yml` could not open this itself. Its run failed on token permissions, not on a conflict:

```
remote: Permission to pipefy/ai-toolkit.git denied to mocha06.
fatal: ... The requested URL returned error: 403
HTTP 403: Resource not accessible by personal access token (labels)
GraphQL: Resource not accessible by personal access token (addComment)
GraphQL: Resource not accessible by personal access token (updateIssue)
```

The merge it computed was clean. The push, the label write, and both issue writes were refused. Issue #611 is closed while the drift still stands, so nothing tracked it either. This pull request is that back-merge, opened by hand.

## Outcome

`dev` carries `main`'s ancestry again, and `git rev-list --count origin/dev..origin/main` reaches zero. `release-pr` can cut the next release.

The merge is clean — no conflict. `CHANGELOG.md` gains 2 lines: the `## [0.5.0-beta.1] - 2026-08-21` heading above the notes it already held, leaving `## [Unreleased]` empty. Every version-bearing file moves to `0.5.0-beta.1`, and `bump_version.py verify` exits 0.

## Notes

**Merge this with a merge commit, never a squash or a rebase.** Both discard the second parent, so the count does not move and the same back-merge is needed again. `RELEASE.md:148` records #577 doing exactly that and leaving the count at 9 for eleven days.

DCO passes on this pull request. Its range is three commits — two merge commits, which `dco-check` skips by default, and `70cd0c4`, whose author and `Signed-off-by` match. The commits with a squash-rewritten author are already on `dev`, so they fall outside the range. A back-merge only goes red on that when `main` has accumulated such commits directly, which a normal release does not produce.

Setting a `BACKMERGE_TOKEN` secret with `contents: write`, `pull-requests: write`, and `issues: write` is what stops the workflow needing a manual back-merge on every release.

PR #640, opened 2026-08-21
docs(release): name the expected DCO failure on a release cut

mocha06:
## Motivation

`RELEASE.md` described the squash-rewritten author only as something that happens to back-merge pull requests. A `dev → main` release cut carries the same commits and fails DCO the same way, and nothing said so.

The failure is easy to misread as a fault in the release commit, and the obvious "fix" makes things worse: re-cutting from somewhere with less history turns the check green by carrying almost nothing, which is also a cut that stamps none of `dev`'s notes.

The bullet was also inaccurate on where such a commit lands. It said `main` and "the default branch", but most feature pull requests squash onto `dev`, which is where the mismatched commits actually accumulate.

## Outcome

The `dev → main` release PR section states the expectation where a reader first meets it, points at the mechanism instead of repeating it, and records the trap about a suspiciously green check.

The back-merge bullet now covers any later pull request that carries the history — a back-merge and a release cut alike — and says the commit is unrepairable once it is on `main` or `dev` rather than on the default branch alone.

## Notes

No `CHANGELOG.md` entry. `## [Unreleased]` on `dev` is where the `v0.5.0-beta.1` notes still sit unstamped, and the pending back-merge (#639) inserts the `## [0.5.0-beta.1]` heading directly above them. An entry added here would land inside the released section and conflict with that merge. It is a documentation clarification with no behavior change.


RELEASE.md
```diff
@@ -160,7 +162,7 @@ Three things to know rather than be surprised by:
...
+- **A red DCO check on a back-merge or a release cut is expected, and does not block it.** A squash merge rewrites the author to the merging account's GitHub profile identity while the `Signed-off-by` trailer, being message text, is carried verbatim. A contributor whose git `user.name` differs from their GitHub display name therefore lands a commit whose author no longer matches its own sign-off, even though it passed DCO on its own pull request. That commit cannot be rewritten once it is on `main` or `dev`, so `dco-check` fails on every later pull request that carries it — a back-merge and a `dev → main` release cut alike. The generated pull request body names the offending commits so the failure does not read as a new problem, and no ruleset requires the check.
```
adriannoes: **Blocking.** This heading now says a red DCO check is expected on a back-merge or a release cut alike.

Those pull requests scan opposite ranges. A `dev` into `main` release cut carries history `dev` has that `main` does not. A `main` into `dev` back-merge carries history `main` has that `dev` does not. They cannot carry the same squash-rewritten commit.

The last cycle matches that split. #638 (the release cut) failed DCO. #639 (the back-merge) passed, because those mismatched commits were already on `dev`. #626 failed only because `7052d6b` was unique to `main`.

The same bullet says the generated pull request body names the offending commits. That list comes from the back-merge workflow. `release_pr` does not compute or render it.

I would keep the new paragraph above the bump steps. That is the right place to say a red DCO check on the release pull request is expected. Narrow this bullet back to back-merges: red when `origin/dev..origin/main` still carries the mismatch. Keep the landing-branch correction (`main` or `dev`). Drop the "alike" pairing. Keep offender naming as a back-merge body fact.

Done when:

- This heading no longer treats a back-merge and a release cut as equivalent expected-red events
- The generated-body sentence applies only to back-merge automation
- A reader of this bullet would not treat a green post-release back-merge as a surprise

adriannoes (changes requested):
Reviewed at `53b29caa`: static review of `RELEASE.md` plus DCO check-run records for the last release cut and back-merge. No worktree, no live MCP.

### Required before merge

- The back-merge DCO bullet now pairs a back-merge with a `dev` into `main` release cut as equivalent expected-red events. Those git ranges are complementary. The last cycle was red on the cut and green on the back-merge. See the thread on `RELEASE.md:165`.

### Decisions I would make

- The Notes still call #639 pending as the reason to skip `CHANGELOG.md`. #639 closed unmerged after #641 stamped current `dev`. I would rebase onto `origin/dev` and either add a short Unreleased docs bullet or restate the skip as a docs-clarification choice. This does not block merge.

### What worked well

- The new paragraph before the bump steps puts the release-cut DCO expectation where a releaser looks first. The green-tiny trap is the right reading of recutting off a short branch.
- Saying the mismatch can land on `main` or `dev`, not only the default branch, matches how feature pull requests squash onto `dev`.

<details>

Tier `lite`, trust `trusted`, reviewed at `723313ff1f20b3d23d960071a8d5af4c45d87a9d...53b29caa830030d96728f0acc7fd6d3b2c75e0c4`.

| Gate | Result |
| :--- | :--- |
| CI on head | green (lint, test, DCO, Shell scripts, install.sh then uninstall.sh) |
| DCO trailer on this commit | pass |
| Worktree / unit tests | n/a (lite) |
| Live smoke | n/a (docs-only) |
| DCO-bullet refute | survives (3 verification Attempts reproduced) |
| Judge | GPT Extra High; blocking finding kept IMPORTANT |

Not covered: executing `release-pr` or opening a live cut. The DCO claim was checked against check-runs on #638, #639, and #626 plus the frozen `backmerge.py` / `release.py` body generators.

</details>

RELEASE.md
mocha06: Done in `2c11fc7`. The bullet is back-merge only again, and its heading now carries the condition: red when the range carries such a commit.

It also states the case you named, so a green one does not read as a surprise: a back-merge that follows a release carries the release commits alone and is green, because the mismatched commits are already on `dev` — the opposite range from the cut that carried them there.

The offender-list sentence stays a back-merge fact. `release.py` has no DCO handling to render one from.

Evidence: `grep -n -i "dco\|sign-off\|signed-off" scripts/release.py` returns nothing. `dco_mismatches()` is defined at `.github/workflows/scripts/backmerge.py:133` and called only at `:801`, `:828`, and `:866`.

The paragraph above the bump steps is unchanged.

mocha06:
The blocking finding is answered in the `RELEASE.md:165` thread. Done in `2c11fc7`.

On the `CHANGELOG.md` note: you are right that the reason was stale. #639 closed unmerged and `## [Unreleased]` on `dev` is live again, so the conflict I was avoiding cannot happen. `2c11fc7` adds the `### Changed` bullet rather than restating the skip, because the skip was a conflict workaround, not a choice about what belongs in the changelog.

The branch is level with `origin/dev` — I merged it in rather than rebasing. A rebase means a force-push, which discards the incremental diff you are reading. `git rev-list --count HEAD..origin/dev` is 0, and the net change against `dev` is still two files: `RELEASE.md` and `CHANGELOG.md`.

adriannoes (approved):
Reviewed at `2c11fc7`, after `53b29caa`.

The blocking DCO bullet is back-merge only. Its heading now says red when that range carries the mismatched commit. A back-merge that follows a release is described as green, because those commits are already on `dev`. The offender-list sentence stays in the back-merge bullet. The release-cut paragraph above the bump steps is unchanged. `CHANGELOG.md` gained an Unreleased bullet for this docs change.

Re-checked at this tip: the three Done when items on `RELEASE.md:165` hold. CI is green (lint, test, DCO, Shell scripts, install.sh then uninstall.sh, skill frontmatter). The two new commits carry matching Signed-off-by trailers.

PR #646, opened 2026-08-25
fix: Set readOnlyHint=False on create_ipaas_connection write tool

### LINKED ISSUE #644: [Bug]: create_ipaas_connection is the only write tool that omits readOnlyHint
## Problem

`create_ipaas_connection` (`packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:402`) passes `ToolAnnotations(openWorldHint=True)` and nothing else. `readOnlyHint` is absent.

Every other write tool on the server sets `readOnlyHint=False` explicitly. This is the only one that leaves it unstated, so a client relying on the annotation has to fall back on a default to learn that the tool writes.

It is not a harmless omission on this particular tool: the docstring at `ipaas_tools.py:442-445` describes the call as an upsert keyed on `external_id`, so passing an existing id overwrites that connection's stored credential in place.

## Evidence

An audit of all 114 write tools under `packages/mcp/src/pipefy_mcp/tools/` found this to be the single case. Every registration passes an `annotations=` keyword and every `ToolAnnotations(...)` is non-empty, so the gap is one missing field rather than a missing annotation block.

## Acceptance criteria

- `create_ipaas_connection` carries `readOnlyHint=False` alongside its existing `openWorldHint=True`.
- No behaviour change.

VedantMadane:
## Summary

Set readOnlyHint=False on create_ipaas_connection write tool

## Changes

- Match other write tools: ToolAnnotations(readOnlyHint=False, openWorldHint=True)
- create_ipaas_connection upserts credentials and must not look read-only

Fixes #644

### DIFF
--- a/packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
+++ b/packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
@@ -399,7 +399,10 @@ class IpaasTools:
 
             return await _run_ipaas_tool(ctx, pipe_id, work)
 
-        @mcp.tool(annotations=ToolAnnotations(openWorldHint=True), meta=REMOTE)
+        @mcp.tool(
+            annotations=ToolAnnotations(readOnlyHint=False, openWorldHint=True),
+            meta=REMOTE,
+        )
         async def create_ipaas_connection(
             pipe_id: PipefyId,
             piece_name: str,
--- a/packages/mcp/tests/tools/test_ipaas_tools.py
+++ b/packages/mcp/tests/tools/test_ipaas_tools.py
@@ -1441,6 +1441,9 @@ async def test_connection_auth_url_is_not_read_only(mock_client, mock_gateway):
     auth_url = by_name["get_ipaas_connection_auth_url"]
     assert auth_url.annotations is not None
     assert auth_url.annotations.read_only_hint is False
+    create_conn = by_name["create_ipaas_connection"]
+    assert create_conn.annotations is not None
+    assert create_conn.annotations.read_only_hint is False
     # The discovery meta-tool stays a genuine read.
     assert by_name["get_ipaas_tools"].annotations.read_only_hint is True
 


mocha06:
Thanks for the contribution!

Can you please review the failed jobs @VedantMadane

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
```diff
@@ -399,7 +399,7 @@ async def work(gateway: IpaasGateway, token: str) -> dict:
 
             return await _run_ipaas_tool(ctx, pipe_id, work)
 
-        @mcp.tool(annotations=ToolAnnotations(openWorldHint=True), meta=REMOTE)
+        @mcp.tool(annotations=ToolAnnotations(readOnlyHint=False, openWorldHint=True), meta=REMOTE)
```
adriannoes: **Blocking.** Adding `readOnlyHint=False` on one line pushed this decorator to 99 columns. `ruff format --check` fails locally and on GitHub (`Would reformat`), so the lint and test jobs never get past format. The sibling `get_ipaas_connection_auth_url` already wraps the same annotations.

I would wrap this decorator the same way.

Done when:

- `uv run ruff format --check packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py` is clean
- the GitHub lint job can proceed past the format step

```suggestion
        @mcp.tool(
            annotations=ToolAnnotations(readOnlyHint=False, openWorldHint=True),
            meta=REMOTE,
        )
```

adriannoes (changes requested):
Thank you for this first-time contribution. You closed the one write tool that still omitted `readOnlyHint`, which is exactly issue 644, and `False` is the right hint for a credential upsert.

Reviewed at `a11fa0c`. DCO passed. GitHub lint and test are red on format.

**Verdict: needs changes** into `dev`. Please switch the base off `main` first.

### Required before merge

- Wrap the `create_ipaas_connection` decorator so `ruff format --check` passes. Detail and a one-click fix are on the thread on `packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py:402`.

### Decisions I would make

- Change this PR's base to `dev`. We will merge `dev` into `main` this week, and that is how this fix should land. Merging to `main` now would skip that train and force a back-merge.

### Optional

Your call on all of these.

- Extend `test_connection_auth_url_is_not_read_only` so `create_ipaas_connection` asserts `read_only_hint is False`. In this repo's Python model an omitted hint is `None`, so `is False` actually catches a regression.

### What worked well

- The annotation matches a real write, `meta=REMOTE` stayed put, and `docs/mcp/tools/ipaas.md` already listed the tool as not read-only.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
VedantMadane: Wrapped the decorator to match `get_ipaas_connection_auth_url` and extended `test_connection_auth_url_is_not_read_only` so `create_ipaas_connection` asserts `read_only_hint is False`. Also retargeted base to `dev`.

Fixed in dc6da8617610733e8abe0efcc123e50adcb63734.

VedantMadane:
Addressed review feedback:

- Wrapped `create_ipaas_connection` `@mcp.tool` decorator so `ruff format --check` passes
- Extended `test_connection_auth_url_is_not_read_only` to assert `create_ipaas_connection.read_only_hint is False`
- Retargeted PR base from `main` to `dev`

Commit: dc6da8617610733e8abe0efcc123e50adcb63734

adriannoes (approved):
Thanks for following up on the review notes, @VedantMadane, and for the contribution.

The decorator wrap matches the sibling iPaaS write tool, `create_ipaas_connection` now pins `read_only_hint is False`, the PR is on `dev`, and CI is green (lint, test, DCO, shell scripts, install/uninstall).

Appreciate you taking a good first issue all the way through. Welcome to the toolkit.

packages/mcp/src/pipefy_mcp/tools/ipaas_tools.py
adriannoes: Thanks. This matches the wrap we asked for and CI is green. Resolving.

PR #651, opened 2026-09-02
refactor(sdk): ship the destructive-confirmation helpers in pipefy_sdk

### LINKED ISSUE #649: sdk: ship the destructive-confirmation token helpers in pipefy_sdk (relocate from pipefy_mcp)
> **Amended 2026-09-01:** original proposal moved two modules into `pipefy-infra`. After an architecture pass, the scope is smaller and the target changed: move **only** `destructive_confirmation_token.py`, into **`pipefy_sdk`**. Second amendment: no re-export shim — the old path is deleted and internal callers updated, since no external Python consumers of these packages exist yet. Third amendment: the export list gains `confirmation_signing_key` — implementation review found the key-derivation half of the protocol still lived in the guard, leaving SDK consumers to reinvent it. Rationale in "Why the SDK (and not pipefy-infra)" below. `toolsets.py` stays in `pipefy_mcp` — it is taxonomy of the MCP tool surface (tool-name strings by domain/profile) and belongs to that layer.

## Problem

The SDK deliberately does not gate destructive operations — its docstrings delegate that to the caller:

- `packages/sdk/src/pipefy_sdk/services/pipe_config_service.py:78` — *"Delete a pipe by ID (permanent). Caller must enforce preview/confirm UX."*
- `packages/sdk/src/pipefy_sdk/services/table_service.py:143` — same sentence for `delete_table`.

The canonical implementation of that preview/confirm protocol — `mint_confirmation_token` / `verify_confirmation_token` / `classify_confirmation_token_failure` in `packages/mcp/src/pipefy_mcp/tools/destructive_confirmation_token.py` — lives in the `pipefy-mcp-server` package. An application that embeds the SDK in process cannot import it without also taking:

- the `mcp[cli]==2.0.*` pin (`packages/mcp/pyproject.toml`), which conflicts with any environment holding `mcp` 1.x (e.g. via `langchain-mcp-adapters`), and
- a package whose identity resolution is coupled to the server request scope by design (`packages/mcp/src/pipefy_mcp/auth/request_identity.py:75` raises outside it).

So every in-process SDK consumer that wants to honor the "caller must enforce preview/confirm" contract has to **reimplement** the token logic — and silently drift from the behavior the MCP server enforces and CI-tests (`packages/mcp/tests/tools/test_destructive_confirm_ticket_drift.py`).

## Why the SDK (and not pipefy-infra)

- The module is **domain safety policy** (canonical resource identity + the preview→confirm ordering), not run-machinery. `pipefy-infra` holds generic infrastructure (config loading, path discovery, SSRF defenses, telemetry headers) and knows nothing about Pipefy resources; parking deletion semantics there mixes layers even though the import graph would allow it.
- The SDK is where the contract already lives half-finished: it ships the ungated `delete_*`/`remove_*` methods (56 across the SDK) **and** the docstring obligation to gate them. Shipping the canonical gating helpers in the same package closes that loop for every SDK consumer.
- Import direction is already right: `pipefy-mcp-server` depends on `pipefy` (the SDK), so the guard keeps importing the same functions from one layer down. The SDK's TID251 bans (`pipefy_mcp`, `pipefy_cli`) are untouched.
- The module is stdlib-only (`base64`, `hashlib`, `hmac`, `json`, `re`, `typing`), so the SDK gains no dependency.

## Proposed change

1. Move `packages/mcp/src/pipefy_mcp/tools/destructive_confirmation_token.py` → `packages/sdk/src/pipefy_sdk/destructive_confirmation.py`, byte-identical (module rename only).
1a. Also export the key-derivation step as `confirmation_signing_key(caller_secret: str | bytes) -> bytes` (the guard's existing `sha256(bearer).digest()`), and have the guard's `signing_key_for` call it. Key derivation is part of the protocol: without a shipped helper, every consumer reimplements it, and one passing a module-level constant key gets tokens that verify **across callers** — the exact cross-user property the per-caller key exists to block. The guard's `_PROCESS_SIGNING_KEY` fallback for no-bearer contexts stays in `pipefy_mcp` (server policy, not protocol). The SDK README gains a worked snippet calling the helper plus an explicit warning against constant keys.
2. Update `destructive_tool_guard.py` and the drift tests to import from `pipefy_sdk`, and **delete the old module path — no shim**. The path is not part of the documented contract (`docs/DEPRECATION.md` scopes MCP deprecation to tool names and argument shapes; `pipefy_mcp.tools.__init__` never exported it), and no external Python consumers of these packages exist yet, so a deprecated alias would be pure liability.
3. `check_destructive_confirmation` (the `ctx`-coupled guard in `destructive_tool_guard.py`) stays in `pipefy_mcp` — only the pure functions move.
4. Do **not** move `toolsets.py` or `PIPEFY_TOOL_NAMES` (the skills lint AST-parses `registry.py` for the latter; the former is MCP-layer taxonomy).
5. Document the new module in the SDK docs next to the delete methods, so "caller must enforce preview/confirm UX" points at the shipped helpers.

## Compatibility

- **MCP protocol: unchanged.** No tool name, schema, annotation, envelope, or preview text changes; hosted/remote users cannot observe the move.
- **Token continuity across a rolling deploy: safe.** The HMAC key derives from the caller's bearer plus the canonicalized resource identity — no server-side secret or state — so a token minted by a pod on the old build verifies on a pod running the relocated, byte-identical functions.
- **Layering: respected.** SDK gains no dependency; `pipefy_mcp → pipefy_sdk` already exists; drift tests keep running against the guard where it is.
- **Release: ordinary.** All five wheels already ship per tag with lockstep `==` pins and version-agreement checks.

## Acceptance criteria

- [ ] `mint_confirmation_token` / `verify_confirmation_token` / `classify_confirmation_token_failure` / `confirmation_signing_key` importable from `pipefy_sdk` with unchanged signatures and behavior (derivation byte-identical to the guard's current `signing_key_for` bearer path, preserving cross-deploy token continuity).
- [ ] `signing_key_for` in the guard delegates to `confirmation_signing_key`; the process-key fallback behavior is unchanged.
- [ ] SDK docs show a worked key-derivation snippet and warn that a shared constant key lets one caller's token confirm another caller's deletion.
- [ ] Old module path removed; no internal reference to `pipefy_mcp.tools.destructive_confirmation_token` remains.
- [ ] SDK dependency list unchanged.
- [ ] Guard/drift tests pass with unchanged assertions (imports updated to the new path).
- [ ] Skills lint (`lint_skill_refs.py`) passes without modification.
- [ ] SDK delete-method docstrings reference the shipped helpers.

mocha06:
Closes #649.

## Motivation

The SDK ships ungated `delete_*` / `remove_*` methods — 56 across the package — plus a docstring obligation to gate them, but not the code that does the gating. `delete_pipe` and `delete_table` both say "Caller must enforce preview/confirm UX".

The canonical implementation of that protocol lived in `pipefy-mcp-server`. An application that embeds the SDK in process cannot import it without taking two things it does not want. First, the `mcp[cli]==2.0.*` pin, which conflicts with any environment holding `mcp` 1.x. Second, a package whose identity resolution raises outside the server request scope by design. So every in-process consumer that wants to honour the contract reimplements the token logic, and drifts from the behaviour the MCP server enforces.

## Outcome

`mint_confirmation_token`, `verify_confirmation_token`, `classify_confirmation_token_failure` and `confirmation_signing_key` are importable from `pipefy_sdk`. An in-process consumer runs the same protocol the MCP server runs, from the same code.

- The module is `pipefy_sdk.destructive_confirmation`, stdlib-only. The SDK dependency list is unchanged: 8 `Requires-Dist` entries, read off the built wheel.
- The token wire format is unchanged and pinned by a golden vector, so a token minted by one build verifies on another during a rolling deploy.
- `confirmation_signing_key` is new. It derives one caller's HMAC key from that caller's credential, which is what stops one caller's token from confirming another's deletion. `signing_key_for` in the MCP guard calls it, so both sides share one derivation.
- `pipefy_mcp.tools.destructive_confirmation_token` is deleted. The guard imports from `pipefy_sdk`.
- The MCP protocol is unchanged. No tool name, schema, annotation, envelope, or preview text moves, so hosted and remote callers cannot observe this.

Read `packages/sdk/src/pipefy_sdk/destructive_confirmation.py` first. Its only change against the old module is the added `confirmation_signing_key`. The three original functions and every private helper are byte-identical.

## Testing

Offline: full suite 4762 passed / 39 skipped, `ruff check`, `ruff format --check`, `lint-imports` (2 contracts kept), `bump_version.py verify`, the skill-reference and Cursor-plugin linters, `uv build --all-packages`.

The live runs below drove the `pipefy-mcp-server` binary built from this branch, over the real MCP protocol, against a prod organization. They used throwaway pipes only. All seven were created and deleted by the runs, and each deletion was confirmed afterwards.

### Local profile, stdio

| 0 | Server built from this branch serves the tool surface | `delete_pipe` registered | 187 tools, present | PASS |
| 1 | `delete_pipe confirm=false` | Preview, token minted, no deletion | `requires_confirmation=true`, token minted | PASS |
| 2 | `get_pipe` after the preview | Pipe still exists | Returned the pipe | PASS |
| 3 | `confirm=true`, token omitted | Previews again, no deletion | `requires_confirmation=true` | PASS |
| 4 | Tampered token, MAC no longer matches | Rejected, no deletion | `requires_confirmation=true` | PASS |
| 5 | Pipe B's token replayed against pipe A | Identity mismatch, no deletion | `requires_confirmation=true` | PASS |
| 6 | `get_pipe` after three rejected confirmations | Pipe still exists | Returned the pipe | PASS |
| 7 | `confirm=true` with the matching token | Pipe is deleted | `Pipe ... was permanently deleted.` | PASS |
| 8 | `get_pipe` immediately after the deletion | Not found | Still resolved | see below |
| 9 | Spent token for A replayed against B | Identity mismatch, B survives | `requires_confirmation=true` | PASS |

Row 8 is a property of the Pipefy API, not of this change. Pipe deletion is eventually consistent: polling shows the pipe stops resolving 6.6 s after the mutation returns success. The deletion mutation itself is untouched SDK code.

A separate stdio run replayed one token across two server processes. Process B rejected process A's token. That confirms the local profile falls back to the per-process random key and never reaches `confirmation_signing_key`.

### Remote profile, Streamable HTTP with a validated bearer

| 0 | RFC 9728 protected-resource metadata | 200 naming this resource | 200, resource matches | PASS |
| 1 | `tools/list` with no bearer | 401 | 401 | PASS |
| 2 | Validated bearer reaches the remote-safe surface | `delete_pipe` present, surface smaller than local's 187 | 184 tools, present | PASS |
| 3 | `delete_pipe confirm=false` | Preview, token minted, no deletion | `requires_confirmation=true`, token minted | PASS |
| 4 | Token minted by process A, confirmed against a fresh process B, same bearer | Verifies and deletes | `Pipe ... was permanently deleted.` | PASS |
| 5 | Pipe A's token replayed against pipe B, same caller | Identity mismatch, B survives | `requires_confirmation=true` | PASS |

Row 4 covers the changed line. The same cross-process replay is rejected on stdio. A token survives a server restart only when its key comes from the caller rather than from the process, so row 4 is what proves `confirmation_signing_key(bearer)` runs.

Not verified: `mcp.pipefy.com`. The hosted deployment pins an exact published release of `pipefy-mcp-server` (`0.5.0b1` today), so it cannot exercise this branch. Hosted coverage follows the release and the pin bump. The remote profile above therefore ran locally against the real issuer.

## Notes

Issue #649 asked for a deprecated re-export shim at the old path. This drops it. `pipefy_mcp.tools.__init__` exports only `PIPEFY_TOOL_NAMES` and `ToolRegistry`, and `docs/DEPRECATION.md` scopes the MCP surface to tool names and argument shapes, so the deleted path was never public.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- **SDK destructive-confirmation helpers**: `mint_confirmation_token`, `verify_confirmation_token`, `classify_confirmation_token_failure`, `confirmation_signing_key`, `DESTRUCTIVE_CONFIRMATION_TTL_SECONDS` and `ConfirmationTokenFailure` are exported from `pipefy_sdk` (module `pipefy_sdk.destructive_confirmation`). Signatures and token wire format are unchanged, so a token minted by one build verifies on the other. An in-process SDK consumer can now honour the "caller must enforce preview/confirm UX" contract without depending on `pipefy-mcp-server`. `confirmation_signing_key` is new: it derives one caller's HMAC key from that caller's credential, which is what stops one caller's token from confirming another's deletion. `signing_key_for` in the MCP guard now calls it, so both sides share one derivation. The internal `pipefy_mcp.tools.destructive_confirmation_token` module is removed; it was never an exported path (`pipefy_mcp.tools.__init__` exports only `PIPEFY_TOOL_NAMES` and `ToolRegistry`) and no MCP tool contract changes.
 - **Claude Code plugin MCP**: `.mcp.json` is the hosted `mcp.pipefy.com` server with in-client OAuth, the same file the Cursor plugin points at. Local stdio (`uvx`) remains the Quick-install / `claude mcp add` path.
 
 ## [0.5.0-beta.1] - 2026-08-21
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -48,6 +48,52 @@ Other exported error types sit outside this root. Catch these by name:
 Transport-level failures (connection refused, timeouts) surface as `gql`'s
 `TransportError`, which the SDK does not wrap.
 
+## Destructive operations
+
+The SDK does not gate deletion. Every `delete_*` and `remove_*` method runs the mutation as soon as you call it. Enforcing a preview/confirm step is the caller's job, for all of them — `delete_pipe` and `delete_table` say so in their docstrings, and the same obligation applies to the rest.
+
+`pipefy_sdk.destructive_confirmation` ships the helpers that step needs, so an in-process consumer runs the same protocol the MCP server runs:
+
+```python
+import time
+
+from pipefy_sdk import (
+    confirmation_signing_key,
+    mint_confirmation_token,
+    verify_confirmation_token,
+)
+
+# Derive the key from THIS caller's own credential, never a shared constant.
+key = confirmation_signing_key(caller_bearer)
+identity = {"pipe_id": pipe_id}
+
+# First call: describe what would be deleted, and hand the token back.
+token = mint_confirmation_token(
+    tool_name="delete_pipe",
+    resource_identity=identity,
+    key=key,
+    now=int(time.time()),
+)
+
+# Second call: the caller returns the token. Delete only when it verifies.
+if verify_confirmation_token(
+    token,
+    tool_name="delete_pipe",
+    resource_identity=identity,
+    key=key,
+    now=int(time.time()),
+):
+    await client.delete_pipe(pipe_id)
+```
+
+The token binds one operation name, one canonicalized resource identity, and one `key`. It expires after `DESTRUCTIVE_CONFIRMATION_TTL_SECONDS`. Verification is stateless HMAC, so no server-side store is needed. `tool_name` is any stable label for the operation: the MCP server passes its tool name, and a library consumer passes its own.
+
+`confirmation_signing_key` derives the `key` from one caller's credential. Use it per caller. A single module-level key defeats the binding, because every token then verifies for every caller, and nothing in the SDK can catch that for you.
+
+`classify_confirmation_token_failure` says **why** verification would fail (`missing`, `invalid_or_expired`, `identity_mismatch`) so you can word the rejection. It is diagnostic only. Never let it authorize the deletion — that stays behind `verify_confirmation_token`.
+
+The token orders the two steps. It is not an authorization control: the API permission on the credential remains the boundary that allows or denies the deletion.
+
 ## Configuration
 
 OAuth and endpoint variables are documented in **[`../config.md`](../config.md)** and **[`../../.env.example`](../../.env.example)**. Integration tests use `@pytest.mark.integration` and the same `PIPEFY_*` keys from local **`.env`** (e.g. `PIPEFY_PORTAL_ORG_UUID` for portal live tests). Unit tests use fictional ids in **[`../../packages/sdk/tests/_shared/fixture_ids.py`](../../packages/sdk/tests/_shared/fixture_ids.py)** — not production org UUIDs.
--- a/packages/mcp/src/pipefy_mcp/tools/destructive_tool_guard.py
+++ b/packages/mcp/src/pipefy_mcp/tools/destructive_tool_guard.py
@@ -13,22 +13,22 @@ control. API permission remains the boundary that allows or denies the deletion.
 
 from __future__ import annotations
 
-import hashlib
 import secrets
 import time
 from collections.abc import Awaitable, Callable, Mapping
 from typing import Any, Literal
 
 from mcp.server.mcpserver import Context
-from typing_extensions import NotRequired, TypedDict
-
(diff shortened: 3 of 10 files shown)


mocha06:
closed because the direction which v1 is moving towards is to not have this helper at all, anywhere in the toolkit

Issue #656, opened 2026-09-02
sdk: typed GraphQL inputs for the relation write methods, removing the last **attrs


mocha06:
Part of #652, which shipped the generator, the schema snapshot, the drift check, and the first batch (pipe configuration). This is the relation batch, and the last one: it removes the final `**attrs: Any` from the SDK.

## Methods

| Method | GraphQL input |
| --- | --- |
| `create_pipe_relation` | `CreatePipeRelationInput` |
| `update_pipe_relation` | `UpdatePipeRelationInput` |
| `create_card_relation` | `CreateCardRelationInput` |

They live in `packages/sdk/src/pipefy_sdk/services/relation_service.py`, with facade wrappers in `client.py`.

## What the batch involves

1. Add the three types to `ROOT_INPUT_TYPES` in `scripts/generate_graphql_inputs.py`.
2. `uv run python scripts/generate_graphql_inputs.py snapshot` then `generate`.
3. Change the service and facade signatures to take the typed input, and drop `**attrs`.
4. Update the MCP tools (`relation_tools.py`), CLI commands, and tests.
5. Add a test asserting no `**attrs: Any` remains anywhere in `packages/sdk/src`, so the surface cannot grow back.

## Contract already proven against the live schema

- **Every boolean on the pipe-relation inputs is `Boolean!`, not `Boolean`.** `allChildrenMustBeDoneToFinishParent`, `allChildrenMustBeDoneToMoveParent`, `autoFillFieldEnabled`, `canConnectExistingItems`, `canConnectMultipleItems`, `canCreateNewItems`, `childMustExistToFinishParent`, `childMustExistToMoveParent` are all required on create *and* on update, as is `name`. The service supplies defaults for them today; the typed input makes the requirement visible, so decide whether the defaults stay in the service or move to the caller.
- `CreatePipeRelationInput` also requires `childId` and `parentId`; `UpdatePipeRelationInput` requires `id`.
- `CreateCardRelationInput` requires all four of `childId`, `parentId`, `sourceId`, `sourceType`. `sourceType` is `String!`, not an enum — a soft enum whose documented values are `PipeRelation` and `Field`.
- Both pipe-relation inputs pull in `FieldMapInput` via `ownFieldMaps`, which the automation batch also touches; whichever lands second gets it for free.

## Acceptance criteria

- [ ] No `**attrs: Any` on the three methods or their facade wrappers.
[...]

Issue #655, opened 2026-09-02
sdk: typed GraphQL inputs for the automation write methods, retiring the hand-written copies


mocha06:
Part of #652, which shipped the generator, the schema snapshot, the drift check, and the first batch (pipe configuration). This is the traditional-automation batch, and the one that retires a duplicated set of hand-written models.

## Methods

| Method | GraphQL input |
| --- | --- |
| `create_automation` | `CreateAutomationInput` |
| `update_automation` | `UpdateAutomationInput` |

They live in `packages/sdk/src/pipefy_sdk/services/automation_service.py`, with facade wrappers in `client.py`.

## Why this batch is the interesting one

`#652` noted that the toolkit keeps a second, hand-synced copy of these shapes. That copy is `packages/sdk/src/pipefy_sdk/models/ai_automation.py`, which hand-writes `ConditionExpressionInput`, `AutomationEventParamsInput`, `AutomationActionParamsInput`, `FieldMapInput`, and (in `models/ai_agent.py`) `AiBehaviorMetadataInput` — every one of them a name the generator also produces from the schema. Today the two live side by side: `pipefy_sdk.ConditionExpressionInput` is the hand-written permissive model, `pipefy_sdk.graphql_inputs.ConditionExpressionInput` is the schema mirror.

This batch is where they converge. Two differences have to be settled first, and neither is mechanical:

- **`extra="allow"` versus `extra="forbid"`.** The hand-written models round-trip unknown keys; the generated ones reject them by name, which is what the API does. Moving the AI-automation surface onto the generated models makes an unknown key an error where it used to pass through.
- **`CONDITION_OPERATIONS`.** The hand-maintained tuple in `models/ai_automation.py` lists the same operations the schema now carries in `ConditionExpressionInput.operation`'s own description. Decide whether the tuple keeps its export (it is public, via `pipefy_sdk.__init__`) and is regenerated, or is deprecated in favour of the generated model.

## What the batch involves

1. Add the two types to `ROOT_INPUT_TYPES` in `scripts/generate_graphql_inputs.py`.
2. `uv run python scripts/generate_graphql_inputs.py snapshot` then `generate`.
3. Change the service and facade signatures to take the typed input, and drop `**attrs`.
4. Decide the convergence above and retire whichever hand-written models it settles.
5. Update the MCP tools (`automation_tools.py`), CLI commands, and tests.

## Contract already proven against the live schema

[...]

Issue #654, opened 2026-09-02
sdk: typed GraphQL inputs for the webhook and inbox-email write methods


mocha06:
Part of #652, which shipped the generator, the schema snapshot, the drift check, and the first batch (pipe configuration). This is the webhook and inbox-email batch.

## Methods

| Method | GraphQL input |
| --- | --- |
| `send_inbox_email` | `CreateAndSendInboxEmailInput` |
| `send_email_with_template` | `CreateAndSendInboxEmailInput` |
| `create_webhook` | `CreateWebhookInput` |
| `update_webhook` | `UpdateWebhookInput` |

They live in `packages/sdk/src/pipefy_sdk/services/webhook_service.py`, with facade wrappers in `client.py`.

## What the batch involves

1. Add the three types to `ROOT_INPUT_TYPES` in `scripts/generate_graphql_inputs.py`.
2. `uv run python scripts/generate_graphql_inputs.py snapshot` then `generate`.
3. Change the service and facade signatures to take the typed input, and drop `**attrs`.
4. Update the MCP tools (`webhook_tools.py`) and CLI commands to build the model with `build_graphql_input` / `graphql_input_or_bad_parameter`.
5. Update the tests.

## Contract already proven against the live schema

- **`CreateAndSendInboxEmailInput.from` is a Python keyword.** This is the batch that exercises the generator's alias path: `field_attribute` renames it `from_` with `alias="from"`, and `GraphQLInput` sets `populate_by_name=True` so either spelling constructs. `tests/test_generate_graphql_inputs.py::test_python_keyword_field_gets_an_alias` already covers the rename; this batch is the first to ship it in a model.
- `CreateAndSendInboxEmailInput.from` and `subject` are `String!`, and `to` is `[String]!` — required.
- `CreateWebhookInput` has the first `defaultValue` fields in any batch: `overridePreviousWithSameUrl` and `identityPropagationEnabled` both default to `false`. The generator records the default in the snapshot but does not yet apply it, since the API supplies it; confirm that is still the right call before shipping.
- `CreateWebhookInput.filters` is `JSON` and `headers` is `Json` — two distinct scalars that differ only in case. Both map to `GraphQLJson`.
- `CreateAndSendInboxEmailInput.emailAttachments` pulls in `EmailAttachmentInput`.

## Note on the SDK's own injection

[...]

Issue #653, opened 2026-09-02
sdk: typed GraphQL inputs for the database-table write methods


mocha06:
Part of #652, which shipped the generator, the schema snapshot, the drift check, and the first batch (pipe configuration). This is the database-table batch.

## Methods

| Method | GraphQL input |
| --- | --- |
| `create_table` | `CreateTableInput` |
| `update_table` | `UpdateTableInput` |
| `create_table_record` | `CreateTableRecordInput` |
| `create_table_field` | `CreateTableFieldInput` |
| `update_table_field` | `UpdateTableFieldInput` |

They live in `packages/sdk/src/pipefy_sdk/services/table_service.py`, with facade wrappers in `client.py`.

## What the batch involves

1. Add the five types to `ROOT_INPUT_TYPES` in `scripts/generate_graphql_inputs.py`.
2. `uv run python scripts/generate_graphql_inputs.py snapshot` then `generate`.
3. Change the service and facade signatures to take the typed input, and drop `**attrs`.
4. Update the MCP tools (`table_tools.py`) and CLI commands to build the model with `build_graphql_input` / `graphql_input_or_bad_parameter`.
5. Update the tests.

## Contract already proven against the live schema

- `UpdateTableFieldInput.table_id` is `ID!` — required on an update, alongside `id`.
- `CreateTableInput.name` and `organization_id` are `ID!` / `String!`.
- `CreateTableRecordInput.table_id` is `ID!`; `due_date` is `DateTime`, the first date scalar any batch pulls in, so `SCALAR_TYPES` needs a decision on whether it stays a passthrough `str`.
- `CreateTableInput` and `UpdateTableInput` both reference `Colors`, already generated.
- `CreateTableInput.labels` / `members` pull in `LabelInput` and `MemberInput`; `CreateTableRecordInput.fields_attributes` pulls in `FieldValueInput`, whose `field_value` is the `UndefinedInput` scalar (mapped to `GraphQLJson`).

## Acceptance criteria

- [ ] No `**attrs: Any` on the five methods or their facade wrappers.
- [ ] `generate_graphql_inputs.py check` passes with the five types in the snapshot.
- [ ] An unknown field in an MCP `extra_input` or a CLI `--extra` is named before the request.
- [ ] No MCP tool name, argument shape, or CLI option changes.

PR #657, opened 2026-09-02
feat(sdk): type the pipe-config write methods from the GraphQL schema

### LINKED ISSUE #652: sdk: replace **attrs: Any with GraphQL-derived typed input models
## Problem

56 `PipefyClient` methods across the SDK accept `**attrs: Any` (e.g. `update_pipe(self, pipe_id, **attrs) -> dict`) and forward the fields into a GraphQL input object. The API itself is strictly typed — `updatePipe(input: UpdatePipeInput!)` rejects unknown fields at the type-system level — so the SDK is the only layer without the shapes:

- The GraphQL schema (ground truth) defines every input type exactly.
- The MCP tool layer hand-maintains typed parameters for the same operations (that is how the server serves proper JSON schemas).
- The SDK exposes `Any`, so consumers get no IDE autocomplete, no client-side validation, no introspectable signature — and any programmatic consumer that derives schemas from signatures (docs generators, function-calling adapters, test factories) gets nothing usable for exactly the write methods where correctness matters most.

## Proposed change

Replace `**attrs: Any` with typed input models derived from the GraphQL input types:

1. Generate pydantic models from schema introspection (`UpdatePipeInput`, `CreatePhaseInput`, ...) into `pipefy_sdk` — codegen, not hand-written, with a CI drift check so a new API field is a regeneration rather than a manual edit.
2. Change the method signatures to take the typed input (e.g. `update_pipe(self, pipe_id: str, input: UpdatePipeInput) -> dict`). Serialization to the wire stays where it is today.
3. Stage by traffic: the highest-use write methods first (pipe/phase/field/card/automation config), the long tail after.

## Breaking change — and why now is the moment

Changing `**attrs` to a typed parameter breaks callers. Two facts make this the cheap moment, the same reasoning as the shim decision in #649:

- The packages are pre-1.0 and ship as pre-releases; the stable contract starts at v1.0.
- No external Python consumers of the SDK exist yet.

After 1.0 this same change costs a deprecation cycle per method.

## Side benefits

- The MCP tool layer's hand-maintained parameter lists can progressively delegate to the generated models, removing a second hand-synced copy of the same shapes.
- Docstrings stop needing prose like "attrs: fields to update" — the model is the documentation.

## Acceptance criteria

- [ ] Generated input models exist for the staged write methods, with a CI check that fails when the GraphQL schema and the generated models drift.
- [ ] The staged methods take typed inputs; no `**attrs: Any` remains on them.
- [ ] Client-side validation errors name the offending field before any network call.
- [ ] SDK docs show one worked example (construct input model → call method).

Related: #649 (same pre-1.0 timing argument; both reduce what in-process consumers must hand-build).

mocha06:
Closes #652. Based on `dev`.

## Motivation

Pipefy's GraphQL input objects are strictly typed and reject an unknown field themselves:

```
mutation { updatePipe(input: {id: "303088927", nmae: "x"}) { pipe { id } } }
→ InputObject 'UpdatePipeInput' doesn't accept argument 'nmae'
```

The SDK was the one layer without those shapes. Twenty-one write operations took `**attrs: Any` and forwarded whatever they were handed, so a consumer got no autocomplete, no client-side check, and no introspectable signature for exactly the methods where a typo costs a round trip. The MCP tool layer hand-maintains the same shapes a second time to serve its JSON schemas.

## What ships

The generator, the schema snapshot, the drift check, and the first batch of seven methods. The remaining fourteen operations are filed as sub-issues of #652 (#653 tables, #654 webhooks and inbox email, #655 automations, #656 relations) so each lands as its own reviewable batch.

### The generator

`scripts/generate_graphql_inputs.py` has three commands, and the split between them is what lets CI check the models without credentials:

| Command | Reaches the network | Run by |
| --- | --- | --- |
| `snapshot` | yes | a person, after an API change |
| `generate` (default) | no | a person, after `snapshot` |
| `check` | no | CI, via `tests/test_generate_graphql_inputs.py` |

`check` catches a hand-edit of a generated file, or a snapshot committed without regenerating. It cannot see the API itself move, because CI has no Pipefy credentials. That half is `test_input_types_snapshot_matches_live`, marked `integration`.

Adding a type to `ROOT_INPUT_TYPES` is the whole of a future batch's codegen step; the transitive closure is resolved from there. Today's seven roots pull in 12 input objects, 108 fields and 2 enums.

`packages/sdk/src/pipefy_sdk/graphql_inputs/` holds `_generated.py` and `__init__.py` (both written by the generator, both carrying a `# @generated` header) plus a hand-written `_base.py`. Everything that needed a decision rather than a mapping lives in `_base.py`, so a regeneration never overwrites it.

### The migrated methods

`update_pipe`, `update_phase`, `create_phase_field`, `update_phase_field`, `update_label`, `create_field_condition`, `update_field_condition` — on `PipeConfigService` and on the `PipefyClient` facade.

```python
from pipefy_sdk.graphql_inputs import UpdatePipeInput

await client.update_pipe(UpdatePipeInput(id=pipe_id, name="Onboarding", color="blue"))
```

## The proven contract

Everything below was read off the live schema by the same introspection the generator runs, and the wire behaviour was checked against the API directly.

**Identifier form.** `ID` maps to `str | int` and is not coerced. That mapping is forced, not chosen: `updatePipe` accepts `{id: 303088927}` as an integer (verified), while `createFieldCondition` answers a string `expressions_structure` entry with an opaque 500 (documented on `normalize_field_condition_payload`). Coercing in either direction would break one of the two.

**Required despite reading as optional.** The schema marks four fields `NON_NULL` on inputs that look like partial updates, and the models now make that visible rather than leaving it to a docstring:

| Input | Required |
| --- | --- |
| `UpdatePhaseInput` | `id`, `name` |
| `UpdatePhaseFieldInput` | `id`, `label` |
| `UpdateLabelInput` | `id`, `name`, `color` |
| `CreatePhaseFieldInput` | `phase_id`, `label`, `type` |

**Unknown fields.** `extra="forbid"` mirrors the API rather than adding a rule on top of it. Nested inputs behave the same way — `updatePipe(input: {preferences: {findabl: true}})` answers `InputObject 'RepoPreferenceInput' doesn't accept argument 'findabl'` — so a nested rejection is not new either. The model only moves both rejections to before the request, where the message can name the field.

**Enums are soft.** A GraphQL enum is typed `str` and the documented values are exported as a tuple (`COLORS_VALUES`), following the `CONDITION_OPERATIONS` convention already in the SDK. A value added server-side keeps working without an SDK release.

**Authorization.** Unchanged. Every one of these mutations is governed by the API permission on the caller's credential, exactly as before; nothing here is a permission check.

## Behaviour boundaries

- **The payload is unchanged.** For all seven methods the typed path builds the same GraphQL input the `**attrs` merge built, normalizers included. Proved by constructing both and comparing (7/7 identical), which is also what rules the change out as the cause of the two API behaviours in the matrix below.
- **A field condition needs its shape repaired before it is parsed.** GraphQL coerces a bare value into a single-item list, so `expressions_structure: [0]` is a legal way to write `[[0]]` on the wire — verified: the API stores it as `[["0"]]`. A model mirroring `[[ID]]` refuses it, so `normalize_field_condition_fields` (new, in `pipefy_sdk.utils`) runs on the raw mapping first, at both boundaries. It is idempotent, and the service still runs the same repair on the serialized payload for a direct SDK caller who skips it.
- **The CLI keeps its reserved-key guard.** Deleting the SDK's `**attrs` guards would have let `--extra '{"phaseId": "999"}'` overwrite the `PHASE_ID` argument and file the rule on another phase, and `--extra '{"id": ...}'` patch a different condition. `reject_reserved_extra_keys` restores the refusal the SDK used to raise, at the layer that now owns it. The MCP tools were already covered by their own reserved sets.
- **A resolution hint is a parameter, not a field.** `update_phase_field` takes `phase_id` / `pipe_id` as keyword arguments. The mutation has no such fields; they narrow the SDK's slug-to-uuid lookup. The MCP tool and the CLI lift them out of `extra_input` / `--extra` rather than rejecting them, so both paths that worked before still work.
- **The reserved-key guards moved rather than vanished.** They lived in the SDK because `**attrs` sat beside a positional argument that the same key could override. One typed input leaves the SDK nothing to reconcile, so the check now sits at the boundary that still has two sources: the CLI's `--extra` and the MCP tools' `extra_input`.
- **No explicit null.** A field left unset is omitted, so a partial update stays partial; no field can be sent as an explicit `null`. That was true of `**attrs` too.
- **The MCP and CLI surfaces do not move.** No tool name, argument shape, annotation, envelope, or CLI option changes, and the CLI help golden is unchanged. The one observable difference is that an unknown `extra_input` key or nested `preferences` key now comes back as `INVALID_ARGUMENTS` naming the field, instead of as the API's rejection one round trip later. The rejected value is never echoed back, so a wrong field carrying a secret stays out of the transcript and the shell log.

## Docs and tests

- `docs/sdk/README.md` — a "Typed mutation inputs" section with the worked example, the `ID` rule, the soft-enum rule, and the resolution-hint rule.
- `packages/sdk/README.md` — how to regenerate, and what `check` can and cannot catch.
- `.gitignore` — `scripts/*` is ignored behind an allowlist; the generator is added to it.
- New tests: `tests/test_generate_graphql_inputs.py` (33), `packages/sdk/tests/test_graphql_inputs.py` (19), `packages/sdk/tests/services/pipefy/test_input_types_snapshot_integration.py` (1, `integration`), four MCP cases (the new rejection, the lifted hint, the nested path, the repaired shape) and three CLI cases (both reserved-key guards and the repaired shape).

## Testing

Offline: full suite green, `ruff check`, `ruff format --check`, `lint-imports` (2 contracts kept), `bump_version.py verify`, the skill-reference and Cursor-plugin linters, `uv build --all-packages`. The built `pipefy` wheel carries `graphql_inputs/`; the snapshot JSON stays out of it, being a dev artifact.

The live drift check ran too: a fresh `snapshot` against the API reproduces the committed file byte for byte, and `check` still passes against it.

The runs below drove the `pipefy-mcp-server` binary built from this branch, over the real MCP protocol on stdio, against a prod organization. Every phase, field, label and condition was created and deleted by the runs.

| 0 | Server built from this branch serves the surface | the seven migrated tools registered | 187 tools, all seven present | PASS |
| 1 | `update_pipe` name | success | `success=True` | PASS |
| 2 | `update_pipe preferences.findabl` | `INVALID_ARGUMENTS` naming `preferences.findabl`, no request | `Invalid arguments for update_pipe: 'preferences.findabl' is not an accepted field.` | PASS |
| 3 | `create_phase` (setup) | a phase id | `344145761` | PASS |
| 4 | `update_phase` description, name resolved by the tool | success | `success=True` | PASS |
| 5 | `create_phase_field` select with options | a field slug | `prioridade_652_19482` | PASS |
| 6 | `create_phase_field extra_input.requred` | `INVALID_ARGUMENTS` naming `requred`, no field created | `'requred' is not an accepted field` | PASS |
| 7 | `update_phase_field` slug + `phase_id` hint | success | `success=True` | PASS |
| 8 | `update_phase_field` hint inside `extra_input` | success, hint lifted rather than sent | `success=True` | PASS |
| 9 | `update_phase_field extra_input.requred` | `INVALID_ARGUMENTS` naming `requred` | `'requred' is not an accepted field` | PASS |
| 10 | `create_label` (setup) | a label id | `318030991` | PASS |
| 11 | `update_label` with a 3-digit hex colour | success, normalized to `#RRGGBB` | `success=True` | PASS |
| 12 | `update_label extra_input.colour` | `INVALID_ARGUMENTS` naming `colour` | `'colour' is not an accepted field` | PASS |
| 13 | `create_phase_field` target (setup) | an internal_id | `433626970` | PASS |
| 14 | `create_field_condition` with integer `expressions_structure` | created; the API places it on the Start form, which the tool reports | `id=307275623 code=FIELD_CONDITION_WRONG_PHASE actual_phase=319036221` | PASS |
| 15 | `update_field_condition` name and actions | success | `success=True` | PASS |
| 16 | `update_field_condition extra_input.nmae` | `INVALID_ARGUMENTS` naming `nmae` | `'nmae' is not an accepted field` | PASS |
| 17 | `delete_field_condition`, preview then confirm | deleted | `success=True` | PASS |
| 18 | `delete_label`, preview then confirm | deleted | `success=True` | PASS |
| 19 | `delete_phase`, cascading both fields (teardown) | deleted | `success=True` | PASS |

Two rows deserve their reasoning written down, because both look like regressions and neither is.

**Row 14.** `createFieldCondition` ignores `phaseId` for a freshly created phase and files the rule under the pipe's Start form. Sending the identical payload through raw GraphQL, with no SDK or MCP layer involved, reproduces it: requested `344145760`, actual `319036221` ("Start form"). This is the behaviour the tool's `_verify_created_field_condition` step and its `FIELD_CONDITION_WRONG_PHASE` code already exist to catch, so the row records the toolkit reporting an API behaviour correctly.

**Row 15.** An update carrying only `name` is rejected with `Validation failed: Condicionais devem ter pelo menos uma ação`. The API requires `actions` on every field-condition update, whatever the schema's nullability says. The scenario sends them.

One known mismatch is left alone deliberately: `update_phase_field`'s `options` argument is annotated `list[Any] | dict[str, Any] | None`, while `UpdatePhaseFieldInput.options` is `[String]`. Neither a dict nor a non-string list was ever usable — no code path handled them and the API refused both — so the effect of this PR is that they now fail before the request instead of after. Narrowing the annotation would change the tool's JSON schema, which this PR otherwise does not touch, so it belongs in its own change.

Not verified: `mcp.pipefy.com`. The hosted deployment pins an exact published release, so it cannot exercise this branch. The SDK is the only layer that changes shape, and it ships in the same wheel.


packages/cli/src/pipefy_cli/commands/field.py
```diff
@@ -62,8 +67,13 @@ def field_create(
         typer.echo("--label and --type must be non-empty.", err=True)
         raise typer.Exit(2)
 
+    create_input = graphql_input_or_bad_parameter(
+        CreatePhaseFieldInput,
+        {"phase_id": phase_id, "label": lab, "type": ft, **extra},
```
adriannoes, line 72: **Blocking.** This flatten puts `**extra` last, so `--extra '{"phase_id": "999"}'` wins over `--phase` and the command still exits 0. Base `create_phase_field(phase_id, label, type, **extra)` raised `TypeError` on a colliding key. MCP still drops `phase_id`/`label`/`type` from `extra_input`. Field-condition CLI in this PR already refuses reserved keys by name.

I would call `reject_reserved_extra_keys` here (or merge dedicated fields last) before `CreatePhaseFieldInput`, naming the refused key and the dedicated argument, and not calling the client.

Done when:

- `pipefy field create --phase 2 --label Owner --type assignee_select --extra '{"phase_id": "999"}'` exits 2 and does not call the client
- a CLI test stays red if that flatten is restored
- `--label` and `--type` in `--extra` are refused the same way

packages/mcp/src/pipefy_mcp/tools/field_condition_tools.py
```diff
@@ -520,7 +535,11 @@ async def update_field_condition(
                 actions: Optional list of ``FieldConditionActionInput`` dicts (same as create).
                     Discover via: ``get_phase_fields(phase_id)[].internal_id`` for ``phaseFieldId``.
                 name: Optional new rule name.
-                extra_input: Additional fields to merge into ``UpdateFieldConditionInput``.
+                extra_input: Additional fields to merge into ``UpdateFieldConditionInput``:
```
adriannoes, line 538: **Optional.** Args now say `index` exists only on create, which matches `UpdateFieldConditionInput`. The unchanged body still cites `index` as an update `extra_input` example, so an agent that copies it gets `INVALID_ARGUMENTS`. Drop `index` from the update example. Your call whether to take this.

docs/sdk/README.md
```diff
@@ -10,6 +10,36 @@ This tree summarizes how to work with **`pipefy`**: the vendor GraphQL client, s
 
 For a short in-repo overview and dev commands, see **[`../../packages/sdk/README.md`](../../packages/sdk/README.md)**.
 
+## Typed mutation inputs
+
+Write methods take one typed input model that mirrors the GraphQL input object, field for field:
```
adriannoes, line 15: **Optional.** The new section opens with "Write methods take one typed input," which reads as repo-wide while fourteen `**attrs` methods remain. CHANGELOG correctly scopes the seven pipe-configuration writes. "Three details" then introduces four bullets. Scope the opener to those seven (or "these writes") and drop or correct the count. Your call.

scripts/generate_graphql_inputs.py
```diff
@@ -0,0 +1,529 @@
...
+
+`ConditionExpressionInput` here is the schema mirror. The separate
+`pipefy_sdk.ConditionExpressionInput` is an older permissive model that the AI
+automation surface still uses; the two converge when those methods take typed
+inputs.
+"""'''
+
+# The input objects an SDK method takes directly. Everything they reference is
+# pulled in transitively, so only the roots are listed. Grows one batch at a
+# time as `**attrs` methods are migrated (#652).
```
adriannoes, line 59: **Optional.** The new comment above `ROOT_INPUT_TYPES` cites `(#652)`. Issue numbers belong in the commit and PR body, not in tracked source. The READMEs already say batches land one at a time without a ticket. Drop `(#652)` from the comment. Your call.

adriannoes (changes requested):
Reviewed at `6a10bafb`: isolated unit gates on a detached worktree, sibling stdio MCP/CLI negatives from that venv, and a runtime refute of the CLI reserved-key hole.

### Required before merge

- `pipefy field create` now lets `--extra` overwrite `--phase` / `--label` / `--type` and still exits 0. Field-condition CLI in this PR already refuses that class of overwrite. See the thread on `packages/cli/src/pipefy_cli/commands/field.py:72`.

### Decisions I would make

- The PR body says the typed dump matches the old `**attrs` merge 7/7. Direct SDK integer primary ids stay ints (`TestIdIsNotCoerced`); MCP/CLI still stringify. I would keep the no-coerce policy and drop or qualify the identity claim, not restore `str(id)`.

### Optional

Your call on all of these.

- 3 smaller pointers are on their lines in the diff.

### What worked well

- Policy lives in `_base.py`; generated files stay `@generated`; the facade still only delegates.
- Unknown extras fail before the network, naming the field, without echoing the rejected value. Sibling stdio at this head proved that path.
- MCP still strips reserved extras on `create_phase_field`.

<details>

Tier `deep`, trust `trusted`, reviewed at `fb63932ccc159a7f52234358e64bd60f7f604f12...6a10bafb4448a70779f10437ce281c7f94c91016`.

| Gate | Result |
| :--- | :--- |
| CI on `dev` | green (lint, test, DCO, shell, install/uninstall) |
| Local ruff + unit pytest | pass (3780 + 33 generator) |
| Live smoke | pass negatives via sibling stdio; mutations blocked (no consent) |
| CLI reserved-key refute | survives (two runtime Attempts) |

Live smoke: worktree stdio, dummy token, no org writes. Not covered: happy-path mutations, hosted `mcp.pipefy.com`.

</details>

Issue #385, opened 2026-07-10
Automations: create accepts event×action combos that never fire


adriannoes:
## Problem
- `create_automation` returns `success: true` for event/action pairs the platform never executes. The rule reads back healthy (`active: true`, `actionEnabled: true`, `disabledReason: null`) and nothing ever runs.
- Two shapes of the same failure: the event against the action (`field_updated` + `move_single_card`), and the event against its own parameters (a `card_moved` trigger whose `to_phase_id` is the pipe's first workflow phase, which cards reach through `card_created`, not through a move).
- Nothing on the read path marks either as dormant. `get_automation_logs` reporting `totalCount: 0` does not separate "structurally dead" from "has not run yet".

## Evidence
- Reproduced live on 2026-08-12: a `field_updated` + `move_single_card` rule was accepted by the API and returned active. `get_automation_actions` on the same pipe reports `move_single_card` with `triggerEvents: ["card_moved", "all_children_in_phase"]`, so `field_updated` is not a valid trigger for it. Re-confirmed against the live catalog on 2026-09-15.
- Reported from product testing: 8 rules of that pairing were created, all returning `success: true`, `actionEnabled: true`, `disabledReason: null`. None fired, the card did not move, 0 log rows.
- `disabledReason` cannot carry the signal. `Automation.disabledReason` is typed `AutomationActionDisabledReason`, whose only three values are `ai_disabled_by_org_settings`, `ai_disabled_by_pipe_settings` and `premium_action`. None expresses an incompatible pair, so a dormant indicator has to be derived client side. `GET_AUTOMATION_QUERY` already selects the field (`packages/sdk/src/pipefy_sdk/queries/automation_queries.py:17`).
- The catalog signals disagree on exactly this case. `move_single_card.eventsBlacklist` is `["scheduler"]` and `field_updated.actionsBlacklist` is `["move_multiple_cards", "schedule_create_card"]`, so only the action's `triggerEvents` allowlist rejects the pair. Verified read-only 2026-09-15.
- `triggerEvents: []` means "no allowlist declared", not "compatible with no event". Live on 2026-09-15 both `send_a_task` and `send_email_template` return `triggerEvents: []` with a populated `eventsBlacklist`, and both create rules that fire. A literal `event_id in triggerEvents` test refuses every valid send-a-task and send-email rule.
[...]

PR #660, opened 2026-09-09
fix(automations): include triggers and conditions in listings

### LINKED ISSUE #612: Automations: get_automations omits condition, event, and trigger params
## Problem
- `get_automations` returns only `id`, `name`, `active` and `action_id`. A caller listing a pipe's rules cannot see what triggers them or whether a trigger condition exists, and concludes there is none.
- `get_automation` (singular) returns all of it, so recovering the data costs one extra call per rule. On a pipe with a dozen rules that is a dozen avoidable round trips.

## Evidence
- `packages/sdk/src/pipefy_sdk/queries/automation_queries.py`: `GET_AUTOMATIONS_BY_ORG_QUERY` at `:152` selects four fields at `:157-160`, and `GET_AUTOMATIONS_FOR_ORG_AND_REPO_QUERY` at `:167` selects the same four at `:172-175`. Re-confirmed on `main` at v0.5.2-beta.1.
- `GET_AUTOMATION_QUERY` at `:7` selects `event_id` (`:14`), `disabledReason` (`:17`), `event_params` (`:23`), `action_params` (`:35`) and `condition` (`:126`).
- The listing row type carries the same four keys: `AutomationRuleSummary` at `packages/sdk/src/pipefy_sdk/services/automation_graphql_types.py:163`.
- No API dependency. The `Automation` type exposes `event_id`, `event_params`, `condition`, `actionEnabled`, `disabledReason`, `schedulerCron`, `scheduler_frequency`, `searchFor`, `updatedAt` and `uuid` on the same nodes the plural queries already reach. Verified read-only 2026-09-15.
- Reproduced live on 2026-08-12: `get_automations(pipe_id=<pipe id>)` on a pipe with one rule returned `{"id": ..., "name": ..., "active": true, "action_id": "move_single_card"}` and nothing else, while `get_automation(automation_id=<automation id>)` on the same rule returned `event_id`, full `event_params`, full `action_params` and `condition`.
- Reported from product testing: a reviewer read the listing, saw no condition, and concluded the automation had none. The build was rejected on that basis. The workaround in use is a hand-maintained GraphQL query.
- #386 closed the equivalent gap on the singular query. The plural queries were not covered.

## Scope
- Extend both plural queries to select `event_id`, `event_params` and `condition`, and add the same keys to `AutomationRuleSummary`.
- Add `actionEnabled` and `disabledReason` while in the query. #385 needs both on the listing for its dormant-rule signal.
- Make the complete shape the default. Keep a compact variant only if a measured payload problem shows up.
- Keep parity across SDK, MCP and CLI, and update `docs/mcp/tools/automations-and-ai.md:16` and the `get_automations` docstring (`packages/mcp/src/pipefy_mcp/tools/automation_tools.py:132`).

## Acceptance
- [ ] `get_automations` on a pipe with a conditional rule returns that rule's `event_id`, `event_params` and `condition`.
- [ ] For the keys both queries select, a listing row matches what `get_automation` returns for the same rule.
- [ ] `AutomationRuleSummary` declares the new keys.
- [ ] `pipefy automation list --json` shows the new fields.
- [ ] Docs and docstring no longer describe the listing as a compact summary.
- [ ] Unit tests assert the new selections on both queries and the resulting payload shape through the MCP tool.

## Notes
- #673 covers the missing pagination on the same two queries. Land both together rather than touching the same selections twice.
- Related: #386 (singular query, closed), #385 (needs `actionEnabled` and `disabledReason` on the listing), #389 (scheduler fields on the read path).

adriannoes:
## Summary

- **Listing fields**: both listing queries select `event_id`, `event_params`, the full `condition`, `actionEnabled`, and `disabledReason`, so `get_automations` / `pipefy automation list` audit triggers, filters, and action state without one `get_automation` call per rule. Empty `condition.expressions` is the API placeholder for "no condition" (confirmed live).
- **Pagination**: the API caps `automations` at 50 rows per page and the toolkit never exposed the cursor. SDK `get_automations` returns the page (`nodes`, `totalCount`, `pageInfo`); MCP `get_automations` and `get_ai_automations` take `first` (1 to 50) / `after` and return `pagination` (`has_more`, `end_cursor`, `page_size`, `total_count`); CLI `automation list` and `ai-automation list` gain `--first` / `--after`. The listing message names the truncation and, on the last page, says so.
- **SDK contract change**: `get_automations` returns `AutomationListPage` instead of a bare list. Four internal callers read `nodes`: `get_ai_automations`, delete-phase dependents, CLI `ai-automation list`, `PipefyClient.get_automation_logs_by_repo` (its empty-pipe guard tests `nodes`). Recorded under Changed in the CHANGELOG.
- **Callers that read one page**: delete-phase preview follows `pageInfo.hasNextPage` to the end; `pipefy automation list` without `--json` prints `nodes` as a table plus `totalCount` / `hasNextPage` / `endCursor`; `--first` on both list commands reuses `validate_cards_page_size(max_size=)`.
- Docs (`automations-and-ai.md`, `parity.md`), the automations skill, and docstrings describe the page contract, including that `get_ai_automations` pagination is of the mixed connection, not the AI subset. Full action payloads stay on `get_automation`.

Closes #612.
Closes #689.

## Event and action compatibility (#689)

Docs-only, in the automations skill and the tools doc. The skill called `field_updated` + `move_single_card` a dead pairing and told agents to gate rule creation on the action's `triggerEvents`. Both are wrong against the shipped catalog, and an agent following them refuses valid rules.

| Claim | What the catalog and API show |
|---|---|
| The constraint is a denylist | A pair is invalid when the event is in the action's `eventsBlacklist`, equivalently the action in the event's `actionsBlacklist`. Cross-checked both directions: 32 pairs each way, identical sets, no orphans |
| `triggerEvents` is not that list | `send_a_task` and `send_email_template` return it empty while the denylist allows them on 7 and 9 of the 10 events; `schedule_create_card` and `move_multiple_cards` list their own `eventsBlacklist` entries inside it |
| `create_automation` enforces the denylist | Rejects with the untranslated key `event_action_blacklist` and writes nothing |
| `update_automation` does not | A rule created as `card_created` + `move_single_card` accepts a patch to `scheduler` + `move_single_card`, and it persists. Checking the pair before an update is the caller's job |

The skill also gains the catalog-to-input spelling map for `event_params` (`trigger_field_ids` is `triggerFieldIds`; `to_phase_id` is the only key that stays snake_case) and the `scheduler_frequency` plus five-field `schedulerCron` shape recurring rules need.

Claims are stated at the level they were verified: the catalog checks are exhaustive, the create and update behaviour was reproduced live, and the one pairing reproduced firing is named as that one pairing rather than generalized across the catalog.

| # | Topic | Choice | Rationale |
|---|---|---|---|
| 1 | Page size cap | `first` validated 1 to 50 on MCP and CLI | The API returns at most 50 whatever `first` says; a larger value would misreport `page_size` |
| 2 | Envelope shape | `data` stays the row list; page info rides in top-level `pagination` | Keeps agents reading `data` working; follows the unified-envelope pagination block |
| 3 | SDK return type | Page dict, not a list plus side channel | A list cannot tell the caller it was truncated |
| 4 | `actionEnabled` / `disabledReason` | Selected on listings, no derived indicator | In the issue's scope and already on `get_automation`; a dormant-rule signal is a separate change |
| 5 | Delete-phase preview | Pages every rule, then one detail call per rule | Listing rows lack `action_params`, where move-card targets live |

## Merge order

This PR and #661, #662, #663 each add bullets under Unreleased in `CHANGELOG.md`, so any two conflict. `origin/dev` is merged here; the clean three-way merge dropped this PR's bullets into `[0.5.1]` (cut after the branch point) and the next commit moves them back. When merging `dev` into the remaining PRs, keep every bullet and check they land under `[Unreleased]`.

## Test plan

- [x] `uv run pytest -m "not integration" -q` (4,771 passed, 5 skipped, 34 deselected)
- [x] Manual smoke, branch MCP server over real stdio and branch CLI against the test org with 210 rules:

| Scenario | Result |
|---|---|
| `get_automations(organization_id)` default page | 50 rows, `total_count 210`, `has_more true`, message says "50 of 210" and how to continue |
| `first=25, after=<end_cursor>` | 25 rows, no overlap with page 1, `has_more true` |
| `first=51` | `INVALID_ARGUMENTS` before any API call |
| `pipe_id` only (37 rules) | 37 rows, `has_more false`, message "37 of 37" |
| CLI `--first 50` then `--after <cursor>` | 50 + 50 rows, zero overlap, `pageInfo.hasNextPage true` |
| CLI `--first 51` | exit 2 |
| Rows on every page | carry `event_id`, `event_params`, `condition` |
| Listing query with `actionEnabled` / `disabledReason`, live schema, `first: 2` on the 37-rule pipe | Accepted; 2 rows, `hasNextPage true`, `totalCount 37` |

Live validation was read-only. The review follow-ups are covered by unit tests in the diff.

## Docs / skills

- [x] `docs/parity.md` updated when MCP ↔ CLI coverage changed
- [x] Affected `skills/` updated in this PR (or a paired PR)
- [ ] No docs/skills update needed

## Legal / contributions


packages/cli/src/pipefy_cli/commands/automation.py
```diff
@@ -38,19 +39,42 @@ def automation_list(
...
+    """
+
+    if first is not None and (first < 1 or first > AUTOMATIONS_LIST_MAX_PAGE_SIZE):
+        raise typer.BadParameter(
+            f"--first must be between 1 and {AUTOMATIONS_LIST_MAX_PAGE_SIZE}."
+        )
+    cursor = after.strip() if after and after.strip() else None
 
     async def factory(client: PipefyClient):
         return await client.get_automations(
```
mocha06, line 89: `pipefy automation list` stops printing a table. It prints raw JSON instead.

How:
1. The command now returns the page object, not the row list.
2. `render` prints a list as a Rich table and a dict as syntax-highlighted JSON.
3. A page is a dict, so every run without `--json` takes the JSON branch.
4. The new query also selects `event_params` and `condition` on every row.
5. A full page is now hundreds of lines where it was a 50-row table.

Fix: render `nodes` as the table and print `totalCount` / `hasNextPage` beside it. Leave `--json` printing the whole page.

Evidence:
- `packages/cli/src/pipefy_cli/output/rich_renderer.py:75-80` @ `2e16eb9a` — `isinstance(data, list)` takes the table branch, `isinstance(data, dict)` takes the JSON branch.
- Probe, the same two rows rendered both ways:

```
dev  (bare list):  6 lines, a table
660  (page dict): 21 lines, JSON
```

- Both new tests in `packages/cli/tests/test_cli_agent_automation_smoke.py` pass `--json`, so nothing covers the human output.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
```diff
@@ -457,10 +457,12 @@ async def _automations_referencing_phase(
...
     """List automations in ``pipe_id`` whose config references ``phase_id`` (summary rows).
 
-    Returns a filtered summary list. Exceptions propagate to the outer gather.
-    Inner per-automation detail fetches are allowed to fail individually.
+    Reads the first page of the pipe's rules (the API caps a page at 50) and returns
+    a filtered summary list. Exceptions propagate to the outer gather. Inner
+    per-automation detail fetches are allowed to fail individually.
     """
-    rows = await client.get_automations(pipe_id=str(pipe_id))
+    page = await client.get_automations(pipe_id=str(pipe_id))
```
mocha06: Deleting a phase on a pipe with more than 50 rules can say no automation is affected, then break one.

How:
1. This reads one page of the pipe's rules. The API sends at most 50.
2. `pageInfo.hasNextPage` arrives on that same response and is not read.
3. Rule 51 onward is never fetched or inspected.
4. The confirm prompt counts only the rules it inspected.
5. The customer confirms, and a rule that pointed at the phase is gone.

This is not new — the API already capped the page before this change. The fix is new.

Fix: loop while `pageInfo.hasNextPage`, passing `after=pageInfo.endCursor`. This change just added both arguments to `get_automations`.

Evidence:
- `packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py:464` @ `2e16eb9a` — one call, no loop; `pageInfo` and `totalCount` are dropped on the next line.
- `packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py:429` @ `2e16eb9a` — the count in the confirm prompt is `len(automations)`.
- Probe: 51 rules on the pipe, only rule 51 references the phase.

```
rules the preview inspected: 50
dependents found           : 0  -> preview says no automations
truth                      : 1 automation (rule 51, inPhaseId=999)
```

mocha06 (commented):

The listing change is sound: both queries now select what `get_automation` selects, and the live schema accepts `first` / `after` on `automations`. Two things regress for people already using the toolkit — a pipe with no automations gets a GraphQL error on its log lookup again, and `pipefy automation list` stops printing a table.

### One blocking finding is not on a changed line

A pipe with no automations fails again when you ask for its automation logs.

How:
1. `get_automations` used to return a list. An empty pipe gave `[]`.
2. It now returns `{"nodes": [], "totalCount": 0, "pageInfo": {…}}`.
3. `packages/sdk/src/pipefy_sdk/client.py:2360` still tests `if not rules`. A three-key dict is never falsy.
4. The guard never fires, so the call reaches `automationLogsByRepo` with no automation ids.
5. The customer reads `MULTIPLE_INVALID_INPUT: Automation_ids can't be blank`.

Fix: test `rules["nodes"]`. The test that covers this still mocks a bare list, so it stays green either way — move it to the page shape in the same commit.

Evidence:
- `packages/sdk/src/pipefy_sdk/client.py:2360` @ `2e16eb9a` — the guard, unchanged by this branch.
- `CHANGELOG.md:317` @ `2e16eb9a` — the guard exists because that call "was returning `MULTIPLE_INVALID_INPUT: Automation_ids can't be blank` from the API".
- `packages/sdk/src/pipefy_sdk/services/automation_service.py:51` @ `2e16eb9a` — `_empty_automation_page()` returns three keys.
- `packages/sdk/tests/services/pipefy/test_automation_service.py:822` @ `2e16eb9a` — `AsyncMock(return_value=[])`, a value the SDK can no longer return.
- Reached from `packages/mcp/src/pipefy_mcp/tools/observability_tools.py:248` and `packages/cli/src/pipefy_cli/commands/automation.py:412`.
- Probe:

```
branch guard `if not rules`        : API called -> MULTIPLE_INVALID_INPUT: Automation_ids can't be blank
fixed  guard `if not rules['nodes']`: short-circuit, API not called
```

### One claim in the description

"the three internal callers (`get_ai_automations`, delete-phase dependents, CLI `ai-automation list`) read `nodes`" — there is a fourth, `PipefyClient.get_automation_logs_by_repo`, and it does not.

### Findings

6 findings. 2 block: the one above, and `packages/cli/src/pipefy_cli/commands/automation.py:73`. The other 4 are inline and non-blocking.

### Question

1 question, posted inline on `packages/sdk/src/pipefy_sdk/queries/automation_queries.py`.

### Read at

- `ai-toolkit` @ `2e16eb9a` (this branch), against `origin/dev`.
- `pipefy-mcp-server` @ `43dbf25c` and `pipefy-ai` @ `c6e531d7`, both at `origin/main`, read-only. Neither calls `get_automations`.
- Live GraphQL schema, read-only: `automations` accepts `first` / `after`; `AutomationConnection` exposes `totalCount` and a non-null `pageInfo`.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
```diff
@@ -231,7 +231,7 @@ async def get_ai_automations(
                     resource_kind="pipe",
                     resource_id=pid,
                 )
-            filtered = filter_ai_automation_summaries(rows)
+            filtered = filter_ai_automation_summaries(page["nodes"])
```
mocha06, line 260: On a pipe with more than 50 rules, the AI automation list can come back short and still read as complete.

How:
1. This keeps the `generate_with_ai` rows out of one page of the pipe's rules.
2. That page holds at most 50 rules, and they are not ordered AI-first.
3. `page["pageInfo"]` and `page["totalCount"]` arrive on the same response and are dropped here.
4. The reply is `AI automations listed.` with no `pagination` block.
5. An agent reads that as the full set and concludes a rule does not exist.

Fix: pass `page["pageInfo"]` through `build_pagination_info` and hand it to `build_automation_read_success_payload(pagination=...)`, the way `get_automations` now does. `pipefy ai-automation list` drops the same two keys.

Evidence:
- `packages/mcp/src/pipefy_mcp/tools/automation_tools.py:198` @ `2e16eb9a` — `get_automations` builds the block from the same page shape.
- `packages/mcp/src/pipefy_mcp/tools/automation_tool_helpers.py:99` @ `2e16eb9a` — the builder already accepts `pagination`.
- `packages/cli/src/pipefy_cli/commands/ai_automation.py:88` @ `2e16eb9a` — same drop on the CLI.
- `skills/automations/pipefy-automations/SKILL.md:22` @ `2e16eb9a` — the skill now tells agents to check `has_more` "before concluding a rule or filter does not exist". This tool gives them nothing to check.

packages/cli/src/pipefy_cli/commands/automation.py
```diff
@@ -38,19 +39,42 @@ def automation_list(
...
     ),
 ) -> None:
-    """List automation rules (``get_automations``)."""
+    """List one page of automation rules (``get_automations``).
+
+    The API caps a page at 50 rules. Output is the page: ``nodes``, ``totalCount``,
+    and ``pageInfo`` (``hasNextPage``, ``endCursor``).
+    """
+
+    if first is not None and (first < 1 or first > AUTOMATIONS_LIST_MAX_PAGE_SIZE):
```
mocha06: `_common.py` already owns the out-of-range `--first` rejection for this CLI. This adds a second copy.

How:
1. `validate_cards_page_size` rejects `--first` outside its bounds and raises `typer.BadParameter`.
2. Its bounds are fixed constants, so this command cannot call it unchanged.
3. The MCP side hit the same wall and solved it by giving `validate_page_size` a `max_size` argument.

Fix: add a `max_size` argument to `validate_cards_page_size` and call it here.

Evidence:
- `packages/cli/src/pipefy_cli/commands/_common.py:78` @ `2e16eb9a` — the existing check, called from `packages/cli/src/pipefy_cli/commands/phase.py:133`.
- `packages/mcp/src/pipefy_mcp/tools/pagination_helpers.py:40` @ `2e16eb9a` — `max_size`, the same problem parameterized once.

packages/mcp/src/pipefy_mcp/tools/automation_tool_helpers.py
```diff
@@ -97,18 +99,41 @@ def build_automation_simulation_success_payload(
...
+
+def build_automations_listed_message(listed: int, total: int, *, has_more: bool) -> str:
+    """``message`` for ``get_automations``: rows in this page vs total, plus how to continue.
+
+    Args:
+        listed: Rows in this page.
+        total: ``totalCount`` of the connection.
+        has_more: ``pageInfo.hasNextPage``; when true the listing is incomplete.
+    """
+    message = f"Automations listed: {listed} of {total}."
```
mocha06, line 131: The last page of an audit reports as a shortfall: `Automations listed: 11 of 61.` and nothing else.

How:
1. `listed` is the rows in this page. `total` is the rows in the whole connection.
2. On any page after the first, those two never match.
3. When `has_more` is false the message stops there, with no "this is the last page".
4. The tool's own docstring tells the agent an audit is complete only when `has_more` is false.
5. The agent reads "11 of 61", takes 50 rules as missing, and pages again from the start.

Fix: when `has_more` is false, say the page is the last one. The bare count cannot be read without a page number this tool does not have.

Evidence:
- `packages/mcp/src/pipefy_mcp/tools/automation_tools.py:202` @ `2e16eb9a` — `listed` is `len(page["nodes"])`, `total` is `page["totalCount"]`.
- Probe:

```
page 1 (50 of 61, more): 'Automations listed: 50 of 61. More rules exist; call again with ...'
page 2 (11 of 61, done): 'Automations listed: 11 of 61.'
```

packages/sdk/src/pipefy_sdk/queries/automation_queries.py
```diff
@@ -151,13 +151,42 @@
...
+            pageInfo {
+                hasNextPage
+                endCursor
+            }
             nodes {
                 id
                 name
                 active
                 action_id
+                event_id
```
mocha06, line 168: The Scope section of #612 asks this change to add `actionEnabled` and `disabledReason` to both listings while it is in the query. Was leaving them out deliberate?

The reason #612 gives for them no longer holds. It cites #385's dormant-rule signal, and that signal rests on `field_updated` + `move_single_card` being a pair that never fires. It fires — reproduced live on 2026-09-16, `status: "success"` in `get_automation_logs`. See #385 and #689.

`disabledReason` may still earn its place on its own three values: `ai_disabled_by_org_settings`, `ai_disabled_by_pipe_settings`, `premium_action`. Those are real, and none of them depends on #385.

So the question changes shape rather than going away: is there a reason to select these two fields that is not #385?

Evidence:
- `packages/sdk/src/pipefy_sdk/queries/automation_queries.py:16-17` @ `2e16eb9a` — `GET_AUTOMATION_QUERY` selects both, so the fields exist on the same `Automation` type these queries reach.
- `AutomationActionDisabledReason` carries exactly those three values, so the field cannot express an incompatible event and action pair.

packages/cli/src/pipefy_cli/commands/automation.py
adriannoes: Fixed in a2a69b50. `automation list` runs the coroutine through `run_pipefy_client_coroutine` and renders on its own: `--json` prints the whole page, otherwise `render_rich(page["nodes"])` draws the table and one line follows with `totalCount=... hasNextPage=... endCursor=...`.

`test_automation_list_human_prints_table_and_page_counts` covers the human output (row present, counts present, no `"nodes"` key in stdout). The existing `--json` tests keep asserting the page.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
adriannoes: Fixed in 645ce7e0. `_list_pipe_automation_summaries` loops on `pageInfo.hasNextPage` with `first=50` and `after=endCursor`, and stops if the cursor comes back empty or unchanged. `_automations_referencing_phase` consumes that list, so the per-rule detail fetch and the confirm count see every rule.

`test_delete_phase_preview_pages_automations_past_the_api_cap` reproduces your probe: 51 rules, only rule 51 references the phase, two listing calls (`after=None`, then `c50`), and the preview lists that one rule with `1 automation` in the hint.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
adriannoes: Fixed in 62b07dc1. `get_ai_automations` takes `first` (1 to 50, via `validate_page_size(max_size=AUTOMATIONS_LIST_MAX_PAGE_SIZE)`) and `after`, and returns `pagination` built from `page["pageInfo"]` plus `total_count`, the same way `get_automations` does. `pipefy ai-automation list` gains `--first` / `--after` and the same block.

The docstring, `automations-and-ai.md`, `parity.md`, and the skill say the block describes the mixed connection, not the AI subset, and to continue while `has_more` is true. Tests: the default page passes `first=50`; a truncated mixed page with no AI rows returns `data []` with `has_more true`; `first=51` is rejected before the API call.

packages/cli/src/pipefy_cli/commands/automation.py
adriannoes: Fixed in 00d57da7. `validate_cards_page_size` takes a keyword `max_size` (the default stays the cards cap of 500) and both `automation list` and `ai-automation list` call it with `max_size=AUTOMATIONS_LIST_MAX_PAGE_SIZE`. The local check is gone. `test_automation_list_rejects_first_above_api_cap` still exits 2 with `between 1 and 50` and no client call.

packages/mcp/src/pipefy_mcp/tools/automation_tool_helpers.py
adriannoes: Fixed in 758b2f68. When `has_more` is false the message reads `Automations listed: 11 of 61. This is the last page.`; when true it keeps the `after=pagination.end_cursor` instruction. Two unit tests pin both branches of `build_automations_listed_message`, and the tool test for a complete page asserts the new text.

packages/sdk/src/pipefy_sdk/queries/automation_queries.py
adriannoes: Not deliberate, an omission. Both scalars are selected in ca477442 (queries, `AutomationRuleSummary`, docs, skill, CHANGELOG), and the live schema accepts them on the connection nodes.

The reason to select them is not the dormant-rule signal. The issue's scope asks for them on both listings, `get_automation` already returns them, and `disabledReason` is useful on its own three values (`ai_disabled_by_org_settings`, `ai_disabled_by_pipe_settings`, `premium_action`): an audit can tell an inactive rule from one whose action the org or plan turned off, without a detail call per rule. This PR derives no indicator from them; that stays a separate change, recorded as row 4 of the design table in the body.

adriannoes:
Thanks for the ground-truth pass. All six findings and the question are addressed on the branch, each on its own commit, and every thread has a reply.

The blocking finding off the changed lines: `get_automation_logs_by_repo` tests `rules["nodes"]` in 044c731f. The test that mocked a bare list moved to the page shape in the same commit, and a second test asserts the logs query runs when the page has rows.

The description claim: you are right, there are four internal callers. The body now lists `PipefyClient.get_automation_logs_by_repo` with the other three.

One thing beyond the findings. `origin/dev` is merged into the branch (a2480c08): `dev` cut 0.5.1 and 0.5.2-beta.1 after this branch's merge-base, and the clean three-way merge put this PR's two CHANGELOG bullets under `[0.5.1]`. 54ef05cf moves them back under `[Unreleased]`, and the Merge order section of the body now warns the sibling PRs about the same placement.

packages/cli/src/pipefy_cli/commands/automation.py
mocha06: Verified. Table renders, the counts line follows, `--json` still prints the page.

```
│ a1 │ Move on approval │ True  │ move_single_card │
totalCount=61 hasNextPage=True endCursor=c50
```
No `nodes` key in human stdout.

One thing my finding brought back, and it is not yours. A rule name with brackets goes through Rich's markup parser:

```
'[on hold] escalate'           printed 'escalate'      <-- ALTERED
'Notify [/marketing] team'     CRASH exit 1  MarkupError
'[HR] Performance Evaluation'  survives
```

`origin/dev` crashes the same way — it also passes a list to `_render_list_of_dicts`. The page-dict version at `2e16eb9a` avoided it by printing JSON, so asking for the table back re-exposed it. Pre-existing, not a regression, and name-dependent so a smoke test will not catch it.

Fix: build the table with `markup=False`, or `rich.markup.escape` each cell in `_render_list_of_dicts`. Your call whether it rides here or goes to its own issue.

Evidence: `packages/cli/src/pipefy_cli/output/rich_renderer.py:49-57` @ `ecb838bb`.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
mocha06: Verified. Two listing calls, `after=[None, 'c50']`, 51 details fetched, rule 51 found, hint reads `1 automation`.

The loop is correct. It costs more than it needs to, and I did not ask for a bound when I asked for the loop. Two separate comments on the lines: `:462` for the round trips, `:492` for the fan-out.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
mocha06: Verified. `first` / `after` on both surfaces, `pagination` built from the mixed page, `--first 51` rejected before the call. `parity.md:106`, `automations-and-ai.md:149` and `SKILL.md:43` all say the block describes the mixed connection.

One contract point on `page_size`. It is 50 while `data` holds only the AI rows, so `len(data) < page_size` means "done" on every other tool and means nothing here. `total_count` cannot cover for it either, since it counts the mixed set. `has_more` is the only correct field, and it is the one that heuristic ignores.

The prose says this in three places. The field still contradicts the prose. Worth considering `ai_count` beside `total_count`, or dropping `page_size` from this tool's block.

Not blocking. `PaginationInfo` holds the same-set invariant everywhere else: `packages/mcp/src/pipefy_mcp/tools/pagination_helpers.py:24-33` @ `ecb838bb`.

packages/cli/src/pipefy_cli/commands/automation.py
mocha06: Verified. `validate_cards_page_size(first, *, max_size=_CARDS_PAGE_SIZE_MAX)`, local check gone, both automation commands pass `max_size=AUTOMATIONS_LIST_MAX_PAGE_SIZE`. `phase.py:133` and `card.py:176` still get 500. The error string interpolates `max_size`, so an override names the right bound.

packages/mcp/src/pipefy_mcp/tools/automation_tool_helpers.py
mocha06: Verified.

```
has_more=True : Automations listed: 50 of 61. More rules exist; call again with after=pagination.end_cursor. …
has_more=False: Automations listed: 11 of 61. This is the last page.
```

packages/sdk/src/pipefy_sdk/queries/automation_queries.py
mocha06: Verified. Both scalars are selected on both plural queries, and row 4 of the design table states the reason without leaning on the dormant-rule signal. That answers the question.

One asymmetry, for the record rather than for this PR: `GET_AUTOMATION_QUERY` and `AutomationRuleRecord` do not carry either field, so `get_automation` now returns less than the listing for those two. Nothing documents otherwise, so nothing is wrong.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
```diff
@@ -452,15 +452,38 @@ def _build_phase_dependents_hint(deps: dict[str, Any]) -> str:
...
 
 
+async def _list_pipe_automation_summaries(
+    client: PipefyClient, pipe_id: str
+) -> list[dict[str, Any]]:
+    """Collect every listing row for ``pipe_id``, following ``pageInfo`` until the end."""
+    rows: list[dict[str, Any]] = []
+    after: str | None = None
+    while True:
+        page = await client.get_automations(
```
mocha06: This asks the API for the pipe's organization again on every page, and one failed page discards every page before it.

How:
1. The call passes `pipe_id` with no `organization_id`.
2. `automation_service.py:240` then runs `GET_PIPE_ORGANIZATION_ID_QUERY` to resolve the org.
3. That repeats per iteration, for the same pipe, with the same answer.
4. When any iteration raises, the exception leaves this function entirely.
5. `resolve_phase_dependents` catches it at `:522` and drops the `automations` key, so the preview shows no automation line at all.

Fix: resolve the org once before the loop and pass it on every call. For step 5, a partial page set is worth more than none — keep what you collected and mark it partial.

Evidence:
- Probe, 200-rule pipe: `listing queries: 4`, `org-lookup queries: 4`, 8 round trips where 5 would do.
- `packages/sdk/src/pipefy_sdk/services/automation_service.py:240` @ `ecb838bb` — the resolve runs whenever `organization_id` is None.
- `packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py:522` @ `ecb838bb` — `gather(..., return_exceptions=True)`, and `:545` omits the key on a `BaseException`.

mocha06 (commented):
## Re-review at `ecb838bb`

All six findings and the question are fixed. Every claim in your six replies and your summary comment verified — by probe where the behaviour is runtime, by citation otherwise. Details in each thread.

The two things I could not see last time also check out: both stale test mocks moved to the page shape, and the CHANGELOG bullets sit under `[Unreleased]` after the `dev` merge.

### One new finding, on a line outside the diff

`packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py:492` — deleting a phase on a busy pipe fires hundreds of simultaneous API calls, and the confirm prompt then under-counts what breaks.

How:
1. The listing now returns every rule, not the first 50. This line starts one detail call per rule.
2. A pipe with 200 rules gives 200 simultaneous requests. Nothing caps them.
3. Under throttling a subset raises.
4. The next line keeps only the calls that returned a dict, so the failures vanish.
5. The prompt says "2 automation(s)" when 9 reference the phase, and nothing says the fetch was partial.

Fix: gate the calls behind `asyncio.Semaphore`. `ai_pipe_validation.py` already does this with `PHASE_FIELD_FETCH_CONCURRENCY = 8`. Separately, count the raised entries and say the preview is partial rather than dropping them.

This one is mine, not yours. I asked for the loop and did not ask for a bound.

Evidence:
- `packages/sdk/src/pipefy_sdk/ai_pipe_validation.py:24` @ `ecb838bb` — the existing cap, used at `:344` and `:518`.
- `grep -rn "Semaphore" packages/sdk/src packages/mcp/src` — only `ai_pipe_validation.py` and `observability_export_csv.py`. No client-level cap absorbs this.
- Probe, 200-rule pipe: `then detail calls, all at once: 200 concurrent get_automation`.
- At `2e16eb9a` the same gather was capped at 50 by the unpaged listing. The fan-out is new.

At ten times the volume this is worse than slow: 5000 rules is 5000 concurrent requests to render a preview the caller may cancel.

### One more on the loop

`packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py:462`, inline. The loop re-resolves the pipe's organization on every page, and a failure on any page discards the pages before it.

### One pre-existing bug my finding re-exposed

A bracketed rule name is mangled or crashes `pipefy automation list`. `origin/dev` does this too. The page-dict rendering at `2e16eb9a` happened to avoid it, and my B2 finding put the table back. In the `automation.py:74` thread, with the fix.

### One contract point

`get_ai_automations` reports `page_size` and `total_count` for the mixed connection while `data` is the AI subset. Documented in three places; the field still reads against the prose. In the `ai_automation_tools.py` thread. Not blocking.

### Read at

- `ai-toolkit` @ `ecb838bb`, against `origin/dev`.
- Live GraphQL schema, read-only, for `actionEnabled` and `disabledReason` on the connection nodes.
- Six probes, all in-process against this branch: the empty-pipe guard, the CLI table, the phase-delete loop, the page message, the round-trip count, and the markup cases.

packages/cli/src/pipefy_cli/commands/automation.py
adriannoes: Taking the markup bug here rather than to its own issue, in 10537fd6. It is a crash, and asking for the table back is what made it reachable again, so it should not outlive the change that re-exposed it.

Reproduced your three cases before and after:

```
'[on hold] escalate'          before: ' escalate'     after: '[on hold] escalate'
'Notify [/marketing] team'    before: MarkupError     after: 'Notify [/marketing] team'
'[HR] Performance Evaluation' before: unchanged       after: unchanged
```

Fix is `Text` around every cell, on both `Table` sites in `rich_renderer.py`, not just the dict one. Rich 14 has no `markup` argument on `Table`, so the choice was `Text` or `rich.markup.escape`; `Text` says what it means and skips the escape round trip. Nothing in the renderer styles a cell and no test asserted on markup, so the blast radius is every CLI table printing its values verbatim, which is what they were always supposed to do.

Two tests in `test_output_renderers.py` cover the dict path and the primitives path. Both fail with the change reverted. Full suite green, 4774 passed.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
adriannoes: Confirmed, both halves. The loop passes a pipe id with no organization id, and `automation_service.py:240` resolves the organization whenever that argument is None, so the lookup repeats per page for the same answer. And any page raising leaves the function entirely, so `:545` drops the key and the preview shows no automation line at all.

Sending this to #690 rather than fixing it here. It is the same loop and the same theme as the fan-out, the cost that came with the paging fix, and the two want one change together: resolve the organization once before the loop, keep the pages already collected when a later one fails, and mark the result partial instead of dropping it.

Your fan-out finding also corrected that issue. I had written it up as the burst making the whole sub-lookup raise. It does not: the gather uses `return_exceptions=True` and `:496` keeps only the dict entries, so individual failures never raise and the preview renders a count that reads as complete. The missing-section path is the listing loop, which is this thread. Both are now in #690 with the right mechanism.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
adriannoes: Agreed, and the prose mitigation is not enough on its own. Filed as #691 with both directions written out: drop `page_size` from this tool's block, or add a count that matches `data`.

Recorded there as the constraint behind it: the API filters nothing server side for `generate_with_ai`, so the block cannot be recomputed against the filtered rows. The note also flags that #673's server-side filters would close this as a side effect, since `actionId` reaching this tool makes the connection itself the AI subset. Worth settling that order before building either direction.

Keeping it out of this PR: the shape is a design choice, it ships documented in three places, and changing the envelope is not something to fold into a fix.

adriannoes:
Re-review addressed at 10537fd6. One fix rode here, the rest went to issues.

| Finding | Where it landed |
|---|---|
| Bracketed rule name mangled or crashing `automation list` | Fixed here in 10537fd6, with tests that fail without it |
| Unbounded detail fan-out on the phase-delete preview | #690 |
| Partial detail failures dropped, count reads as complete | #690, and it corrected that issue: see below |
| Organization re-resolved on every listing page, failed page discards the rest | #690 |
| `page_size` describes the mixed connection while `data` is the AI subset | #691 |
| `get_automation` does not carry the two new scalars | No action, recorded in the thread |

The markup one is the only change to the branch. `Text` around every table cell, both `Table` sites, so a name like `[on hold] escalate` prints whole and `Notify [/marketing] team` stops raising `MarkupError`. Rich 14 has no `markup` argument on `Table`, so it was `Text` or `escape`. Full suite 4774 passed, lint clean.

Your fan-out finding also fixed a mistake in #690. I had written it up as the burst making the whole sub-lookup raise and the preview drop its automations section. The gather uses `return_exceptions=True` and keeps only the dict entries, so individual failures never raise and the preview shows a confident undercount instead. The missing-section path belongs to the listing loop, which is your `:462` finding. Both are now in #690 with the right mechanism, along with the organization lookup.

mocha06 (changes requested):
The six findings and the question from the first pass are fixed and verified. Three of those threads are resolved.

Holding on one thing: `pipe_config_tool_helpers.py:492` fires one detail call per rule with no cap, and drops the ones that fail, so the delete-phase confirm prompt can under-count what the delete removes. A destructive prompt that under-reports is the reason this is a block rather than a note. Full write-up in the review summary.

Also open, neither blocking:
- `pipe_config_tool_helpers.py:462` — the org lookup repeats per page, and one failed page discards the pages before it.
- `ai_automation_tools.py` — `page_size` and `total_count` describe the mixed connection while `data` is the AI subset.

The bracketed-name crash is not yours and not this PR's. It is pre-existing on `dev` and now tracked as #692.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
adriannoes: Fixed in 2c907af2, both halves, together with the `:492` block since they share the plumbing.

The organization is resolved once. The SDK gained `get_pipe_organization_id`, which is the lookup `get_automations` was already doing internally per call; `get_automations` now calls it too, so there is one implementation rather than two. The loop resolves up front and passes `organization_id` on every page. A two-page walk awaits the resolver once and sends the same org on both calls, pinned by a test.

A failed page no longer discards the pages before it. The loop catches, returns what it collected, and reports the walk as incomplete. That flows to `automations_partial` in the payload and to the hint.

This changes a contract two existing tests encoded, so I moved them rather than leaving them green: a failed listing used to make the automations section disappear, and the all-lookups-fail case used to return no `dependents` at all. Both now assert the preview says it could not read the rules.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
adriannoes: Fixed in 2c907af2. You were right to block: I had sent this to #690 as follow-up, and a destructive prompt that under-counts is not a follow-up.

The bound: detail reads go through `asyncio.Semaphore(AUTOMATION_DETAIL_FETCH_CONCURRENCY)`, 8, the same number and the same pattern as `PHASE_FIELD_FETCH_CONCURRENCY`, rather than a second convention. A test drives 40 rules through a counting mock and asserts peak concurrency never passes the bound while still exceeding 1, so it fails both if the bound goes away and if the calls accidentally serialize.

The silent drop mattered more than the bound, and it is the part I had described wrongly in #690. `return_exceptions=True` plus the keep-only-dicts filter means a throttled read never raises, so the old code could not have known it was short. Now `_automations_referencing_phase` returns the matches and whether the walk completed, `resolve_phase_dependents` sets `automations_partial`, and the hint reads:

```
Deleting this phase will remove at least 1 automation. This action is irreversible.
Some automation rules could not be read, so the automation count is a lower bound.
```

Tests cover a failed detail read, a failed listing page, and the all-fail case. Each fails with its own piece of the plumbing removed, verified by reverting them one at a time. Full suite 4781 passed.

adriannoes:
Blocker fixed at 2c907af2, with the non-blocking `:462` item in the same commit since the two share the partial-reporting plumbing.

You were right to block rather than take my referral to #690. I had treated the fan-out as a cost problem, which is why follow-up felt reasonable. It is a correctness problem in a destructive prompt, and those are not the same thing.

What changed:

- Detail reads run behind `asyncio.Semaphore`, bound 8, the number and pattern already used by `PHASE_FIELD_FETCH_CONCURRENCY`.
- `_automations_referencing_phase` returns the matches and whether the walk completed, instead of silently keeping whatever came back as a dict. `resolve_phase_dependents` surfaces that as `automations_partial`, and the hint switches to "at least N automation" plus a sentence naming the caveat.
- A failed listing page keeps the pages already collected rather than discarding the whole section.
- The organization resolves once per preview through a new SDK `get_pipe_organization_id`, which is the same lookup `get_automations` did internally; that method now calls it too, so there is one implementation.

Two existing tests encoded the old contract, where a failure made the automations section vanish. I moved them to the new one rather than leaving them passing on a behaviour we just decided was wrong.

On the bracketed-name crash and #692: that one is already fixed on this branch, in 10537fd6, before you filed the issue. The fix wraps cells in `Text` in `rich_renderer.py`, so it covers every command that renders a list of dicts, not just `automation list`. Both your failing cases are pinned by tests that fail without it. One residual your write-up is right about and my fix does not cover: column headers still go through the markup parser. I left it because no command builds rows with data-derived keys, so there is no path to it today. Your call whether #692 closes with this PR or stays open narrowed to the headers.

#691 has the live numbers now: a real call returning one row on `page_size` 5 with `has_more` true.

packages/cli/src/pipefy_cli/commands/automation.py
mocha06: Verified at `2c907af2`. All six cases print whole, including the one that raised:

```
'[on hold] escalate'       -> '[on hold] escalate'
'Notify [/marketing] team' -> 'Notify [/marketing] team'
```

Agreed on taking it here. Wrapping cells at both `Table` sites covers every command that renders a list of dicts, which is wider than the issue asked for.

Your header caveat is right, and I checked the reachability claim too. `Table(*keys)` still parses headers — `Notify [/x] col` raises. No command can reach it: nested payloads take the JSON branch, and the rest get keys from GraphQL field names and aliases, which cannot contain `[`. #692 narrowed to the headers.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
mocha06: Verified at `2c907af2`. The block is clear.

```
40 rules -> peak concurrent get_automation = 8  (cap 8)
3 reads throttled -> matches 2 -> 1, complete=False
hint: Deleting this phase will remove at least 1 automation. … the automation count is a lower bound.
```

Peak is 8 and above 1, so the bound holds and the calls do not serialize. The throttled rule was one of the two that matched the phase, so this is the exact case the old code reported as a confident 2.

You are right that the silent drop mattered more than the bound, and right that my write-up in the issue had the mechanism wrong. Reporting the lower bound is the part that makes the prompt honest.

packages/mcp/src/pipefy_mcp/tools/pipe_config_tool_helpers.py
mocha06: Verified at `2c907af2`, both halves.

```
200 rules   -> listing queries=4, org lookups=1   (was 4)
page 3 fails -> rows kept=100, complete=False      (was: all discarded)
```

Folding it in with the fan-out was the right call: the partial flag is the same plumbing, and shipping the bound without it would have left the prompt confident and wrong.

Moving the two tests that encoded the old contract is the right handling. A test that pins behaviour we just decided is wrong is worth less than no test.

packages/mcp/src/pipefy_mcp/tools/ai_automation_tools.py
mocha06: Settled. #691 carries both directions and the constraint that the API filters nothing server side, so the block cannot be recomputed against the filtered rows.

Agreed on keeping it out of this PR. The shape is a design choice and it ships documented in three places.

The #673 ordering point is the useful part: if `actionId` reaches this tool, the connection becomes the AI subset and the mismatch disappears without an envelope change. Worth settling that before building either direction.

mocha06 (dismissed):
Approving at `2c907af2`. Every finding from both passes is fixed and verified against this head, and all seven threads are resolved.

The blocker, re-run here:

```
40 rules   -> peak concurrent get_automation = 8  (cap 8, and above 1)
3 throttled -> matches 2 -> 1, complete=False, hint says "at least 1 automation"
200 rules  -> listing queries=4, org lookups=1   (was 4)
page 3 fails -> rows kept=100, complete=False     (was: all discarded)
```

The throttled rule in that second line was one of the two matching the phase. That is the exact case the old code reported as a confident 2, so the lower-bound wording is doing real work.

Two things I want on the record.

You pushed back on your own referral and fixed the blocker instead of shipping it to #690. That was the right read: a destructive prompt that under-counts is a correctness bug, not a cost one.

You also corrected #690 after my finding showed the mechanism differed from your write-up, and you flagged a residual in your own markup fix that I had not asked about. I checked that residual — the headers do still parse, and nothing can reach them today, because every table path either takes the JSON branch or draws its keys from GraphQL names and aliases, which cannot contain `[`. #692 is narrowed to it.

`f353ee28` is not something I raised and it is the right call. 144 lines to 29 for a six-row page, with both structures still on `--json`, which is where #612's acceptance put them.

adriannoes:
New scope on the branch since your approval, plus a correction I found while checking it.

`a7e2d1c3` added the #689 docs work: the skill called `field_updated` + `move_single_card` a dead pairing and told agents to gate on `triggerEvents`. The PR body now carries a section for it and closes #689, which it did not before.

Then I verified its claims against the live catalog rather than shipping them, and three went past the evidence. `3a9a3534` corrects them.

The one that matters: **`update_automation` does not enforce the denylist.** Both surfaces said the API rejects a blacklisted pair on create and update. Create does. Update does not.

```
create  scheduler + move_single_card  -> rejected, event_action_blacklist, nothing written
create  card_created + move_single_card -> id 308108419
update  that rule's event_id -> scheduler -> accepted
re-read -> event_id: scheduler, action_id: move_single_card   (blacklisted pair, persisted)
```

The probe rule was deleted after the read-back. An agent trusting the old wording would rely on a guard that is not there on update, and the troubleshooting step that ruled out the pairing because it is "refused at write time" rested on the same premise. Both now say create enforces it and update does not.

The other two were runtime claims backed only by catalog structure. "Every action with a non-empty `triggerEvents` runs with events outside it" became "the denylist allows it", since one pair was reproduced firing, not twelve. And the empty `triggerEvents` of `send_a_task` and `send_email_template` now carry the counts the denylist permits, 7 and 9 of 10, instead of asserting those rules fire.

What does hold, checked exhaustively: `eventsBlacklist` and `actionsBlacklist` are symmetric across the catalog, 32 pairs from each direction with no orphans.

Everything from both of your earlier passes is untouched. Full suite 4781 passed.

mocha06 (approved):

Issue #692, opened 2026-09-17
CLI: bracketed column headers are mangled or crash the table renderer


mocha06:
## Problem

`render_rich` passes cell values to Rich without escaping, so any user-authored string containing square brackets is read as Rich markup. Bracket prefixes are a common naming convention for pipes and automation rules, so this hits real data.

## Evidence

`pipefy automation list` on a pipe whose rules use bracketed names:

```
'[on hold] escalate'           printed 'escalate'      <-- prefix silently deleted
'[blue] team notify'           printed 'team notify'   <-- prefix silently deleted
'Notify [/marketing] team'     CRASH exit 1  MarkupError
'[HR] Performance Evaluation'  survives
'[SLA/2h] warn'                survives
```

`rich.errors.MarkupError: closing tag '[/marketing]' at position 7 doesn't match any open tag`.

Whether a name breaks depends on whether the bracketed text happens to parse as a Rich tag, so a smoke test with one sample name will not catch it.

- `packages/cli/src/pipefy_cli/output/rich_renderer.py:49-57` — `_render_list_of_dicts` calls `table.add_row(*(str(row.get(k, "")) for k in keys))`.
- Any command that renders a list of dicts is affected, not only `automation list`.

## Scope

- Build the table with `markup=False`, or run each cell through `rich.markup.escape`.
- One test with a name that deletes a tag and one with a name that raises.

## Acceptance

- [ ] `[on hold] escalate` prints with its prefix intact.
- [ ] `Notify [/marketing] team` prints instead of raising.

adriannoes:
This is fixed on the branch of #660, at 10537fd6, which predates this issue.

`_render_list_of_dicts` and the primitives table both wrap cells in `rich.text.Text`, so values are never parsed as markup. That is in the shared renderer, so it covers every command that renders a list, matching your note that this is not only `automation list`. Rich 14 has no `markup` argument on `Table`, so the choice was `Text` or `rich.markup.escape`; `Text` avoids the escape round trip. Both acceptance cases are pinned by tests in `test_output_renderers.py` that fail with the change reverted:

```
'[on hold] escalate'        before: ' escalate'   after: '[on hold] escalate'
'Notify [/marketing] team'  before: MarkupError   after: 'Notify [/marketing] team'
```

One residual, which your write-up covers and the fix does not. Column headers are still passed as bare strings, so a header containing brackets is parsed as markup:

```
[...]

PR #693, opened 2026-09-17
fix(cli): escape bracketed table column headers

abhinyaay:
## Summary

- `render_rich` passed column keys to `rich.table.Table` as bare strings, so a header containing square brackets was read as console markup: `[on hold] col` printed as ` col` and `Notify [/marketing] col` raised `MarkupError`. Cell values were already made literal by wrapping them in `Text` (#660); this extends the same protection to the header row.
- `Table.__init__` rejects a `Text` header (`AttributeError: 'Text' object has no attribute '_index'`), so the keys are passed through `rich.markup.escape` instead of wrapped.
- Adds a regression test asserting a bracketed header prints intact and does not raise, matching the acceptance criteria in #692.

As noted in the issue, no shipped command builds table rows from data-derived keys today (keys are GraphQL field names, which cannot contain `[`), so this is defensive: it stops the next command that renders data-derived columns from reintroducing the crash.

## Test plan

- [x] `uv run pytest -m "not integration"` (packages/cli: 449 passed)
- [ ] Manual smoke (if applicable): callable via Cursor MCP

## Docs / skills

## Legal / contributions

Closes #692


adriannoes (approved):
Reviewed at `2e419a69`: static read of the two-file diff against Rich 14's markup parser, plus the green GitHub Actions run on this head (DCO, shell, install, lint, test). Local pytest was not repeated. The author is a first-time fork contributor; the diff is header escaping and an in-memory test, with no auth, workflow, or dependency change.

### What worked well

- Headers go through `rich.markup.escape` because `Table` will not take a `Text` header, and cell lookup still uses the raw key.
- The new test calls public `render_rich` with both leftover cases from #692: a tag that would be swallowed, and a closing tag that would raise.
- The write-up matches the narrowed issue and leaves the #660 cell fix alone.

<details>

Tier standard, trust sandbox-required (static-only, no sandbox). Reviewed at `fda5835630bffe8d8341d1e713cdf42df7fd8641...2e419a694ee6cc21a638d40fb19ec85dc3f5e265`.

| Gate | Result |
| :--- | :--- |
| DCO | pass |
| CI (lint, test, shell, install) | pass on this head |
| Local pytest | not run (fork, no sandbox) |
| Contract, tests, architecture, API shape, claims, security, structure, disclosure | pass on a static read |
| Live smoke | not run; bracketed headers are not reachable from live commands |

Not covered: a local revert of the new test. Upstream test is green.

</details>

adriannoes:
Thanks for picking up the narrowed #692 header fix — the `rich.markup.escape` change and regression test were exactly what we needed. Hope to see you around on future contributions!

